Issue #23, opened 2026-08-06
Run the conformance suite from committed vectors on every pull request


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/00-foundation.md`](docs/specs/features/00-foundation.md):

> FR-foundation-4 — The conformance suite runs on every pull request.

> FR-foundation-5 — The conformance suite fails when a vector produces an unexpected fingerprint.

> FR-foundation-6 — The conformance suite fails when it collects zero vectors.

## What to build

The suite currently reads `tests/foxio_vectors/` and skips silently when the directory is empty, which is why the gap survived. A vector that exercises no method this project implements is reported as not applicable and does not count as a pass. The download script must never run during a test: a test reads a committed file or it fails and names the missing file.

Relevant files: `tests/test_spec_validation.py`, `tests/download_test_vectors.py`, `.github/workflows/test.yml`.

## Acceptance criteria

- [ ] `pytest tests/ -m spec_validation` passes with no network access.
- [ ] `pytest tests/ -m spec_validation` fails when a vector file is deleted.
- [ ] The conformance suite fails when it collects zero vectors.

## Links

- Feature: [`docs/specs/features/00-foundation.md`](docs/specs/features/00-foundation.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #11
Depends on #22

Crank-Git:
## A sequencing decision this issue must resolve

With the vectors committed (#22), `pytest tests/` is now **red** on `epic/11-foundation`: 5 of 74 conformance vectors fail. That is correct behaviour — `FR-foundation-5` requires the suite to fail when a vector produces an unexpected fingerprint — but it means Epic 0 cannot merge green while the fingerprinters are still wrong, and Epic 1 is what fixes them.

Three ways out. The second is recommended.

1. **Let Epic 0 merge red.** Honest, but it removes the signal: every later change lands against a red suite and nobody can tell a new break from an old one.

2. **Mark the five known-failing vectors `xfail`, each naming the issue that will fix it.** The suite goes green, the known non-conformance stays visible and counted, and an `xpass` tells Epic 1 exactly when a fix has landed. Epic 1 removes each marker as it fixes the method. This is the standard way to record known non-conformance without losing the gate.

[...]

Issue #78, opened 2026-08-06
Diagnose and fix the 35 JA4X conformance failures


Crank-Git:
## Context

The conformance baseline measured on `epic/11-foundation` at `b9a19a3` reports
**35 JA4X failures** across nine vectors. The full measurement is on #12.

No issue owns JA4X today. Epic 1 covers JA4SSH, JA4_o and JA4L. Epic 2 covers JA4H,
JA4 and JA4S. JA4X falls between them, which is why the gap survived.

The old harness hid these failures twice. It compared one expected value against a
flat list of every fingerprint in the capture, and `KNOWN_DEVIATIONS` skipped the
whole method with this reason:

> JA4X certificate chain extraction depends on full TLS dissection which is not
> always available from pcapng captures

**That claim is now untested.** The vectors are committed and the suite reads them
offline, so the reference values exist for every one of these streams. FoxIO produced
a JA4X value from the same capture. Either this project can too, or the reason it
cannot is a specific, nameable difference.

## What to build

Diagnose first, then decide. Do not change `ja4plus/fingerprinters/ja4x.py` without a
vector that proves the change is correct.

1. Take one failing stream and read the FoxIO expected value beside the produced one.
2. Establish which of these it is: this project produces no value where the reference
   holds one, this project produces a different value, or this project produces an
   extra value the reference does not hold. 35 failures across the three classes are
   unlikely to share one cause.
3. Fix what is a defect. Where the FoxIO material is ambiguous, record the reading in
   the ambiguity register that #31 builds, and register the case as a deviation
   naming this issue.

Affected vectors, from the measurement: `tls-handshake.pcapng`, `ssh2.pcapng`,
`latest.pcapng`, `socks-https-example.pcap`, `browsers-x509.pcapng`,
`http2-with-cookies.pcapng`, `chrome-cloudflare-quic-with-secrets.pcapng`,
`https-connect.pcap`.

## Acceptance criteria

- [ ] Every JA4X failure is classified as a defect, an ambiguity, or a limitation of
      the capture, and the classification names the evidence.
- [ ] Every JA4X case that is a defect passes.
- [ ] Every JA4X case that is not a defect has a register entry naming this issue and
      a one-line cause.
- [ ] The `KNOWN_DEVIATIONS` claim about TLS dissection is replaced by a statement
      that the measurement supports, or removed.

## Links

- Feature: [`docs/specs/features/01-spec-conformance.md`](docs/specs/features/01-spec-conformance.md)
[...]

Issue #35, opened 2026-08-06
Build both JA4H cookie hashes from one cookie list


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/02-correctness-audit.md`](docs/specs/features/02-correctness-audit.md):

> FR-correctness-audit-6 — JA4H builds both cookie hashes from the same list of cookies.

> FR-correctness-audit-7 — JA4H keeps every occurrence of a repeated cookie name.

## What to build

Two defects in `ja4plus/fingerprinters/ja4h.py`:

- Line 163 writes `cookies[k] = v`, which drops a repeated cookie name, while `cookie_fields` at line 164 keeps it. Part c is built from the list and part d from the dictionary, so the two hashes then describe different cookie sets. A repeated cookie name is legal in HTTP. Build both from one ordered list of pairs.
- Line 131 requires `HTTP/\d+\.\d+` in the request line, so a request line reading `HTTP/2` never matches, although `_http_version_to_str` at line 34 handles `2` and returns `20`. Accept an optional minor version.

`http2-with-cookies.pcapng` and `http1-with-cookies.pcapng` are the vectors to check against.

## Acceptance criteria

- [ ] An HTTP request with a repeated cookie name produces a cookie-name hash and a cookie-value hash that both describe two cookies.
- [ ] An HTTP request line that reads `GET / HTTP/2` produces a fingerprint whose version code is `20`.

## Links

- Feature: [`docs/specs/features/02-correctness-audit.md`](docs/specs/features/02-correctness-audit.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #13

Crank-Git:
## Plan

Batch 13, on `epic/13-batch13`, off `dev` at `845065c`. This issue is first of the JA4H
chain. #170 follows it, and #170 also writes `ja4plus/fingerprinters/ja4h.py`.

**The line numbers in the issue body are stale. The defects are not.** The project manager
checked both against the current file before scheduling.

- The cookie dictionary write is at **line 187**, not 163. `cookies[k] = v` drops a
  repeated cookie name while `cookie_fields.append(k)` at line 188 keeps it. Part c reads
  the list and part d reads the dictionary, so the two hashes describe different cookie
[...]

Issue #30, opened 2026-08-06
Read the JA4L propagation factor from the FoxIO hop-count table


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/01-spec-conformance.md`](docs/specs/features/01-spec-conformance.md):

> FR-spec-conformance-9 — `calculate_distance` reads the propagation factor from the FoxIO hop-count table.

> FR-spec-conformance-10 — `calculate_distance` accepts an explicit propagation factor that overrides the table.

## What to build

`ja4plus/fingerprinters/ja4l.py:101` hardcodes `propagation_factor=1.6`. The FoxIO JA4L material gives a table keyed on hop count: `<=21` maps to `1.5`, `22` to `1.6`, `23` to `1.7`, `24` to `1.8`, `25` to `1.9`, and `>=26` to `2.0`.

The same material gives the speed of light in fiber as `0.128 miles or 0.206 km per µs`, which matches the constants already in the file.

Verified against https://github.com/FoxIO-LLC/ja4/tree/main/technical_details (retrieved 2026-08-06).

## Acceptance criteria

- [ ] `calculate_distance` returns a different result for a 20-hop path and a 30-hop path with the same latency.
- [ ] `calculate_distance(latency, propagation_factor=1.6)` uses 1.6.
- [ ] A TTL that implies a hop count above 26 uses a factor of 2.0.
- [ ] A TTL that implies a negative hop count clamps to zero hops and uses 1.5.

## Links

- Feature: [`docs/specs/features/01-spec-conformance.md`](docs/specs/features/01-spec-conformance.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #12
Depends on #27

Crank-Git:
## Plan

The base is `epic/12-spec-conformance`, Epic 1 batch 3, off `dev` at `9ef9ba7`.

**#88 rewrote this file.** `ja4plus/fingerprinters/ja4l.py` now halves every measured
round-trip time, and `JA4L-C` measures to the client's first data packet after the
SYN-ACK. Read the file as it stands before you plan the change.

**No fingerprint value reads `calculate_distance`.** That is why this issue was held out
of batch 2 and why it changes no register entry. If your change moves a JA4L value,
[...]

PR #79, opened 2026-08-06
Run the conformance suite from committed vectors on every pull request

Crank-Git:
Closes #23.

## What this changes

**A deviation register replaces two skip tables.** `KNOWN_DEVIATIONS` and
`STREAM_DEVIATIONS` emitted `pytest.skip` for a whole method. They hid 226 of the 263
conformance failures behind a green summary line, and they reported nothing when a
method started working. Both are deleted.

`tests/foxio_deviations.json` holds 263 entries, one for each conformance case that
this project fails today. Every entry names the issue that fixes it and states one line
of cause. The suite reads the register at collection and marks the matching case
`pytest.mark.xfail(strict=True)`.

- A registered case that fails reports as `xfailed`.
- A registered case that **passes** fails the suite and names itself.
- A failing case the register does not hold fails normally. That is the regression gate.

**The register is measured, not written by hand.**
`tests/generate_foxio_deviations.py` runs the suite with `JA4PLUS_IGNORE_DEVIATIONS=1`,
reads the parameters of each failed case from the pytest item, and writes one entry for
each. It reads the item, not the report text, so a vector name that holds a hyphen never
confuses it.

| Method | Entries | Issue named |
|---|---|---|
| JA4L-S | 82 | #30 |
| JA4L-C | 80 | #30 |
| JA4X | 35 | #78 |
| JA4SSH | 24 | #28 |
| JA4H | 21 | #35 |
| JA4 | 19 | #13 |
| JA4S | 2 | #13 |
| **Total** | **263** | |

The totals reproduce the baseline measurement on #12.

**The conformance job runs in continuous integration.** `FR-foundation-4`. The new
`conformance` job runs `pytest tests/ -m spec_validation` on every pull request into
`dev` and into `master`. It pins `actions/checkout` to
`11d5960a326750d5838078e36cf38b85af677262` and `actions/setup-python` to
`a26af69be951a213d495a4c3e4e4022e16d87065`, the identifiers the `test` and `lint` jobs
already hold. The vectors are committed, so the job reads no network and never runs
`tests/download_test_vectors.py`.

## Two readings this pull request settles

**The register file is JSON, not TOML.** The `test` job runs on Python 3.9, which holds
no `tomllib`. A TOML register needs a new dependency for one file. The issue allows
either form.

**The key holds the source port.** The issue gives the form
`<vector>/<stream>/<method>.<occurrence>`. That form is not unique:
`chrome-cloudflare-quic-with-secrets.pcapng` gives the stream index 0 to two streams,
which differ only by source port, and two `JA4L` cases collided. The key is now
`<vector>/<stream>:<port>/<method>.<occurrence>`, which matches the test identifier the
suite already prints. `test_the_register_key_of_every_case_is_unique` holds the property.

## Evidence

```
pytest tests/ -m spec_validation
424 passed, 119 skipped, 601 deselected, 263 xfailed  (offline)

pytest tests/ -m "not spec_validation"
589 passed, 12 skipped, 86 subtests passed
(572 before; the 17 new cases test the register and its reader)

coverage report --format=total -> 70   (floor 70, unchanged)
ruff check ja4plus/ tests/           -> All checks passed!
ruff format --check ja4plus/ tests/  -> 71 files already formatted
mypy ja4plus/                        -> Success: no issues found in 25 source files
```

**A deleted vector fails the suite.** Moved `tests/foxio_vectors/ssh2.pcapng` aside:

```
FAILED test_every_register_entry_matches_a_collected_case
1 failed, 402 passed, 233 xfailed
```

**Zero vectors fails the suite.** Moved `tests/foxio_vectors/` aside:

```
FAILED test_the_suite_collects_at_least_one_vector
FAILED test_every_register_entry_matches_a_collected_case
2 failed, 11 passed
```

**An unregistered failing case fails.** Removed
`tls-handshake.pcapng/38:50159/JA4.2` from the register:

```
FAILED test_the_produced_fingerprint_equals_the_reference[tls-handshake.pcapng-stream38:50159-JA4.2]
1 failed, 424 passed, 262 xfailed
```

**A registered case that passes fails and names itself.** Registered the passing case
`tls12.pcap/0:36372/JA4.1`:

```
[XPASS(strict)] issue #13: A case that passes, added to prove the strict marker.
FAILED test_the_produced_fingerprint_equals_the_reference[tls12.pcap-stream0:36372-JA4.1]
```

Each of the four working-copy edits above was reverted. The committed tree is the green
one.

## Documentation

The module docstring of `tests/foxio_deviations.py` states the two key forms, how to add
an entry, how to remove one, and how to measure a new baseline. The docstring of
`tests/test_spec_validation.py` points at it.

## Note for #78 and #30

The register is the record of what each issue owns. When a fix lands, the strict marker
turns the pass into a failure that names the case. Delete the entry in the same change
that lands the fix.

Verified against: https://docs.pytest.org/en/stable/how-to/skipping.html (pytest 8, strict xfail)


Crank-Git:
## Both findings are fixed on `3ad0fdb`

### 1. A deleted vector now fails the suite and names the file

`tests/foxio_vector_manifest.json` names all 37 vectors and the number of cases each
one carries. `test_the_collected_vector_set_equals_the_manifest` compares the manifest
against what the suite collected.

Your exact probe:

```
$ mv tests/foxio_vectors/tls12.pcap /tmp/
$ pytest tests/ -m spec_validation -q
E   AssertionError: the vector set differs from tests/foxio_vector_manifest.json:
E     tls12.pcap is absent from tests/foxio_vectors, and it carries 6 case(s)
FAILED tests/test_spec_validation.py::test_the_collected_vector_set_equals_the_manifest
1 failed, 421 passed, 115 skipped, 263 xfailed
```

The check reports the other direction too, so a vector cannot arrive unnoticed either:

```
$ cp tests/foxio_vectors/tls12.pcap tests/foxio_vectors/zz-new.pcap   # plus its .json
E     zz-new.pcap is not in the manifest
FAILED tests/test_spec_validation.py::test_the_collected_vector_set_equals_the_manifest
```

A vector is now a two-file change, which is the cost you asked for.
`tests/foxio_manifest.py` documents how to add one.

### 2. The JA4L attribution was wrong, and so was the correction

You are right that #30 alone does not fix 162 failures. **#30 fixes none of them, and
neither does #34 on its own.**

A JA4L fingerprint is `<latency_us>_<ttl>`. #30 reads the propagation factor into
`calculate_distance`, and no fingerprint value reads that function. #30 cannot change
any JA4L conformance case.

The real cause is in the harness. `tests/conformance_index.py:29` sets

```python
FINGERPRINTER_METHODS = ("JA4", "JA4S", "JA4H", "JA4X", "JA4SSH")
```

and `index_produced` builds those five fingerprinters. **JA4L is not one of them**, so
every JA4L case fails as `produced=<none>` before any value is compared. The comment
above that line says "JA4L is absent because ja4plus produces no JA4L value". That
claim is false, and it is the same stale claim #12 flagged in the old deviation table.
Measured on the committed vectors:

| Vector | JA4L values ja4plus produces |
|---|---|
| `ssh2.pcapng` | 610 |
| `v6.pcap` | 79 |
| `tls12.pcap` | 0 |

There is a second defect. A JA4L entry carries a connection key of the form
`udp_172.16.224.1:53_172.16.225.48:60983`. `_connection_identity` reads the form
`<address>:<port>-<address>:<port>` and partitions on the hyphen, so every JA4L value
would land under the unknown stream even once the fingerprinter runs.

**I opened #80, "Run the JA4L fingerprinter in the conformance harness".** All 162 JA4L
entries name it, and each cause line reads
`The conformance harness does not run the JA4L fingerprinter.` #80 must land before
this project can say anything true about JA4L conformance, and #34 is the expected
owner of the JA4L-C multiplicity defect once it can be measured — `ssh2.pcapng`
produces 610 values where the reference holds a few.

I did not wire JA4L into the harness in this pull request. It changes the measured case
set and the register that #12 published, which is its own reviewable change, the same
way #78 is for JA4X. Say the word if you want it here instead.

Register attribution now:

| Method | Entries | Issue |
|---|---|---|
| JA4L-S | 82 | #80 |
| JA4L-C | 80 | #80 |
| JA4X | 35 | #78 |
| JA4SSH | 24 | #28 |
| JA4H | 21 | #35 |
| JA4 | 19 | #13 |
| JA4S | 2 | #13 |

### Verification on `3ad0fdb`

```
pytest tests/ -m spec_validation      425 passed, 119 skipped, 263 xfailed, 0 failed
pytest tests/ -m "not spec_validation"  601 passed, 12 skipped, 86 subtests passed
coverage report --format=total        70   (floor 70, unchanged)
ruff check / ruff format --check      clean, 73 files
mypy ja4plus/                         Success: no issues found in 25 source files
```

The three probes you ran still hold:

```
removed CVE-2018-6794.pcap/JA4H from the register
  FAILED ...test_the_produced_occurrence_keys_equal_the_reference[CVE-2018-6794.pcap-JA4H]

registered the passing case tls12.pcap/0:36372/JA4.1
  [XPASS(strict)] issue #13: A case that passes, added to prove the strict marker.
  FAILED ...test_the_produced_fingerprint_equals_the_reference[tls12.pcap-stream0:36372-JA4.1]

moved tests/foxio_vectors/ aside
  FAILED test_the_suite_collects_at_least_one_vector
  FAILED test_the_collected_vector_set_equals_the_manifest
  FAILED test_every_register_entry_matches_a_collected_case
```

Every working-copy edit above was reverted. The pushed tree is the green one.

Issue #94, opened 2026-08-07
The conformance suite gives different results on Linux and macOS


Crank-Git:
## Context

Pull request #93, the Epic 1 batch, is **green on the project manager's machine and red
in continuous integration**. The `conformance` job reports:

```
17 failed, 437 passed, 145 skipped, 636 deselected, 240 xfailed
```

The project manager's run on the same commit reports:

```
0 failed, 454 passed, 145 skipped, 240 xfailed
```

**This is the first thing provider continuous integration has caught that a local run
did not.** It is also the reason #93 is not merged.

## The failures

All 17 are on an IPv6 stream, and every one says the fingerprinter produced nothing.

```
ipv6.pcapng      stream=0 2001:4998:ef83:14:8000::100d:64034 -> 2606:4700::6811:d209:443
                 JA4.1  expected='t12d4605h2_85626a9a5f7f_aaf95bb78ec9'  produced=<none>
                 JA4S.1 expected='t1204h2_cca9_1428ce7b4018'             produced=<none>
                 JA4X.1 expected='7d5dbb3783b4_ba7ce0880c07_7bf9a7bf7029' produced=<none>
                 JA4X.2 expected='7d5dbb3783b4_7d5dbb3783b4_41a019652939' produced=<none>
tls-alpn-h2.pcap same stream, same four values, same result
http-empty-useragent.pcap  JA4L-S.1 missing on stream=0 ::1:57722 -> ::1:9200
```

Plus the matching occurrence-key failures for JA4, JA4S, JA4X and JA4L-S on both
vectors.

## What is already ruled out

- **Not a scapy version difference.** Both environments resolve `scapy-2.7.0`.
- **Not a stale checkout.** Both runs are on the same commit.
- **Not the deviation register.** None of the 17 is registered, in either environment,
  and none is a strict `xpass`.

The project manager's environment is macOS on Python 3.14. The `conformance` job is
`ubuntu-latest` on Python 3.13.

## Why it matters more than 17 cases

**The deviation register is generated by running the suite.** If the suite produces a
different result on a different machine, the register records one machine's opinion.
Every conformance number this project has published today was measured on macOS. The
register may be understating the true deviation count, and no reviewer can tell.

A conformance suite whose result depends on the host is not a conformance suite.

## What to build

Diagnose first, and do not register the 17 to make the suite green. That would encode
the defect.

1. Reproduce on Linux. Establish whether the fingerprinters produce nothing, or whether
[...]

PR #93, opened 2026-08-07
Epic #12: Spec conformance, batch 1

Crank-Git:
Epic 1 batch 1. Three members. Two are breaking behaviour changes.

| Issue | What landed |
|---|---|
| Closes #29 | The `JA4_o` value on JA4 and JA4S |
| Closes #80 | The conformance harness runs the JA4L fingerprinter |
| Closes #28 | JA4SSH emits one fingerprint every 200 SSH packets |
| Closes #94 | A loopback capture that carries IPv6 reads the same way on every host |

## Verified on the integration branch, offline

```
ruff check / ruff format --check   clean
mypy ja4plus/                      Success: no issues found in 25 source files
unit                               624 passed, 12 skipped, 86 subtests passed
conformance                        454 passed, 145 skipped, 240 xfailed, 0 failed
coverage                           71%, floor 70
```

At the start of Epic 1 the suite reported 425 passed and 263 xfailed. **29 more cases
pass and 23 fewer are registered as deviations.** No strict `xpass`, so nothing was
fixed silently.

## Two breaking changes

**JA4SSH emits at 200 SSH packets, not at 10.** The window triggered at
`min(packet_count, 10)` above a comment reading `# For testing purposes`. FoxIO states
the interval verbatim: `(runs every 200 SSH packets by default)`. `ssh.pcapng` now
produces exactly one fingerprint, `c36s36_c76s124_c0s0`, equal to the reference. A
caller who relied on a fingerprint at 10 packets gets none.

**JA4L values are now measured.** #80 found that the harness never ran the JA4L
fingerprinter, so no JA4L value this project ever produced had been compared against
anything. The measurement is on #12.

## What the honest harness exposed

`ja4plus/fingerprinters/ja4l.py:218` claims `Per FoxIO spec: use raw time difference
(not divided by 2)`. The vectors contradict it: FoxIO halves the round-trip time, and
`JA4L-C` measures to the client's first data packet after the SYN-ACK. **Zero of 114
JA4L value cases pass.** #88 owns them.

Five issues exist only because the harness stopped hiding failures: #78 JA4X,
#88 JA4L values, #92 JA4SSH on long connections, #89 the specification correction, and
#80 itself.

A defect in `tests/generate_foxio_baseline.py` also made every value-form cause line in
the register false; it read pytest's `longrepr`, which holds the source of the test,
which holds the string `produced=<none>`. Fixed, with 11 unit tests.

## The register

```
240 entries, 0 without an issue number
[...]

Issue #12, opened 2026-08-06
Epic 1: Spec conformance


Crank-Git:
## Goal

Make every method match the FoxIO reference on every vector.

## Exit criteria

All 37 vectors pass. The JA4SSH window is 200 packets. `JA4_o` is exposed. The JA4L propagation factor follows the FoxIO table. The six surface-area tests in `tests/test_parity.py` are replaced by the vector suite.

## Spec

- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Feature: [`docs/specs/features/01-spec-conformance.md`](docs/specs/features/01-spec-conformance.md)

Depends on: #11

## Sub-issues

- [x] #27
- [x] #28
- [x] #29
- [x] #30
- [x] #31

Crank-Git:
## Conformance baseline, measured 2026-08-06

The vectors are committed (#22), so this epic now has a real starting number rather than an estimate.

```
pytest tests/ -m spec_validation -q
5 failed, 69 passed  in 3.53s, no network access
```

**69 of 74 vectors already conform.** Failures, with what each points at:

| Vector | Mismatches | Points at |
|---|---|---|
| `http2-with-cookies.pcapng` | 15 of 17 | JA4H. Matches the HTTP/2 request-line pattern and the duplicate-cookie-name defects registered in Epic 2 (#13). |
| `tls-sni.pcapng` | 5 of 64 | JA4 extension count. Expected `t13d1515h2_8daaf6152771_f37e75b10bcc`, produced `t13d1516h2_8daaf6152771_e5627efa2ab1`. The count differs by one and the extension hash differs, so one extension is counted that the reference excludes. |
| `chrome-cloudflare-quic-with-secrets.pcapng` | 1 of 3 | QUIC. |
| `tls-non-ascii-alpn.pcapng` | 1 of 2 | ALPN encoding, which is FoxIO PR #277 territory. |
| `http-empty-useragent.pcap` | 1 of 1 | JA4H edge case with an empty header value. |

### A defect in the harness, not in a fingerprinter

The suite compares one expected fingerprint against a flat list of every fingerprint found in the capture, rather than against the fingerprint for that stream. A failure line reads:

    stream=38 JA4: expected=t13d1515h2_... got=[80 values from every stream in the file]

So the mismatch counts above are not yet a per-method scorecard. **#27 must land first**, because it makes the suite compare one method on one stream at a time. Re-measure after #27 and treat that number as the true baseline.

[...]

PR #95, opened 2026-08-07
Read a loopback capture the same way on every host

Crank-Git:
## What

`ja4plus` binds the three BSD address family values of a loopback frame to the IPv6
layer. A loopback capture that carries IPv6 now produces the same fingerprints on
Linux and on macOS.

Refs #94. Part of #12.

## Why

Pull request #93 was green on macOS and red on `ubuntu-latest` with 17 failures, all
on an IPv6 stream.

The cause is the `DLT_NULL` link type, not the harness and not a fingerprinter. A
capture whose link type is `DLT_NULL` starts each frame with a four-byte address
family value, and libpcap writes the value of the host that captured the frame. The
value of `AF_INET6` is 24 on NetBSD and OpenBSD, 28 on FreeBSD, and 30 on Darwin.

scapy names all three values as IPv6 in `LOOPBACK_TYPES`, and then binds one value,
`socket.AF_INET6`. That value is 30 on Darwin and 10 on Linux. No capture holds 10,
so **no `DLT_NULL` IPv6 capture dissects on Linux at all**.

Three vectors hold `DLT_NULL` with the value 30: `ipv6.pcapng`, `tls-alpn-h2.pcap`
and `http-empty-useragent.pcap`. On Linux each frame stopped at `Loopback / Raw`, so
every fingerprinter received a packet with no IPv6 layer and no TCP layer, and
produced nothing.

## Evidence

The suite ran in `python:3.13-slim` on the merge base, and it reproduced the
continuous-integration result exactly:

```
17 failed, 437 passed, 145 skipped, 636 deselected, 240 xfailed
```

`index_produced("tests/foxio_vectors/ipv6.pcapng")` returned every method on macOS
and an empty map on Linux. The Linux dissection of the first frame:

```
packets: 9
0 Loopback / Raw
has IPv6: 0 has TCP: 0
JA4 fps: []
```

The first byte of that frame is `\x1e`, which is 30.

**macOS is the correct environment.** The reference values in
`tests/foxio_vectors/ipv6.pcapng.json` do not depend on a host, and macOS produces
them. Linux produced nothing, which is neither the reference value nor a deviation
from FoxIO. **No case earned a register entry**; `tests/foxio_deviations.json` is
unchanged and the registered count stays 240.

## How tested

Test-first. `tests/test_loopback_link_type.py` holds two tests:

- `test_dissects_every_bsd_address_family_value_as_ipv6` builds a loopback frame for
  each of 24, 28 and 30, and asserts the IPv6 layer. Before the fix it failed for 24
  and 28 on macOS, and for all three on Linux. This test reproduces the platform
  difference on any host.
- `test_the_ipv6_vector_produces_the_reference_ja4_fingerprint` reads `ipv6.pcapng`
  and asserts the JA4 value the reference names. Before the fix it failed on Linux
  and passed on macOS.

Full local suite, macOS on Python 3.14:

```
pytest tests/ -m spec_validation     454 passed, 145 skipped, 640 deselected, 240 xfailed
pytest tests/ -m "not spec_validation"  628 passed, 12 skipped, 86 subtests passed
coverage                             71% (floor 70)
ruff check                           All checks passed!
ruff format --check                  78 files already formatted
mypy ja4plus/                        Success: no issues found in 26 source files
```

Full local suite, Linux `python:3.13-slim` on the same commit:

```
pytest tests/ -m spec_validation     454 passed, 145 skipped, 640 deselected, 240 xfailed
pytest tests/ -m "not spec_validation"  628 passed, 12 skipped, 86 subtests passed
```

The two hosts report the same counts.

## Notes

`bind_bottom_up` reaches the dissection path alone, so a caller that builds a
loopback frame keeps the scapy default. This project writes no packet.

Verified against: https://github.com/the-tcpdump-group/libpcap/blob/f98637ad7f086a34c4027339c9639ae1ef842df3/gencode.c#L3333-L3354 (libpcap, retrieved 2026-08-06)
Verified against: `scapy/layers/inet6.py:4226-4228` and `scapy/layers/l2.py:720-724` (scapy 2.7.0)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,18 @@ project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Fixed
+
+- **A loopback capture that carries IPv6 now reads the same way on every host**
+  (#94). A capture whose link type is `DLT_NULL` starts each frame with the
+  address family value of the host that captured it. That value is 24 on NetBSD
+  and OpenBSD, 28 on FreeBSD, and 30 on Darwin. scapy binds one value,
+  `socket.AF_INET6`, which is 10 on Linux, and no capture holds 10. `ipv6.pcapng`,
+  `tls-alpn-h2.pcap` and `http-empty-useragent.pcap` therefore produced
+  fingerprints on macOS and none on Linux. `ja4plus` binds all three BSD values
+  when a caller imports it, and the conformance suite now reports the same counts
+  on both hosts.
+
 ### Changed
 
 - **BREAKING — JA4SSH emits one fingerprint for every 200 SSH packets** (#28).
--- a/docs/specs/features/01-spec-conformance.md
+++ b/docs/specs/features/01-spec-conformance.md
@@ -3,7 +3,7 @@ id: spec-conformance
 feature: Spec conformance
 epic: "Epic 1: Spec conformance"
 status: issued
-issues: [12, 27, 28, 29, 30, 31, 78, 80]
+issues: [12, 27, 28, 29, 30, 31, 78, 80, 94]
 mockups: []
 ---
 
@@ -60,6 +60,12 @@ ambiguous part of the specification.
 FR-spec-conformance-12 — `tests/test_parity.py` no longer asserts that a name
 exists.
 
+FR-spec-conformance-13 — The conformance suite reports the same counts on every
+host.
+
+FR-spec-conformance-14 — `ja4plus` reads the IPv6 layer of a loopback capture on
+every host.
+
 ## User flows
 
 **A maintainer fixes a failing method.**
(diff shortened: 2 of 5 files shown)


Crank-Git:
## Self-review

I reviewed the diff for correctness, for the scapy interface, and against the repository rules. Findings:

- `bind_bottom_up(Loopback, ...)` also covers `LoopbackOpenBSD`, the class scapy registers for `DLT_LOOP`. On a host that is not OpenBSD, `LoopbackOpenBSD` subclasses `Loopback` and defines no `payload_guess` of its own. On OpenBSD the two names are one class. `scapy/layers/l2.py:744-749` and `scapy/packet.py:2051-2070`.
- `bind_bottom_up` changes the dissection path alone. `bind_layers` calls `bind_top_down` as well, and `bind_top_down` changes the build path. `scapy/packet.py:2095-2111`. This project writes no packet, so the narrower call is the right one.
- The values 24, 28 and 30 collide with no other entry of `LOOPBACK_TYPES`. The table holds 2 for IPv4, 7 for OSI, 16 for Appletalk and 23 for Netware. `scapy/layers/l2.py:722-726`.
- The `_bound` flag is a check and then a set, and the two are not one operation. The call site is a module-level statement, and the Python import lock serialises it. A second bind appends one redundant entry and changes no result. I leave it.
- No fingerprinter changed. The one fingerprint the diff changes on Linux is the value `tests/foxio_vectors/ipv6.pcapng.json` already names, and the new test asserts it.

One point a reader needs: `test_the_ipv6_vector_produces_the_reference_ja4_fingerprint` passes on Darwin with the fix reverted, because the vector holds the Darwin value 30. It guards Linux. The parametrized test carries the weight on every host, because it covers 24 and 28 as well.

libpcap reads three values from a savefile: 24, 28 and 30. It never writes 10 into a savefile, so the fix leaves the Linux `socket.AF_INET6` binding of 10 in place and adds nothing for it.

PR #99, opened 2026-08-07
Read every handshake message of a TLS record for JA4X

Crank-Git:
## What

JA4X failed 35 conformance cases. It now fails 9, and each of the 9 carries a
register entry that names a cause the measurement supports.

Two defects caused 26 of the 35.

1. **The scan read only the first handshake message of a record.** One TLS record
   carries more than one handshake message. `latest.pcapng` stream 9 holds one
   7136-byte record that carries a ServerHello and a Certificate message, and the
   scan stopped at the ServerHello. The scan now joins the payload of every
   complete handshake record that follows without a gap, then reads the handshake
   messages of the joined bytes.
2. **The processed certificate set was keyed by the certificate alone.** The
   second stream that carried a chain produced nothing.
   `socks-https-example.pcap` streams 2 and 4 exposed it: both hold the two values
   of stream 0, and `ja4plus` held none. The key now names the stream and the
   certificate.

## Why the key names the stream

`python/ja4x.py` of the FoxIO repository states "JA4X does not use any caching
from common.py", and it computes one JA4X value for each certificate of the
stream it reads.

Verified against: https://github.com/FoxIO-LLC/ja4/blob/main/python/ja4x.py
(FoxIO `main` at `27f0cbf`, retrieved 2026-08-06)

JA4X is published as `technical_details/JA4X.png`, so the expected-output files
decide. `docs/implementation_notes.md` records both readings with the vectors
that support them.

## The nine cases that remain

Seven state that the reference decrypts the handshake with the secrets the
capture carries. `http2-with-cookies.pcapng` and
`chrome-cloudflare-quic-with-secrets.pcapng` each hold a decryption secrets block
with a `SERVER_HANDSHAKE_TRAFFIC_SECRET` entry, and the certificate of both
reaches the wire encrypted. `ja4plus` decrypts nothing.

Two state that the reference reads no TLS on the proxy port. `tshark` reads the
tunnel of `socks-https-example.pcap` on port 1080 and holds a JA4X value, and it
reads no TLS on port 8080 of `https-connect.pcap` or on port 9901 of
`socks4-https.pcap`. `ja4plus` reads the record layer by content, so it holds a
value on all three.

**`socks4-https.pcap/JA4X` is a new register entry.** The scan now reads three
real certificates there — `CN=login.live.com`, `CN=VeriSign Class 3 Extended
Validation SSL SGC CA` and `CN=VeriSign Class 3 Public Primary Certification
Authority - G5` — and the reference holds no JA4X for that capture. Its
expected-output file holds only `JA4L-S` and `JA4L-C`, so the reference reads no
TLS on the stream at all. The register already holds `https-connect.pcap/JA4`,
`/JA4S` and `/JA4X` for the same reason. To suppress the three values would need
a port allow-list that FoxIO does not specify, and it would take
`socks-https-example.pcap` with it.

## Performance

The scan resumes where the last packet of the stream left it. The contiguous
bytes of a stream only grow, so a byte the scan already read never changes. A
scan that starts at zero on every packet costs the square of the stream length,
and `http2-with-cookies.pcapng` then takes 18 seconds. Measured on that capture:
0.29 s before this branch, 18.09 s without the resumption, 0.19 s now.

## How it is tested

- `tests/test_ja4x_record_scan.py` holds 12 new tests. Four failed on the first
  commit of the branch.
- The conformance suite reports 481 passed, 144 skipped, 214 xfailed, 0 failed.
  The register holds 214 entries where it held 240.

## Local checks

```
pytest tests/ -m "not spec_validation"   640 passed, 12 skipped, 86 subtests passed
pytest tests/ -m spec_validation         481 passed, 144 skipped, 214 xfailed
ruff format --check ja4plus/ tests/      79 files already formatted
mypy ja4plus/                            Success: no issues found in 26 source files
python -m build                          Successfully built ja4plus-0.6.0
line coverage                            72%, floor 70. It reads 71% on the base.
```

Refs #78


Crank-Git:
## Self-review

Three specialist passes over the diff: correctness, security, and the writing
standard. `c1ba76e` carries the fixes.

### Accepted — a defect the first commit introduced

**The resumed scan skipped bytes a late segment carried.** `get_stream` starts at
the lowest sequence number the stream holds, so a segment that arrives after a
later one lowers that number and moves every offset. The saved offset then named
another byte, and a Certificate message in the bytes that moved was never read.
`ja4plus/utils/tcp_stream.py` gains `base_seq`, the offset is stored beside it, and
the scan starts at zero when the two differ.
`test_the_scan_reads_a_certificate_that_a_late_segment_carries` fails without the
fix.

### Accepted — state the diff let grow

**Both tables now hold a maximum entry count that no wall clock gates.** The key of
`processed_certs` names the stream, so the table fills faster than before, and the
30-second wall-clock gate let it grow without a limit between two runs.
`.claude/rules/conformance.md` warns about the wall clock for the same reason: a
capture replays faster than real time. The eviction now runs on each new entry.
`processed_certs` is a dict, so the eviction takes the oldest entry and not an
arbitrary one. `scan_offsets` drops the streams the reassembler no longer holds,
which bounds it at the reassembler's 50.

### Accepted — the writing standard

Seven sentences ran past the 25-word limit, two comments personified a packet
("a length field the packet lies about"), two dropped an article, and the register
causes stated the effect before the condition. All rewritten.

### Rejected — out of scope, filed instead

**The segment list of one stream grows without a limit.**
`TCPStreamReassembler.add_segment` appends with no cap on the entry count, and
`get_stream` sorts the list on every packet. This is quadratic in the packet count
of one stream. It is not a defect this branch introduced: the `trim_stream` call
this branch removed passed a byte offset where the method expects an absolute
sequence number, so it removed nothing for a realistic initial sequence number. The
reassembler is shared with JA4H, so the fix belongs to its own issue. Filed as #103.

Issue #98, opened 2026-08-07
Count an SSH message, not a TCP segment, in a JA4SSH window


Crank-Git:
## Context

#92 measured the surviving JA4SSH deviations and found this defect beside them.

`ja4plus` counts one SSH packet for every TCP segment that carries a payload. The FoxIO
reference counts the packets `tshark` labels `ssh`, and `tshark` reassembles an SSH
message that spans two TCP segments. It labels the segment that completes the message,
and it labels the earlier segment `tcp`. The two implementations therefore disagree by
one packet for each SSH message that spans two segments.

`ssh-r.pcap` stream 1 shows it. The stream holds 12 payload packets, and packet 303 and
packet 304 carry one KEXINIT message of 1496 bytes across two segments.

```
idx 303 client len 1448 seq 4052927673 flags A
idx 304 client len   48 seq 4052929121 flags PA
```

The reference reports `c6s5` for that stream, and `ja4plus` reports `c7s5`.

`ssh-r.pcap` stream 2 shows the same difference. The reference reports 931 SSH packets
on the stream, and `ja4plus` reports 932.

Verified against: https://github.com/FoxIO-LLC/ja4/blob/main/python/ja4ssh.py
(retrieved 2026-08-06)

## Warning

The count decides where a window closes. `ssh-r.pcap` stream 2 matches the reference on
window 2, window 3 and window 4 today. A change to the count moves every boundary after
the first message that spans two segments, so those three windows can break.

Measure before and after on every SSH vector, and read `ssh.pcapng` first, because that
vector conforms today.

## What to build

1. Establish how many SSH messages span two TCP segments on each SSH vector.
2. Count one SSH packet for the segment that completes an SSH message.
3. Keep `ssh.pcapng` at `c36s36_c76s124_c0s0`.

`ja4plus/utils/tcp_stream.py` holds the reassembly the project already uses for JA4H and
JA4X.

Part of #12

Crank-Git:
Triaged. This issue is scheduled for Epic 1 batch 3, sequenced after #92.

Both change `ja4plus/fingerprinters/ja4ssh.py`, and the warning in the body is the
reason for the order: the packet count decides where a window closes, and #92 is
settling the window boundaries now. A change to the count before #92 lands moves the
boundaries under it.

The batch 3 members are #31, #89 and this issue.

Crank-Git:
## Plan

[...]

Issue #97, opened 2026-08-07
Emit the JA4SSH occurrence a bare ACK produces after a window closes


Crank-Git:
## Context

#92 measured the surviving JA4SSH deviations. Three of them come from one cause.

`python/ja4ssh.py` tests `(entry['count'] % ssh_sample_count) == 0` on every packet of
the connection. A bare ACK does not advance `count`, so the test stays true for every
bare ACK that arrives after a window closes and before the next SSH packet. Each of
those packets calls `to_ja4ssh` again, which writes the next occurrence key from a
window that holds no SSH packet.

The reference shows the result on two vectors.

- `ssh-r.pcap` stream 0 holds `JA4SSH.2` equal to `c64s64_c0s0_c0s1`.
- `ssh2.pcapng` stream 14 holds `JA4SSH.2` equal to `c36s36_c0s0_c2s0`.

Both values report zero SSH packets in each direction. A simulation of
`python/ja4ssh.py` reproduces both.

Verified against: https://github.com/FoxIO-LLC/ja4/blob/main/python/ja4ssh.py
(retrieved 2026-08-06)

## The decision

The value depends on the number of bare ACKs between the window boundary and the next
SSH packet, and the last one of those packets decides it. A JA4SSH value that reports
zero SSH packets describes no traffic.

This issue picks one reading.

1. `ja4plus` reproduces the occurrence. Three cases leave the register.
2. `ja4plus` emits one value for each full window only. Three cases keep an entry that
   names this issue.

Read this issue together with #96. Both ask whether `ja4plus` reproduces a defect of the
reference, and one answer should cover both.

## Register entries this issue owns

```
ssh-r.pcap/0:64980/JA4SSH.2
ssh2.pcapng/14:57377/JA4SSH.2
ssh2.pcapng/JA4SSH
```

Part of #12

Crank-Git:
## Decision: `ja4plus` does not reproduce this defect

The user decided on 2026-08-07.

`(entry['count'] % ssh_sample_count) == 0` stays true for every bare ACK after a window
boundary in the FoxIO reference, so each bare ACK writes the next occurrence key from a
window that holds zero SSH packets. `ssh-r.pcap` stream 0 `JA4SSH.2` reads
`c64s64_c0s0_c0s1`.

**A fingerprint that describes no traffic is worse than no fingerprint.** A consumer
cannot tell it apart from a real quiet window. `ja4plus` emits a JA4SSH fingerprint only
[...]

Issue #96, opened 2026-08-07
Decide how JA4SSH reads the FoxIO mode field


Crank-Git:
## Context

#92 measured the surviving JA4SSH deviations. Four of them come from one cause, and that
cause is a defect in the FoxIO reference implementation.

`python/ja4ssh.py` builds each window bucket with `dict(ja4sh_stats)`. That call copies
the dictionary and not the two lists inside it, so every window on every connection in
one program run shares one `client_payloads` list and one `server_payloads` list. The
mode field of a JA4SSH value therefore reads the packet lengths of every connection the
tool has seen, and not the lengths of the window.

Two vectors show the result.

- `ssh-scp-1050.pcap` stream 0 holds no client packet in window 3 and window 4. The
  reference still reports `c112`, which is the client mode of window 1.
- `ssh-r.pcap` stream 2 window 1 holds packets whose own mode is 76. The reference
  reports `c64s64`, which is the mode of `ssh-r.pcap` stream 0.

A simulation of `python/ja4ssh.py` reproduces every one of these values with the shared
lists, and reproduces none of them with one list per window.

Verified against: https://github.com/FoxIO-LLC/ja4/blob/main/python/ja4ssh.py
(retrieved 2026-08-06)

## The decision

A user compares a JA4SSH value against a value another tool produced. The mode field of
the reference is not a property of the connection. It changes when an unrelated
connection appears in the same capture, and it changes with the order the tool reads the
connections in.

Two readings are open, and this issue picks one.

1. `ja4plus` reproduces the shared list. Every value in the register below matches the
   reference. The mode field stops describing the connection.
2. `ja4plus` keeps one payload list for each window. The mode field describes the
   window. Four cases keep a register entry that names this issue.

Rule 1 of `CLAUDE.md` says a FoxIO vector decides. `.claude/rules/conformance.md` says
the expected-output file decides an ambiguous case. Neither text covers a case where the
reference is reproducible and wrong, so a person decides it.

## Register entries this issue owns

```
ssh-r.pcap/2:46396/JA4SSH.1
ssh-scp-1050.pcap/0:49237/JA4SSH.3
ssh-scp-1050.pcap/0:49237/JA4SSH.4
```

Part of #12

Crank-Git:
## Decision: `ja4plus` does not reproduce this defect

The user decided on 2026-08-07.

`dict(ja4sh_stats)` shallow-copies the window bucket in the FoxIO reference, so every
[...]

Issue #92, opened 2026-08-06
Emit every JA4SSH window on a long connection


Crank-Git:
## Context

#28 set the JA4SSH window to 200 SSH packets and made `ssh.pcapng` produce exactly one
fingerprint equal to the reference. It removed 12 of the 24 JA4SSH register entries.

**12 entries survive**, and they are a different defect. Every one is on a capture that
holds more than one window:

```
ssh-r.pcap/0:64980/JA4SSH.1        The produced JA4SSH value differs from the reference.
ssh-r.pcap/0:64980/JA4SSH.2        ja4plus produces no JA4SSH value on this stream.
ssh-r.pcap/1:46394/JA4SSH.1        ja4plus produces no JA4SSH value on this stream.
ssh-r.pcap/2:46396/JA4SSH.1        The produced JA4SSH value differs from the reference.
ssh-r.pcap/2:46396/JA4SSH.5        ja4plus produces no JA4SSH value on this stream.
ssh-r.pcap/JA4SSH                  ja4plus produces no JA4SSH fingerprint the reference holds.
ssh-scp-1050.pcap/0:49237/JA4SSH.1 The produced JA4SSH value differs from the reference.
ssh-scp-1050.pcap/0:49237/JA4SSH.3 The produced JA4SSH value differs from the reference.
ssh-scp-1050.pcap/0:49237/JA4SSH.4 The produced JA4SSH value differs from the reference.
ssh2.pcapng/14:57377/JA4SSH.1      The produced JA4SSH value differs from the reference.
ssh2.pcapng/14:57377/JA4SSH.2      ja4plus produces no JA4SSH value on this stream.
ssh2.pcapng/JA4SSH                 ja4plus produces no JA4SSH fingerprint the reference holds.
```

The pattern is consistent. A capture with one window now conforms. **A capture with
several windows does not.** The reference holds occurrences `.1` through `.5` on
`ssh-r.pcap` stream 2; this project produces fewer, and the ones it does produce differ.

#28 fixed *when* the first window closes. This issue owns *what happens after it*.

## What to build

Diagnose before changing the fingerprinter.

1. Take `ssh-r.pcap` stream 2, which the reference gives five occurrences. Establish
   how many this project produces and where the boundaries fall.
2. Establish whether the counters reset correctly at a window boundary. The two counts
   in a JA4SSH value sum to the window size, so a value whose counts do not sum to 200
   names its own defect.
[...]

PR #100, opened 2026-08-07
Emit every JA4SSH window on a long connection

Crank-Git:
## What

`ja4plus` reads a bare ACK the way FoxIO reads one, and it emits the window a
connection holds open when it closes. `ssh-r.pcap` now produces the occurrence keys the
reference holds on all three of its streams.

Three changes to `ja4plus/fingerprinters/ja4ssh.py`:

1. A bare ACK carries the ACK flag alone. FoxIO counts one only when the TCP flags equal
   `0x0010`. `ja4plus` read the ACK flag alone, so it counted a SYN+ACK, a FIN+ACK and a
   RST+ACK as bare ACKs.
2. The ACK that completes the TCP handshake is a bare ACK, and it arrives before the
   first SSH packet. `ja4plus` created its state table entry on the first SSH packet, so
   it dropped that ACK.
3. A connection that closes emits the window it holds open. An empty window emits
   nothing, so the second FIN packet of a close adds no value.

## Why

#92 asked what JA4SSH does after the first window closes. The 12 surviving JA4SSH
register entries had six distinct causes. This pull request fixes the three that belong
to `ja4plus`, and it records the three that belong to the reference.

### The rule for the window a connection holds open

FoxIO states it above `finalize_ja4ssh`: `If the SSH connection is not terminated or the
last sample is less than 200 the finalize function just cleans up and prints the last
JA4SSH hash`. That function runs on a packet that carries the FIN flag and the ACK flag.

The reference appears to contradict its own rule, and `tshark` settles it. `ssh-r.pcap`
stream 1 gets a trailing occurrence, while `gre-sample.pcap`, `sshv1.pcap` and `v6.pcap`
hold no JA4SSH value at all for a connection of exactly the same shape.

The measurement ran the FoxIO python implementation at the pinned commit, against
`tshark` 4.6.7, with a print added to `finalize_ja4ssh`.

```
$ python ja4.py --ja4ssh -J tests/foxio_vectors/ssh-r.pcap
FINALIZE called stream=1 truthy=True  in_cache=True protos='eth:ethertype:ip:tcp:ssh'
FINALIZE called stream=2 truthy=True  in_cache=True protos='eth:ethertype:ip:tcp:ssh'
FINALIZE called stream=0 truthy=False in_cache=True protos='eth:ethertype:ip:tcp:ssh'

$ python ja4.py --ja4ssh -J tests/foxio_vectors/gre-sample.pcap
FINALIZE called stream=0 truthy=False in_cache=True protos='eth:ethertype:ip:gre:ip:tcp:ssh'

$ python ja4.py --ja4ssh -J tests/foxio_vectors/sshv1.pcap
FINALIZE called stream=0 truthy=False in_cache=True protos='eth:ethertype:ipv6:tcp:ssh'
```

`finalize_ja4ssh` opens with `if stream:`, and the stream index `0` is false in Python.
Every entry above is in the cache and carries a protocol chain that ends with `:ssh`, so
the guard is the only thing that stops the trailing window. `gre-sample.pcap`,
`sshv1.pcap` and `v6.pcap` each hold their SSH connection at index 0.

The same run, with that guard read as `if stream is not None:` and nothing else changed:

```
gre-sample.pcap  stream 0  JA4SSH.1  c24s23_c4s4_c5s4     (the reference holds none)
sshv1.pcap       stream 0  JA4SSH.1  c20s12_c18s23_c10s1  (the reference holds none)
ssh-r.pcap       stream 0  JA4SSH.2  c76s76_c33s48_c41s2  (the reference holds c64s64_c0s0_c0s1)
ssh-r.pcap       stream 1  JA4SSH.1  c64s64_c6s5_c4s5     (unchanged)
ssh-r.pcap       stream 2  JA4SSH.5  c76s76_c66s65_c9s51  (unchanged)
```

The rule holds. The exceptions are one integer-zero defect.

### The three results this project declines to reproduce

Each describes the capture and not the connection, so a user cannot compare it against
another tool's output. The user decided each one.

| Issue | The reference | `ja4plus` |
|---|---|---|
| #96 | `dict(ja4sh_stats)` shares one payload list, so the mode field reads the packet lengths of the whole capture. | Reads the lengths of the window. |
| #97 | A bare ACK after a window boundary writes another occurrence from a window of zero SSH packets. | Emits a value only for a window that holds SSH packets. |
| #105 | `finalize_ja4ssh` guards with `if stream:`, so no trailing window reaches the connection at stream index 0. | Emits the trailing window for every connection that closes. |

### The register

240 entries on the base, 239 now. Four leave and three join.

**Removed, because the case passes:**

```
ssh-r.pcap/0:64980/JA4SSH.1        c64s64_c107s93_c74s10
ssh-scp-1050.pcap/0:49237/JA4SSH.1 c112s1460_c52s148_c41s4
ssh2.pcapng/14:57377/JA4SSH.1      c36s36_c76s124_c74s5
ssh-r.pcap/JA4SSH                  the occurrence keys now equal the reference
```

**Added, recording the #105 divergence:** `gre-sample.pcap/JA4SSH`, `sshv1.pcap/JA4SSH`,
`v6.pcap/JA4SSH`. Each states that `ja4plus` emits one value the reference does not
hold, on a connection the reference indexes as stream 0. No entry was added to make a
red case green.

The six entries that name #96 and #97 now state that the cause is a reference defect
this project declines to reproduce, and that the decision is made.

`ssh-r.pcap/2:46396/JA4SSH.5` names #98, which owns the one remaining `ja4plus` defect:
`tshark` labels only the TCP segment that completes an SSH message, and `ja4plus` counts
both. That issue is out of this batch.

## How tested

```
pytest tests/ -m "not spec_validation"   644 passed, 12 skipped, 86 subtests passed
pytest tests/ -m spec_validation         458 passed, 142 skipped, 239 xfailed, 0 failed
ruff format --check ja4plus/ tests/      79 files already formatted
mypy ja4plus/                            Success: no issues found in 26 source files
coverage                                 72%, up from 71% on the base
```

The base measured 628 unit tests, 454 conformance passed and 240 xfailed. No case that
passed before fails now. `ja4plus/fingerprinters/ja4ssh.py` coverage rose from 88% to
97%.

`tests/test_ja4ssh_windows.py` is new and holds 16 tests: five on the bare ACK, six on
the window a connection holds open, and five that read the FoxIO vectors.

Verified against: https://github.com/FoxIO-LLC/ja4/blob/main/python/ja4.py (retrieved 2026-08-06)
Verified against: https://github.com/FoxIO-LLC/ja4/blob/main/python/ja4ssh.py (retrieved 2026-08-06)
Verified against: https://github.com/FoxIO-LLC/ja4/tree/main/python/test/testdata (retrieved 2026-08-06)
Verified against: TShark (Wireshark) 4.6.7

Refs #92, #96, #97, #105

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -14,6 +14,7 @@
       "Bash(mkdocs build:*)",
       "Bash(mkdocs serve:*)",
       "Bash(python tests/download_test_vectors.py:*)",
+      "Bash(tshark:*)",
       "Bash(git:*)",
       "Bash(gh:*)"
     ]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,27 @@ project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   `ssh-scp-1050.pcap` and `ssh2.pcapng` now equal the reference on their first
   window. `ssh.pcapng` holds no bare ACK, and it stays at `c36s36_c76s124_c0s0`.
 
+- **JA4SSH emits the window a connection holds open when it closes** (#92). A
+  connection that carries fewer than 200 SSH packets produced no fingerprint at
+  all, and a connection that closed part way through a window lost those packets.
+  `ja4plus` now emits that window on a packet that carries the FIN flag and the ACK
+  flag, which is the rule `python/ja4.py` states above `finalize_ja4ssh`. An empty
+  window emits nothing. `ssh-r.pcap` now produces the occurrence keys the reference
+  holds on all three of its streams.
+
+### Divergence from the FoxIO reference
+
+- **JA4SSH declines three results of the reference** (#96, #97, #105). Each one
+  describes the capture and not the connection, so it cannot be compared against
+  the output of another tool. The reference reads its mode field from the packet
+  lengths of every connection in the capture, because `dict(ja4sh_stats)` shares
+  one payload list (#96). It writes an extra occurrence from a window of zero SSH
+  packets whenever a bare ACK follows a window boundary (#97). It emits no trailing
+  window for the connection it holds at stream index 0, because `finalize_ja4ssh`
+  guards with `if stream:` (#105). `ja4plus` reads the window alone, emits a value
+  only for a window that holds SSH packets, and emits the trailing window for every
+  connection that closes. `docs/implementation_notes.md` holds the measurement.
+
 - **A loopback capture that carries IPv6 now reads the same way on every host**
   (#94). A capture whose link type is `DLT_NULL` starts each frame with the
   address family value of the host that captured it. That value is 24 on NetBSD
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,15 @@ project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 
+- **BREAKING — JA4SSH counts a bare ACK the way FoxIO counts one** (#92). A bare
+  ACK carries the ACK flag alone and no payload. `ja4plus` read the ACK flag
+  alone, so it counted a SYN+ACK, a FIN+ACK and a RST+ACK as bare ACKs. It also
+  created its state table entry on the first SSH packet, so it dropped the ACK
+  that completes the TCP handshake. The two ACK counts of a JA4SSH fingerprint
+  therefore change on any connection that carries a bare ACK. `ssh-r.pcap`,
+  `ssh-scp-1050.pcap` and `ssh2.pcapng` now equal the reference on their first
+  window. `ssh.pcapng` holds no bare ACK, and it stays at `c36s36_c76s124_c0s0`.
+
 - **A loopback capture that carries IPv6 now reads the same way on every host**
   (#94). A capture whose link type is `DLT_NULL` starts each frame with the
   address family value of the host that captured it. That value is 24 on NetBSD
--- a/docs/implementation_notes.md
+++ b/docs/implementation_notes.md
@@ -100,12 +100,66 @@ Verified against: https://github.com/FoxIO-LLC/ja4/tree/main/technical_details
 it. A capture of 10 SSH packets produced a fingerprint that no reference
 implementation matches.
 
+### The bare ACK
+
+A bare ACK is one packet that carries the ACK flag alone and no payload. FoxIO
+counts one only when the TCP flags equal `0x0010`, so a SYN+ACK, a FIN+ACK and a
+RST+ACK never reach the bare-ACK counter.
+
+The ACK that completes the TCP handshake is a bare ACK, and it arrives before the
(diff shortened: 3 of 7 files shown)


Crank-Git:
## Self-review

Three points a reviewer should check first.

**The refactor beside the fix.** `process_packet` built the fingerprint inline, and
`_generate_ja4ssh` held the same construction and no caller. The change removes the
inline copy and gives `_generate_ja4ssh` its caller through a new `_close_window`. The
emitted string is unchanged: both forms read `_mode(values) if values else 0` for each
direction and write the same format string. The conformance suite reproduces every
value that passed on the base, which is the evidence.

**The wider state table.** The fingerprinter now creates an entry for a bare ACK on a
stream where one port is 22. `self.connections` carries no maximum entry count and no
maximum age, and it did not carry one before this change either.
`.claude/rules/conformance.md` requires both. The gap is pre-existing, and this change
makes the entry easier to create. A separate issue owns the bound.

**The rule that was removed.** The first attempt emitted the open window on a FIN+ACK
packet. It gave `ssh-r.pcap` stream 1 and stream 2 their trailing occurrence, and it
also gave `gre-sample.pcap`, `sshv1.pcap` and `v6.pcap` a value the reference does not
hold. `tests/test_ja4ssh_windows.py::test_a_connection_that_closes_emits_no_partial_window`
pins the current behaviour and records why.

Crank-Git:
The bound on `self.connections` already has an owner. #38 adds the bounded state table with count and age limits, and #39 moves every fingerprinter onto it. No new issue is needed, and this change files none for it.

Crank-Git:
## Self-review, second pass

A correctness review of the close rule found no defect. It checked six paths, and each
one is covered by a test.

- A packet that fills a window and carries FIN+ACK emits one fingerprint, not two.
- A FIN+ACK on an unknown connection creates no state and emits nothing, because a
  FIN+ACK is not a bare ACK and so opens no entry.
- Two FIN packets emit one fingerprint. The connection key is direction-free, so the
  second packet finds the window the first emptied.
- A FIN+ACK that carries a payload counts its SSH packet before the window closes,
  which is the order `python/ja4.py` runs `update_ssh_entry` and `finalize_ja4ssh` in.
- A RST packet leaves the window open, which matches the FoxIO rule.
- The change adds no path that raises on a malformed packet.

Two review notes are fixed in 451b31a. The close rule had no test for the flags it does
not read, so `test_a_reset_packet_leaves_the_window_open` now pins that. The
`_close_window` docstring named one of the two states that reach its empty-window
guard, and it now names both.

## The JA4SSH occurrence-key cases

```
ssh.pcapng           PASSED
ssh-r.pcap           PASSED   was a register entry, now removed
ssh-scp-1050.pcap    PASSED
ssh2.pcapng          XFAIL    #97, the occurrence a bare ACK writes
gre-sample.pcap      XFAIL    #105, the stream index 0
sshv1.pcap           XFAIL    #105
v6.pcap              XFAIL    #105
```

Every remaining case names a decided divergence, and no case fails.

Issue #102, opened 2026-08-07
Read the JA4L server point from the QUIC Initial that completes the ServerHello


Crank-Git:
## Context

#88 landed the JA4L value fix. Two `JA4L-S` cases still differ, and both are QUIC.

The reference reads the server measurement point from the Initial packet that
completes the TLS ServerHello. `python/ja4.py` in the FoxIO repository records the
point when `packet_type` is `0` and `tls.handshake.type` is `2`. A server that splits
the ServerHello across two Initial packets therefore gives the timestamp of the second
one.

`ja4plus/fingerprinters/ja4l.py` reads the first server Initial packet instead,
because `ja4plus/utils/quic_utils.py` cannot decrypt these server Initial packets. The
decryption raises on the AEAD tag, so the fingerprinter cannot see which Initial
carries the ServerHello.

The measurement:

- `chrome-cloudflare-quic-with-secrets.pcapng` stream 0, port 50280. The reference
  holds `10990_56`, and ja4plus produces `9285_56`. The server sends two Initial
  packets, at `+18570 us` and at `+21981 us`.
- `tls3.pcapng` stream 25, port 61884. The reference holds `3583_57`, and ja4plus
  produces `3051_57`.

`ja4plus/utils/quic_utils.py` holds `decrypt_quic_initial_crypto` and
`reassemble_crypto_fragments` for the client side. The server side holds
`parse_quic_server_initial`, which reads one packet only.

## What to build

1. Find why `parse_quic_server_initial` fails on these packets. Read RFC 9001
   Section 5 before you change the key derivation.
2. Add a server-side form of `decrypt_quic_initial_crypto`, so that a caller collects
   the CRYPTO fragments of several server Initial packets.
3. Record the JA4L server point on the packet that completes the ServerHello.

## Acceptance criteria

- [ ] `chrome-cloudflare-quic-with-secrets.pcapng` produces `JA4L-S=10990_56` on
      stream 0, port 50280.
- [ ] `tls3.pcapng` produces `JA4L-S=3583_57` on stream 25, port 61884.
- [ ] `tests/foxio_deviations.json` holds no entry for either case.
- [ ] No vector that passes today fails.

## Links

- Value defect: #88
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Crank-Git:
## A second QUIC gap, found while #88 was reviewed

[...]

Issue #101, opened 2026-08-07
Group a mirrored tunnel capture by its inner connection


Crank-Git:
## Context

#88 landed the JA4L value fix and #88 added the scapy tunnel dissectors, so
`tcpdump-geneve.pcap` now conforms. `gre-erspan-vxlan.pcap` still produces no JA4L
value, and the cause is the connection key, not the dissector.

The capture is an ERSPAN mirror. Every packet travels from `100.20.9.2` to
`100.20.9.1`, in both directions of the session the mirror carries. The inner session
is `10.16.27.12:65174` to `10.16.27.131:80`.

`ja4plus/fingerprinters/ja4l.py` builds the connection key from the outer address pair
and the inner port pair. The outer addresses do not change direction, so the SYN and
the SYN-ACK reach two different keys:

- SYN: `tcp_100.20.9.1:80_100.20.9.2:65174`
- SYN-ACK: `tcp_100.20.9.1:65174_100.20.9.2:80`

No key holds both measurement points, so the fingerprinter reports nothing.

The reference groups the packets by the inner session and reports the outer address.
`tests/foxio_vectors/gre-erspan-vxlan.pcap.json` holds `JA4L-S` `997_64` and `JA4L-C`
`953_64` on the stream `100.20.9.2:65174` to `100.20.9.1:80`.

## What to build

Group a connection by the innermost address pair and the innermost port pair. Report
the outer address pair and the outer TTL, because the reference reports those.

The connection key of JA4L is also the stream identity the conformance suite reads, so
the fingerprint entry needs the reported address pair as well as the grouping key.
`tests/conformance_index.py` documents the two key forms.

## Acceptance criteria

- [ ] `gre-erspan-vxlan.pcap` produces `JA4L-S=997_64` on the stream
      `100.20.9.2:65174` to `100.20.9.1:80`.
- [ ] `gre-erspan-vxlan.pcap` produces `JA4L-C=953_64` on the same stream.
- [ ] `tests/foxio_deviations.json` holds no entry for `gre-erspan-vxlan.pcap`.
- [ ] No vector that passes today fails.

## Links

- Value defect: #88
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Crank-Git:
## Plan

The base is `epic/12-spec-conformance`, Epic 1 batch 4, off `dev` at `4e335f3`.

[...]

Issue #89, opened 2026-08-06
Correct the JA4L measurement point in the specification


Crank-Git:
## Context

A FoxIO vector contradicts a functional requirement. #80 measured it and #34 records
the evidence.

`docs/specs/features/02-correctness-audit.md` states that the JA4L client fingerprint
comes from the first ACK after the SYN-ACK. #34 quotes that requirement as an
acceptance criterion.

`browsers-x509.pcapng` stream 0 shows otherwise:

- SYN-ACK `+0.003815s`, bare ACK `+0.003927s`, client Client Hello `+0.004371s`.
- The reference `JA4L-C` is `278_128`, and `(4371 - 3815) / 2 = 278`.
- Measured from the bare ACK the value is 56. This project produces 111.

The reference measures to the client's **first data packet** after the SYN-ACK, and it
halves the round-trip time. `ja4plus/fingerprinters/ja4l.py:218` also carries the
comment `Per FoxIO spec: use raw time difference (not divided by 2)`, which the same
vectors contradict.

`CLAUDE.md` rule 1 governs: the vector decides, so the specification is wrong.

## What to change

- Correct the JA4L requirement in `docs/specs/features/02-correctness-audit.md` to
  state the measurement point the vectors show, and to state that the round-trip time
  is halved.
- Move the measurement-point requirement to
  `docs/specs/features/01-spec-conformance.md` if that file is the better home, since
  #88 owns the value and #34 now owns multiplicity only.
- Add a dated line to the `## Changelog` in `docs/specs/spec.md`.
- Cite the FoxIO material for JA4L. Do not write the correction from the vector alone
  where the published material states the rule.

## Acceptance criteria

- [ ] No feature file states that the JA4L client fingerprint comes from the first ACK
      after the SYN-ACK.
- [ ] A feature file states the measurement point and the halving, and cites the FoxIO
      material or the vector that proves it.
- [ ] `docs/specs/spec.md` carries a dated changelog line.

## Links

- Evidence: #34, #80, #12
- Value defect: #88

Part of #13

Crank-Git:
## Plan

The base is `epic/12-spec-conformance`, Epic 1 batch 3, off `dev` at `9ef9ba7`.

**#88 already landed the code change this issue describes.** `ja4l.py` now halves every
measured round-trip time, and `JA4L-C` measures to the client's first data packet after
the SYN-ACK. The comment at `ja4l.py:218` is deleted. **This issue changes documents
[...]

Issue #88, opened 2026-08-06
Diagnose and fix the 114 JA4L conformance value failures


Crank-Git:
## Context

#80 made the conformance harness run the JA4L fingerprinter for the first time. The
measurement is on #80 and on #12. **Zero JA4L value cases pass. All 114 fail.**

No issue owned this. #30 changes `calculate_distance`, and no fingerprint value reads
it. #34 emits one JA4L client fingerprint per connection, which is a multiplicity
defect, not a value defect. This issue owns the value.

## The evidence

`ja4plus/fingerprinters/ja4l.py:218` carries the comment
`Per FoxIO spec: use raw time difference (not divided by 2)`. **The vectors say that
comment is false.**

`badcurveball.pcap` stream 0, read from the capture:

- SYN at `+0.000000s`, SYN-ACK at `+0.001563s`. The difference is 1563 us. The
  reference `JA4L-S` is `781_238`, and `1563 / 2 = 781`.
- Bare ACK at `+0.005918s`, client data packet at `+0.005925s`. The reference
  `JA4L-C` is `2181_64`, and `(5925 - 1563) / 2 = 2181`.

`browsers-x509.pcapng` stream 0 fixes the client measurement point:

- SYN-ACK `+0.003815s`, bare ACK `+0.003927s`, client Client Hello `+0.004371s`. The
  reference `JA4L-C` is `278_128`, and `(4371 - 3815) / 2 = 278`. Measured from the
  bare ACK the value is 56. This project produces 111.

So the reference computes two things this project does not.

1. **Both values are halved.** This project reports the raw round-trip time.
2. **`JA4L-C` measures to the client's first data packet after the SYN-ACK**, not to
   the bare ACK.

## What to build

The three groups this issue owns, from the #80 measurement:

- **58 cases** where the latency is exactly twice the reference. The round-trip time
  is not halved.
- **52 cases** where the `JA4L-C` measurement point differs, and where the QUIC path
  differs.
- **4 cases** where JA4L produces no value on a tunnelled capture
  (`gre-erspan-vxlan.pcap`, `tcpdump-geneve.pcap`), because the fingerprinter reads
  the outer address layer.

Read the FoxIO material for JA4L before changing the fingerprinter, and cite it. Where
the material is ambiguous, record the reading in the register that #31 builds rather
than guessing. The comment at `ja4l.py:218` is removed or corrected either way,
because it currently states the opposite of what the vectors show.

[...]

PR #104, opened 2026-08-07
Report the one-way JA4L latency and the reference measurement point

Crank-Git:
## What this changes

JA4L reported the whole round-trip time, and it measured the client value to the bare
ACK of the handshake. Both are wrong against the FoxIO vectors. This pull request
reports the one-way latency and the measurement points the reference uses.

- **The latency is halved.** `technical_details/JA4L.png` states
  `One-way TCP latency in us`. `badcurveball.pcap` stream 0 sends the SYN at
  `+0.000000s` and the SYN-ACK at `+0.001563s`, and the reference `JA4L-S` is
  `781_238`.
- **The client point is the last packet that carries the relative sequence number 1
  and the relative acknowledgement number 1.** `browsers-x509.pcapng` stream 0 gives
  the bare ACK at `+0.003927s` and the Client Hello at `+0.004371s`. The reference
  `JA4L-C` is `278_128`, and `(4371 - 3815) / 2 = 278`.
- **A complete HTTP request does not move the client point.** See the reading below.
- **One connection carries one `JA4L-C` value.** The reference overwrites the value
  while the point moves, so a later packet replaces the value rather than adding one.
- **The QUIC form reads the Initial packets and the Handshake packets on port 443.**
  It read the first UDP packet of any flow before, so a DNS flow produced a JA4L
  value that the reference does not hold.
- **`ja4plus/utils/tunnels.py` imports the scapy Geneve, VXLAN and ERSPAN
  dissectors.** scapy leaves them unbound and stops at the tunnel header, so
  `tcpdump-geneve.pcap` produced nothing.
- The comment at `ja4l.py:218`, `Per FoxIO spec: use raw time difference (not divided
  by 2)`, is gone. The vectors state the opposite.

This is a breaking change. Every JA4L value a caller stored before this release
differs from the value this release produces. `CHANGELOG.md` records it.

## The measurement

This branch is rebased onto `epic/12-spec-conformance` at `d26ad4f`, which already
carries #99 and #100.

| Measurement | Before | After |
|---|---|---|
| Register entries | 213 | 74 |
| Entries that name #88 | 114 | 0 |
| Entries that name #34 | 37 | 6 |
| JA4L conformance cases that fail | 151 | 12 |

108 of the 114 value cases this issue owns now conform. The six that remain carry a
register entry that names a real cause, and two new issues own them:

- #101 — `gre-erspan-vxlan.pcap` mirrors both directions of one session between one
  outer address pair, so `ja4plus` groups the two directions as two connections.
  2 value cases and 2 occurrence-key cases.
- #102 — the reference reads the server point from the QUIC Initial packet that
  completes the ServerHello. `ja4plus` cannot decrypt those server Initial packets,
  so it reads the first one. 2 value cases on
  `chrome-cloudflare-quic-with-secrets.pcapng` and `tls3.pcapng`.

27 of the 37 entries that named #34 also conform now, and this pull request removes
them. That overlap is unavoidable: the client value case passes only when one
connection carries one value. #34 keeps the six occurrence-key cases that remain, and
it keeps the removal of `_src_is_client`.

## Where the FoxIO material is ambiguous

FoxIO publishes JA4L as a PNG image, so `.claude/rules/conformance.md` makes the
expected-output files the authority. `docs/implementation_notes.md` now records every
reading. Three of them need a reader's attention.

**1. A complete HTTP request does not move the client point.** `python/ja4.py` in the
FoxIO repository selects its cache with `get_cache(x)`, which returns a separate cache
when the highest layer is `http` or `http2`. A packet that carries a whole HTTP
request therefore never reaches the JA4L timestamps. Two vectors prove both halves:
`latest.pcapng` stream 6 sends one complete `GET` request and the reference `JA4L-C`
is `32_128`, which is the bare ACK; `http-empty-useragent.pcap` sends the request
line, the header and the blank line in three packets, and the reference `JA4L-C` is
`177863_64`, which is the request line. `ja4plus` reads the request line and the blank
line that ends the header block. **Reading:** this is a property of the reference
program, not a stated rule, and the vectors are the only evidence for it.

**2. The three reference implementations disagree, and the vectors follow the Python
one.** `rust/ja4/src/time/tcp.rs` takes the first ACK after the SYN-ACK.
`wireshark/source/packet-ja4.c:1302` takes the first bare ACK with `tcp.len == 0`.
`python/ja4.py:571` takes the last packet with the relative sequence number 1 and the
relative acknowledgement number 1. The expected-output files under
`python/test/testdata/` come from the Python program, and
`.claude/rules/external-apis.md` names those files as the authority. **Reading:**
follow the Python program.

**3. A handshake of one second or more.** `common.py:179` reads the difference as a
`timedelta` and takes its `microseconds` attribute, which holds only the part below
one second. A handshake of 1.2 seconds therefore gives 100000 rather than 600000. No
FoxIO vector holds a difference of one second or more, so no vector requires that
behaviour and `ja4plus` divides the whole difference. **Reading:** treat the reference
behaviour as a defect until a vector proves otherwise.

`docs/implementation_notes.md` also records the reading that the reference reads the
address and the TTL of the outer layer of a tunnelled capture, and the port of the
inner layer. `gre-sample.pcap` proves it.

## How this was tested

`tests/test_ja4l_foxio.py` is new. Each test names one vector, one stream and the
value the FoxIO expected-output file holds. The tests were written first and they
failed against the unchanged fingerprinter.

Local suite, in the worktree:

```
pytest tests/ -m "not spec_validation"   674 passed, 11 skipped, 86 subtests passed
pytest tests/ -m spec_validation         624 passed, 141 skipped, 74 xfailed
ruff format --check ja4plus/ tests/      83 files already formatted
coverage                                 74%, and the floor is 70. It was 73 before.
```

No conformance case outside JA4L changed. The measurement with
`JA4PLUS_IGNORE_DEVIATIONS=1` reports the same non-JA4L failure list before and
after, and the same list again after the three self-review fixes.

`tests/foxio_deviations.json` comes from `tests/generate_foxio_baseline.py`, not from
a hand edit. No entry was added to make the suite green. The JA4L slice of that
measurement is merged into the register the integration branch already carries,
because the generator still reads #92 for every JA4SSH case while #99 and #100 point
those cases at #96, #97, #98 and #105. A whole regeneration here would name the wrong
issue on eleven entries this issue does not own. The project manager still regenerates
once on the integration branch.

Verified against: https://github.com/FoxIO-LLC/ja4/tree/main/technical_details (JA4L.png, retrieved 2026-08-06)
Verified against: https://github.com/FoxIO-LLC/ja4/blob/27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8/python/ja4.py (the program that writes the expected-output files, retrieved 2026-08-06)
Verified against: https://github.com/FoxIO-LLC/ja4/blob/27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8/python/common.py (retrieved 2026-08-06)
Verified against: https://github.com/FoxIO-LLC/ja4/blob/main/rust/ja4/src/time/tcp.rs (retrieved 2026-08-06)
Verified against: https://github.com/FoxIO-LLC/ja4/blob/main/wireshark/source/packet-ja4.c (retrieved 2026-08-06)

## Out of scope

- Multiplicity, which #34 owns. This pull request changes the `JA4L-C` count only
  because the value case cannot pass otherwise.
- `calculate_distance`, which #30 owns. The function is untouched.
- The specification wording, which #89 owns.

Part of #12. Refs #88.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,12 +14,12 @@ project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   FoxIO material states `One-way TCP latency in us`, and the vectors hold half of
   the measured time. The client value now measures to the last packet that
   carries the relative sequence number 1 and the relative acknowledgement number
-  1, and one connection now carries one `JA4L-C` value. The QUIC form now reads
-  the Initial packets and the Handshake packets on port 443, and a UDP flow that
-  carries no QUIC long header produces no value. `ja4plus` also imports the scapy
+  1. One connection now carries one `JA4L-C` value. The QUIC form now reads the
+  Initial packets and the Handshake packets on port 443. A UDP flow that carries
+  no QUIC long header gives no value. `ja4plus` also imports the scapy
   Geneve, VXLAN and ERSPAN dissectors, so a mirrored capture reads its inner TCP
   layer. Every JA4L value a caller stored before this release differs from the
-  value this release produces. 108 of the 114 registered JA4L value deviations
+  value this release emits. 108 of the 114 registered JA4L value deviations
   now conform, and #101 and #102 own the six that remain.
 
 - **BREAKING — JA4SSH counts a bare ACK the way FoxIO counts one** (#92). A bare
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,20 @@ project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 
+- **BREAKING — JA4L reports the one-way latency and the measurement point the
+  FoxIO vectors hold** (#88). The fingerprinter reported the whole round-trip
+  time, and it measured the client value to the bare ACK of the handshake. The
+  FoxIO material states `One-way TCP latency in us`, and the vectors hold half of
+  the measured time. The client value now measures to the last packet that
+  carries the relative sequence number 1 and the relative acknowledgement number
+  1, and one connection now carries one `JA4L-C` value. The QUIC form now reads
+  the Initial packets and the Handshake packets on port 443, and a UDP flow that
+  carries no QUIC long header produces no value. `ja4plus` also imports the scapy
+  Geneve, VXLAN and ERSPAN dissectors, so a mirrored capture reads its inner TCP
+  layer. Every JA4L value a caller stored before this release differs from the
+  value this release produces. 108 of the 114 registered JA4L value deviations
+  now conform, and #101 and #102 own the six that remain.
+
 - **BREAKING — JA4SSH counts a bare ACK the way FoxIO counts one** (#92). A bare
   ACK carries the ACK flag alone and no payload. `ja4plus` read the ACK flag
   alone, so it counted a SYN+ACK, a FIN+ACK and a RST+ACK as bare ACKs. It also
--- a/docs/implementation_notes.md
+++ b/docs/implementation_notes.md
@@ -70,10 +70,78 @@ JA4L fingerprints include a direction prefix:
 `JA4L-S={latency_us}_{ttl}` and `JA4L-C={latency_us}_{ttl}`.
 The spec describes `{latency_microseconds}_{ttl}` without a prefix.
 
-### Latency is raw time difference, not RTT/2
+### The latency is half the measured time
 
-The latency value is the raw time difference between handshake points,
-not the round-trip time divided by 2.
+`technical_details/JA4L.png` states `One-way TCP latency in us`, and every FoxIO
+vector holds half the time the capture shows. `badcurveball.pcap` stream 0 sends
+the SYN at `+0.000000s` and the SYN-ACK at `+0.001563s`, and the reference
+`JA4L-S` is `781_238`.
+
+The division truncates toward zero, and it produces `0` for a difference of one
+microsecond.
+
+### The client measurement point
+
+FoxIO publishes JA4L as an image, so the expected-output files decide the
+measurement point. `python/ja4.py` in the FoxIO repository records the client
+point on every TCP packet that carries the relative sequence number `1` and the
+relative acknowledgement number `1`, and it keeps the last one. That is the bare
+ACK of the handshake first, and then the first packet of the application
+handshake.
+
+`browsers-x509.pcapng` stream 0 proves it. The SYN-ACK is at `+0.003815s`, the
+bare ACK at `+0.003927s` and the Client Hello at `+0.004371s`. The reference
+`JA4L-C` is `278_128`, and `(4371 - 3815) / 2 = 278`.
+
+The point moves in either direction. `http1-with-cookies.pcapng` stream 0 puts it
+on the bare ACK the server sends.
+
+### A complete HTTP request does not move the client point
+
+The FoxIO program keeps the timestamps of a packet under the protocol the tshark
+dissector reports, and it holds a separate cache for `http` and `http2`. A packet
+that carries a whole HTTP request therefore never moves the client point, and a
+packet that carries the first part of a request does move it.
+
+Two vectors prove both halves:
+
+- `latest.pcapng` stream 6 sends one complete `GET` request. The reference
+  `JA4L-C` is `32_128`, which is the bare ACK.
+- `http-empty-useragent.pcap` sends the request line, the header and the blank
+  line in three packets. The reference `JA4L-C` is `177863_64`, which is the
+  request line.
+
+`ja4plus` reads the request line and the blank line that ends the header block.
+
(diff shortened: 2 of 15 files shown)


Crank-Git:
## Self review

I reviewed the diff by two lenses in parallel, correctness and the writing standard.
Both found real defects. `8790cda` carries the fixes.

### Fixed — a reused connection corrupted an already reported value

`self.connections` keys on the endpoints alone. The measurement points were recorded
once, so a later connection on the same endpoints read the points of the earlier one.
The new single-entry rule made that worse than noise: the later connection replaced
the client value the fingerprinter had already reported.

A SYN that carries another initial sequence number now restarts the connection state.
That is how the reference counts the two, because tshark gives the later connection
its own index. `test_a_second_connection_on_the_same_endpoints_gets_its_own_values`
covers it.

### Fixed — a reordered capture lost the server value

The server value was reported only while the SYN-ACK was in hand. A capture that
merges two interfaces holds the SYN-ACK first, and the value never appeared. The SYN
now reports it when the SYN-ACK came first.
`test_a_reordered_capture_still_reports_the_server_value` covers it.

### Fixed — a QUIC flow whose two ports are 443

`to_server` and `from_server` were read apart, so both were true on such a flow and a
client retransmission reported a server value of the client's own delay. A flow that
names no server now gives no value.
`test_the_fingerprinter_reports_nothing_when_both_ports_are_443` covers it.

### Not fixed — a coalesced QUIC datagram

`long_header_packet_type` reads the first QUIC packet of a datagram only. A server
that coalesces an Initial packet and a Handshake packet into one datagram therefore
hides the Handshake packet, and the client measurement point never moves to it. Every
QUIC case in the vector set passes without this, so no vector proves what the change
should be, and `CLAUDE.md` rule 1 forbids a fingerprint change that no vector
requires. I recorded it on #102, which owns the QUIC measurement point.

### Not fixed — a negative latency on a reordered capture

A capture whose SYN-ACK timestamp precedes its SYN timestamp gives a negative
latency. The reference gives a large positive value there, because it reads the
`microseconds` attribute of a negative `timedelta`. Neither value means anything, and
no vector holds the case. I left the arithmetic faithful rather than guess a third
answer.

### Writing

Seven sentences ran past the 25-word limit of `.claude/rules/ste.md`, and the prose
used `produce` where the `## Terms` table of `docs/specs/spec.md` names `emit`. It
also used `cache`, which the table names `state table`, and one British spelling. All
are corrected. `docs/usage.md` stated the client measurement point too simply, and it
now names the bare-ACK case and points at `docs/implementation_notes.md`.

### Verification after the fixes

```
pytest tests/ -m "not spec_validation"   639 passed, 11 skipped, 86 subtests passed
pytest tests/ -m spec_validation         593 passed, 145 skipped, 101 xfailed
ruff check / ruff format --check         clean
coverage                                 72%, and the floor is 70
```

The conformance failure list with `JA4PLUS_IGNORE_DEVIATIONS=1` is identical before
and after the three fixes, so no vector changed.

Crank-Git:
## Rebased onto `d26ad4f`, and one note for the merge gate

#99 and #100 landed while this branch was open, so `tests/foxio_deviations.json`
conflicted. The branch is rebased and the pull request is mergeable again.

**The register needs a reader's attention.** `tests/generate_foxio_baseline.py` still
maps every JA4SSH case to #92, and #99 and #100 point those cases at #96, #97, #98 and
#105. A whole regeneration on this branch therefore renames the owner of eleven
entries that this issue does not own. `a551f8f` keeps the JA4L slice this issue
measured and every other entry the integration branch already carries. The key set is
identical either way, so only the JA4L entries differ from the branch point.

The generator needs the same correction for JA4SSH that this pull request makes for
JA4L, and that belongs to whoever owns #96, #97, #98 and #105. I did not make it here.

`.claude/rules/conformance.md` gained the rule on a FoxIO reference defect after this
branch started. I read it against the two readings in the body of this pull request:

- **A complete HTTP request does not move the client point.** Reproducing this is not
  declined by the rule. The value still depends on the packets of the connection
  alone, not on the composition of the capture, the read order or the position in the
  file, and it still describes traffic. The vectors require it.
- **A time of one second or more.** The reference reads only the sub-second part of
  the difference. No vector holds such a difference, so nothing proves what the value
  should be and `CLAUDE.md` rule 1 keeps the behaviour out. No register entry is
  affected, so there is nothing to point at a divergence issue.
  `docs/implementation_notes.md` records it.

Local suite after the rebase:

```
pytest tests/ -m "not spec_validation"   674 passed, 11 skipped, 86 subtests passed
pytest tests/ -m spec_validation         624 passed, 141 skipped, 74 xfailed
ruff check / ruff format --check         clean
coverage                                 74%, and the floor is 70
```

PR #107, opened 2026-08-07
Read the JA4L propagation factor from the FoxIO hop-count table

Crank-Git:
## What

`calculate_distance` hardcoded the propagation factor 1.6 at `ja4l.py:101`. It now
accepts the observed TTL and reads the factor of the FoxIO table row for the hop count
that TTL implies. `calculate_distance_km` reads the same table.

Satisfies `FR-spec-conformance-9` and `FR-spec-conformance-10`.

## Why

The FoxIO JA4L material gives the factor as a table keyed on hop count. One fixed
factor gives a distance the reference does not give for any path other than a 22-hop
path.

## How

- `PROPAGATION_FACTOR_TABLE` holds the five bounded rows. `MAXIMUM_PROPAGATION_FACTOR`
  holds the last row, which covers 26 hops or more.
- `_propagation_factor` resolves one call: the caller's factor first, then the table
  row for the TTL, then `DEFAULT_PROPAGATION_FACTOR`.
- A TTL above the initial TTL implies a negative hop count. The lookup clamps it to
  zero hops, so the factor is 1.5.
- A call that passes no TTL gives no hop count, so the factor stays 1.6. The interface
  therefore stays compatible for every caller that passes latency alone.
- `MILES_PER_MICROSECOND` and `KILOMETERS_PER_MICROSECOND` replace the two literals.

## How tested

`tests/test_ja4l.py::TestJA4LPropagationFactor` holds seven tests, and they were
written first and committed red in `8d0d983`.

- `pytest tests/ -m "not spec_validation"` — 681 passed, 11 skipped, 93 subtests
  passed. The base gave 674 passed.
- `pytest tests/ -m spec_validation` — 624 passed, 141 skipped, 74 xfailed. The base
  gave the same counts. The divergence register holds 74 entries before and after, so
  this change moves no JA4L conformance value.
- `ruff check` clean. `ruff format --check` clean. `mypy ja4plus/` clean.
- `python -m build` builds the wheel and the source distribution.
- Coverage is 74 percent in total, against the floor of 70.
  `ja4plus/fingerprinters/ja4l.py` is 91 percent.

## Interface

Verified against: https://github.com/FoxIO-LLC/ja4/blob/main/technical_details/JA4L.png
(retrieved 2026-08-06). The image gives the table `<=21` -> `1.5`, `22` -> `1.6`,
`23` -> `1.7`, `24` -> `1.8`, `25` -> `1.9`, `>=26` -> `2.0`, the formula `D = jc/p`,
and the speed of light in fiber as `0.128 miles or 0.206 km per µs`. The FoxIO Python
reference computes no distance, so the image is the only source.

## Docs

- `docs/usage.md` states the new interface and the table.
- `docs/specs/spec.md` gains the terms `hop count` and `propagation factor`. It moves
  no register entry.

Refs #30

### DIFF
--- a/docs/specs/spec.md
+++ b/docs/specs/spec.md
@@ -74,6 +74,8 @@ that user, because comparison is the only thing a fingerprint is for.
 | conform | verb | To produce the same output as the FoxIO reference. | comply, match spec |
 | port | noun | The Go implementation at `Crank-Git/ja4plus-go`. | Go version, sibling, twin |
 | parity | noun | The state where this project and the port expose the same interface and emit the same fingerprint. | alignment, sync, consistency |
+| hop count | noun | The count of routers a packet crossed, read as the initial TTL minus the observed TTL. | hop distance, TTL delta |
+| propagation factor | noun | The JA4L divisor that the FoxIO hop-count table gives for one hop count. | propagation delay, terrain factor |
 
 ## Goals
 
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -176,10 +176,25 @@ for packet in handshake_packets:
 - 64 = Linux / macOS / mobile
 
 **Distance estimation:**
+
+`calculate_distance` returns miles and `calculate_distance_km` returns kilometers. The
+formula is `latency_us * 0.128 / propagation_factor` for miles, and it uses `0.206` for
+kilometers.
+
+Pass the observed TTL, and the method reads the propagation factor from the FoxIO
+hop-count table. A hop count of 21 or fewer gives 1.5, then 1.6, 1.7, 1.8 and 1.9 for
+22, 23, 24 and 25 hops, and 26 hops or more give 2.0. A TTL above the initial TTL
+implies a negative hop count, which clamps to zero hops.
+
 ```python
-# JA4L latency can estimate physical distance
-# distance = latency_us * 0.128 / propagation_factor
-# propagation_factor: 1.5 (good terrain) to 2.0 (poor terrain)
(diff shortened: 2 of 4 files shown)


Crank-Git:
## Self-review

Two Sonnet reviewers read the diff: one for correctness, one for the writing standard.

**Correctness — no finding.** The reviewer traced every table boundary by hand: hop
count 0, 21, 22, 25 and 26 all read the row the FoxIO image gives. `is not None` guards
the caller's factor, so an explicit `0` reaches the division rather than falling through
to the table. Every existing caller passes latency alone or passes `propagation_factor=`
by keyword, so the new second parameter breaks none of them. No packet path changed.

**Writing standard — five findings, five accepted.** `10a5f3a` fixes them:

- Three sentences ran above the 25-word limit and each carried more than two
  conditions. A vertical list replaces each one, in `docs/usage.md`, in the test class
  docstring, and in the `Returns` block of `_propagation_factor`.
- The table text was stated three times. The test docstring now points at
  `PROPAGATION_FACTOR_TABLE` instead of restating it.
- The comment said `km`, which the project defines nowhere. It says `kilometers`.
- One test docstring said "wins over", which is idiom. It says "replaces".

## Verification after the fixes

- `pytest tests/ -m "not spec_validation"` — 681 passed, 11 skipped, 93 subtests passed.
- `pytest tests/ -m spec_validation` — 624 passed, 141 skipped, 74 xfailed.
- `ruff check` clean. `ruff format --check` clean. `mypy ja4plus/` clean.

The register holds 74 entries, and it held 74 on the base.

Issue #105, opened 2026-08-07
Record the JA4SSH divergence on the connection the reference indexes as stream zero


Crank-Git:
## Context

#92 settled the condition FoxIO applies to the window a connection holds open when it
closes, and it found one more defect of the reference beside it.

`python/ja4.py` runs `finalize_ja4ssh` on a packet that carries the FIN flag and the ACK
flag, and that function opens with `if stream:`. The stream index `0` is false in
Python, so the reference emits no trailing window for the connection it holds at index
0. The index describes the position of a connection in a capture. It describes nothing
about the connection.

A measurement with `tshark` 4.6.7 and the FoxIO python implementation at the pinned
commit shows it. The instrumented `finalize_ja4ssh` reports the guard value for each
call.

```
$ python ja4.py --ja4ssh -J tests/foxio_vectors/ssh-r.pcap
FINALIZE called stream=1 truthy=True  in_cache=True protos='eth:ethertype:ip:tcp:ssh'
FINALIZE called stream=2 truthy=True  in_cache=True protos='eth:ethertype:ip:tcp:ssh'
FINALIZE called stream=0 truthy=False in_cache=True protos='eth:ethertype:ip:tcp:ssh'

$ python ja4.py --ja4ssh -J tests/foxio_vectors/gre-sample.pcap
FINALIZE called stream=0 truthy=False in_cache=True protos='eth:ethertype:ip:gre:ip:tcp:ssh'

$ python ja4.py --ja4ssh -J tests/foxio_vectors/sshv1.pcap
FINALIZE called stream=0 truthy=False in_cache=True protos='eth:ethertype:ipv6:tcp:ssh'
```

Each call holds its entry in the cache, and each entry carries a protocol chain that
ends with `:ssh`. The guard is the only thing that stops the trailing window.

One character settles it. The same run, with `if stream:` read as `if stream is not
None:` and nothing else changed, emits the window for every one of them.

```
gre-sample.pcap  stream 0  JA4SSH.1  c24s23_c4s4_c5s4     (the reference holds none)
sshv1.pcap       stream 0  JA4SSH.1  c20s12_c18s23_c10s1  (the reference holds none)
ssh-r.pcap       stream 0  JA4SSH.2  c76s76_c33s48_c41s2  (the reference holds c64s64_c0s0_c0s1)
ssh-r.pcap       stream 1  JA4SSH.1  c64s64_c6s5_c4s5     (unchanged)
[...]

PR #111, opened 2026-08-07
Count an SSH message, not a TCP segment, in a JA4SSH window

Crank-Git:
Refs #98. Epic 1 batch 3.

## What

`ja4plus` counted one SSH packet for every TCP segment that carried a payload. The
FoxIO reference counts the packets `tshark` labels `ssh`. `update_ssh_entry` reads
`has_ssh = ('ssh' in x['protos'])`, and `x['protos']` is the `frame.protocols` field
of `tshark -T ek`. `tshark` reassembles an SSH message that spans two TCP segments,
labels the segment that completes the message, and labels the earlier segment `tcp`.

## Why

The count decides where a window closes. The two implementations disagreed by one
packet for each SSH message that spans two segments, and every later window boundary
moved with it.

## Evidence

`tshark` 4.6.7 names the boundary. Frame 399 carries no `ssh` layer, and frame 400
carries the reassembled message:

```
$ tshark -r tests/foxio_vectors/ssh-r.pcap -Y "tcp.stream==2 && tcp.len>0" \
    -T fields -e frame.number -e tcp.srcport -e tcp.len -e frame.protocols \
    -e tcp.segment -e tcp.reassembled.length
395	46396	21	eth:ethertype:ip:tcp:ssh
397	22	21	eth:ethertype:ip:tcp:ssh
399	46396	1448	eth:ethertype:ip:tcp
400	46396	48	eth:ethertype:ip:tcp:ssh	399,400	1496
```

This command counts, for every SSH vector, the payload segments the reference skips:

```
$ tshark -r tests/foxio_vectors/<vector> -Y "tcp.len>0 && (tcp.port==22)" \
    -T fields -e tcp.stream -e frame.protocols \
  | awk -F'\t' '{ssh = ($2 ~ /:ssh$/) ? "ssh":"notssh"; c[$1" "ssh]++} \
                END {for (k in c) print k, c[k]}' | sort
```

| Vector | Stream | Segments outside a message end |
|---|---|---|
| `ssh-r.pcap` | 0 | 0 |
| `ssh-r.pcap` | 1 | 1 |
| `ssh-r.pcap` | 2 | 1 |
| `ssh-scp-1050.pcap` | 0 | 0 |
| `ssh.pcapng` | 0 | 0 |
| `ssh2.pcapng` | 14 | 0 |

`sshv1.pcap`, `v6.pcap`, `ssh2-malformed.pcap` and `ssh2-moloch-crash.pcap` each hold
two payload segments the SSH dissector rejects outright. Those segments carry no
`tcp.segment` field, so they are not a reassembly case. The reference holds no JA4SSH
value on any of those four vectors, and none of their values changed.

## How

`ja4plus/utils/ssh_utils.py` gains `SSHMessageTracker`. One tracker follows one
direction of one connection, and it reads the four-byte length field of each message
while the direction sends plaintext. It counts every segment before the version banner
and after `SSH_MSG_NEWKEYS`, because neither phase carries a length a reader can
trust. A length outside 2 and 65536 bytes stops the walk, so a hostile packet degrades
the tracker to the previous behaviour rather than dropping packets.

`JA4SSHFingerprinter` passes every payload segment to the tracker of its direction,
and it counts the segment only when the tracker reports a message end.

## Measurement before and after

| Vector | Stream | Reference | Before | After |
|---|---|---|---|---|
| `ssh-r.pcap` | 0 | `c64s64_c107s93_c74s10` | same | same |
| `ssh-r.pcap` | 1 | `c64s64_c6s5_c4s5` | `c48s21_c7s5_c4s5` | `c48s21_c6s5_c4s5` |
| `ssh-r.pcap` | 2, window 1 | `c64s64_c104s96_c19s82` | `c76s76_c104s96_c19s82` | unchanged |
| `ssh-r.pcap` | 2, window 2 | `c76s76_c108s92_c0s105` | same | same |
| `ssh-r.pcap` | 2, window 3 | `c76s76_c106s94_c0s107` | same | same |
| `ssh-r.pcap` | 2, window 4 | `c76s76_c111s89_c0s102` | same | same |
| `ssh-r.pcap` | 2, window 5 | `c76s76_c66s65_c9s51` | `c76s76_c67s65_c9s51` | `c76s76_c66s65_c9s51` |
| `ssh-scp-1050.pcap` | 0 | four windows | same | same |
| `ssh.pcapng` | 0 | `c36s36_c76s124_c0s0` | same | same |
| `ssh2.pcapng` | 14 | `c36s36_c76s124_c74s5` | same | same |

The three windows the warning named — window 2, window 3 and window 4 of `ssh-r.pcap`
stream 2 — still match. `ssh.pcapng` holds `c36s36_c76s124_c0s0`.

The remaining part A difference on stream 1 and on stream 2 window 1 is #96, which is
closed as a decided divergence. Its register entries stay. #97 and #105 are untouched.

## Register

The register falls from 74 entries to 73. `ssh-r.pcap/2:46396/JA4SSH.5` is removed,
because the case now passes and its `xfail(strict=True)` marker reports it. The cause
line of `ssh-r.pcap/1:46394/JA4SSH.1` drops the sentence about the packet count, which
no longer differs. The entry stays, and #96 still owns it.

## How tested

Local suite, because this PR is a batch member and skips provider CI.

- `pytest tests/ -m "not spec_validation"` — 688 passed, 11 skipped, 86 subtests passed.
- `pytest tests/ -m spec_validation` — 625 passed, 141 skipped, 73 xfailed. The base
  reports 624 passed and 74 xfailed.
- `ruff format --check ja4plus/ tests/` — 84 files already formatted.
- `mypy ja4plus/` — Success: no issues found in 27 source files.
- `python -m build` — Successfully built `ja4plus-0.6.0.tar.gz` and the wheel.
- Line coverage 74 percent, against a floor of 70. The base measures 74 percent.
  `ja4plus/utils/ssh_utils.py` rises from 78 percent to 85 percent.

`tests/test_ja4ssh_message_count.py` holds 14 new tests. They were committed before the
implementation, in `787b54c`, and they failed on that commit.

Verified against: https://github.com/FoxIO-LLC/ja4/blob/main/python/ja4ssh.py (retrieved 2026-08-06)
Verified against: https://github.com/FoxIO-LLC/ja4/blob/main/python/ja4.py (retrieved 2026-08-06)
Verified against: https://www.rfc-editor.org/rfc/rfc4253 (sections 6 and 7.3)


Crank-Git:
## Self-review

Two Sonnet children reviewed the diff in parallel: one for correctness, one for the
writing standard and the documented facts. Six findings. Two were defects, and both are
fixed in `0ff0675`.

### Fixed — a version banner that spans two TCP segments turned the tracker opaque

`_read_banner` ran on every segment while the state was `banner`. The second segment of
a split banner does not start with `SSH-`, so the tracker read it as a capture that
starts after the banner. It then counted one SSH packet for every segment of that
direction, which is the behaviour this PR removes.

The tracker now matches the prefix of the line it holds so far, and it turns opaque only
above the 255-byte limit of RFC 4253 section 4.2. Two tests cover it:
`test_a_banner_that_spans_two_segments_keeps_the_message_boundary` and
`test_a_banner_longer_than_the_limit_makes_the_tracker_count_every_segment`.

No vector holds a split banner, so no measured value changed.

### Fixed — the tracker missed SSH_MSG_NEWKEYS when the message code arrived late

The tracker read the message code at the segment offset, right after it parsed the
length field. A segment that ends within the first two body bytes carries no code, and
the tracker recorded `None`. It then missed `SSH_MSG_NEWKEYS`, stayed in the
length-reading state, and read lengths out of encrypted bytes.

The tracker now reads the code at the body offset, so a body that spans two segments
still delivers it. `test_a_new_keys_message_that_spans_two_segments_turns_the_tracker_opaque`
covers it.

### Filed, not fixed — a retransmitted or out-of-order segment desyncs the tracker

The tracker reads segments in capture order and holds no sequence number, so a
retransmission reaches it twice. That is a real gap on live traffic, and it is wider
than this issue: the fix reads the SSH payload through `TCPStreamReassembler`, which
JA4H and JA4X already use, and it needs a capture that holds a retransmission to prove
it. #113 owns it. No FoxIO vector exercises the path, so no conformance case depends on
it today.

### Rejected — nothing

### Accepted — three writing-standard nits

One `-ing` form used as a noun at `ja4plus/fingerprinters/ja4ssh.py:184`, one passive
sentence at `ja4plus/utils/ssh_utils.py:21`, and one word used with two meanings in one
sentence at `ja4plus/utils/ssh_utils.py:124`. All three are corrected.

### Confirmed by the review, with no change needed

- The tracker survives a window boundary. `_close_window` clears `ssh_packets` and
  `bare_acks` alone, which is correct, because the message framing continues across a
  window.
- `cleanup_connection` releases the tracker with the rest of the connection state.
- No input makes the tracker raise, loop forever, or grow without a limit.
  `_length_field` holds fewer than four bytes, `_banner` holds 255 bytes at most, and
  `_remaining` never goes below zero.
- No non-SSH code path changed.
- Every fact the documentation states matches the measurement, including the register
  count of 73 and the three remaining JA4SSH owners.

### The gate after the fixes

- `pytest tests/ -m "not spec_validation"` — 691 passed, 11 skipped, 86 subtests passed.
- `pytest tests/ -m spec_validation` — 625 passed, 141 skipped, 73 xfailed.
- `ruff check`, `ruff format --check`, `mypy ja4plus/` — clean.
- Line coverage 74 percent. `ja4plus/utils/ssh_utils.py` is at 87 percent.
- Every vector measurement is identical to the table in the PR body.

Issue #115, opened 2026-08-07
Run the three tests that name a missing fixture path


Crank-Git:
## What is wrong

Three tests skip on every run. Each one names a capture under
`tests/foxio_vectors/pcap/`, and this repository holds no such directory. The capture
sits at the top level of `tests/foxio_vectors/`.

## Evidence

```
SKIPPED [1] tests/test_ja4_alpn.py:75: fixture missing: tests/foxio_vectors/pcap/tls-non-ascii-alpn.pcapng
SKIPPED [1] tests/test_ja4h_spec.py:95: FoxIO http2-with-cookies fixture missing
SKIPPED [1] tests/test_quic_multipacket.py:132: quic-with-several-tls-frames fixture missing
```

```
tests/test_ja4_alpn.py:73        path = "tests/foxio_vectors/pcap/tls-non-ascii-alpn.pcapng"
tests/test_ja4h_spec.py:96       not os.path.exists("tests/foxio_vectors/pcap/http2-with-cookies.pcapng"),
tests/test_quic_multipacket.py:133   not os.path.exists("tests/foxio_vectors/pcap/quic-with-several-tls-frames.pcapng"),
```

`tests/foxio_vectors/tls-non-ascii-alpn.pcapng`, `tests/foxio_vectors/http2-with-cookies.pcapng`
and `tests/foxio_vectors/quic-with-several-tls-frames.pcapng` are all committed.

## Why it matters

A skipped test reads as a pass in every summary. This is the same defect as #109, which
covered `tests/test_ja4d_foxio.py` and `tests/test_ja4d6_foxio.py` alone.

## What to build

Point each of the three tests at the committed capture, and remove the skip. Read the
result. A test that then fails states a real gap, and it is not a reason to restore the
skip.

Found while #109 was built.

Crank-Git:
## Plan

The base is `epic/12-spec-conformance`, Epic 1 batch 5, off `dev` at `46c5266`.

**#109 solved the same problem. Read what it did first.** Both JA4D test files named
`tests/foxio_vectors/pcap/` and `tests/foxio_vectors/ja4_expected/`, which this
repository does not hold, and the captures were at the top level of
`tests/foxio_vectors/` all along. #109 also found that the reference values existed in a
directory nobody had read.

**Establish, per test, why it skips.** The capture may be present under another name. The
expected output may be absent from the FoxIO Python material and present in the Wireshark
or Zeek material. Say which for each of the three.

[...]

Issue #109, opened 2026-08-07
Run the JA4D and JA4D6 reference tests


Crank-Git:
## What is wrong

`tests/test_ja4d_foxio.py` and `tests/test_ja4d6_foxio.py` skip on every run. Each names a
capture path and an expected-output path that this repository does not hold.

## Evidence

```
tests/test_ja4d_foxio.py:13   PCAP_PATH = "tests/foxio_vectors/pcap/dhcp.pcapng"
tests/test_ja4d_foxio.py:14   EXPECTED_PATH = "tests/foxio_vectors/ja4_expected/dhcp.pcapng.ja4d.json"
```

Neither `tests/foxio_vectors/pcap/` nor `tests/foxio_vectors/ja4_expected/` exists. The
capture is at `tests/foxio_vectors/dhcp.pcapng`, and no file in this repository holds an
expected JA4D value.

`pytest tests/ -m "not spec_validation"` reports both files as skipped.

## Why it matters

A test that always skips reads as a pass. No other reference test covers JA4D or JA4D6.
`tests/foxio_vectors/dhcp.pcapng.json` and `tests/foxio_vectors/dhcpv6.pcap.json` each
hold an empty array, so the conformance suite compares no value for either method.

## What to build

Take one of two paths. Commit the expected-output files that the two tests name, or delete
the two files and keep the record in `docs/implementation_notes.md` that no reference test
covers JA4D or JA4D6.

Found by #31.

Crank-Git:
Triaged as `priority:high`.

**Nothing tests JA4D or JA4D6 today.** Two of the twelve methods carry no reference test
that runs, and a skipped test reads as a passing test in every summary this project has
published. That is the same class of defect as #80, where the conformance harness never
ran the JA4L fingerprinter and 162 register entries stated a harness defect rather than
non-conformance.

Scheduled for Epic 1 batch 4 with #101, #102 and #108.

Crank-Git:
## Plan

The base is `epic/12-spec-conformance`, Epic 1 batch 4, off `dev` at `4e335f3`.

**Establish what runs before you change anything.** Run
`pytest tests/test_ja4d_foxio.py tests/test_ja4d6_foxio.py -rs` and read the skip
reason. State the path each file names and whether this repository holds it.

**The vectors may hold nothing to compare.** #31 measured that
`tests/foxio_vectors/dhcp.pcapng.json` and `dhcpv6.pcap.json` each hold `[]`, and that
[...]

Issue #31, opened 2026-08-06
Record the ambiguous spec readings and replace the parity tests


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/01-spec-conformance.md`](docs/specs/features/01-spec-conformance.md):

> FR-spec-conformance-11 — The project documents every reading it made of an ambiguous part of the specification.

> FR-spec-conformance-12 — `tests/test_parity.py` no longer asserts that a name exists.

## What to build

Seven of the twelve FoxIO methods are published as PNG images rather than text: JA4L, JA4S, JA4SSH, JA4T, JA4X, JA4D and JA4D6. Where an image is ambiguous, the expected-output file decides, and the reading goes into `docs/implementation_notes.md` with the vector that supports it.

`tests/test_parity.py` holds six tests that assert names exist, such as `"ja4d" in VALID_TYPES` and `hasattr(fp, 'last_raw')`. None compares an actual fingerprint. Two implementations can pass every one of them and disagree on every byte. Replace them with the shared vector suite, and keep one test that proves the shard-key format still matches the port's `"%s:%s:%d->%s:%d"`.

This is the last issue in the epic: it confirms all 37 vectors pass.

## Acceptance criteria

- [ ] `docs/implementation_notes.md` records one entry for each ambiguous reading.
- [ ] No test in `tests/test_parity.py` asserts only that an attribute exists.
- [ ] `pytest tests/ -m spec_validation` passes on all 37 vectors.

## Links

- Feature: [`docs/specs/features/01-spec-conformance.md`](docs/specs/features/01-spec-conformance.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #12
Depends on #28

Crank-Git:
## Carried forward from #29

#29 exposed the `JA4_o` value on JA4 and on JA4S.

**`JA4_o` on JA4 is proven by a vector.** `tls12.pcap` gives
`t13d1715h2_5b234860e130_014157ec0da2`, and the fingerprinter produces it.

**`JA4S_o` is proven by nothing.** No FoxIO expected-output file carries a `JA4S_o`
key; every one holds `None`. The value this project publishes for JA4S in original
order is therefore unvalidated in both directions.

[...]

PR #116, opened 2026-08-07
Run the JA4D and JA4D6 reference tests

Crank-Git:
## What

`tests/test_ja4d_foxio.py` and `tests/test_ja4d6_foxio.py` skipped on every run. Each one
named a capture path and an expected-output path this repository does not hold, so two of
the twelve methods carried no reference test that runs. Both files now run and compare ten
FoxIO reference values.

Refs #109.

## Why

A skipped test reads as a pass. Nothing tested JA4D or JA4D6.

The issue offered two paths. The honest deliverable depended on whether the FoxIO material
holds a reference value for either method, so that question came first.

**The FoxIO Python implementation emits neither method.** `tests/foxio_vectors/dhcp.pcapng.json`
and `tests/foxio_vectors/dhcpv6.pcap.json` each hold `[]`, which #31 measured. The FoxIO
Rust snapshots for the same two captures also hold `[]`.

**The FoxIO Wireshark dissector does write a reference value for both.**
`wireshark/test/testdata/dhcp.pcapng.json` holds four values and
`wireshark/test/testdata/dhcpv6.pcap.json` holds six. Both files carry the `ja4.ja4d` key
that the two test files already parsed, and their docstrings already named the Wireshark
dissector as the source. This pull request takes path 1 of the issue: it commits those two
files, without change, from the pinned upstream commit.

**No expected value comes from `ja4plus` output.** Every one of the ten values is a
verbatim copy of FoxIO material.

## The authority question — please read

`.claude/rules/external-apis.md` states that the files under `wireshark/test/testdata/`
are not the authority, because `wireshark/test/testdata/tls12.pcap.json` is an empty array
where the Python file of the same name holds four fingerprints. JA4D and JA4D6 are the
reverse case: the Python file is the empty one, and the Wireshark file is the only FoxIO
reference output that exists for either method.

A second FoxIO implementation corroborates the JA4D values. The Zeek baseline
`zeek/tests/Traces/Scripts.ja4-dhcp/ja4d.log` holds the same four values, byte for byte:

```
disco0000in_61-55_1-3-6-42
offer0000nn_1-58-59-51-54_00
reqst0000in_61-54-55_1-3-6-42
dpack0000nn_58-59-51-54-1_00
```

No second FoxIO implementation emits JA4D6, so the Wireshark dissector stands alone for
those six values. The rule now records the exception and keeps the existing precedence
everywhere else.

## Result

`ja4plus` matches all ten reference values, and it emits no fingerprint the reference does
not hold. The two fingerprinters are unchanged. This pull request adds coverage; it moves
no fingerprint.

| Capture | Frame | Reference value |
|---|---|---|
| `dhcp.pcapng` | 1 | `disco0000in_61-55_1-3-6-42` |
| `dhcp.pcapng` | 2 | `offer0000nn_1-58-59-51-54_00` |
| `dhcp.pcapng` | 3 | `reqst0000in_61-54-55_1-3-6-42` |
| `dhcp.pcapng` | 4 | `dpack0000nn_58-59-51-54-1_00` |
| `dhcpv6.pcap` | 2 | `solct0014nn_1-6-8-25_23-24` |
| `dhcpv6.pcap` | 5 | `advrt0014nn_25-26-1-2_00` |
| `dhcpv6.pcap` | 7 | `reqst0014nn_1-2-6-8-25-26_23-24` |
| `dhcpv6.pcap` | 8 | `reply0014nn_25-26-1-2_00` |
| `dhcpv6.pcap` | 11 | `relse0014nn_1-2-6-8-25-26_23-24` |
| `dhcpv6.pcap` | 12 | `reply0014nn_1-2-13_00` |

## How it is tested

**The tests came first.** Commit `6d8d336` rewrites both files and fails with
`FileNotFoundError`: `3 failed, 3 passed`, and zero skipped. Commit `ee6846a` adds the
expected output, and the same command gives `6 passed`.

**Each test compares a value, and a mutated expected-output file proves it.**

- A changed value in each file: `2 failed, 4 passed`, and both reference tests failed.
- One frame dropped from `dhcpv6.pcap.json`: `2 failed, 3 passed`.

**The comparison runs in both directions.** The assertion compares the whole map of frame
number to fingerprint, so a method that emits more fingerprints than the reference fails,
and so does one that emits fewer.

**`_load_expected` fails on an empty file.** The refresh script also fails when either
Wireshark file holds an empty array. Neither one can quietly return to a pass on nothing.

**The refresh script reproduces the whole vector set.** Running
`python tests/download_test_vectors.py` rewrote all 37 committed vectors byte-identically
and added the two new files, so the only change to `tests/foxio_vectors/` is the new
subdirectory and the NOTICE text.

## Local checks

```
pytest tests/ -m "not spec_validation"   706 passed, 3 skipped, 93 subtests passed
pytest tests/ -m spec_validation         625 passed, 141 skipped, 73 xfailed
ruff format --check ja4plus/ tests/      84 files already formatted
```

The deviation register holds 73 entries on the base and 73 entries here. The conformance
suite globs only `tests/foxio_vectors/*.json`, so the subdirectory adds no case and no
register entry.

Coverage rises. Two test files moved from zero cases that run to six that run.

## Out of scope

Three more tests skip for the same reason, and each names a capture under
`tests/foxio_vectors/pcap/`. They are filed as #115, not fixed here.

## Verified against

```
Verified against: https://github.com/FoxIO-LLC/ja4/tree/27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8/wireshark/test/testdata (retrieved 2026-08-06)
Verified against: https://github.com/FoxIO-LLC/ja4/blob/27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8/zeek/tests/Traces/Scripts.ja4-dhcp/ja4d.log (retrieved 2026-08-06)
Verified against: https://github.com/FoxIO-LLC/ja4/tree/27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8/rust/ja4/src/snapshots (retrieved 2026-08-06)
```

### DIFF
--- a/.claude/rules/external-apis.md
+++ b/.claude/rules/external-apis.md
@@ -44,7 +44,16 @@ under `python/test/testdata/` decide. Record the reading in
 
 **The files under `wireshark/test/testdata/` are not the authority.**
 `wireshark/test/testdata/tls12.pcap.json` is an empty array, while the file with the
-same name under `python/test/testdata/` holds four fingerprints.
+same name under `python/test/testdata/` holds four fingerprints. Where both directories
+carry a value for a method, `python/test/testdata/` decides.
+
+**JA4D and JA4D6 are the one exception.** The FoxIO Python implementation emits neither
+method, so `python/test/testdata/dhcp.pcapng.json` and
+`python/test/testdata/dhcpv6.pcap.json` each hold an empty array. The Wireshark dissector
+is the only FoxIO implementation that writes a reference value for the two methods, and
+`tests/foxio_vectors/wireshark_expected/` holds a copy of its two files. The Zeek
+baseline `zeek/tests/Traces/Scripts.ja4-dhcp/ja4d.log` holds the same four JA4D values.
+`docs/implementation_notes.md` records the reading.
 
 **`ja4db.com` publishes no versioned document.** Treat any unexpected response shape as
 a miss. Never let its response shape reach a caller unchecked.
--- a/.claude/rules/external-apis.md
+++ b/.claude/rules/external-apis.md
@@ -50,7 +50,7 @@ carry a value for a method, `python/test/testdata/` decides.
 **JA4D and JA4D6 are the one exception.** The FoxIO Python implementation emits neither
 method, so `python/test/testdata/dhcp.pcapng.json` and
 `python/test/testdata/dhcpv6.pcap.json` each hold an empty array. The Wireshark dissector
-is the only FoxIO implementation that writes a reference value for the two methods, and
+is the only FoxIO implementation that writes a reference value for the two methods.
 `tests/foxio_vectors/wireshark_expected/` holds a copy of its two files. The Zeek
 baseline `zeek/tests/Traces/Scripts.ja4-dhcp/ja4d.log` holds the same four JA4D values.
 `docs/implementation_notes.md` records the reading.
--- a/docs/implementation_notes.md
+++ b/docs/implementation_notes.md
@@ -128,21 +128,22 @@ The list keeps the wire order. `ja4plus` never sorts it.
 The FoxIO Python implementation emits no JA4D and no JA4D6, so the expected-output file
 of each DHCP capture holds an empty array. `tests/foxio_vectors/dhcp.pcapng.json` and
 `tests/foxio_vectors/dhcpv6.pcap.json` each hold `[]`. The FoxIO Rust implementation
-emits neither method either, and both of its snapshots hold `[]`.
+emits neither method. Both of its snapshots hold `[]`.
 
 The FoxIO Wireshark dissector does write a reference value for both methods.
 `tests/foxio_vectors/wireshark_expected/` holds a copy of the two files, taken without
 change from `wireshark/test/testdata/` at the pinned upstream commit.
 
 `.claude/rules/external-apis.md` states that the files under `wireshark/test/testdata/`
-are not the authority, because one of them holds an empty array where the Python file of
-the same name holds four fingerprints. These two methods are the reverse case, and the
+are not the authority. `wireshark/test/testdata/tls12.pcap.json` holds an empty array
+where the Python file of the same name holds four fingerprints. These two methods are the
+reverse case, and the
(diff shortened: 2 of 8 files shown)


Crank-Git:
Self-review by two lenses, correctness and the writing standard. Both ran over the whole diff.

**Correctness — no finding.** The reviewer checked five categories and confirmed each one against the running code.

- Neither test can pass without a comparison. A missing expected-output file raises `FileNotFoundError`, and an empty array trips the `assert expected` guard.
- The frame numbering is right for both captures. `dhcpv6.pcap` holds 12 packets and carries MLDv2 and neighbour-discovery traffic between the DHCPv6 messages, and the 1-based enumeration lands on frames 2, 5, 7, 8, 11 and 12 exactly. Neither the pcapng format nor the pcap format shifts the number.
- `tests/download_test_vectors.py` reuses `capture` across two sequential loops, which is reuse and not shadowing. `WIRESHARK_DIR` exists before the write.
- `Path.glob` does not descend, so `tests/foxio_vectors/wireshark_expected/` is invisible to `tests/test_spec_validation.py` and to the manifest. The conformance suite reports the same shape as the base.
- `assert actual == expected` is stronger than the assertion it replaces, which tested only that each expected frame is present and never caught an extra fingerprint.

**Writing standard — six findings, all fixed in 7c71f03.** Four sentences ran past the 25-word limit, one used a double negative, and one used an `-ing` form as a noun.

**One finding rejected.** The reviewer flagged `docs/specs/spec.md:493` as a stale claim, because that register entry still reads that both files skip on every run and nothing tests JA4D or JA4D6. That entry is row 20 of the dated decision log, and it records the state Epic 1 batch 3 measured. A log entry describes the day it was written. Rewriting it destroys the record of what the batch found, so this pull request leaves it alone. The batch 4 entry is the place to record the close, and the pull request that ships the batch owns it.

The local suite is green after the fix: 706 passed and 3 skipped in the unit suite, 625 passed and 73 xfailed in the conformance suite, `ruff` clean, `mypy` clean.

PR #117, opened 2026-08-07
Group a mirrored tunnel capture by its inner connection

Crank-Git:
## What

JA4L holds two keys for one connection.

- The connection key groups the packets. It reads the inner address pair and the inner port pair.
- The reported key names the stream. It reads the outer address pair and the inner port pair, because the reference reports those.

`ja4plus/utils/tunnels.py` gains `innermost_layer`, which reads the deepest address layer or port layer of one packet. `cleanup_connection` reads a map of the reported key to the grouping key, so a caller that names the outer address pair still removes the state.

## Why

`gre-erspan-vxlan.pcap` is an ERSPAN mirror. Every packet travels from `100.20.9.2` to `100.20.9.1`, so the outer address pair separates no direction. The SYN reached `tcp_100.20.9.1:80_100.20.9.2:65174` and the SYN-ACK reached `tcp_100.20.9.1:65174_100.20.9.2:80`, so no key held both measurement points and the fingerprinter reported nothing.

## How tested

Each test came before the code, and each one failed on the base:

- `test_the_fingerprinter_groups_a_mirrored_capture_by_its_inner_connection`
- `test_the_mirrored_capture_holds_one_connection`
- `test_cleanup_removes_a_tunnelled_connection_by_the_reported_address`
- Five cases in `tests/test_tunnels.py` cover `innermost_layer`.

`gre-erspan-vxlan.pcap` now produces `JA4L-S=997_64` and `JA4L-C=953_64` on the stream `100.20.9.2:65174` to `100.20.9.1:80`.

**The whole conformance suite is measured before and after.** A script dumped every produced fingerprint of all 37 vectors, grouped by stream identity. The only difference is the intended one:

```
147c147,156
<  "gre-erspan-vxlan.pcap": {},
---
>  "gre-erspan-vxlan.pcap": {
>   "100.20.9.1:80|100.20.9.2:65174": {
>    "JA4L-C": [ "953_64" ],
>    "JA4L-S": [ "997_64" ]
>   }
>  },
```

No vector that conformed before stopped conforming.

| Check | Base | This branch |
|---|---|---|
| `pytest tests/ -m spec_validation` | 625 passed, 141 skipped, 73 xfailed | 629 passed, 141 skipped, 69 xfailed |
| `pytest tests/ -m "not spec_validation"` | 700 passed, 9 skipped | 709 passed, 9 skipped, 93 subtests passed |
| `ruff check` and `ruff format --check` | clean | clean, 85 files already formatted |
| `mypy ja4plus/` | clean | Success: no issues found in 27 source files |
| coverage | 75% | 76% |

The deviation register falls from 73 entries to 69. The four entries this issue owned are deleted, and each of the four cases now passes.

`docs/implementation_notes.md` records the two keys and the measurement.

## Verified against

- https://github.com/FoxIO-LLC/ja4/tree/main/python/test/testdata (retrieved 2026-08-06). `tests/foxio_vectors/gre-erspan-vxlan.pcap.json` holds `JA4L-S` `997_64` and `JA4L-C` `953_64`.
- https://scapy.readthedocs.io/en/latest/api/scapy.packet.html (scapy 2.7.0). The page documents `NoPayload` as a `Packet` subclass whose `add_payload` returns `NoReturn`. The page states nothing about the end of the layer chain, so `innermost_layer` was measured against the vector: the walk over `gre-erspan-vxlan.pcap` frame 1 reads 11 layers and stops at `NoPayload`.

## Follow-up filed

- #119: `_src_is_client` reads the outer address and compares it against the inner one. The function has no caller, so nothing reads a wrong answer today.

Refs #101

### DIFF
--- a/docs/implementation_notes.md
+++ b/docs/implementation_notes.md
@@ -230,6 +230,29 @@ inner ports.
 `ja4plus/utils/tunnels.py` imports the scapy dissectors for Geneve, VXLAN and
 ERSPAN, because scapy leaves them unbound and stops at the tunnel header.
 
+### The connection key of a mirrored capture
+
+A mirror sends both directions of one session from one outer address to one
+other outer address. The outer address pair then separates no direction, and one
+key cannot hold both measurement points of the connection.
+
+`gre-erspan-vxlan.pcap` is such a capture. Every packet travels from
+`100.20.9.2` to `100.20.9.1`, and the inner session is `10.16.27.12:65174` to
+`10.16.27.131:80`. The SYN reached the key
+`tcp_100.20.9.1:80_100.20.9.2:65174`, and the SYN-ACK reached the key
+`tcp_100.20.9.1:65174_100.20.9.2:80`.
+
+`ja4plus/fingerprinters/ja4l.py` holds two keys for one connection:
+
+- The connection key groups the packets. It reads the inner address pair and the
+  inner port pair, which name both endpoints of a mirrored session.
+- The reported key names the stream. It reads the outer address pair and the
+  inner port pair, because the reference reports those. The SYN pairs the source
+  address with the source port, and a later packet does not move that pair.
+
+The expected-output file holds `JA4L-S` `997_64` and `JA4L-C` `953_64` on the
+stream `100.20.9.2:65174` to `100.20.9.1:80`. Read #101 for the measurement.
+
 ### The QUIC measurement points
 
 The reference reads four QUIC packets, and it reads the direction from port 443:
--- a/ja4plus/fingerprinters/ja4l.py
+++ b/ja4plus/fingerprinters/ja4l.py
@@ -75,6 +75,10 @@ class JA4LFingerprinter(BaseFingerprinter):
         """Initialize the fingerprinter."""
         super().__init__()
         self.connections = {}
+        # A caller of `cleanup_connection` holds the address pair this fingerprinter
+        # reports, and a tunnelled connection groups under its inner address pair. This
+        # map reads the grouping key from the reported key.
+        self.grouping_keys = {}
 
     def process_packet(self, packet):
         """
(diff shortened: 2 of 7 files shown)


Crank-Git:
## Self-review

I reviewed the diff for correctness and robustness. Two findings.

**1. `cleanup_connection` reached no tunnelled connection. Fixed in 526170f.**

`cleanup_connection` builds its key from the addresses the caller holds, and
`Processor._packet_endpoints` reads the outer address layer. The grouping key of the
first commit reads the inner address pair, so the call reached no entry and the state of
every tunnelled connection stayed. That defeats the contract the method states, and it
breaks the traffic shape this pull request adds support for.

The fingerprinter now keeps a map of the reported key to the grouping key, and
`cleanup_connection` reads it. `reset` clears it.
`test_cleanup_removes_a_tunnelled_connection_by_the_reported_address` failed before the
fix and passes after it.

**2. `_src_is_client` compares an outer address against an inner one. Filed as #119.**

The function reads the outer address with `get_ip_layer` and compares it against the
address the connection key holds, which is now the inner one. No caller exists, so
nothing reads a wrong answer today. The code is out of scope here, so #119 owns it.

**Checked with no finding.**

- `innermost_layer` walks an already dissected packet, so it raises on no packet shape.
  `NoPayload` ends the chain, a `NoPayload` argument returns None, and no loop repeats.
- `_reported_key` gives one key for both directions of a connection that is not
  mirrored, because both directions give the same two endpoints and the function sorts
  them.
- `conn["reported_key"]` holds a value before any entry reaches `self.fingerprints`. The
  first packet of a connection sets it, and both layer reads are checked for None first.
- `tests/conformance_index.py` reads the key form `<protocol>_<address>:<port>_<address>:<port>`,
  which this change keeps.
- The new code adds no raise path. It reads dissected fields only.

Issue #118, opened 2026-08-07
Tell a QUIC server Initial packet from a client one in JA4S


Crank-Git:
## Context

Found while #102 was built. Nothing in #102 changes it, and no vector requires a
change, so `CLAUDE.md` rule 1 keeps it out of that issue.

`ja4plus/fingerprinters/ja4s.py` reads every QUIC Initial packet as a client Initial
packet. `_is_quic_client_initial` tests the long-header packet type alone, and a
server Initial packet carries the same type. The fingerprinter then stores the
connection ID of that packet under the reverse connection key and returns, so it never
reaches `parse_quic_server_initial`.

## The measurement

`chrome-cloudflare-quic-with-secrets.pcapng` stream 0, port 50280, after #102:

```
dcids {'2001:db8:1::1:50280-2606:4700:10::6816:826:443': '203f9e9f68698274',
       '2606:4700:10::6816:826:443-2001:db8:1::1:50280': ''}
```

The second entry is the server Initial packet. It carries an empty connection ID,
because a server Initial packet names the client with the connection ID the client
chose as its source. The stored value is therefore empty, and the server path never
runs on this connection.

The same capture proves that the decryption works. #102 made both server Initial
packets of that connection decrypt, and the packet at `+21981 us` holds a 90-byte
ServerHello.

## Why no vector requires the change

The reference holds no `JA4S` value for stream 50280. A fingerprinter that read the
server path there would emit a value the reference does not hold. Whoever builds this
issue measures the reference first and states what it produces.

## What to build

1. Read the direction of a QUIC Initial packet from the port, as `ja4l.py` does, or
   from the presence of a stored connection ID for the reverse key.
2. Keep the stored connection ID of the client Initial packet. A server Initial packet
   never replaces it.
3. Run the conformance suite. No vector that passes today may fail.

## Links

- Found by: #102
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Crank-Git:
Triaged for Epic 1 batch 5.

**No vector requires this change**, which is why #102 correctly left it alone. The
project holds no JA4S QUIC vector, so the conformance suite reports nothing here and the
register holds no entry for it.

That is the reason to fix it deliberately rather than to wait: **a defect no vector
reports is a defect no gate will ever catch.** The measurement in the body is the
evidence, and a unit test is the gate.

Crank-Git:
## Plan

[...]

PR #120, opened 2026-08-07
Read the JA4L QUIC server point from the completing Initial packet

Crank-Git:
Refs #102.

## What

The JA4L server measurement point of a QUIC connection now reads the Initial packet
that completes the ServerHello. `chrome-cloudflare-quic-with-secrets.pcapng` stream 0
port 50280 reports `JA4L-S=10990_56`, and `tls3.pcapng` stream 25 port 61884 reports
`JA4L-S=3583_57`. The deviation register falls from 73 entries to 71.

## Why

`ja4plus` read the first server Initial packet. The reference reads the Initial packet
whose TLS handshake type is `2`, and a server sends an Initial packet that holds an
ACK frame before it sends the ServerHello.

#88 recorded that both server Initial packets of stream 50280 fail AEAD decryption.
The plan asked whether the measurement point can be found without decryption.
**Decryption is the prerequisite, and it works.** The cause is one line of
`decrypt_initial_payload`:

```python
ciphertext = packet_bytes[pn_offset + pn_length :]
```

It read to the end of the UDP datagram. RFC 9000 Section 17.2 states, verbatim:
`Length: This is the length of the remainder of the packet (that is, the Packet Number
and Payload fields) in bytes, encoded as a variable-length integer (Section 16).` RFC
9001 Section 5.3 names the AEAD plaintext, verbatim: `The payload of the QUIC packet,
as described in [QUIC-TRANSPORT].` A sender coalesces packets in one datagram under
RFC 9000 Section 12.2, so the reader covered bytes the tag does not, and the tag
failed. The client Initial packet escaped this, because its PADDING frames sit inside
the QUIC payload and its Length field reaches the end of the datagram.

The measurement, from `chrome-cloudflare-quic-with-secrets.pcapng` stream 50280:

```
pkt#46 client Initial dcid=203f9e9f68698274
pkt#47 server Initial t=+18569us
  old parse_quic_server_initial: None
  NEW ok pn=0 len=23 plaintext=6B frames=[]
pkt#48 server Initial t=+21981us
  old parse_quic_server_initial: None
  NEW ok pn=1 len=111 plaintext=94B frames=[(0, 90)]
    crypto off=0 first_byte=0x02 len=90
```

`tls3.pcapng` stream 61884 has the same shape: an ACK-only Initial packet at
`+6102 us`, and a 96-byte ServerHello at `+7166 us`. `7166 / 2 = 3583`.

## How

- `_initial_packet_end` bounds the ciphertext by the Length field. It returns the
  datagram length for a header it cannot read, so a hostile packet keeps the old
  behaviour rather than raising.
- `decrypt_quic_server_initial_crypto` returns the CRYPTO fragments of one server
  Initial packet. `parse_quic_server_initial` now calls it, so one decryption path
  serves both readers.
- `server_hello_is_complete` reports whether the collected fragments hold a whole
  ServerHello.
- `ja4l.py` stores the client connection ID, decrypts each server Initial packet, and
  records the point on the packet that completes the ServerHello. The fragment buffer
  of one connection stops at 16384 bytes.

**A connection whose server Initial packets do not decrypt now emits no `JA4L-S`
value.** That is what the reference does: no `packet_type == 0` packet with handshake
type `2` means no `B` point. No vector regressed.

**Neither vector splits a ServerHello across two Initial packets.** The issue text
expected a split; the measurement shows an ACK-only first packet. The accumulation
across packets is built because RFC 9001 allows the split, and it is proven by
synthetic packets in `tests/test_ja4l_quic_server_hello.py`, not by a vector.

## How tested

Local, in the worktree. `ci: skip`, so this replaces provider CI.

- `pytest tests/ -m "not spec_validation"` — 733 passed, 9 skipped, 93 subtests
  passed. It was 697 passed before.
- `pytest tests/ -m spec_validation` — 627 passed, 141 skipped, 71 xfailed, 0 failed.
  It was 625 passed and 73 xfailed. The two cases that moved are the two acceptance
  criteria.
- `ruff format --check ja4plus/ tests/` — 87 files already formatted.
- `mypy ja4plus/` — Success: no issues found in 27 source files.
- Line coverage 76%, against a floor of 70. It was 75%.

TDD: `e970637` lands the tests and removes the two register entries. With that commit
alone, 5 behaviour tests and both conformance cases fail. `480c1f3` lands the fix.
`adba8c0` and `78a9d34` answer the self-review; both are described in the PR comments.

## Note

`ja4s.py` reads a server Initial packet as a client Initial packet, records its empty
connection ID, and never reaches the server path. No JA4S value changed here as a
result. Filed as #118 rather than fixed, because no vector requires the change and the
reference holds no `JA4S` value for stream 50280.

Verified against: https://www.rfc-editor.org/rfc/rfc9000.txt (RFC 9000, retrieved 2026-08-07)
Verified against: https://www.rfc-editor.org/rfc/rfc9001.html (RFC 9001, retrieved 2026-08-07)
Verified against: https://github.com/FoxIO-LLC/ja4/blob/main/python/ja4.py (retrieved 2026-08-07)


Crank-Git:
## Self-review finding: an unbounded CRYPTO fragment offset (fixed in `adba8c0`)

A decrypted CRYPTO frame carries an offset, and RFC 9000 Section 16 lets that offset
reach `4611686018427387903`. `reassemble_crypto_fragments` allocates a buffer that
reaches the highest offset:

```python
total_len = max(off + len(data) for off, data in sorted_frags)
buf = bytearray(total_len)
```

Nothing checked it. A server that named an offset of `2**40` made the fingerprinter
allocate a terabyte inside `process_packet`. Measured:

```
$ python -c "from ja4plus.utils.quic_utils import reassemble_crypto_fragments; reassemble_crypto_fragments([(2**34, b\"x\")])"
(allocates 16 GiB and returns)

$ python -c "from ja4plus.utils.quic_utils import reassemble_crypto_fragments; reassemble_crypto_fragments([(2**40, b\"x\")])"
(the process dies)
```

`CLAUDE.md` states: `Every packet is hostile input. No parser trusts a length field it
read from the packet.` An offset is such a field, and the change above newly reached
this call from JA4L.

`_quic_server_initial` now drops a fragment that reaches past
`MAXIMUM_SERVER_CRYPTO_BYTES`, and it still reads a later Initial packet that carries
the ServerHello. Two tests state both halves.

**#122 filed, not fixed here.** `JA4Fingerprinter._try_quic_multi_packet` reaches the
same call, it is older than this change, and it is reachable by anyone who sends one
UDP datagram: a client Initial packet carries in cleartext the connection ID its keys
derive from. Fixing it changes a path this issue does not own, so it is filed.

Also filed: #118, `ja4s.py` reads a server Initial packet as a client Initial packet.

## Adversarial sweep

`_initial_packet_end`, `initial_packet_dcid`, `decrypt_quic_server_initial_crypto`,
`parse_quic_server_initial` and `server_hello_is_complete` were run against roughly
4000 random byte strings plus targeted truncated and malformed headers. No call
raised, so the parsers return nothing rather than raising.

Crank-Git:
## Self-review round two, and the replies to each finding

Two Sonnet reviewers read the diff: one for correctness, one for the writing standard
and test quality. `78a9d34` answers them.

### Accepted and fixed

**Blocking — the fragment bound protected the JA4L path alone.** `adba8c0` bounded
`_quic_server_initial`. It did not bound `parse_quic_server_initial`, which
`ja4plus/fingerprinters/ja4s.py:73` calls on every server Initial packet. One datagram
that named a CRYPTO frame offset of `2**40` made that parser allocate a terabyte. This
branch also removed the `except Exception` that used to swallow the `MemoryError`, so
the failure newly reached the caller. `collect_crypto_fragments` now holds the bound,
and both paths call it. Measured: the parser returns None in `0.000s` on that packet,
where it did not return within 10s before.

The bound also stops *before* it passes 16384 bytes rather than after, so a sender that
repeats a fragment cannot overshoot by one fragment. A test found that off-by-one.

**Missing test for the byte cap.** `MAXIMUM_CRYPTO_BUFFER_BYTES` now has three tests:
the offset guard, the byte guard, and the fragment that fits.

**Writing standard.** Split ten sentences above the 25-word limit across
`quic_utils.py`, `quic_builder.py`, both new test modules, the feature file and the
decision register. Replaced the four-word noun cluster in the
`test_ja4l_quic_server_hello.py` title. Turned the three-function list in `CHANGELOG.md`
into a vertical list, per rule 16. Added `Retry packet` to the `## Terms` table, because
the `_quic_client_initial` docstring uses it as a defined QUIC concept.

### Rejected, with the reason

**Restore the `try/except Exception` around `parse_tls_handshake`.** Declined. The
reviewer found no input that makes it raise, `client_hello_from_crypto_fragments` in
this same module already calls it unguarded, and `CLAUDE.md` states that a parser
catches the errors it expects rather than bare `Exception`. Re-adding a bare catch to
match a pre-existing bare catch makes the convention worse, not better.

**Bare `except Exception` in `decrypt_quic_server_initial_crypto`.** Kept. It is the
pre-existing pattern of `decrypt_quic_initial_crypto` beside it, and narrowing one of
the two is a half-change. Worth its own issue.

**`bytearray(total_len)` raises on a negative offset.** Not reachable:
`_decode_varint` returns no negative number. Left alone.

**QUIC reordering and connection reuse have no restart path.** Confirmed pre-existing,
identical before and after this diff, and no vector covers it. `CLAUDE.md` rule 1 keeps
it out. Filed as #123.

**The builder shares key-derivation functions with the code under test.** Accepted as a
known limit, and `tests/quic_builder.py` states it in its own docstring. The two
acceptance-criterion tests read the real FoxIO captures and compare against the
reference values, so a key-derivation defect still fails the suite.

### Gates after the round

- `pytest tests/ -m "not spec_validation"` — 733 passed, 9 skipped, 93 subtests passed.
- `pytest tests/ -m spec_validation` — 627 passed, 71 xfailed, 0 failed.
- `ruff check` clean, `ruff format --check` clean, `mypy` clean, coverage 76%.

Issue #121, opened 2026-08-07
Compare the raw forms against the FoxIO reference


Crank-Git:
## What is wrong

`tests/conformance_index.py:29` drops every key that ends with `_r`, `_ro`, `_o` or
`_raw` before the comparison. **No raw form of any method is ever compared against the
FoxIO reference.**

The FoxIO expected-output files carry those keys. `tls-alpn-h2.pcap.json` holds
`JA4S_r`, and #108 found that `ja4plus` does not match it, by reading the file directly
rather than by a suite failure.

## Why it matters

This is the same shape as #80 and #109. A comparison that is never made reads as a
comparison that passes.

- #80: the harness never ran the JA4L fingerprinter, and 162 register entries stated a
  harness defect rather than non-conformance.
- #109: two test files named paths this repository does not hold, so nothing tested JA4D
  or JA4D6.
- This issue: every raw form is dropped before the comparison.

A raw form exists so that a person can read what produced a hash. A raw form that no
vector checks is a value this project has never verified.

## What to build

1. Compare each raw key the expected-output files carry, per method and per occurrence,
   the same way the hashed form is compared.
2. Measure the result before you fix anything, and register each failure against the
   issue that owns it. **Expect the count to rise**, and say by how much.
3. #108 owns the JA4S extension order. Do not fix a method's raw form in this issue.

## Acceptance criteria

- [ ] The conformance suite compares every raw key the FoxIO expected-output files hold.
- [ ] The measurement states how many raw cases pass and how many fail, per method.
- [ ] Every failing raw case carries a register entry naming a real cause and its issue.
- [ ] No hashed-form case changes state.

Part of #12

Crank-Git:
## Plan

The base is `epic/12-spec-conformance`, Epic 1 batch 5, off `dev` at `46c5266` plus
#119.

**This issue was sequenced after #115 and now runs beside it.** #115 makes three skipped
tests run, and it may add register entries. Measure your delta against **your own base
commit**, and state that commit, so the two numbers stay attributable. A conflict on
`tests/foxio_deviations.json` is expected and the project manager resolves it at
sub-merge.

**Measure before you fix.** The first honest number will look like a regression and will
not be one. Report, per method: how many raw cases the expected-output files hold, how
many pass, and how many fail.

**Do not fix a method's raw form in this issue.** #108 landed the JA4S extension order and
recorded the sort rule of each method in `docs/implementation_notes.md` with the evidence
[...]

Issue #108, opened 2026-08-07
Match the FoxIO JA4S_r extension order in the raw form


Crank-Git:
## What is wrong

`ja4plus` sorts the JA4S extensions in the `raw` value. The FoxIO `JA4S_r` value holds
them in wire order, so the two do not match.

## Evidence

`tests/foxio_vectors/tls-alpn-h2.pcap.json` holds:

```
"JA4S_r": "t1204h2_cca9_0000,ff01,000b,0010"
```

`ja4plus` produces:

```
raw                 t1204h2_cca9_0000,000b,0010,ff01
raw_original_order  t1204h2_cca9_0000,ff01,000b,0010
```

The command that produced it:

```
python -m ja4plus.cli --format json --types ja4s analyze tests/foxio_vectors/tls-alpn-h2.pcap
```

`ja4plus/fingerprinters/ja4s.py:252` sorts the extensions for the `raw` value.

The JA4S fingerprint is correct. It hashes the extensions in wire order and it equals the
reference `JA4S` value. The difference is in the raw form alone.

No vector reports this. `tests/conformance_index.py:29` drops every key that ends with
`_r`, `_ro`, `_o` or `_raw`, so the conformance suite compares no raw form.

## What to build

Decide which value the `raw` key of JA4S holds. JA4 carries two distinct raw forms,
because JA4 sorts its ciphers and its extensions for the fingerprint. JA4S sorts neither,
so both raw forms of JA4S carry one value.

Found by #31.

Crank-Git:
## Plan

The base is `epic/12-spec-conformance`, Epic 1 batch 4, off `dev` at `4e335f3`.

**This changes a published field, so state it as a breaking change.** A caller who reads
`raw` on a JA4S result gets a different string after this issue. The `CHANGELOG.md` entry
says so, in the form #28 used for the JA4SSH window.

**Establish the rule per method before you change anything.** JA4 sorts the extensions in
`JA4_r` and holds wire order in `JA4_ro`. The evidence in this body shows JA4S does not
sort. Read the FoxIO material for each method that exposes a raw form, and state which
ones sort and which do not. **Do not apply the JA4S finding to another method without
its own evidence.**

**Parity.** The port at `Crank-Git/ja4plus-go` may expose the same field. FoxIO specifies
`JA4S_r` here, so FoxIO decides and the port follows. Read the port to see what it
exposes; never build, run or import it, and never edit it.

[...]

Issue #28, opened 2026-08-06
Emit a JA4SSH fingerprint every 200 SSH packets


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/01-spec-conformance.md`](docs/specs/features/01-spec-conformance.md):

> FR-spec-conformance-4 — JA4SSH emits one fingerprint for every 200 SSH packets on a connection.

> FR-spec-conformance-5 — The JA4SSH window size is the value the caller passed to the constructor.

> FR-spec-conformance-6 — JA4SSH emits no fingerprint before the window fills.

## What to build

`ja4plus/fingerprinters/ja4ssh.py:170` reads `if total_packets >= min(self.packet_count, 10)`, above a comment that says `# For testing purposes`. FoxIO states the interval verbatim in `technical_details/JA4SSH.png`: `(runs every 200 SSH packets by default)`.

FoxIO's expected output for `ssh.pcapng` is a single fingerprint, `c36s36_c76s124_c0s0`. The two counts sum to 200, which confirms the window size.

The window counts SSH packets sent by either side. A bare ACK is not an SSH packet and does not advance the window. The bare-ACK counters reset with the window.

This is a breaking behaviour change for every current user of JA4SSH. Say so plainly in the changelog. The Go port at `Crank-Git/ja4plus-go` carries the same defect at `ja4ssh.go:176`; do not edit that repository from here.

Verified against https://github.com/FoxIO-LLC/ja4/tree/main/technical_details (retrieved 2026-08-06).

## Acceptance criteria

- [ ] A JA4SSH fingerprinter built with default arguments emits nothing at 10 packets.
- [ ] A JA4SSH fingerprinter built with default arguments emits one fingerprint at 200 SSH packets.
- [ ] `ssh.pcapng` produces exactly one JA4SSH fingerprint, and it equals `c36s36_c76s124_c0s0`.
- [ ] A caller who passes `packet_count=10` gets a window of 10.

## Links

- Feature: [`docs/specs/features/01-spec-conformance.md`](docs/specs/features/01-spec-conformance.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #12
Depends on #27

Crank-Git:
## Plan

[...]

PR #124, opened 2026-08-07
Match the FoxIO JA4S_r extension order in the raw form

Crank-Git:
## What

The `raw` key of a JA4S result sorted the extensions into numeric order. The FoxIO
`JA4S_r` value holds them in the order the ServerHello carries them. Both raw keys of a
JA4S result now hold that one wire-order value. #108

## Why

FoxIO publishes `JA4S_r` and no `JA4S_ro`, because JA4S sorts no list. The sorted `raw`
value matched 49 of the 84 `JA4S_r` values the 37 committed vectors hold, so a caller who
compared it against another tool got a mismatch on 35 of them.

**This is a breaking change to a published field.** A caller who stored the `raw` value of
a JA4S result before this change gets a different string now. `badcurveball.pcap` reports
`t1205h1_c02b_0000,ff01,000b,0023,0010` where it reported
`t1205h1_c02b_0000,000b,0010,0023,ff01`. The CHANGELOG entry states it, in the form #28
used for the JA4SSH window.

The JA4S fingerprint does not change. It already hashed the extensions in wire order, and
it already equals the reference `JA4S` value.

## The sort rule of each method

The rule of one method proves nothing about another method, so each row below carries its
own evidence. `docs/implementation_notes.md` holds the same table.

| Method | Raw keys FoxIO publishes | Sort rule | Evidence |
|---|---|---|---|
| JA4 | `JA4_r`, `JA4_ro` | `JA4_r` sorts the ciphers and the extensions, and holds the signature algorithms in wire order. `JA4_ro` holds every list in wire order. | 160 `JA4_r` values. All 156 that carry a signature-algorithm section hold the ciphers and the extensions in numeric order, and the signature algorithms in an order that is not numeric. No `JA4_ro` value equals its `JA4_r` value. |
| JA4S | `JA4S_r` | The extensions stay in wire order. | 84 `JA4S_r` values. 35 hold the extensions in an order that is not numeric order. No file carries a `JA4S_ro` key. |
| JA4H | `JA4H_ro` | Not measured. | 89 `JA4H_ro` values and no `JA4H_r` value. `ja4plus` exposes no JA4H raw form. |
| JA4X, JA4SSH, JA4L, JA4T, JA4TS, JA4D, JA4D6 | None | Not applicable. | No expected-output file carries a raw key for these methods. |

**JA4 does not change in this pull request.**

## Parity

`Crank-Git/ja4plus-go` publishes the `Raw` and `RawOriginalOrder` field names on its
result struct, and `ja4plus` keeps both names under parity rule 2. Its `ja4s.go` computes
no raw value, so parity rule 1 decides the value and FoxIO holds the wire order. The port
was read, never built, run, imported or edited.

## How tested

The four tests came first, and all four failed on the test-only commit `7ba540a`.

- `tests/test_ja4s_raw_order.py` — 4 tests. `badcurveball.pcap` is the vector, because its
  extension order is not numeric order. One test sweeps every committed vector: each
  `JA4S_r` value equals the `raw` key of one result, across the 14 vectors that carry
  the key.
- `tests/test_parity.py::test_the_ja4s_raw_forms_match_the_reference` now asserts `raw` as
  well as `raw_original_order`.

Local suite, in the worktree:

```
pytest tests/ -m "not spec_validation"   719 passed, 3 skipped, 93 subtests passed
pytest tests/ -m spec_validation         629 passed, 141 skipped, 69 xfailed
ruff check ja4plus/ tests/               clean
ruff format --check ja4plus/ tests/      86 files already formatted
coverage (both suites)                   78% total, ja4plus/fingerprinters/ja4s.py 84%
                                         78% and 84% on the base, so coverage holds
```

The deviation register holds 69 entries before this change and 69 after it. A raw-form
change moves no entry, because `tests/conformance_index.py:29` drops every raw key before
the comparison. #121 owns that harness gap, and this pull request does not touch it.

Verified against: https://github.com/FoxIO-LLC/ja4/tree/main/python/test/testdata
(FoxIO commit 27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8, as `tests/foxio_vectors/NOTICE`
records, retrieved 2026-08-06)
Verified against: https://github.com/Crank-Git/ja4plus-go (`types.go`, `ja4s.go`,
`processor.go`, default branch `master`, retrieved 2026-08-06)


Crank-Git:
## Self-review

Two Sonnet children reviewed the diff: one for correctness, one for the writing standard and for hostile input.

**Correctness — no defect.** The reviewer confirmed that no other path produces a sorted JA4S raw value, that `_generate_ja4s_raw_from_tls_info` has one caller and it passes no removed argument, that `ja4plus/fingerprinters/ja4.py` is absent from the diff, and that the new tests assert a hardcoded non-numeric order so a returned sort would fail them. It re-measured every number in the prose from the committed vectors and confirmed each one: 84 `JA4S_r` values, 35 not in numeric order, 49 matched before, 160 `JA4_r` values, 156 with a signature-algorithm section, 89 `JA4H_ro` values.

**Hostile input — no defect.** The `try` block and its five caught exception classes are unchanged. The helper returns `None` and does not raise.

### Findings I fixed in 7643a0d

1. `ja4plus/fingerprinters/ja4s.py` — the new docstring named one failure mode, and the helper holds two. The `Returns` section now names the parse failure as well as the missing cipher.
2. `CHANGELOG.md` — two description sentences passed the 25-word limit of `.claude/rules/ste.md`. Split.
3. `docs/implementation_notes.md` — one description sentence passed the limit. Split.
4. `tests/test_ja4s_raw_order.py` — the comment called 29 the count of distinct values. It is the sum of the distinct-value counts of each vector, and the cross-file distinct count is 21. The comment now says so. The assertion is unchanged, and it guards against a vector leaving the set.

### Finding I reject

`tests/test_parity.py:140` — "records the reading" was read as an `-ing` noun under STE rule 14. "reading" here is the project's term for a decision the project made about an ambiguous FoxIO image, and `.claude/rules/conformance.md` and `docs/implementation_notes.md` both use it that way. Renaming it in one file alone would rotate a term, which STE rule 7 forbids.

Issue #130, opened 2026-08-07
Collect the JA4S ServerHello across two QUIC Initial packets


Crank-Git:
## Context

Found while #118 was built. Nothing in #118 changes it, and no vector requires a
change, so `CLAUDE.md` rule 1 keeps it out of that issue.

`ja4plus/fingerprinters/ja4s.py` calls `parse_quic_server_initial` on one server
Initial packet at a time. The function reads the CRYPTO fragments of that packet
alone. A server that splits the ServerHello across two Initial packets therefore
produces no JA4S value, because neither packet holds the whole message.

`ja4plus/fingerprinters/ja4l.py` holds the reader for this case. #102 added
`collect_crypto_fragments` and `server_hello_is_complete`, and the JA4L QUIC path
collects the fragments of several packets before it reads the point.

## Why no vector requires the change

Round 21 of `docs/specs/spec.md` states it: neither vector splits a ServerHello
across two Initial packets. The conformance suite is therefore silent, and the
register holds no entry.

## What to build

1. Collect the CRYPTO fragments of the server Initial packets of one connection,
   as `ja4l.py` does.
2. Read the ServerHello when `server_hello_is_complete` reports a whole message.
3. Give the fragment table a maximum entry count and a maximum age, and drop the
   entry of a connection that `cleanup_connection` names.
4. Run the conformance suite. No vector that passes today may fail.

## Links

- Found by: #118
- The JA4L form: `ja4plus/fingerprinters/ja4l.py`
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Crank-Git:
## Plan — Epic 1 batch 8

**Branch** `issue/130-quic-split-serverhello` off `epic/12-batch8`. The pull request is a
draft into that integration branch and every commit carries `[skip ci]`.

**Approach.** `ja4plus/fingerprinters/ja4l.py` already holds the reader for this case. #102
added `collect_crypto_fragments` and `server_hello_is_complete`. Reuse them rather than
write a second reader.

1. Collect the CRYPTO fragments of the server Initial packets of one connection in
   `ja4plus/fingerprinters/ja4s.py`.
2. Read the ServerHello when `server_hello_is_complete` reports a whole message.
3. Give the fragment table a maximum entry count and a maximum age, and drop the entry of
   a connection that `cleanup_connection` names. `ja4plus/fingerprinters/ja4.py` holds the
[...]

Issue #13, opened 2026-08-06
Epic 2: Correctness audit


Crank-Git:
## Goal

Remove the defects that produce a wrong fingerprint, or a crash, on input the vectors do not cover.

## Exit criteria

The malformed-input suite passes. TCP sequence wraparound is handled. JA4L emits one client fingerprint per connection. JA4H handles a repeated cookie name.

## Spec

- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Feature: [`docs/specs/features/02-correctness-audit.md`](docs/specs/features/02-correctness-audit.md)

Depends on: #12

## Sub-issues

- [x] #32
- [x] #33
- [x] #34
- [x] #35
- [x] #36
- [x] #37

Crank-Git:
**This issue grew from 21 register entries to 24.**

#118 made JA4S read a QUIC server Initial packet. JA4S now emits on nine QUIC streams the
reference holds no value for, and they are the exact nine streams on which `ja4plus`
already emits a JA4 value the reference does not hold. The three new entries are
`chrome-cloudflare-quic-with-secrets.pcapng/JA4S`, `ssh2.pcapng/JA4S` and
`tls3.pcapng/JA4S`.

**The question this issue must answer is now larger and clearer:** does `ja4plus` emit a
fingerprint on a QUIC stream that the FoxIO reference ignores, or does the reference miss
it? Read both methods together. A rule that resolves one resolves the other.

Crank-Git:
## This epic now owns 73 register entries, and it has no sub-issue that owns any of them

`tests/foxio_deviations.json` holds 226 entries. **73 of them name #13**, which is an
epic. An epic is not schedulable, so no worker can be assigned to any of those 73, and
the register points at a container rather than at work.

The 73 split into two questions, and they are not the same question:

1. **Does `ja4plus` emit a fingerprint on a QUIC stream that the reference ignores, or
   does the reference miss it?** This covers the JA4, JA4S and raw-form entries on
   `chrome-cloudflare-quic-with-secrets.pcapng`, `ssh2.pcapng` and `tls3.pcapng`. #118
   widened it from JA4 alone to JA4 and JA4S, and #121 widened it again to the raw forms.
   One rule resolves every one of them.
2. **The remaining hashed-form deviations** that predate this session.

**Decompose this epic before the next batch reads it.** Each group needs a sub-issue that
a worker can be assigned to, and every register entry needs to name that sub-issue rather
than #13.

[...]

Issue #39, opened 2026-08-06
Move every fingerprinter onto the bounded state table


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/03-concurrency-safety.md`](docs/specs/features/03-concurrency-safety.md):

> FR-concurrency-safety-7 — Every state table holds no more than a maximum entry count.

> FR-concurrency-safety-8 — Every state table evicts an entry that has not been read for longer than a maximum age.

## What to build

The stateful fingerprinters are JA4 (QUIC CRYPTO fragments keyed by connection identifier), JA4H and JA4X (reassembled TCP segments), JA4L (handshake timestamps and TTL values) and JA4SSH (packet lengths, bare-ACK counts and banner strings).

`ja4plus/fingerprinters/ja4.py:299` holds `_quic_fragments` and `_quic_dcid_to_tuple`, which grow without limit when a QUIC handshake never completes. A flood of QUIC Initials with distinct connection identifiers is an unbounded allocation today.

`cleanup_connection` stays. It is the caller's way to evict early, and the port has the same method.

## Acceptance criteria

- [ ] Feeding 1000000 packets across 100000 distinct connections holds resident memory below the stated ceiling.
- [ ] Every state table reports an entry count no greater than `max_connections`.
- [ ] The conformance suite still passes on all 37 vectors.

## Links

- Feature: [`docs/specs/features/03-concurrency-safety.md`](docs/specs/features/03-concurrency-safety.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #14
Depends on #38

Crank-Git:
#92 makes the JA4SSH state table cheaper to fill, and the bound this issue adds is what
limits it.

`JA4SSHFingerprinter.connections` created an entry only for a packet that carried SSH
data. `ja4plus/utils/ssh_utils.py` needs at least four bytes that match an SSH banner or
a plausible length field before it reports one. FoxIO holds an entry for every packet
whose source port or destination port is 22, and it counts the bare ACK that completes
the TCP handshake. #92 adopts that reading, because three vectors each report one client
ACK too few without it.

The cost of one entry therefore drops from one SSH-shaped payload to one packet that
carries the ACK flag alone and no payload. Port 22 carries a high volume of scan
[...]

Issue #38, opened 2026-08-06
Add a bounded state table with count and age limits


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/03-concurrency-safety.md`](docs/specs/features/03-concurrency-safety.md):

> FR-concurrency-safety-7 — Every state table holds no more than a maximum entry count.

> FR-concurrency-safety-8 — Every state table evicts an entry that has not been read for longer than a maximum age.

> FR-concurrency-safety-9 — The maximum entry count and the maximum age are constructor arguments.

> FR-concurrency-safety-10 — A state table evicts the least recently used entry when it reaches its maximum entry count.

## What to build

New file `ja4plus/utils/state_table.py`. This is the shared structure every stateful fingerprinter moves onto in the next issue.

Eviction runs on packet arrival, not on a timer. The library starts no thread.

Eviction by age uses the packet timestamp when the packet carries one, and the wall clock otherwise. A capture file replays faster than real time, and a wall clock would evict entries that the capture still needs.

An eviction pass runs at most once per 1000 packets, so eviction cost stays proportional to traffic. Defaults: 10000 entries, 300 seconds. Measure a baseline and state the memory ceiling in the pull request.

## Acceptance criteria

- [ ] A state table holds no more than its maximum entry count.
- [ ] An entry that receives no packet for longer than the maximum age is absent afterwards.
- [ ] The least recently used entry is evicted when the table reaches its maximum count.
- [ ] `max_connections` of zero or a negative number raises `ValueError`.

## Links

- Feature: [`docs/specs/features/03-concurrency-safety.md`](docs/specs/features/03-concurrency-safety.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #14

Crank-Git:
## Plan — Epic 3, integration branch `epic/14-concurrency-and-resource-safety`

**Changelog round 81.** It is assigned. Do not choose one, and never write `TBD`.

**This is the first issue of Epic 3 and it gates the rest.** #39 moves every fingerprinter
onto the structure you build, and #40, #42 and #43 build on that. **A shape that is wrong
here is repaired six times later.** Nothing else of Epic 3 is scheduled until this
sub-merges.

### What the repository already knows about these bounds

[...]

PR #133, opened 2026-08-07
Tell a QUIC server Initial packet from a client one in JA4S

Crank-Git:
Refs #118.

## What

`ja4plus/fingerprinters/ja4s.py` read every QUIC Initial packet as a client Initial
packet. `_is_quic_client_initial` tested the long-header packet type alone, and a
server Initial packet carries the same type. The fingerprinter stored the connection
ID of the server packet under the reverse connection key and returned, so it never
reached the server path.

JA4S now reads the port that names the server, as `ja4l.py` does. A server Initial
packet replaces no stored client connection ID.

## Why

A server Initial packet names the client with the connection ID the client chose as
its source. The stored value was therefore empty, and the server path never ran. The
measurement on #118 states it:

```
dcids {"2001:db8:1::1:50280-2606:4700:10::6816:826:443": "203f9e9f68698274",
       "2606:4700:10::6816:826:443-2001:db8:1::1:50280": ""}
```

No vector reports this defect, because the project holds no JA4S QUIC vector. A unit
test is the gate.

## How it is tested

`tests/test_ja4s_quic_direction.py` holds nine tests. They were written first and
they failed against the base: six of the nine failed, and the other three cover
paths the change keeps.

The expected value comes from the FoxIO material.
`chrome-cloudflare-quic-with-secrets.pcapng` holds `JA4S`
`t130200_1301_234ea6891581` with `JA4S_r` `t130200_1301_0033,002b`. The synthetic
ServerHello carries the same TLS version, the same cipher and the same two
extensions in the same order, so only the transport letter differs. The test expects
`q130200_1301_234ea6891581`, and the same value comes out of the real capture.

Local checks, all green:

```
pytest tests/ -m "not spec_validation"   761 passed, 3 skipped (752 on the base)
pytest tests/ -m spec_validation         628 passed, 141 skipped, 70 xfailed
ruff format --check ja4plus/ tests/      90 files already formatted
python -m build                          Successfully built ja4plus-0.6.0-py3-none-any.whl
coverage                                 79% total, floor 70; ja4plus/fingerprinters/ja4s.py 88%
```

## One decision to confirm at the merge gate

**No JA4S value moved.** Every JA4S value case still passes, and no vector reports a
missing occurrence key.

The fix does make JA4S emit on nine QUIC streams that the reference holds no value
for:

- `chrome-cloudflare-quic-with-secrets.pcapng` stream 0:50280
- `ssh2.pcapng` streams 33 and 36
- `tls3.pcapng` streams 21, 22, 23, 24, 25 and 28

The FoxIO reference reads no QUIC handshake in the committed vectors. It holds no
JA4 value for the same nine streams, and `tests/foxio_deviations.json` already
registered that divergence for JA4 under #13 on all three vectors. Three entries now
record the JA4S form of it, with the same issue and the same cause. The register
rises from 67 entries to 70, and the conformance suite is green.

## Out of scope, filed

#130: `ja4s.py` reads one server Initial packet at a time, so a ServerHello that
spans two Initial packets produces no JA4S value. Round 21 of `docs/specs/spec.md`
records that no vector splits a ServerHello, so no vector requires the change.

## Documentation

- `CHANGELOG.md` holds the entry.
- `docs/specs/features/01-spec-conformance.md` gains `FR-spec-conformance-26` and
  `FR-spec-conformance-27`, plus the behaviour rules.
- `docs/specs/spec.md` holds round 23.

Verified against: https://www.rfc-editor.org/rfc/rfc9000.html (RFC 9000 Sections 7.2
and 17.2, retrieved 2026-08-07) and
https://github.com/FoxIO-LLC/ja4/tree/main/python/test/testdata (the pinned FoxIO
vectors, as committed under `tests/foxio_vectors/`).


Crank-Git:
## Self-review

A review by lens found three items. One is rejected, and two are context.

**Rejected — `if dcid:` drops a zero-length client connection ID.** The review read
`ja4plus/fingerprinters/ja4s.py:110` and stated that the deleted `_extract_dcid`
caller used `if dcid is not None:`, so an empty connection ID no longer reaches the
table. The change is deliberate and it costs nothing:

1. RFC 9000 Section 7.2 states it: "This Destination Connection ID MUST be at least 8
   bytes in length." A client Initial packet with an empty connection ID is invalid.
2. `decrypt_quic_server_initial_crypto` returns None for an empty `client_dcid`
   (`ja4plus/utils/quic_utils.py:417`), so a stored empty value produces no
   fingerprint on any path.
3. `ja4l.py:469` holds the same test, so the two QUIC readers agree.

The empty value is exactly what the defect on #118 stored, so the test
`test_the_fingerprinter_reads_no_server_initial_packet_before_the_client_one`
asserts that the table stays empty.

**Context — a QUIC server on a port other than 443 gives no JA4S value.** The port
names the direction, and `ja4l.py` reads it the same way. JA4S produced no QUIC
server value at all before this change, so no behaviour regresses. A capture of a
QUIC server on another port is out of scope here.

**Context — `_quic_dcids` has no maximum entry count and no maximum age.** The table
predates this change, and #38 and #39 own it. This change reduces what reaches the
table: only a client Initial packet writes to it now.

Verified against: https://www.rfc-editor.org/rfc/rfc9000.txt (RFC 9000 Section 7.2,
retrieved 2026-08-07).

Issue #123, opened 2026-08-07
Restart the JA4L QUIC measurement points when a connection repeats


Crank-Git:
Found by the self-review of #102. The behaviour predates that issue and is identical
before and after it, so `CLAUDE.md` rule 1 keeps it out.

## The gap

`_quic_ja4l` in `ja4plus/fingerprinters/ja4l.py` guards the two server points with
`"A" not in timestamps` and `"B" in timestamps`. Nothing clears them.

1. A second QUIC handshake over the same address pair and port pair reads the
   measurement points of the first one. The TCP path handles this: `_tcp_ja4l` calls
   `_restart_connection` when a repeated SYN carries another initial sequence number.
   The QUIC path has no equivalent.
2. A server Initial packet that arrives before any client Initial packet is dropped
   for the life of the connection. The TCP path handles the mirrored case: a SYN that
   arrives after its SYN-ACK still completes the server value.

#102 added `conn["quic_dcid"]`, which is the natural signal for the first case. A
client Initial packet that carries another connection ID, and that follows no Retry
packet, opens another connection.

## Why no vector requires the change

Every QUIC case in the vector set passes without it. Whoever builds this issue finds a
capture that holds the case, or builds one, and measures the reference first.

## Links

- Found by the self-review of #102.

Crank-Git:
## Plan — Epic 1 batch 9

Integration branch `epic/12-batch9`, forked from `dev` at `b7c1140`. The pull request is a
draft into that branch, and every head commit carries `[skip ci]`.

### Approach

1. **Measure before you change anything.** Rule 1 of `CLAUDE.md` binds this issue: no
   fingerprint moves without a vector. Read `_quic_ja4l` and `_restart_connection` in
   `ja4plus/fingerprinters/ja4l.py` and state the two cases as failing tests first.
2. **Find or build the capture.** Search `tests/foxio_vectors/` for a capture that holds a
   repeated QUIC connection over one address pair and port pair, and for one that holds a
   server Initial packet before its client Initial packet. No vector holds either case,
   which the issue already records — then build a synthetic capture with `scapy` under
   `tests/`, and say in the pull request that it is synthetic.
3. **Mirror the TCP path.** `_tcp_ja4l` already answers both cases. Restart on a client
   Initial packet that carries another `conn["quic_dcid"]` and that follows no Retry packet.
   Complete the server value on a mirrored arrival, as the TCP path does for a SYN that
   follows its SYN-ACK.
[...]

PR #154, opened 2026-08-07
Hold the JA4L QUIC measurement points against the reference

Crank-Git:
Refs #123.

## Verdict: the reference rejects both changes this issue asks for

#123 asks for two changes to `_quic_ja4l`:

1. Restart the measurement points when a client repeats the handshake over one address
   pair and port pair.
2. Complete the server value when a server Initial packet arrives before its client
   Initial packet.

`CLAUDE.md` rule 1 binds this issue: a fingerprint changes only when a FoxIO vector
requires it. No vector holds either case, which the issue records. So I measured the
reference instead of reading it.

The FoxIO Rust implementation is the only FoxIO implementation that reads a QUIC
handshake, and `.claude/rules/external-apis.md` makes it decide where the Python file
holds no value for a stream. I built it at the pinned commit
`27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8` and ran it on two synthetic captures.

I first confirmed the binary reproduces a committed vector, so the harness is sound:

```
$ ja4 tests/foxio_vectors/chrome-cloudflare-quic-with-secrets.pcapng
  ja4l_c: 113_64
  ja4l_s: 9285_56
```

That matches `tests/foxio_vectors/rust_expected/ja4__insta@chrome-cloudflare-quic-with-secrets.pcapng.snap` exactly.

### Case 1 — a repeated QUIC handshake

The capture holds two complete four-packet handshakes over one address pair and port
pair, 1 second apart, with different connection IDs.

```
$ ja4 quic_repeat.pcap
- stream: 0
  transport: udp
  src: 10.0.0.1
  dst: 10.0.0.2
  src_port: 44444
  dst_port: 443
  ja4s: q000000_0000_000000000000
  ja4l_c: 500_64
  ja4l_s: 5000_56

$ ja4plus   ->  ['JA4L-S=5000_56', 'JA4L-C=500_64']
```

The two readers agree. The reference reports one value pair, and it reads the first
handshake. It does not restart. The restart this issue asks for would make `ja4plus`
report a second value pair that no FoxIO implementation reports.

### Case 2 — a server Initial packet before its client Initial packet

```
$ ja4 quic_mirrored.pcap
[]

$ ja4plus   ->  ['JA4L-C=500_64']
```

The reference reports nothing at all. It discards the leading server Initial packet,
exactly as `ja4plus` does, and it then never reaches its terminal state. Completing the
server value would move `ja4plus` further from the reference, not closer.

`rust/ja4/src/time/udp.rs` states the mechanism, and the measurement confirms it. The
state machine holds a terminal `Done` state that ignores every later packet. Its
`Handshake` state discards a later `ClientInitial`, and its first state discards a
`ServerInitial`.

## What this pull request contains

No fingerprinter changes. No fingerprint moves.

- `tests/test_ja4l_quic_repeated_connection.py` — two tests that hold the measured
  reference behaviour, so a later restart cannot land without a vector.
- `tests/quic_builder.py` — a `handshake_packet` builder the tests need.
- `docs/implementation_notes.md` — the reading, under JA4L.

I proved the first test binds. With the restart the issue describes applied to
`_quic_client_initial`, it fails:

```
E       AssertionError: assert ['JA4L-S=5000...A4L-C=500_64'] == ['JA4L-S=5000...A4L-C=500_64']
E         Left contains 2 more items, first extra item: 'JA4L-S=5000_56'
```

## The open question

Case 2 leaves one real divergence, and it is not the one the issue names. On a capture
whose server Initial packet leads, the reference reports no value, and `ja4plus`
reports `JA4L-C=500_64`. `ja4plus` emits each value as it reads it, and the reference
emits a value pair only when it collects all four points. This pull request does not
change that, because suppressing a partial value is a wider decision than #123 covers.

## Gates

```
pytest tests/ -m "not spec_validation"   919 passed, 8 xfailed, 93 subtests passed
pytest tests/ -m spec_validation         1342 passed, 141 skipped, 103 xfailed
ruff format --check ja4plus/ tests/      97 files already formatted
```

The register invariant holds. `pytest tests/ -m spec_validation` reports `103 xfailed`,
and `tests/foxio_deviations.json` holds 103 keys. Neither moved.

Verified against: https://github.com/FoxIO-LLC/ja4/blob/27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8/rust/ja4/src/time/udp.rs (retrieved 2026-08-07)

### DIFF
--- a/docs/implementation_notes.md
+++ b/docs/implementation_notes.md
@@ -505,14 +505,20 @@ discards that packet, and it reports no value at all for the flow.
 `ClientInitial`, and its first state discards a `ServerInitial`.
 
 The FoxIO Rust implementation is the only FoxIO implementation that reads a QUIC
-handshake, so it decides both cases. No FoxIO vector holds either case. Two synthetic
-captures measure them, and `tests/test_ja4l_quic_repeated_connection.py` holds the
-measurement:
+handshake, so it decides both cases. No FoxIO vector holds either case.
+`tests/build_quic_ja4l_captures.py` writes the two synthetic captures that measure
+them. The reference ran at the commit `tests/download_test_vectors.py` pins:
 
 ```
-ja4 quic_repeat.pcap     -> ja4l_c: 500_64   ja4l_s: 5000_56
-ja4 quic_mirrored.pcap   -> []
+$ ja4 quic_repeat.pcap
+  ja4l_c: 500_64
+  ja4l_s: 5000_56
+$ ja4 quic_mirrored.pcap
+[]
 ```
+
+`tests/test_ja4l_quic_repeated_connection.py` reads the same packets from the same
+builder, so the tests and the measurement cover one capture each.
 
 `ja4plus` reports the same value pair for the first capture. For the second capture it
 reports no server value, and it reports `JA4L-C=500_64` where the reference reports
--- a/docs/implementation_notes.md
+++ b/docs/implementation_notes.md
@@ -491,6 +491,33 @@ The reference reads four QUIC packets, and it reads the direction from port 443:
 `JA4L-S` is half the time from `A` to `B`, and `JA4L-C` is half the time from `C`
 to `D`.
 
+### The QUIC measurement points never restart
+
+A client repeats a QUIC handshake over one address pair and port pair. The reference
+reads the first handshake, and it reports one value pair for the whole flow. It reads
+no later handshake.
+
+A server Initial packet arrives before its client Initial packet. The reference
+discards that packet, and it reports no value at all for the flow.
+
+`rust/ja4/src/time/udp.rs` states the mechanism. The state machine holds a terminal
+`Done` state that ignores every later packet. Its `Handshake` state discards a later
+`ClientInitial`, and its first state discards a `ServerInitial`.
+
+The FoxIO Rust implementation is the only FoxIO implementation that reads a QUIC
+handshake, so it decides both cases. No FoxIO vector holds either case. Two synthetic
+captures measure them, and `tests/test_ja4l_quic_repeated_connection.py` holds the
+measurement:
+
+```
+ja4 quic_repeat.pcap     -> ja4l_c: 500_64   ja4l_s: 5000_56
+ja4 quic_mirrored.pcap   -> []
+```
+
+`ja4plus` reports the same value pair for the first capture. For the second capture it
+reports no server value, and it reports `JA4L-C=500_64` where the reference reports
+nothing. #123 carries that open reading.
+
 ### A time of one second or more
 
 The FoxIO program reads the difference of two timestamps as a `timedelta`, and it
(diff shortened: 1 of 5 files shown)


Crank-Git:
## Self-review

I reviewed the diff by correctness lens and by writing-standard lens. Five findings, all
addressed in `82fb97b`.

**1. Blocking — the measurement was not reproducible.** The two captures existed only in
`/tmp` on my machine. `.claude/rules/conformance.md` asks for the command and its output,
and a reader could run neither. `tests/build_quic_ja4l_captures.py` now writes both
captures, and the test module reads the same packets from the same builder. I re-ran the
reference against the captures the committed builder writes:

```
$ ja4 /tmp/ja4l123/quic_repeat.pcap
  ja4l_c: 500_64
  ja4l_s: 5000_56
$ ja4 /tmp/ja4l123/quic_mirrored.pcap
[]
```

Same values. The tests and the measurement now cover one capture each.

**2. Blocking — the second test hid the divergence.** It asserted only that no `JA4L-S=`
value appears. Two independent guards in `_quic_server_initial` produce that result, so
the assertion could not tell them apart, and it hid the `JA4L-C=500_64` value that
diverges from the reference. The test now states the whole reported value list, and its
docstring names #123 as the open reading.

**3. Should-fix — two sentences exceeded the 25-word limit.** The module docstring opener
conjoined both cases, and the `handshake_packet` docstring conjoined three clauses. Both
are split.

**4. Nit — `Handshake packet` was not in the `## Terms` table.** It now sits beside
`Initial packet` and `Retry packet`, with the RFC 9000 section.

**5. Nit — the doc and the test named the captures with different paths.** Both now use
the bare file names, and the doc points at the builder.

### Rejected findings

None.

### Checked and clean

- `handshake_packet()` is a correct QUIC v1 long-header Handshake packet per RFC 9000
  Section 17.2. The first byte is `0xE0`, the packet-number length bits are `00` and one
  packet-number byte follows, and the Length varint covers the packet number plus the
  payload. A Handshake packet carries no Token field, unlike an Initial packet.
  `long_header_packet_type` returns `QUIC_HANDSHAKE` for it, and tshark agrees:

  ```
  $ tshark -r quic_repeat.pcap -T fields -e frame.number -e ip.src -e quic.long.packet_type
  1  10.0.0.1  0
  2  10.0.0.2  0
  3  10.0.0.2  2
  4  10.0.0.1  2
  5  10.0.0.1  0
  6  10.0.0.2  0
  7  10.0.0.2  2
  8  10.0.0.1  2
  ```

- The float timestamps are exact. `1000.010`, `1001.010` and `2000.012` all pass through
  `int(round(seconds * 1000000))` without crossing a rounding boundary, so `5000` and
  `500` are deterministic.

- The first test binds. With the restart #123 describes applied to
  `_quic_client_initial`, it fails:

  ```
  E       AssertionError: assert ['JA4L-S=5000...A4L-C=500_64'] == ['JA4L-S=5000...A4L-C=500_64']
  E         Left contains 2 more items, first extra item: 'JA4L-S=5000_56'
  ```

### Gates after the fixes

```
pytest tests/ -m "not spec_validation"   919 passed, 8 xfailed, 93 subtests passed
pytest tests/ -m spec_validation         1342 passed, 141 skipped, 103 xfailed
ruff format --check ja4plus/ tests/      98 files already formatted
```

`tests/foxio_deviations.json` still holds 103 keys, and the suite still reports
`103 xfailed`. No fingerprint value moved, because no fingerprinter changed.

Issue #132, opened 2026-08-07
Settle the JA4_o extension hash when only SNI is present


Crank-Git:
## What is wrong

Four `JA4_o` values differ from the reference, and the `JA4` value of the same stream
matches. The four streams are:

- `https3-301-get.pcap` stream 0, source port 62599
- `socks-https-example.pcap` stream 0, source port 53554
- `socks-https-example.pcap` stream 2, source port 53555
- `socks-https-example.pcap` stream 4, source port 53556

Each one carries one extension, and that extension is SNI (`0x0000`).

```
JA4.1    = t10d230100_6a57a6f57151_000000000000
JA4_r.1  = t10d230100_0003,...,00ff_
JA4_o.1  = t10d230100_ce175d585f73_000000000000
JA4_ro.1 = t10d230100_0039,...,00ff_0000
```

`ja4plus` produces `t10d230100_ce175d585f73_9af15b336e6a`. The cipher hash matches. The
extension hash does not.

The reference `JA4_ro` extension field holds `0000`, and the reference `JA4_o` extension
hash holds `000000000000`. The reference therefore does not hash its own `JA4_ro`
extension field on these four streams. `ja4plus` does hash it, which is the rule that
matches the other 156 values.

#121 made the comparison run for the first time. The other 11 `JA4_o` failures mirror the
11 `JA4` failures on the same streams, and #13 owns those.

## Why it matters

`JA4_o` is a fingerprint. A value that matches 156 of 160 reference values and differs on
four is worse than a value that always differs, because the four look usable.

## What to build

1. Prove the mechanism before you change any code. Instrument the FoxIO Python
   implementation at the pinned commit, or measure the four client hellos directly, and
   put the command and its output in this issue.
2. If FoxIO removes SNI and ALPN from the `JA4_o` extension hash, state why the other 156
   values still match.
3. If the reading is a proven FoxIO defect, apply `.claude/rules/conformance.md`. Record
   the decision, and point the four register entries at it.
4. Record the reading in `docs/implementation_notes.md`, under the section
   `The original-order hashed value`.

Do not change `generate_ja4` before step 1 gives an answer.

## Acceptance criteria

- [ ] The issue holds a command and its output that prove what the reference computes.
- [ ] `docs/implementation_notes.md` records the reading and names the vector.
- [ ] The four register entries either disappear, or they name the decision this issue
      records.
- [ ] The 156 `JA4_o` values that match today still match.

The four register entries are:

```
[...]

PR #155, opened 2026-08-07
Read the sorted extension list for the JA4_o zero sentinel

Crank-Git:
## What

`JA4_o` reads the sorted extension list for the zero sentinel. A client hello whose only
extensions are SNI (`0x0000`) and ALPN (`0x0010`) now gives the extension hash
`000000000000`, and `JA4_ro` still shows the extension in wire order.

Refs #132.

## Why

The issue named four `JA4_o` values that differ from the reference. Step 1 of the plan
measured the reference before any code changed. #132 holds the command and the full
output.

`python/ja4.py` line 248 holds one guard for two hashes:

```python
    if x['sorted_extensions']:
        sorted_extensions = sha_encode(x['sorted_extensions'])
        original_extensions = sha_encode(x['original_extensions'])
    else:
        sorted_extensions = '000000000000'
        original_extensions = '000000000000'
```

`python/common.py` line 144 removes SNI and ALPN from the sorted string alone. The guard
reads the sorted string, and it selects the zero sentinel for both hashes.

A probe on the reference at the pinned commit reads both strings:

```
$ python foxio/ja4.py tests/foxio_vectors/https3-301-get.pcap --ja4 -J -r -o -s 0
PROBE stream=0 sorted_extensions='' original_extensions='0000' guard=bool(sorted_extensions)=False sha_encode(original_extensions)=9af15b336e6a
    "JA4_o.1": "t10d230100_ce175d585f73_000000000000",
    "JA4_ro.1": "t10d230100_0039,...,00ff_0000"
```

The reference computes `sha256('0000')[:12] = 9af15b336e6a`, and then it discards that
value. `9af15b336e6a` is the value `ja4plus` emitted before this change.

**Why the other 156 values still match.** The guard fails only when the sorted string is
empty, and the sorted string is empty only when every extension is SNI or ALPN. On the
other 156 values the guard passes, and the reference hashes the original-order string.

**This reading is a rule, not a defect.** `technical_details/JA4.md` states it:

> If there are no extensions in the sorted extensions list, then the value of JA4_c is
> set to `000000000000`

The specification keys the zero sentinel on the sorted extensions list, and it names one
field. It describes `JA4_o` by example alone, and it states no separate rule for the
original-order extension hash. `.claude/rules/conformance.md` therefore does not apply,
because that rule covers a proven defect. No entry joins the declined set.

## How

`ja4plus/fingerprinters/ja4.py` now builds the sorted extension string on both paths, and
it guards the hash on that string. `get_raw_fingerprint` is unchanged, because `JA4_ro`
already matched the reference.

The four register entries disappear. `tests/foxio_deviations.json` falls from 103 keys to
99, and `tests/foxio_deviation_owners.json` drops the `132` owner that nothing names now.

## How tested

Test first, then the code. Two tests failed against the old behaviour:

```
FAILED tests/test_ja4_empty_ext.py::test_ja4_original_order_yields_the_zero_hash_when_only_sni_is_present
FAILED tests/test_ja4_original_order.py::test_ja4_gives_the_zero_sentinel_when_sni_is_the_only_extension
```

`tests/test_ja4_original_order.py` names the vector `https3-301-get.pcap` and the expected
value `t10d230100_ce175d585f73_000000000000`.

The five gates:

```
.venv/bin/pytest tests/ -m "not spec_validation"   921 passed, 8 xfailed, 93 subtests passed
.venv/bin/pytest tests/ -m spec_validation         1346 passed, 141 skipped, 99 xfailed
.venv/bin/ruff format --check ja4plus/ tests/      96 files already formatted
```

Baseline before the change: `1342 passed, 141 skipped, 103 xfailed`.

**The 156 held.** A per-value tally of the 160 `JA4_o` cases, before and after:

```
before:  156 PASSED    4 XFAIL
after:   160 PASSED
```

**The register invariant holds.** 99 keys in `tests/foxio_deviations.json`, and 99
`xfailed`. The count moved from 103 because the four entries this issue names disappear.

Follow-up commit `744c7bd` adds `zero sentinel` to the `## Terms` table of
`docs/specs/spec.md`, because the first commit coined a term the project does not
hold. The five gates run again after it, all green.

Coverage of `ja4plus/fingerprinters/ja4.py` is 94%, and the project total is 80%. The
change adds two statements, and the suite runs both.

Verified against: https://github.com/FoxIO-LLC/ja4/blob/27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8/python/ja4.py (retrieved 2026-08-07)
Verified against: https://github.com/FoxIO-LLC/ja4/blob/27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8/python/common.py (retrieved 2026-08-07)
Verified against: https://github.com/FoxIO-LLC/ja4/blob/27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8/technical_details/JA4.md (retrieved 2026-08-07)


Crank-Git:
## Self-review

Two lenses ran over the diff: correctness against the FoxIO reference logic, and the
project rules in `.claude/rules/ste.md` and `.claude/rules/conformance.md`.

### Correctness — no finding

The reviewer traced four cases and found no divergence from the reference.

- GREASE filtering happens once, at `ja4plus/fingerprinters/ja4.py:123`, before both
  derivations. The reference filters GREASE inside `get_hex_sorted` for both. Same point,
  same effect.
- The signature-algorithm suffix goes on both strings, at
  `ja4plus/fingerprinters/ja4.py:166` and `:167`. The reference appends it to both at
  `python/ja4.py:245` and `:246`.
- `sorted_extensions` is a subset of `extensions`, because the filter only removes. An
  empty `ext_str` therefore forces an empty `sorted_ext_str`, so no input reaches the
  guard from the other direction.
- When `original_order` is False, `ext_str` and `sorted_ext_str` hold the same string. The
  `JA4` value cannot move, and the conformance suite confirms it.

The edge case of an empty extension list with signature algorithms present gives
`sorted_ext_str = "_<sigalgs>"`, which is true, so both implementations hash. No physical
client hello reaches that state, because the signature algorithms come from extension
`0x000d`, and `0x000d` survives the SNI and ALPN filter.

### Writing standard — three findings, two fixed

**1. Fixed. The first commit coined a term the project does not hold.** The commit used
"zero marker" in five files, and it never added the word to the `## Terms` table.
`.claude/rules/ste.md` requires the addition.

The finding is worse than the lens reported. The repository already calls the value a
**sentinel**, at `ja4plus/fingerprinters/ja4x.py:64`, `tests/test_ja4x_deep.py:246` and
`tests/test_comprehensive.py:996`. "zero marker" therefore rotated a synonym, which rule 7
forbids. `marker` also sits in the "Do not use" column of `measurement point` at
`docs/specs/spec.md:79`.

`744c7bd` adds `zero sentinel` to the `## Terms` table, and it renames every use. One test
name changed with it.

**2. Fixed. Two sentences ran long.** The changelog row held a 38-word sentence, and
`docs/implementation_notes.md` held a 26-word sentence. Rule 1 caps a description sentence
at 25 words. `744c7bd` splits both.

**3. Rejected. The multi-sentence code comment stands.** The lens read "A code comment.
One sentence, one fact" as a cap of one sentence per comment. It is a cap of one fact per
sentence. The standard's own good example is two sentences:

```python
# FoxIO emits one fingerprint per 200 SSH packets. A smaller window produces a
# fingerprint that no reference implementation will match.
SSH_WINDOW = 200
```

Rule 3 sets the real cap, at six sentences per paragraph. The comment holds four.

### After the fixes

The five gates run again, all green. The failing-test evidence in the pull-request body
now carries the renamed test:

```
FAILED tests/test_ja4_empty_ext.py::test_ja4_original_order_yields_the_zero_hash_when_only_sni_is_present
FAILED tests/test_ja4_original_order.py::test_ja4_gives_the_zero_sentinel_when_sni_is_the_only_extension
```

Issue #151, opened 2026-08-07
Read the JA4S ServerHello on three streams the FoxIO Rust implementation holds a value for


Crank-Git:
## What is wrong

`ja4plus` produces no JA4S value on three streams that the FoxIO Rust implementation holds
a value for. This is the opposite direction from #138, which records the streams where
`ja4plus` produces more than the reference.

## Evidence

Measured by the #138 parity work, against FoxIO commit
`27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8`.

| Capture | Stream | FoxIO Rust JA4S | `ja4plus` |
|---|---|---|---|
| `ssh2.pcapng` | `57374 <-> 52.178.17.3:443` | `t120300_c030_09f674154ab3` | none |
| `ssh2.pcapng` | `57375 <-> 204.79.197.220:443` | `t1205h2_c030_015e35fdd027` | none |
| `tls-handshake.pcapng` | `50167 <-> 40.126.24.84:443` | `t120400_c030_4e8089b08790` | none |

All three carry TLS over TCP on port 443.

## Why the suite is silent

The FoxIO Python material omits these streams too, and the conformance suite compares
against that material. The register therefore holds no entry, and no test fails today.
`tests/test_foxio_rust_parity.py`, which #138 added, is where the measurement lives.

## What to build

1. Read one of the three streams and state why `ja4plus` reads no ServerHello from it.
   The reason is the deliverable. Do not change a fingerprinter before the reason is known.
2. If `ja4plus` cannot read a ServerHello that the FoxIO Rust implementation reads, that is
   a defect in `ja4plus`, and a vector now proves it. Fix it.
3. If the three streams carry no readable ServerHello, and the Rust value comes from
   another source, record that reading in `docs/implementation_notes.md` and add the
   register entries.
4. Extend `tests/test_foxio_rust_parity.py` to cover the three streams either way.

## Acceptance criteria

- [ ] The issue states, with a measurement, why `ja4plus` reads no ServerHello on each of
      the three streams.
- [ ] Each stream either produces the FoxIO Rust value, or carries a register entry with a
      cause the evidence supports.
- [ ] `tests/test_foxio_rust_parity.py` covers all three streams.
- [ ] The conformance suite passes, and the `xfailed` count still equals the key count of
      `tests/foxio_deviations.json`.

Found by: #138. Part of #12.

Crank-Git:
## Plan — Epic 1 batch 10

[...]

Issue #141, opened 2026-08-07
Settle the JA4 ALPN trigger condition that no vector separates


Crank-Git:
Found by the self-review of #127. The value written is settled. The condition that triggers it is not.

## The gap

Two rules decide when a JA4 ALPN field reads `99`, and the vector set separates neither from the other.

| Source | The condition |
|---|---|
| FoxIO `technical_details/JA4.md` | The first byte or the last byte of the first ALPN value is outside `0x30-0x39`, `0x41-0x5A` and `0x61-0x7A` |
| FoxIO `python/ja4.py` | `ord(alpn[0]) > 127`, tested on the value after it is truncated to two characters |
| `ja4plus` after #127 | The prose condition, applied to both bytes |

#127 settled the **value**: `ja4plus` writes `99`, because two FoxIO implementations write `99`. It did not settle the **condition**, because the only vector that reaches the branch, `tls-non-ascii-alpn.pcapng`, holds `0xba 0xad`, and every one of the three rules fires on it.

## Where they disagree

A first ALPN value of `b"h\x20"`. The last byte is a space, which is not alphanumeric and is below 128.

- The prose condition and `ja4plus` write `99`.
- `python/ja4.py` writes `h `, because `ord("h")` is 104.

They also disagree on any value whose last byte is non-ASCII and whose first byte is alphanumeric.

## Why it matters

This is the shape this project has now found four times: **a comparison that is never made reads as a comparison that passes.** #80, #109 and #121 were each found by reading material by hand, never by a failure. This one was found the same way. The vector set is silent, so the suite is green either way, and the two tools will disagree on live traffic that no capture in the set holds.

## What to build

1. Build a capture, or synthesise a ClientHello, for each case the three rules separate. At minimum `b"h\x20"` and a value whose last byte alone is non-ASCII.
2. Run the FoxIO Python implementation at the pinned commit against it and record the output. Run the FoxIO Rust implementation too, because #127 showed the Rust snapshot agreeing with Python against the prose.
3. Adopt the condition the implementations produce, or record the divergence under `.claude/rules/conformance.md`. `CLAUDE.md` rule 1 holds: a measurement decides, never a reading of the source.
4. Add the capture to the vector set so the condition is compared from then on.

## Acceptance criteria

- A capture in `tests/foxio_vectors/` reaches the branch with a value the three rules do not agree on.
- The measured FoxIO output for that capture is recorded in this issue with the command that produced it.
[...]

Issue #138, opened 2026-08-07
Settle whether ja4plus emits a fingerprint on a stream the FoxIO reference ignores


Crank-Git:
## What is wrong

`ja4plus` emits a fingerprint on streams the FoxIO reference holds no value for. 21
register entries record it, and every one of them names #13, which is an epic. An epic is
not schedulable, so no worker owns them.

## Evidence

| Capture | Methods |
|---|---|
| `chrome-cloudflare-quic-with-secrets.pcapng` | `JA4`, `JA4_r`, `JA4_o`, `JA4_ro`, `JA4S`, `JA4S_r` |
| `ssh2.pcapng` | `JA4`, `JA4_r`, `JA4_o`, `JA4_ro`, `JA4S`, `JA4S_r` |
| `tls3.pcapng` | `JA4`, `JA4_r`, `JA4_o`, `JA4_ro`, `JA4S`, `JA4S_r` |
| `https-connect.pcap` | `JA4`, `JA4S` |
| `quic-tls-handshake.pcapng` | `JA4` |
| `quic-with-several-tls-frames.pcapng` | `JA4` |

The cause reads the same on all of them: "ja4plus produces a fingerprint the reference
does not hold".

#118 widened this from JA4 alone to JA4 and JA4S. #121 widened it again to the raw forms.
**One rule resolves every one of them**, because the extra JA4S streams are the exact
streams that already carry an extra JA4 value.

#128 records that the FoxIO Rust implementation holds the exact value `ja4plus` produces
for `quic-with-several-tls-frames.pcapng/JA4`. Read the Rust snapshots before you treat
any of these as a defect in `ja4plus`.

## What to build

1. Read all three FoxIO implementations, the Python one, the Rust one and the Wireshark
   dissector, on the same streams. A value that one implementation holds and another
   omits is a gap in that implementation, not a defect in `ja4plus`.
2. State one rule that explains every stream in the table. If `ja4plus` is right, record
   the divergence and keep the values. If `ja4plus` is wrong, state which packet it must
   stop reading.
3. Apply `.claude/rules/conformance.md` where the reading is a proven reference defect.
   Proof is a measurement, never a reading of the source.
4. Repoint or remove the 21 register entries.

## Acceptance criteria

- One stated rule covers every capture in the table above.
- Every one of the 21 entries is removed, or names this issue with a cause the evidence
  supports.
- The conformance suite passes.
[...]

PR #159, opened 2026-08-07
Bound a TLS hello on its own length field, not on the record length

Crank-Git:
Refs #151.

## What

`parse_tls_handshake` bounded a ClientHello or a ServerHello on the length field of the
TLS record that carries it. The bound now reads the three-byte length field of the
handshake message, and the reader slices the buffer at the end of that message.

## Why

A TLS handshake record carries one or more handshake messages. A TLS 1.2 server puts the
ServerHello, the Certificate, the ServerKeyExchange and the ServerHelloDone in one
record, and that record is longer than one TCP segment. The record bound rejected the
first segment, so the reader never looked at the ServerHello inside it.

The first server data packet of each affected stream, read with `scapy`:

| Capture | Stream | Packet | Segment bytes | Record needs | ServerHello ends at byte |
|---|---|---|---|---|---|
| `ssh2.pcapng` | `57374 <-> 52.178.17.3:443` | 232 | 1452 | 6296 | 94 |
| `ssh2.pcapng` | `57375 <-> 204.79.197.220:443` | 254 | 1452 | 7036 | 107 |
| `tls-handshake.pcapng` | `50167 <-> 40.126.24.84:443` | 148 | 1460 | 3955 | 98 |
| `browsers-x509.pcapng` | `54524 <-> 13.107.21.239:443` | 5 | 1458 | 5947 | 111 |
| `latest.pcapng` | `52940 <-> 52.249.29.248:443` | 158 | 1452 | 7141 | 94 |
| `latest.pcapng` | `52941 <-> 52.249.29.248:443` | 192 | 1452 | 7141 | 94 |
| `socks4-https.pcap` | `50606 <-> 10.0.0.2:9901` | 8 | 1360 | 6776 | 90 |

Every row reads the same way. The record needs more bytes than the segment carries, and
the ServerHello ends inside the segment.

The three streams #151 names are the first three rows. The other four are the same cause,
and one reader change settles all seven.

## The vector that proves it

`.claude/rules/external-apis.md` reads the Rust snapshot tie-break at stream granularity.
Six of the seven streams now produce the exact JA4S value the FoxIO Rust snapshot holds:

| Stream | FoxIO Rust JA4S | ja4plus |
|---|---|---|
| `ssh2.pcapng` `57374` | `t120300_c030_09f674154ab3` | same |
| `ssh2.pcapng` `57375` | `t1205h2_c030_015e35fdd027` | same |
| `tls-handshake.pcapng` `50167` | `t120400_c030_4e8089b08790` | same |
| `browsers-x509.pcapng` `54524` | `t1206h2_c030_044dc9b3196d` | same |
| `latest.pcapng` `52940` | `t120300_c030_09f674154ab3` | same |
| `latest.pcapng` `52941` | `t120300_c030_09f674154ab3` | same |

`tls-handshake.pcapng` stream `50167` proves the FoxIO Python gap on one capture. It runs
on port 443, so neither of the two gaps #138 names explains the omission.

The snapshots of `browsers-x509.pcapng` and `latest.pcapng` join
`tests/foxio_vectors/rust_expected/`, taken without change from the pinned upstream
commit. `tests/download_test_vectors.py` and the `NOTICE` list them, so the copy stays
reproducible.

## The seventh stream

`socks4-https.pcap` stream `50606 <-> 10.0.0.2:9901` is the exception. Its Rust snapshot
holds `ja4t`, `ja4l_c` and `ja4l_s`, and no `ja4s` and no `ja4`. `ja4plus` reads the
record layer without regard to the tunnel protocol that carries it, which is the
behaviour #138 decided to keep for the three JA4X values on the same stream. The register
records the JA4S value under that decision.

## The register

99 keys to 104 keys, and the conformance suite moves from 99 xfailed to 104 xfailed. The
two counts stay equal. The five new keys are `browsers-x509.pcapng/JA4S`,
`browsers-x509.pcapng/JA4S_r`, `latest.pcapng/JA4S`, `latest.pcapng/JA4S_r` and
`socks4-https.pcap/JA4S`. The existing `ssh2.pcapng` and `tls-handshake.pcapng` JA4S keys
hold the capture, so they absorb the three streams #151 names, and their cause text now
records the second reason.

## How tested

Written test-first. Three unit tests in `tests/test_utils.py` fix the reader contract, and
`TestTheStreamsThatCoalesceTheServerHelloRecord` in `tests/test_foxio_rust_parity.py`
names each of the six streams and its value. All five gates run in the worktree:

```
.venv/bin/pytest tests/ -m "not spec_validation"   926 passed, 8 xfailed
.venv/bin/pytest tests/ -m spec_validation         1358 passed, 104 xfailed
.venv/bin/ruff format --check ja4plus/ tests/      98 files already formatted
```

Coverage did not fall: 650 missed of 3299 statements on the base, 650 missed of 3301 on
this branch.

## Note for the merge gate

`docs/specs/spec.md` gains no changelog round here. #141 is live on the same integration
branch, and a numbered row would conflict. The full reading sits in
`docs/implementation_notes.md`, appended as a new section.

Verified against: https://www.rfc-editor.org/rfc/rfc5246#section-6.2.1 (TLS 1.2, retrieved 2026-08-07)
Verified against: https://github.com/FoxIO-LLC/ja4/tree/main/rust/ja4/src/snapshots (commit 27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8)

### DIFF
--- a/.claude/rules/external-apis.md
+++ b/.claude/rules/external-apis.md
@@ -61,11 +61,12 @@ streams and omits six, and the Rust snapshot holds all thirteen. Where the Pytho
 holds no value for a stream and the Rust snapshot holds one, the Rust snapshot decides.
 Where both carry a value for one method on one stream, `python/test/testdata/` decides.
 
-The FoxIO Python implementation reads no QUIC handshake, and it reads no TLS on a port it
-does not know. Those two gaps produce every case. `tests/foxio_vectors/rust_expected/`
-holds a copy of the eight snapshots that cover them, and
-`tests/test_foxio_rust_parity.py` measures the match. #138 records the reading, and
-`docs/implementation_notes.md` holds the table.
+The FoxIO Python implementation has three gaps. It reads no QUIC handshake, it reads no
+TLS on a port it does not know, and it reads no ServerHello whose handshake record spans
+several TCP segments. Those three gaps produce every case.
+`tests/foxio_vectors/rust_expected/` holds a copy of the ten snapshots that cover them,
+and `tests/test_foxio_rust_parity.py` measures the match. #138 records the first two
+readings, #151 records the third, and `docs/implementation_notes.md` holds both tables.
 
 Read this rule at stream granularity. An earlier form read "only where the Python file
 holds an empty array", and that form covered one capture of the eight.
--- a/docs/implementation_notes.md
+++ b/docs/implementation_notes.md
@@ -859,3 +859,79 @@ all other non-zero versions use the v1 salt.
 
 **Integration:** `extract_tls_info` checks for QUIC on UDP packets
 before falling through to standard TLS parsing.
+
+---
+
+## A hello is bounded on its own length field, not on the record length
+
+Issue #151 asked why `ja4plus` read no ServerHello on three streams for which the FoxIO
+Rust snapshot holds a JA4S value. The three share one cause, and the cause was a defect
+in this project.
+
+### The mechanism
+
+A TLS handshake record carries one or more handshake messages. A TLS 1.2 server puts the
+ServerHello, the Certificate, the ServerKeyExchange and the ServerHelloDone in one
+record. That record is longer than one TCP segment, so the first segment holds the whole
+ServerHello and only part of the record.
+
+`parse_tls_handshake` bounded the hello on the record length field. It returned nothing
+whenever the segment held less than the record declared, so it never read a ServerHello
+that a coalesced record carries. The bound now reads the three-byte length field of the
+handshake message at offset 6, and the reader slices the buffer at the end of that
+message. The slice keeps the reader inside the hello, because the bytes that follow
+belong to the Certificate message.
+
+### The measurement
+
+The first server data packet of each stream, read with `scapy`:
+
+| Capture | Stream | Packet | Segment bytes | Record needs | ServerHello ends at byte |
+|---|---|---|---|---|---|
+| `ssh2.pcapng` | `57374 <-> 52.178.17.3:443` | 232 | 1452 | 6296 | 94 |
+| `ssh2.pcapng` | `57375 <-> 204.79.197.220:443` | 254 | 1452 | 7036 | 107 |
+| `tls-handshake.pcapng` | `50167 <-> 40.126.24.84:443` | 148 | 1460 | 3955 | 98 |
+| `browsers-x509.pcapng` | `54524 <-> 13.107.21.239:443` | 5 | 1458 | 5947 | 111 |
+| `latest.pcapng` | `52940 <-> 52.249.29.248:443` | 158 | 1452 | 7141 | 94 |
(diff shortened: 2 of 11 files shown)


Crank-Git:
## Self-review

Two lenses ran over the diff: security and hostile input, and correctness and test quality.

### Security and hostile input — no finding

The loop advance is unchanged and still moves at least five bytes per iteration, because
`record_length` is a 16-bit field. `handshake_length` reaches no allocation and no loop
advance. The reader tests `end > len(raw_data)` before it slices, so the slice never
passes the buffer. The new `offset + 9 <= len(raw_data)` guard is stronger than the reads
it protects at `tls_utils.py:70` and `tls_utils.py:78`.

The false-positive question is the one that mattered, and the answer is that the change
narrows the window. The record-type test and the handshake-type test are unchanged, and
the reader now also needs the three length bytes to name a message that fits the buffer.

### Correctness and test quality — two findings, both fixed

**Finding 1, should-fix.** `test_parse_server_hello_rejects_a_truncated_hello_message`
passed against the old parser. It declared `record_length` 6291 on a 37-byte buffer, so
the old record bound rejected the record for its own reason and the test proved nothing.

Fixed. The record now declares the length it carries, so the old bound reaches the
parser and returns a partial ServerHello. The message declares 38 bytes and the record
carries 28, so the new bound returns nothing.

**Finding 2, nit.** `test_parse_server_hello_reads_no_byte_after_the_hello_message`
claimed to prove that the reader stops at the end of the hello. It did not. The extension
list declared 8 bytes and the hello carried 8, so the extension walk was already bounded
by the packet's own field and the slice changed nothing.

Fixed and renamed to `test_parse_server_hello_reads_no_extension_after_the_hello_message`.
The extension list now declares 16 bytes and the hello carries 8. The eight bytes that
follow the hello read as extension `0x0010`, and the old parser reports
`[0x002b, 0x0010]` where the new one reports `[0x002b]`.

**Verification.** Each of the three unit tests was run against the parser of the parent
commit:

```
git checkout HEAD~1 -- ja4plus/utils/tls_utils.py
.venv/bin/pytest tests/test_utils.py -k "record_the_segment or truncated_hello or after_the_hello"

FAILED tests/test_utils.py::TestTLSParsing::test_parse_server_hello_reads_a_record_the_segment_truncates
FAILED tests/test_utils.py::TestTLSParsing::test_parse_server_hello_reads_no_extension_after_the_hello_message
FAILED tests/test_utils.py::TestTLSParsing::test_parse_server_hello_rejects_a_truncated_hello_message
3 failed
```

The reviewer ran the same check on the six parity cases in
`TestTheStreamsThatCoalesceTheServerHelloRecord`. All six fail against the parent parser
with `assert () == ('t120300_c030_09f674154ab3',)` and the matching values, so no case
passes on an empty produced list.

### The ClientHello path — no finding

`_parse_client_hello` bounds its extension walk on `extensions_len` from the packet, so
the narrower slice changes no well-formed ClientHello. No JA4 value in the vector set
moved, and the conformance suite holds its counts.

PR #160, opened 2026-08-07
Pass an ASCII ALPN byte through the JA4 value

Crank-Git:
Settles the printable ASCII part of the JA4 ALPN trigger condition, by measurement. Part of #141.

## What

The FoxIO prose writes `99` when the first byte or the last byte of the first ALPN
value is not alphanumeric. The measurement contradicts the prose. Both FoxIO
implementations pass a **printable ASCII** byte through, whether or not that byte is
alphanumeric. `ja4plus` now does the same, so `h\x20` reads `h ` and not `99`.

The range is `0x20-0x7E`. It stops at `0x7E`, because the two implementations agree only
inside it. `rust/ja4/src/tls.rs` reads a control byte as the escape text tshark writes,
so it reads `h\x1f` as five characters and writes `hf`, while `python/ja4.py` reads the
byte. A control byte therefore keeps `99`.

`ja4s.py` reads the same function, so JA4 and JA4S carry one rule.

Four inputs move, each from `99` to the value both implementations write: `h\x20`,
`\x20h`, `h\x21` and `h\x7e`.

## Why the measurement needed a new capture

No FoxIO capture separates the rules. `tls-non-ascii-alpn.pcapng` holds `0xba 0xad`,
and every rule fires on it. This branch builds the separating capture.

- `tests/build_alpn_condition_capture.py` writes `tests/foxio_vectors/alpn-condition.pcap`.
- Stream 0 holds the first ALPN value `h\x20`. Stream 1 holds `\x20h`.
- Each stream carries one TLS ClientHello and no TCP handshake, so the capture adds no
  JA4T value and no JA4L value to the comparison.
- The two ClientHellos differ in the first ALPN value alone.

## How it was tested

Both FoxIO implementations were built from a clean clone at the pinned upstream commit
`27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8`:

```
python python/ja4.py tests/foxio_vectors/alpn-condition.pcap -J
rust/target/release/ja4 tests/foxio_vectors/alpn-condition.pcap
```

Both write `t13d0304h _55b375c5d22e_ef5f37ab036a` for stream 0 and
`t13d0304 h_55b375c5d22e_ef5f37ab036a` for stream 1. `ja4plus` now produces both.

Two wider probes reached 21 inputs in total. #141 carries the full output, and
`docs/implementation_notes.md` holds the tables.

## The invariant

A script checked the final function against all 21 measured inputs. Two properties hold.

- Every input where the two implementations **agree**, `ja4plus` produces the agreed
  value.
- Every input where they **disagree**, `ja4plus` produces exactly the value it produced
  before this branch. Nothing moves.

There is one deliberate exception. `0x09` is the only byte outside `0x20-0x7E` where the
two agree, and that agreement is an accident of the tshark text form rather than a rule,
because `0x0A` beside it makes the Rust implementation write no value.
`test_the_tab_byte_keeps_the_value_99_although_the_references_agree` pins the decision.

## What this branch does NOT adopt

The two FoxIO implementations **disagree** on every byte outside `0x20-0x7E` that sits in
a position other than the first, and on a one-byte value.

| The first ALPN value | FoxIO Python | FoxIO Rust |
|---|---|---|
| `h\xab` | `h?` | `h9` |
| `\xabh` | `99` | `9h` |
| `\x30\x31\xab\xcd` | `0?` | `09` |
| `h` | `h` | `h0` |
| `\xab` | `99` | `90` |
| `\x20` | ` ` | ` 0` |
| `h\x00` | `h` | `h0` |
| `h\x1f` | `h\x1f` | `hf` |
| `h\x7f` | `h\x7f` | `hf` |
| `\x01h` | `\x01h` | `\h` |
| `h\x0a` | `h\n` | no value |

`?` is `U+FFFD`, which tshark writes for a byte it cannot decode.

This branch adopts neither side. It holds the value the project wrote before, so **no
fingerprint moves on a case the reference disputes**. `.claude/rules/conformance.md`
states that a defect outside its two named shapes is a question for the user, and #141
carries the question.

## Provenance

`alpn-condition.pcap` is not FoxIO material, so the blanket claim in
`tests/foxio_vectors/NOTICE` would have become false. This branch adds a
`Captures the ja4plus authors built` section to that file, and the same paragraph to
the `NOTICE_TEMPLATE` in `tests/download_test_vectors.py` that rewrites it.
`alpn-condition.pcap.json` holds a measurement, not a copy.

## Gates

All five gates pass in the worktree.

```
pytest tests/ -m "not spec_validation"   942 passed, 8 xfailed, 93 subtests passed
pytest tests/ -m spec_validation         1359 passed, 147 skipped, 99 xfailed
ruff format --check ja4plus/ tests/      100 files already formatted
```

The deviation register holds **99** keys and the suite reports **99 xfailed**, both
before and after. The new vector adds no register entry. The manifest moves from 37 to
38 vectors.

Coverage does not fall: base `origin/epic/12-batch10` is 3299 statements with 650
missed, and this branch is 3303 statements with the same 650 missed.

Out-of-scope discovery filed as #161, not fixed here.

Verified against: https://github.com/FoxIO-LLC/ja4/tree/27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8 (FoxIO JA4+ python and rust implementations, retrieved 2026-08-07)


Crank-Git:
Out-of-scope discovery, filed as #161 and not fixed here.

`tests/generate_foxio_baseline.py` is the documented way to add a vector. It drops the `decided` field from every register entry that carries one: 44 entries on `epic/12-batch10` at `83e8e01` become 0. The entry count stays at 99, so no invariant catches the loss.

This branch ran the generator to get the one new manifest line, then restored `tests/foxio_deviations.json` with `git checkout`. The register in this diff is byte-identical to the base, and it still holds 44 `decided` markers.

Crank-Git:
## Self-review

Three findings, all fixed on this branch.

**1. Blocker — the passthrough range was too broad.** The first commit adopted "both bytes are ASCII". A second probe of the control bytes and the top of ASCII shows the two FoxIO implementations agree only inside `0x20-0x7E`. `rust/ja4/src/tls.rs` reads a control byte as the escape text tshark writes, so it reads `h\x1f` as five characters and writes `hf`, and `h\x0a` makes it write no value at all. Passing a control byte through would have moved a fingerprint on an input the reference disputes, which is the one thing this issue must not do. Fixed in `9c0e284`.

**2. Minor — one agreement is declined without a record.** `0x09` is the only byte outside `0x20-0x7E` where both implementations agree. The agreement is an accident of the tshark text form, not a rule, because `0x0A` beside it breaks the Rust implementation. Declining it silently would read as an oversight later, so `99dcdfe` pins it in `test_the_tab_byte_keeps_the_value_99_although_the_references_agree` and `docs/implementation_notes.md` states the reasoning.

**3. Minor — two line references went stale.** The change adds 24 lines above the extension hash, so the `Location:` reference for #132 in `docs/implementation_notes.md` and row 10 of `docs/specs/features/02-correctness-audit.md` no longer named the statement they describe. Fixed in `580d8ee`.

### The invariant this branch is built on

A script checked the final function against all 21 measured inputs. Two properties hold:

- Every input where the two implementations **agree**, `ja4plus` produces the agreed value. The one exception is `0x09`, declined above with a reason.
- Every input where they **disagree**, `ja4plus` produces exactly the value it produced before this branch. Nothing moves.

Four inputs move in total: `h\x20`, `\x20h`, `h\x21` and `h\x7e`, each from `99` to the value both implementations write.

Issue #162, opened 2026-08-07
Record the settled JA4 ALPN readings and measure the divergence


Crank-Git:
#141 settled the half of the JA4 ALPN condition that the two FoxIO implementations agree on,
and it left the half they dispute. The user settled the disputed half on 2026-08-07: **the
values stay as they are.** This issue records that reading where a reader looks for it, and it
makes the divergence a measured comparison rather than an invisible one.

## The two readings the user settled

**1. A byte outside `0x20-0x7E` in a position other than the first keeps `99`.**

No measurement settles it, because the two FoxIO implementations disagree with each other:

| First ALPN value | FoxIO Python | FoxIO Rust | `ja4plus` |
|---|---|---|---|
| `h\xab` | `h�` | `h9` | `99` |
| `\xabh` | `99` | `9h` | `99` |
| `h\x1f` | `h\x1f` | `hf` | `99` |
| `h\x0a` | `h\n` | no value | `99` |

`CLAUDE.md` rule 1 forbids a move that no vector requires. Two further reasons support the
reading. The FoxIO Rust results on a control byte are an artifact: `rust/ja4/src/tls.rs` reads
the tshark escape text, so it reads `h\x1f` as the five characters `h`, `\`, `x`, `1`, `f`.
And the FoxIO Python result puts `U+FFFD` into a fingerprint, which is a character no byte of
the packet holds.

**2. A one-byte first ALPN value keeps the repeated byte.**

| First ALPN value | FoxIO Python | FoxIO Rust | `ja4plus` |
|---|---|---|---|
| `h` | `h` | `h0` | `hh` |

All three differ, and no vector separates them. The FoxIO Python value holds one character,
which cannot fill a two-character field. `ja4plus` holds `hh`, which is well formed and stable.

## Why this issue exists

The measurements sit in two issue comments and in `docs/implementation_notes.md`. Neither is
where a reader looks for a settled reading, and neither makes the comparison run.

**A comparison that is never made reads as a comparison that passes.** This project has found
that shape six times. The disputed inputs are the seventh candidate: the suite is green today
because no capture holds one.

## What to build

1. **Record both readings in `docs/specs/spec.md` § Divergence register**, in the form the
   `socks4-https.pcap/JA4X` row and the #127 row already use. State that each reading is
   reversible, and name the measurement.
2. **Add a capture that holds the disputed inputs** to `tests/foxio_vectors/`, beside
[...]

Issue #161, opened 2026-08-07
Keep the decided field when the baseline generator rewrites the register


Crank-Git:
Found by #141, which ran the generator to add one manifest line.

## The defect

`tests/generate_foxio_baseline.py` rewrites `tests/foxio_deviations.json` from a
measurement. The rewrite drops the `decided` field from every entry that carries it.

On `epic/12-batch10` at `83e8e01` the register holds 44 entries with `"decided": true`.
After the generator runs, it holds 0. The entry count stays at 99, so the count
invariant does not catch the loss.

## How to reproduce

```
$ git checkout epic/12-batch10
$ grep -c '"decided"' tests/foxio_deviations.json
44
$ python tests/generate_foxio_baseline.py
$ grep -c '"decided"' tests/foxio_deviations.json
0
```

The generator also reorders the keys of each entry, which turns a one-line change into a
diff of about 370 lines. That hides the real change from a reviewer.

## Why it matters

`decided` records that a person settled a deviation. It is the difference between "this
case fails and nobody has looked" and "this case fails and the user accepted the
reason". The generator is the documented way to add a vector, so the next person who
adds a vector silently deletes 44 such decisions.

`tests/foxio_manifest.py` documents that path:

> 2. Run `python tests/generate_foxio_baseline.py`. It rewrites the manifest and the
>    register from the vectors it finds.

#141 worked around the defect. It ran the generator, kept the one new manifest line, and
restored `tests/foxio_deviations.json` with `git checkout`.

## Acceptance criteria

- `python tests/generate_foxio_baseline.py` keeps the `decided` field of every entry it
  rewrites.
- The generator keeps the key order of an entry it does not change, so a run that adds
  no deviation produces no diff in `tests/foxio_deviations.json`.
- A test proves both, and it fails against the current generator.

Crank-Git:
## Plan

Batch 11, member 1 of 2, on `epic/12-batch11`. Member 2 is #162, and it runs after this
one merges, because it adds vectors through the generator this issue repairs.

**Approach.** Make `tests/generate_foxio_baseline.py` merge into the existing register
rather than rewrite it. Read `tests/foxio_deviations.json` first. For a key that already
exists, keep the stored entry and its key order, and keep every field the measurement
does not produce, `decided` among them. Write a new key only for a deviation the
measurement finds and the register does not hold. Remove a key only when the measurement
no longer produces that deviation.

**Write the test first.** A test that runs the generator against a register holding a
[...]

PR #164, opened 2026-08-07
Keep the stored register entry when the baseline generator runs

Crank-Git:
Fixes the defect #161 records. Part of Epic 1 batch 11.

## What

`tests/generate_foxio_baseline.py` rebuilt each register entry as
`{"issue": held.issue, "cause": held.cause}`. That form drops every field the
`Deviation` class does not name, and it states its own field order. A run therefore
deleted the `decided` field of 49 entries and rewrote the field order of all 104, which
turned a one-line change into a diff of about 370 lines.

The register run now merges:

- A key the register already holds keeps its stored entry, byte for byte.
- The generator writes a new entry only for a deviation the register does not hold.
- A key that stops failing still disappears.

`_committed` reads the raw JSON, because a parsed entry carries neither the extra fields
nor the field order. It still calls `load_register` first, so a malformed register stops
the run as it did before.

## Why

`decided` records that a person settled a deviation. The generator is the documented way
to add a vector, so the next person who added one deleted 49 such decisions silently.
The entry count stayed at 104, so no count invariant caught the loss. #141 met this and
worked around it with `git checkout`.

## How tested

The tests came first, and they failed against the generator as it stood:

```
FAILED tests/test_generate_foxio_baseline.py::TestTheCommittedEntry::test_a_committed_entry_keeps_the_decided_field
FAILED tests/test_generate_foxio_baseline.py::TestTheCommittedEntry::test_a_committed_entry_keeps_the_order_of_its_fields
FAILED tests/test_generate_foxio_baseline.py::TestTheCommittedRegisterAfterARun::test_a_run_that_finds_the_same_deviations_writes_the_same_bytes
FAILED tests/test_generate_foxio_baseline.py::TestTheCommittedRegisterAfterARun::test_the_run_keeps_every_decided_entry
6 failed, 16 passed in 0.05s
```

A direct measurement against the generator as it stood reported the loss:

```
committed keys 104 decided 49
rewritten keys 104 decided 0
field order of first rewritten entry ['issue', 'cause']
```

`test_a_run_that_finds_the_same_deviations_writes_the_same_bytes` drives every one of the
104 committed keys through `_register` and `_write`, then compares the written text
against the committed file. It is a whole-file comparison, not a sample.

The generator itself also ran end to end. The MD5 of both written files is the same
before and after, and `git status` reports neither file as changed:

```
93c8cd5d6800c994c2b5421ac5aa2a91  tests/foxio_deviations.json
720586af8ddd29e7728e6912a7540b3c  tests/foxio_vector_manifest.json
wrote 38 vectors to tests/foxio_vector_manifest.json
wrote 104 entries to tests/foxio_deviations.json
93c8cd5d6800c994c2b5421ac5aa2a91  tests/foxio_deviations.json
720586af8ddd29e7728e6912a7540b3c  tests/foxio_vector_manifest.json
```

The five gates:

```
pytest tests/ -m "not spec_validation"  950 passed, 8 xfailed, 93 subtests passed
pytest tests/ -m spec_validation        1371 passed, 146 skipped, 104 xfailed
ruff format --check ja4plus/ tests/     100 files already formatted
mypy ja4plus/                           Success: no issues found in 27 source files
```

The register still holds 104 keys and 49 entries that carry `decided`.

Coverage of `tests/generate_foxio_baseline.py` rises from 64% to 65%. Coverage of
`ja4plus/` stays at 80%, because this change touches no file under `ja4plus/`.

## Scope

No fingerprinter changes. No register content changes. No new vector, which is #162.

`tests/foxio_manifest.py` states the procedure that this change alters, so its step 2 and
step 3 change in the same commit.

## Notes for the reviewer

- I chose no `docs/specs/spec.md` Changelog round number. The project manager owns the
  round for this batch.
- Out of scope, and not fixed here: the `## How to measure a new baseline` section of
  `tests/foxio_deviations.py` names `tests/generate_foxio_deviations.py`, and the file is
  `tests/generate_foxio_baseline.py`. The same section says the program "writes a
  register from that measurement", which now reads as a merge. Both are stale prose in a
  file this issue does not own.

Verified against: repository code only. This change calls no external interface.

### DIFF
--- a/tests/foxio_manifest.py
+++ b/tests/foxio_manifest.py
@@ -15,10 +15,11 @@ today, so a deleted vector that passes leaves nothing behind to notice.
 ## How to add a vector
 
 1. Put the capture and its expected-output file under `tests/foxio_vectors/`.
-2. Run `python tests/generate_foxio_baseline.py`. It rewrites the manifest and the
-   register from the vectors it finds.
+2. Run `python tests/generate_foxio_baseline.py`. It rewrites the manifest from the
+   vectors it finds, and it merges the measurement into the register.
 3. Read the difference. The manifest gains one line, and the register gains one entry
-   for each case the new vector fails.
+   for each case the new vector fails. An entry the register already holds keeps every
+   field, so the `decided` field of a settled deviation survives the run.
 
 A vector is a two-file change on purpose. A capture that arrives without a manifest
 line, or leaves without one, fails the suite.
--- a/tests/generate_foxio_baseline.py
+++ b/tests/generate_foxio_baseline.py
@@ -24,10 +24,10 @@ is not allowed.
 
 The register run merges. A key the committed register already holds keeps its stored
 entry, with every field and in the field order the file holds. A cause this program
-reads from a failure message states the symptom, and a person states the mechanism, and
-the `decided` field records that a person settled the deviation. This program writes a
-new entry only for a deviation the register does not hold. A key that stops failing
-still disappears.
+reads from a failure message states the symptom, and a person states the mechanism. The
+`decided` field records that a person settled the deviation. This program writes a new
+entry only for a deviation the register does not hold. A key that stops failing still
+disappears.
 
 A run that measures the deviations the register already holds therefore changes no byte
 of `tests/foxio_deviations.json`.
@@ -290,7 +290,7 @@ def _register(collector, committed):
     cause this program reads from a failure message states the symptom, and a person
     states the mechanism. #78, #96, #97 and #105 hold mechanisms no failure message
     carries, and a mechanical rewrite would destroy them. The `decided` field records
-    that a person settled the deviation, and this program measures no such field, so a
+    that a person settled the deviation. This program measures no such field, so a
     rewrite of a stored entry would delete it. A key that stops failing still disappears.
 
     Args:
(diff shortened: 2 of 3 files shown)


Crank-Git:
Self-review, three lenses over the diff.

**Correctness.** No defect found in `_committed` or in the changed `_register`. `_register` reuses the stored entry object, so the field order and every field survive. A key the measurement no longer produces still leaves the register.

**Test rigor.** No test passes vacuously. The review reproduced the pre-fix logic and confirmed it fails `test_a_run_that_finds_the_same_deviations_writes_the_same_bytes` and `test_the_run_keeps_every_decided_entry`. Both tests read the real committed register, all 104 keys, and compare the written text against the file.

**A note, not a defect.** A committed `cause` is now never re-derived, even when the failure message of that key later changes shape. That is the intent of #161: it protects the mechanisms #78, #96, #97 and #105 hold. Both docstrings state it.

**STE.** The review found three sentences longer than the 25-word limit, at `tests/generate_foxio_baseline.py:26`, `tests/generate_foxio_baseline.py:292` and `tests/test_generate_foxio_baseline.py:262`. `f29c01a` splits each one. The five gates ran again after that commit and all five pass.

No thread stays open.

PR #165, opened 2026-08-07
Record the settled JA4 ALPN readings and measure the divergence

Crank-Git:
Records the JA4 ALPN readings the user settled on 2026-08-07, and turns the divergence
into a comparison the conformance suite runs. Part of #162. Epic 1 batch 11, member 2.

## What

#141 measured the disputed region of the JA4 ALPN condition, and the user settled it: the
`ja4plus` values stay. The readings then sat in two issue comments and in
`docs/implementation_notes.md`. No capture held a disputed input, so the suite was green
because the comparison was never made. That is the seventh instance of this shape in the
project.

Four changes:

1. `tests/build_alpn_condition_capture.py` writes seven streams instead of two.
2. `tests/foxio_vectors/alpn-condition.pcap.json` gains five expected records.
3. `tests/foxio_deviations.json` gains sixteen decided entries that #162 owns.
4. `docs/specs/spec.md` § Divergence register gains two rows, and Changelog round 35.

| Stream | First ALPN value | FoxIO Python | FoxIO Rust | `ja4plus` | Register entry |
|---|---|---|---|---|---|
| 0 | `h\x20` | `h ` | `h ` | `h ` | no, it conforms |
| 1 | `\x20h` | ` h` | ` h` | ` h` | no, it conforms |
| 2 | `h\xab` | `h?` | `h9` | `99` | yes |
| 3 | `\xabh` | `99` | `9h` | `99` | no, it conforms |
| 4 | `h\x1f` | `h\x1f` | `hf` | `99` | yes |
| 5 | `h\x0a` | `h\n` | no value | `99` | yes |
| 6 | `h` | `h` | `h0` | `hh` | yes |

## Why stream 3 holds no register entry

`\xabh` is disputed between the two FoxIO implementations, and `ja4plus` writes the FoxIO
Python value there, so the case conforms. A strict `xfail` on a case that passes fails the
suite, so that stream cannot carry an entry. The issue asks for an entry per disputed
value, and its own text scopes the entry to a value where "ja4plus matches neither". Four
streams meet that test, and stream 3 does not.
`test_the_agreed_streams_carry_no_register_entry` states the rule, and the capture still
carries the input so a later change of the condition moves it.

## How the expected-output file was built

The ALPN characters come from the two measurement tables in
`docs/implementation_notes.md`. No value is re-derived. The rest of each fingerprint is
copied from stream 0, because the seven client hellos carry the same ciphers and the same
extensions, so the JA4 prefix and both hashes are invariant.
`test_every_stream_differs_from_the_others_in_the_alpn_characters_alone` proves that
invariance rather than assuming it.

## Out of scope

`compute_alpn_value` is unchanged: `git diff accab01..HEAD -- ja4plus/` is empty. No
fingerprint value moved, and the printable-ASCII range #141 landed is untouched.

## How this is tested

The new assertions are proved to fail, because an assertion that cannot fail is not
evidence.

**The tests came first.** Commit `38dcaa2` holds them alone, and 17 cases fail against the
two-stream capture:

```
17 failed, 19 passed
FAILED test_the_capture_carries_every_alpn_value_the_measurement_names
FAILED test_the_produced_value_matches_neither_foxio_implementation[stream2-68ab]
FAILED test_the_deviation_register_holds_each_disputed_stream[stream6-68]
...
```

**Mutation A — the register entries bite.** Move the stream 6 expected value from `h` to
`hh`, so the reference agrees with `ja4plus`:

```
4 failed, 1375 passed, 116 xfailed
[XPASS(strict)] issue #162: The first ALPN value is the single byte 0x68 ...
FAILED test_the_expected_file_holds_the_foxio_python_alpn_characters
FAILED test_the_produced_value_matches_neither_foxio_implementation[stream6-68]
```

**Mutation B — the recorded FoxIO Rust value bites.** Change `hf` to `hX` in `STREAMS`:

```
1 failed, 35 passed
FAILED test_each_register_entry_states_the_output_of_both_foxio_implementations[stream4-681f]
```

**The five gates, all green in the worktree:**

```
pytest tests/ -m "not spec_validation"   967 passed, 8 xfailed, 93 subtests passed
ruff format --check ja4plus/ tests/      100 files already formatted
```

**The invariant.** `tests/foxio_deviations.json` holds 120 keys, and the conformance suite
reports 120 xfailed. The register rose from 104 keys, and the decided count rose from 49
to 65. `tests/generate_foxio_baseline.py` runs twice and changes no stored entry, which
confirms the #161 merge behaviour.

**Coverage.** 80% at the base commit `accab01`, and 80% on this branch. It does not fall,
and no line under `ja4plus/` changed.

Verified against: https://github.com/FoxIO-LLC/ja4/tree/main/python/test/testdata (FoxIO
test vectors at the pinned commit `27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8`, retrieved
2026-08-07). The FoxIO measurements themselves are the #141 readings the user settled;
this change re-derives none of them.


Crank-Git:
## Self-review, lens 1: correctness

A review agent read `git diff accab01..HEAD` and recomputed the values independently,
rather than trusting the artifacts this branch adds. Findings:

- **No vacuous assertion.** Every assertion compares two independently sourced things:
  the value computed from the capture, the committed expected-output file, or the
  committed register. None compares a table against itself.
- **The stream and port mapping is correct.** A direct recomputation gives ports 44401
  to 44407 for streams 0 to 6, and the register keys `2:44403`, `4:44405`, `5:44406` and
  `6:44407` match. Stream 3 correctly holds no entry.
- **The hash-half invariance is true, not merely asserted.** A direct computation of all
  seven streams gives `55b375c5d22e_ef5f37ab036a` on every one.
- **`ja4plus/` is unchanged.** `git diff accab01..HEAD -- ja4plus/` is empty.
- **Each cause text matches the measurement tables** in `docs/implementation_notes.md`.

### The one finding, and the response

> In `STREAMS` the FoxIO Rust readings are hand-transcribed literals with no backing
> artifact file in this repo, unlike `foxio_python`, which is checked against
> `alpn-condition.pcap.json`.

This is correct, and it is a limitation this repository cannot remove. The capture is
synthetic, so no FoxIO Rust snapshot exists for it: `tests/foxio_vectors/rust_expected/`
holds copies of upstream snapshots, and upstream holds none for a capture this project
built. `.claude/rules/external-apis.md` allows a Rust snapshot to decide only where one
exists.

The literal is not unchecked, though. `test_each_register_entry_states_the_output_of_both_foxio_implementations`
asserts the literal appears in the cause text of the register entry, which is a second
committed artifact written from the #141 measurement. A mutation confirms the assertion
bites: changing `hf` to `hX` in `STREAMS` fails that test on stream 4.

Accepted as a documented limitation. No change.

Crank-Git:
## Self-review, lens 2: the writing standard

A review agent read `.claude/rules/ste.md` and the `## Terms` table, then checked every
line of prose this branch adds. Fixed in `b4a5c45`:

**Sentences above the 25-word cap, each split into single facts.**

- `docs/specs/spec.md` Divergence register, the disputed-byte row: 27 words joined by
  "because ... and".
- `docs/specs/spec.md` Changelog round 35: one 36-word sentence and one 31-word
  comma-splice.
- `docs/implementation_notes.md`: a 32-word sentence that held two facts.
- `tests/build_alpn_condition_capture.py`: a 26-word sentence and a 28-word sentence.
- `tests/test_ja4_alpn_condition.py`: a 32-word module docstring sentence and a 27-word
  docstring sentence.
- `tests/foxio_deviations.json`: the templated closing sentence of all sixteen new
  causes, at 26 words, and the 29-word stream 4 sentence.

**Passive voice, rule 11.**

- "The field is read by position" becomes "This function reads the field by position".
- "No value here is derived" becomes "Nothing here derives a value".
- "The rest of each fingerprint is copied from stream 0" becomes "The build copies the
  rest of each fingerprint from stream 0".

**One `-ing` noun, rule 14.** "moving the stream 6 expected value to `hh` makes four
cases report ..." becomes "Move the stream 6 expected value to `hh`, and four cases
report ...".

**Consistency.** The agent found no synonym rotation and no term used two ways.

### Two findings not taken, with the reason

> `docs/specs/spec.md:533` (Changelog row 35) runs roughly 14 sentences, past the
> six-sentence paragraph cap.

> The disputed-byte register row runs 8 sentences in one cell.

Both are table cells, not paragraphs, and the house form for both tables is one dense
cell. Round 28 and round 33 of the Changelog each run longer, and the
`socks4-https.pcap/JA4X` row runs 8 sentences. A six-sentence cap applied to a cell would
make every existing row of both tables a violation. The sentence-level findings inside
those cells are fixed; the cell length matches the form the tables already use.

### Verification after the rewrite

The register causes changed text and kept their meaning. The generator ran again and
preserved all sixteen entries. All five gates stay green, and the invariant holds:

```
pytest tests/ -m "not spec_validation"   967 passed, 8 xfailed, 93 subtests passed
ruff format --check ja4plus/ tests/      100 files already formatted
```

Issue #33, opened 2026-08-06
Bound the bytes one TCP stream may hold


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/02-correctness-audit.md`](docs/specs/features/02-correctness-audit.md):

> FR-correctness-audit-2 — The TCP reassembler holds no more than a fixed number of segments for one connection.

## What to build

`max_stream_bytes` at `ja4plus/utils/tcp_stream.py:41` limits the reassembled output, not the stored segments. A sender that emits many distinct small segments grows one stream without bound. Cap the stored bytes per stream, not only the output.

Related: `ja4plus/fingerprinters/ja4h.py:82`. When the buffer is not an HTTP request, the segment stays in the reassembler forever. Remove the stream when the buffer cannot become an HTTP request.

## Acceptance criteria

- [ ] One stream holds no more than the configured byte cap, whatever the segment count.
- [ ] A sender that emits 100000 distinct one-byte segments on one connection does not grow the reassembler without limit.
- [ ] A buffer that cannot become an HTTP request is removed from the reassembler.

## Links

- Feature: [`docs/specs/features/02-correctness-audit.md`](docs/specs/features/02-correctness-audit.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #13
Depends on #32

Crank-Git:
## Plan

Batch 12, on `epic/13-batch12`. #32 merged first as `c4fec95`. **This issue absorbs #103**
by the user's decision on 2026-08-07, and it inherits one defect #32 found and did not fix.

### What #32 already did, so you do not repeat it

- `get_stream` orders segments with modular arithmetic, so the 32-bit wrap is handled.
- `add_segment` detects a duplicate through a set of `(seq, length)` pairs, so the
  quadratic scan #103 measured is gone. The cost of one segment no longer grows with the
  segment count.
- The dead `base_seq` dictionary value is removed. The `base_seq` **method** stays, because
  `ja4plus/fingerprinters/ja4x.py:166` reads it.

**What #103 leaves for you is the bound itself.** `max_stream_bytes` limits the
reassembled output, not the stored segments, and `max_streams` bounds only the count of
streams. Nothing bounds the list of `(seq, data)` tuples one stream holds.

[...]

Issue #32, opened 2026-08-06
Order TCP segments correctly across a sequence-number wrap


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/02-correctness-audit.md`](docs/specs/features/02-correctness-audit.md):

> FR-correctness-audit-1 — The TCP reassembler orders segments correctly when the sequence number wraps past its 32-bit maximum.

> FR-correctness-audit-3 — The TCP reassembler detects a duplicate segment without scanning every stored segment.

## What to build

Two defects in `ja4plus/utils/tcp_stream.py`:

- Line 53 sorts raw sequence numbers. A TCP sequence number is 32 bits and wraps. A connection that crosses the boundary reassembles in the wrong order. Compare with modular arithmetic.
- Line 37 scans every stored segment to detect a duplicate, so cost grows with the square of the segment count. Use a set of `(seq, length)` pairs.

Also remove `base_seq`, stored at line 33 and never read, or use it.

Write the failing test before the fix.

## Acceptance criteria

- [ ] A reassembly test that crosses the 32-bit sequence boundary produces the segments in the correct order.
- [ ] Adding 10000 segments to one stream completes in under one second.

## Links

- Feature: [`docs/specs/features/02-correctness-audit.md`](docs/specs/features/02-correctness-audit.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #13

Crank-Git:
The self-review of #113 measured this defect and it is real.

#113 gave `SSHMessageTracker` its own sequence-aware comparison, and it found that a plain
comparison reads a wrapped sequence number as a number far behind the base. Measured
against the plain comparison, before the fix in that issue:

```
out of order, held:   [] (asserts [])
predecessor:          [] (asserts [48])
retransmission:       [] (asserts [])
held after it:        dict_keys([4294967286]) (asserts empty)
```

A retransmission that precedes the wrap point was held until the buffer reached its bound,
and an out-of-order segment that follows the wrap point was dropped.

**#113 fixed `ssh_utils.py` alone and did not touch `ja4plus/utils/tcp_stream.py`, which
holds the same defect.** That was correct, because `tcp_stream.py` is the subject of this
[...]

PR #168, opened 2026-08-07
Order TCP segments correctly across a sequence-number wrap

Crank-Git:
Refs #32. Part of #13.

## What

`ja4plus/utils/tcp_stream.py` held three defects from the audit register:

- **Row 1 — the wrap.** `get_stream` sorted raw sequence numbers. A TCP sequence number
  is 32 bits and wraps back to zero, so a connection that crossed the boundary
  reassembled backwards.
- **Row 2 — the duplicate scan.** `add_segment` scanned every stored segment, so its
  cost grew with the square of the segment count.
- **Row 4 — `base_seq`.** The stored `base_seq` value was written and never read.

## How

`_seq_before` orders two sequence numbers on the difference between them, as RFC 1982
does. `get_stream` reads it for the gap test and the overlap arithmetic, and it masks
`next_seq` to 32 bits.

`_ordered_segments` gives the segment order. A stream occupies one arc of the sequence
space, and one step between two neighbours closes that arc. The method sorts on the raw
number, finds the widest step, and starts the stream after it. `get_stream` and
`base_seq` both read that order.

`add_segment` detects a duplicate against a set of `(seq, length)` pairs.

The stored `base_seq` value is removed. The `base_seq` **method** stays, because
`ja4plus/fingerprinters/ja4x.py:166` reads it.

## Why the first ordering attempt was wrong

The self-review rejected a simpler form that compared each segment against a running
earliest value. `_seq_before` stops being transitive once the segments span more than
half the sequence space, so that form returned a different first segment for a different
arrival order. Three segments at `0x00000000`, `0x60000000` and `0xC0000000` gave three
different answers across the six arrival orders. `.claude/rules/conformance.md` forbids a
value that depends on the order the capture is read.
`test_the_segment_order_does_not_depend_on_the_arrival_order` holds that line.

Nothing bounds the spread of the stored sequence numbers today, because
`max_stream_bytes` bounds the reassembled output and not the stored segments. That is
row 3 of the register, which is #33.

## How this was tested

Every new test was proven to fail against the code as it stood. The failures:

| Test | Failure before the fix |
|---|---|
| `test_reassembly_across_the_sequence_wrap_keeps_the_byte_order` | `AssertionError: b'efgh' != b'abcdefgh'` |
| `test_a_late_segment_across_the_sequence_wrap_keeps_the_byte_order` | `AssertionError: b'efgh' != b'abcdefgh'` |
| `test_a_gap_across_the_sequence_wrap_stops_the_stream` | `AssertionError: b'efgh' != b'abcd'` |
| `test_an_overlap_across_the_sequence_wrap_drops_the_repeated_bytes` | `AssertionError: b'def' != b'abcdef'` |
| `test_the_base_sequence_across_the_sequence_wrap_names_the_first_byte` | `AssertionError: 2 != 4294967294` |
| `test_the_cost_of_a_segment_does_not_grow_with_the_segment_count` | `AssertionError: 1.77643425000133 not less than 0.8767423359677196` |
| `test_the_segment_order_does_not_depend_on_the_arrival_order` | `AssertionError: 3 != 1` |
| `test_a_trimmed_segment_reassembles_when_it_arrives_again` | fails without the `seen` rebuild in `trim_stream` |

`WRAP_SEQ` is `0xFFFFFFFE`, so the four bytes of the first segment sit at `0xFFFFFFFE`,
`0xFFFFFFFF`, `0x00000000` and `0x00000001`. The segments cross the boundary.

### The timing criterion, measured

The issue asks for 10000 segments in under one second. **That criterion already passed
against the unfixed code**, so it is weak evidence on its own:

| Segments | Before | After |
|---|---|---|
| 5000 | 0.1096 s | 0.0010 s |
| 10000 | 0.4511 s | 0.0018 s |
| 20000 | 1.7652 s | 0.0038 s |

Before, four times the segments cost sixteen times the time. After, it costs four times.
`test_the_cost_of_a_segment_does_not_grow_with_the_segment_count` measures the growth
rather than one absolute number, and it fails against the unfixed code.

## Gates

| Gate | Result |
|---|---|
| `pytest tests/ -m "not spec_validation"` | 977 passed, 8 xfailed, 93 subtests passed |
| `pytest tests/ -m spec_validation` | 1375 passed, 146 skipped, 120 xfailed |
| `ruff format --check ja4plus/ tests/` | 100 files already formatted |

**No fingerprint moved.** The conformance suite reports `1375 passed, 146 skipped, 120
xfailed` both before and after the change, against 120 keys in
`tests/foxio_deviations.json`.

**Coverage rose.** `ja4plus/utils/tcp_stream.py` goes from 91% to 99%. The project total
holds at 80%, with 670 uncovered statements against 674 before.

## Two things the reviewer should check

1. **I touched `trim_stream`, which the brief fenced off for #33.** One statement only:
   it rebuilds the `seen` set from the surviving segments. Without it a trimmed segment
   that arrives again is dropped as a duplicate, which is a defect my own change would
   have introduced. I did **not** bound the segment list, which is #33's work. Revert
   that one statement if you would rather #33 own it.
2. **`trim_stream` still compares raw sequence numbers** at `seq + len(data) > up_to_seq`,
   so it holds the same wrap defect. It has no caller and no test today. I left it, and
   flag it here for whoever takes #33.

No `Closes #` keyword: this is a batch member, and the batch PR closes the issue.

Verified against: https://datatracker.ietf.org/doc/html/rfc1982 (RFC 1982, serial number arithmetic, retrieved 2026-08-07)

### DIFF
--- a/docs/implementation_notes.md
+++ b/docs/implementation_notes.md
@@ -905,13 +905,22 @@ numbers puts every segment after the wrap point before every segment before it, 
 connection that crosses the boundary reassembles backwards.
 
 `_seq_before` orders two sequence numbers on the difference between them, as RFC 1982
-does. `_ordered_segments` finds the earliest sequence number of the stream, then sorts
-on the distance from it. `get_stream` and `base_seq` both read that order, so the gap
-test, the overlap arithmetic and the reported base all hold across the wrap.
+does. `get_stream` reads it for the gap test and the overlap arithmetic, so both hold
+across the wrap.
 
-The method needs every sequence number of one stream to lie within 2**31 of the others.
-`max_stream_bytes` bounds a stream at 1048576 bytes by default, which meets that
-condition with a wide margin.
+`_ordered_segments` gives the segment order. A stream occupies one arc of the sequence
+space, and one step between two neighbours closes that arc. The method sorts the
+segments on the raw number, finds the widest step, and starts the stream at the segment
+after it. `get_stream` and `base_seq` both read that order.
+
+A comparison of each segment against a running earliest value looks equivalent and is
+not. `_seq_before` stops being transitive once the segments span more than half the
+sequence space, so that form returns a different first segment for a different arrival
+order. Nothing bounds the spread of the stored sequence numbers today, because
+`max_stream_bytes` bounds the reassembled output and not the stored segments. Row 3 of
+the audit register holds that defect. The widest-step reading depends only on the
+sequence numbers, so the reassembled bytes never depend on the order the capture
+delivered the segments.
 
 `add_segment` detects a duplicate against a set of `(seq, length)` pairs. The earlier
 scan of every stored segment cost the square of the segment count: 10000 segments took
--- a/docs/implementation_notes.md
+++ b/docs/implementation_notes.md
@@ -898,6 +898,27 @@ have no counterpart here.
 for correct ordering. Prior versions appended data in arrival order,
 which could corrupt streams with out-of-order TCP segments.
 
+### The order holds across a sequence wrap
+
+A TCP sequence number is 32 bits and wraps back to zero. An order that compares the raw
+numbers puts every segment after the wrap point before every segment before it, so a
+connection that crosses the boundary reassembles backwards.
+
+`_seq_before` orders two sequence numbers on the difference between them, as RFC 1982
+does. `_ordered_segments` finds the earliest sequence number of the stream, then sorts
+on the distance from it. `get_stream` and `base_seq` both read that order, so the gap
+test, the overlap arithmetic and the reported base all hold across the wrap.
+
+The method needs every sequence number of one stream to lie within 2**31 of the others.
+`max_stream_bytes` bounds a stream at 1048576 bytes by default, which meets that
+condition with a wide margin.
+
+`add_segment` detects a duplicate against a set of `(seq, length)` pairs. The earlier
+scan of every stored segment cost the square of the segment count: 10000 segments took
+0.451 s, and 20000 took 1.765 s. The set gives 0.0019 s and 0.004 s.
+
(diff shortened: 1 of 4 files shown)


Crank-Git:
## Self-review

Two lenses ran against the diff: correctness, and hostile input plus resource safety.

### Correctness — one defect found and fixed

The first ordering attempt compared each segment against a running earliest value.
`_seq_before` stops being transitive once the segments span more than half the sequence
space, so the answer moved with the arrival order. Three segments at `0x00000000`,
`0x60000000` and `0xC0000000` gave three different first segments across the six arrival
orders.

`.claude/rules/conformance.md` forbids a value that depends on the order the capture is
read. `_ordered_segments` now reads the widest step between two neighbours in raw order,
which depends only on the sequence numbers.
`test_the_segment_order_does_not_depend_on_the_arrival_order` fails against the rejected
form with `AssertionError: 3 != 1`.

Commit `5948db8` holds the fix.

### Hostile input and resource safety — clean

- No new path raises. `add_segment` returns early on empty data, and the mask arithmetic
  accepts any Python integer.
- The `max_stream_bytes` truncation in `get_stream` is untouched and still bounds the
  return value.
- No infinite loop. The duplicate check falls from O(n) per segment to O(1).

### One finding rejected, with the reason

The lens reported that the segment count within one stream has no bound, which reads
against the `CLAUDE.md` rule that nothing grows without a limit.

**Rejected as out of scope, not as wrong.** The defect is pre-existing: `max_stream_bytes`
bounds the reassembled output and not the stored segments, and `trim_stream` has no
caller. It is row 3 of the audit register in
`docs/specs/features/02-correctness-audit.md`, which #33 owns and which runs on this file
after this PR. The `seen` set grows in step with the segment list and adds no new class of
growth.

Issue #156, opened 2026-08-07
Suppress a JA4L value until every measurement point is read


Crank-Git:
Found by the measurement in #123. It is not the divergence #123 named, and #123 closed
without it.

## The gap

`ja4plus` emits each JA4L value as it reads it. The FoxIO Rust implementation emits a value
pair only when it has collected all four measurement points. On a connection that never
completes, the two disagree: `ja4plus` reports a partial value and the reference reports
nothing.

## The measurement

The FoxIO Rust implementation at commit `27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8`, the
commit `tests/download_test_vectors.py:17` pins. On a synthetic QUIC capture whose server
Initial packet leads its client Initial packet:

```
ja4 quic_mirrored.pcap
  []
```

`ja4plus` reports `JA4L-C=500_64` on the same capture.

`rust/ja4/src/time/udp.rs` states the mechanism: the first state discards a `ServerInitial`,
and `finish()` returns `None` unless the state machine reaches `Done`.

The capture builder is committed at `tests/build_quic_ja4l_captures.py`, from #123.

## Why it is wider than #123

The reference rule is not about QUIC. It is about when a JA4L value exists at all. Suppressing
a partial value touches **every** JA4L path, the TCP one included, and every connection that
a capture cuts short.

## What to build

1. **Measure the TCP path before you change it.** The QUIC case above is measured. Build the
   matching TCP case — a capture holding a SYN with no SYN-ACK, and one holding a SYN-ACK with
   no ACK — and record what the reference reports on each.
2. **Check the committed vectors first.** A capture in `tests/foxio_vectors/` that holds a
   cut-short connection already decides this, and it decides it better than a synthetic one.
   Search before you build.
3. **Decide the rule from the measurement**, and state it: does a JA4L value exist only when
   all four measurement points are read, on every transport.
4. **Count the vectors the change moves.** If it moves a reference value that matches today,
   stop and return `needs-feedback`.

## Acceptance criteria

- [ ] The issue holds the reference output for the TCP cut-short cases, with the command.
- [ ] Any committed vector that holds a cut-short connection is named, and its reference
      value is quoted.
- [ ] The rule is stated in `docs/implementation_notes.md`, under `JA4L`, and it names every
      transport it binds.
- [ ] The two tests #123 landed in `tests/test_ja4l_quic_repeated_connection.py` still pass,
      or the change to them is justified from the reference.
[...]

PR #169, opened 2026-08-07
Suppress a JA4L value until every measurement point is read

Crank-Git:
## What

Two things, and only one of them changes a fingerprint.

1. **A fix.** `ja4plus` reported a JA4L client value on a QUIC connection whose server
   Initial packet leads its client Initial packet. The reference reports no value there.
   The QUIC client value now needs point `B`.
2. **A divergence made measurable.** `process_packet` returns a client value for every
   packet that moves the client measurement point, and the reference reports one per
   connection. The user decided on 2026-08-07 to record that divergence and to leave the
   return path as it is. `tests/test_ja4l_return_path.py` now measures it.

Part of #156.

## Why the issue's stated rule did not land

The body of #156 asks for one rule: a JA4L value exists only when the fingerprinter
reads all four measurement points. The committed vectors reject it.

**The reference gates the two values apart.** Two committed QUIC connections hold a
`JA4L-S` and no `JA4L-C`:

| Vector | Stream | Reference `JA4L-S` | Reference `JA4L-C` |
|---|---|---|---|
| `ssh2.pcapng` | 33, port 51810 | `16192_57` | none |
| `tls3.pcapng` | 25, port 61884 | `3583_57` | none |

I implemented the literal rule to measure its cost. It **breaks 80 conformance cases
that pass today** and deletes both values above:

```
$ pytest tests/ -m spec_validation -q
FAILED tests/test_spec_validation.py::test_the_produced_occurrence_keys_equal_the_reference[v6.pcap-JA4L-S]
80 failed, 1295 passed, 149 skipped, 980 deselected, 117 xfailed
```

A JA4L value is gated per side. `docs/implementation_notes.md` now states that, with
this output as the evidence.

**The TCP cut-short case already matched.** `ssh2.pcapng` holds 11 TCP connections that
carry a SYN and no SYN-ACK. The expected-output file names none of them, and `ja4plus`
produces no value for any of them. No committed vector holds a TCP connection that
carries a SYN-ACK and no client measurement point.

## The return-path divergence, and why it is now a comparison that runs

`tests/conformance_index.py` reads the stored list. The stored list holds 60 client
values and is correct, so **no conformance case read the return value of
`process_packet` at all**. The divergence was invisible.

`tests/test_ja4l_return_path.py` holds the returned count and the stored count of every
committed vector, and the totals of 105 against 60. It fails when either number moves.

The three candidate end-of-connection points I measured against the 60 values:

| The end-of-connection point | Client values it reports |
|---|---|
| A FIN or a RST after the final client point | 26, and it loses 34 |
| Every measurement point read | 105, which is the behaviour today |
| A read of the stored list at the end of a capture | 60, and it loses none |

34 of the 60 connections never close inside their capture. `macos_tcp_flags.pcap` holds
no FIN packet and no RST packet at all.

`docs/specs/spec.md` § Divergence register carries the row, and the decision is
reversible.

## Why `tests/foxio_deviations.json` gains no entry

The harness reads the stored list, and the stored list is correct, so every case passes.
A strict entry reports `XPASS(strict)` and fails the suite. I proved it with a probe
entry for `latest.pcapng/JA4L-C`:

```
[XPASS(strict)] issue #156: Probe: the return path reports 11 client values and the stored list holds 6.
FAILED tests/test_spec_validation.py::test_the_produced_occurrence_keys_equal_the_reference[latest.pcapng-JA4L-C]
```

The register holds 120 keys before and after. The spec row and the new test carry the
divergence instead.

## How I tested it

Every new test fails against a mutation of the code as it stands. The failures are
quoted in the issue comments.

```
pytest tests/ -m "not spec_validation"   1026 passed, 8 xfailed
ruff format --check ja4plus/ tests/      101 files already formatted
mypy ja4plus/                            no issues found in 27 source files
coverage                                 81% total, ja4l.py 97%
```

120 xfailed against the 120 keys of `tests/foxio_deviations.json`. No reference value
that matched before this change stops matching: the conformance result is byte-identical
to the base.

Rebased onto `3e3aa57` (#33). This branch touches neither `ja4plus/utils/tcp_stream.py`
nor `ja4plus/fingerprinters/ja4h.py`.

Verified against: https://github.com/FoxIO-LLC/ja4/tree/main/python/test/testdata (retrieved 2026-08-07)


Crank-Git:
## Self-review, and what it changed

I reviewed the diff by two lenses in parallel.

**Correctness.** The reviewer asked the question that matters for the new guard: can
`if "B" not in timestamps` suppress a client value on a QUIC connection whose server
Initial packet arrives normally but does not decrypt? It swept every expected-output
file and cross-checked each one against the produced output.

- 56 reference `JA4L-C` entries across the vectors: 54 TCP, 2 QUIC.
- Both QUIC ones decrypt, so both set point `B`, and the guard does not fire on either.
- 0 connections where the reference holds a `JA4L-C` and `ja4plus` cannot set `B`.

The early return of `_quic_server_initial` on a decryption failure could strand a
legitimate point `B`, but no vector exercises it. `.claude/rules/conformance.md` states
that a defect is proven and never asserted, so it stays unfiled.

Verdict: no defect. The guard sits after the `QUIC_HANDSHAKE` check and before the code
that records point `C` or point `D`, and it touches no TCP path.

**The writing standard.** The reviewer found four real violations, and `49580f6` fixes
all of them.

- Three sentences above the 25-word cap for a description, at
  `ja4plus/fingerprinters/ja4l.py:542` and two places in `docs/implementation_notes.md`.
- `stream` used where the `## Terms` table names `connection`, in one test name and two
  sentences. A `stream` that names a reference index stays, because the expected-output
  file holds that key verbatim.
- The dangling `-ing` phrase "The four measurement points completing" in a table cell,
  now "Every measurement point read".

The gate is green after the fixes: 972 passed and 8 xfailed on the unit suite, 1375
passed and 120 xfailed on the conformance suite, and ruff, ruff format and mypy clean.

Issue #170, opened 2026-08-07
Give a TCP reassembler stream a maximum age


Crank-Git:
Found during the self-review of #33. It is not a defect that #33 introduced, and it is outside the scope of that issue.

## What is wrong

`CLAUDE.md` states the rule: a state table has a maximum entry count and a maximum age,
and nothing grows without a limit. `ja4plus/utils/tcp_stream.py` now meets the first half
and not the second.

- `max_streams` bounds the count of streams, and the oldest stream leaves first.
- `max_stream_bytes` and `max_stream_segments` bound one stream, which #33 built.
- No field holds a timestamp, and no method removes a stream on its age.

A connection that sends one segment and then stops holds its slot until 100 further
connections push it out. A fingerprinter that reads a long capture therefore holds state
for connections the capture finished with long before.

## What to build

Give a stream entry the packet timestamp of its most recent segment, and remove an entry
that passes a maximum age. `.claude/rules/conformance.md` names the reading:
"Eviction uses the packet timestamp when the packet carries one. A capture file replays
faster than real time, and a wall clock would evict state the capture still needs."

`add_segment` takes no timestamp today, so the change reaches its two callers,
`ja4plus/fingerprinters/ja4h.py` and `ja4plus/fingerprinters/ja4x.py`.

## Acceptance criteria

- [ ] A stream that receives no segment for the maximum age leaves the reassembler.
- [ ] The eviction reads the packet timestamp, never the wall clock.
- [ ] The FoxIO conformance suite reports the same counts before the change and after it.

Crank-Git:
## Plan

Batch 13, on `epic/13-batch13`. #35 merged first as `6915bad`, and it changed
`ja4plus/fingerprinters/ja4h.py`, so branch off the updated integration branch.

**This issue completes the `CLAUDE.md` state rule.** #33 built the maximum entry count for
one stream, and #78 and #32 built the rest of the bounds. No field holds a timestamp yet,
so the maximum age is the half that is missing.

**Approach.**

1. Give a stream entry the packet timestamp of its most recent segment.
2. Remove an entry that passes a maximum age.
3. **The eviction reads the packet timestamp, never the wall clock.**
   `.claude/rules/conformance.md` states the reading: a capture file replays faster than
   real time, and a wall clock would evict state the capture still needs.
4. `add_segment` takes no timestamp today, so the change reaches its two callers,
[...]

Issue #103, opened 2026-08-07
Bound the segment list of one TCP stream


Crank-Git:
## What is wrong

`TCPStreamReassembler.add_segment` appends to `stream["segments"]` with no limit on
the number of entries. `max_streams` bounds the number of streams, and
`max_stream_bytes` bounds only the bytes `get_stream` returns. Neither bounds the
list of `(seq, data)` tuples one stream holds.

`ja4plus/utils/tcp_stream.py:25` holds the append. `ja4plus/utils/tcp_stream.py:44`
holds `get_stream`, which sorts that list on every call.

The cost is quadratic in the packet count of one stream:

- `add_segment` scans every existing segment to reject a duplicate.
- `get_stream` sorts every segment on every packet.

A measurement of 4000 single-byte segments on one stream shows the marginal cost of
one packet rise from 0.06 ms to 0.12 ms.

`trim_stream` exists to close this, and nothing calls it. The one call site passed a
byte offset where the method expects an absolute sequence number, so it removed
nothing for a realistic initial sequence number. #78 removed that call because it
was a no-op.

## What to build

Give one stream a maximum segment count, or call `trim_stream` with an absolute
sequence number after a reader consumes a prefix. The state rule in `CLAUDE.md`
applies: a state table has a maximum entry count and a maximum age, and nothing
grows without a limit.

JA4H and JA4X both use the reassembler, so a change here affects both.

## Evidence

Found during the self-review of #78. It is not a defect that #78 introduced, and it
is outside the scope of that issue.

Crank-Git:
Folded into #33 by the user's decision on 2026-08-07.

#33 and this issue name the same defect in `ja4plus/utils/tcp_stream.py`. #33 quotes `FR-correctness-audit-2`, which requires that the reassembler holds no more than a fixed number of segments for one connection. This issue holds the sharper diagnosis: the quadratic cost, the measurement of 4000 single-byte segments, and the `trim_stream` call site that #78 removed because a byte offset where the method expects an absolute sequence number removed nothing.

#33 stays the issue of record, because the requirement traces through it. The plan for #33 carries the measurement and the `trim_stream` history from this issue, so nothing here is lost. Two workers on one file would have raced.

This issue closes when #33 lands in Epic 2 batch 12.

Crank-Git:
Closed by #33, which landed as `3e3aa57` in Epic 2 batch 12.

[...]

PR #171, opened 2026-08-07
Bound the bytes and the segments one TCP stream holds

Crank-Git:
Refs #33. This issue absorbed #103, which named the same defect.

## What is wrong

`max_stream_bytes` bounded the bytes `get_stream` returns, and nothing bounded the bytes
one stream stores. `max_streams` bounded the count of streams only. A sender of many
distinct small segments therefore grew one stream without a limit. `ja4h.py` left a
buffer that is not an HTTP request in the reassembler for the life of the capture.
`trim_stream` compared raw sequence numbers, which is the defect #32 removed from
`get_stream`.

## What this builds

1. `add_segment` refuses a segment once the stream holds `max_stream_bytes` bytes, or
   `max_stream_segments` segments. The default segment cap is 4096.
2. A refused segment enters neither the segment list nor the set of seen segments, so the
   stream accepts it again after a trim frees the room.
3. `trim_stream` reads `_seq_before`, so it orders the way `get_stream` does. It
   recomputes the stored byte count beside the set of seen segments. Its docstring states
   that it takes an absolute sequence number, never a byte offset, because #78 removed
   its one call site over that mistake. This change gives it no new caller.
4. `get_stream` no longer truncates its result. The stream stores no more than
   `max_stream_bytes`, and the result never holds more than the stream stores, so the
   truncation is unreachable.
5. JA4H removes a stream whose buffer no later byte turns into an HTTP request.
   `can_become_http_request` keeps a buffer that is a prefix of a method token, because a
   request line spans two segments.

## The measurement of the vectors

The largest stream of the FoxIO vectors holds **1853328 bytes and 1336 segments**, both
on key `142.250.187.206:443-192.168.2.200:58847` of `http2-with-cookies.pcapng`. The
measurement runs `Processor` over every capture under `tests/foxio_vectors/` and reads
the stored segments after each call to `add_segment`.

The byte cap of 1048576 cuts that one stream, and no reference value moves, because
`get_stream` already truncated its result at the same number. After the change the peak
readings are 1048554 bytes and 788 segments. The segment cap of 4096 sits three times
above the larger segment reading, so no vector reaches it.

## The out-of-order defect this pull request found in itself

The first form of the JA4H removal dropped the stream on the first packet whose buffer
cannot become an HTTP request. Two review passes found the same defect: a segment that
arrives out of order puts the middle of a request at the start of the buffer, and that
buffer starts no HTTP request. The removal then dropped a segment the request needs, and
the request never reassembled when the head arrived.

The measurement, with the removal on the first packet:

```
tail first  -> None
head second -> None
```

The head lowers the base sequence number when it arrives, so `_drop_an_unusable_stream`
removes the stream only on a packet that leaves the base sequence number where the
previous packet left it. `tests/test_ja4h_reassembly.py` holds both cases:
`test_a_reordered_request_head_still_produces_a_fingerprint` and
`test_a_buffer_that_cannot_become_an_http_request_leaves_the_reassembler`.

`self.unusable_base` holds one base sequence number for each waiting stream, and it drops
the keys the reassembler no longer holds once it passes `max_streams` entries, the way
`ja4x.py` prunes `scan_offsets`.
`test_the_table_of_waiting_streams_stays_inside_the_stream_cap` measures that bound over
300 connections.

## Every new test fails against the code as it stands

```
TypeError: TCPStreamReassembler.__init__() got an unexpected keyword argument 'max_stream_segments'. Did you mean 'max_stream_bytes'?
AssertionError: 2000 not less than or equal to 100
AttributeError: 'TCPStreamReassembler' object has no attribute 'max_stream_segments'. Did you mean: 'max_stream_bytes'?
KeyError: 'bytes'
AssertionError: b'abcdefgh' != b'efgh'
AssertionError: OrderedDict({'10.0.0.1:12345-10.0.0.2:443[240 chars])}}}) != {}
```

`test_max_stream_size` already read `assertLessEqual(len(result), 20)` against a cap of
10, so it passed whatever the reassembler did. It now names 10.

## How this was tested

```
.venv/bin/pytest tests/ -m "not spec_validation"   994 passed, 8 xfailed, 93 subtests passed
.venv/bin/pytest tests/ -m spec_validation         1375 passed, 146 skipped, 120 xfailed
.venv/bin/ruff format --check ja4plus/ tests/      100 files already formatted
```

The conformance suite reports 120 xfailed against the 120 keys of
`tests/foxio_deviations.json`, unchanged from the base commit `c4fec95`.

Coverage holds: `tcp_stream.py` 99 percent before and after, `http_utils.py` 89 to 90,
`ja4h.py` 84 to 86, and the project total 80 percent with 670 missed statements before
and after.

## What this does not build

- **The reassembler still has no maximum age.** The project state rule names a maximum
  entry count and a maximum age. This change bounds the count, the bytes and the
  segments. A connection that never closes holds its slot until `max_streams` evicts it.
  That gap is outside this issue, and #170 records it.
- **JA4X inherits the segment cap of 4096 and overrides no value.** A certificate scan
  reads at most `MAX_SCAN_BYTES`, which is 200000 bytes, so the cap bites only when the
  stored segments average under about 49 bytes. That is the hostile case the bound exists
  for. The largest vector stream holds 788 segments.

Verified against: https://github.com/FoxIO-LLC/ja4/tree/main/python/test/testdata
(retrieved 2026-08-07)

### DIFF
--- a/docs/implementation_notes.md
+++ b/docs/implementation_notes.md
@@ -916,9 +916,8 @@ after it. `get_stream` and `base_seq` both read that order.
 A comparison of each segment against a running earliest value looks equivalent and is
 not. `_seq_before` stops being transitive once the segments span more than half the
 sequence space, so that form returns a different first segment for a different arrival
-order. Nothing bounds the spread of the stored sequence numbers today, because
-`max_stream_bytes` bounds the reassembled output and not the stored segments. Row 3 of
-the audit register holds that defect. The widest-step reading depends only on the
+order. `max_stream_bytes` now bounds the stored segments, so the spread of the stored
+sequence numbers stays inside one arc. The widest-step reading depends only on the
 sequence numbers, so the reassembled bytes never depend on the order the capture
 delivered the segments.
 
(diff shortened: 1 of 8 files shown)


Crank-Git:
Self-review, two passes over the diff: one for correctness and one for resource bounds.

**Blocking, accepted and fixed.** `ja4plus/fingerprinters/ja4h.py` removed a stream on the first packet whose buffer cannot become an HTTP request. Both passes reached the same case: a segment that arrives out of order puts the middle of a request at the start of the buffer, so the removal drops a segment the request needs. The removal now waits for one packet that leaves the base sequence number where the previous packet left it. The pull-request body holds the measurement.

**Not a defect.** `_seq_before(up_to_seq, (seq + len(data)) & _SEQ_MASK)` in `trim_stream` holds both boundaries. A segment that ends exactly at `up_to_seq` leaves, because `_seq_before` reports False on two equal numbers. A segment that ends one byte later stays.

**Not a defect.** The removed truncation in `get_stream` is unreachable. The result grows by `len(data) - overlap` for each segment, which is never more than `len(data)`, so it never passes the stored byte count. `add_segment` holds the stored byte count at or under `max_stream_bytes`.

**Not a defect.** A refused segment enters neither `segments` nor `seen`, because both cap checks return before the two statements that write them. `test_a_refused_segment_reassembles_when_the_stream_has_room_again` measures it.

**Minor, filed as #170.** The reassembler has a maximum entry count and no maximum age. The state rule in `CLAUDE.md` names both. The gap is older than this change and outside the scope of #33.

**Minor, declined.** JA4X inherits the segment cap of 4096 and overrides no value, so a heavily fragmented certificate scan reaches the cap before it reaches `MAX_SCAN_BYTES`. That needs a mean stored segment under about 49 bytes, which is the hostile case the bound exists for. The largest vector stream holds 788 segments. A separate cap for JA4X adds a second number to tune and no vector asks for it.

PR #174, opened 2026-08-07
Build both JA4H cookie hashes from one cookie list

Crank-Git:
## What

JA4H builds the cookie-name hash and the cookie-value hash from one ordered list of
`(name, value)` pairs. The request-line pattern accepts an optional minor version.

Part of #35. Part of #13.

## Why

Two defects in `ja4plus/fingerprinters/ja4h.py`, rows 7 and 8 of the audit register:

- The cookie-value hash read the `cookies` dictionary, which kept the last value of a
  repeated cookie name. The cookie-name hash read a list that kept every occurrence. The
  two hashes then described different cookie sets. HTTP permits a repeated cookie name.
- The request-line pattern required `HTTP/\d+\.\d+`, so the parser read no request line
  that reads `GET / HTTP/2`, although `_http_version_to_str` maps `2` to the version
  code `20`.

`ja4plus/utils/http_utils.py` held the same request-line pattern, and
`parse_http_request` reported cookies as a dictionary alone. Both now match.

## How the change keeps the fingerprint

The FoxIO Python implementation collects the cookies into a list and sorts that list on
the cookie name alone:

```python
sorted_pairs = sorted(cookie_pairs, key=lambda p: p[0])
x['cookie_fields'] = [pair[0] for pair in sorted_pairs]
x['cookie_values'] = [pair[1] for pair in sorted_pairs]
```

`ja4plus` sorts the same way. The sort is stable, so two cookies that carry one name
keep their wire order.

Verified against: https://github.com/FoxIO-LLC/ja4/blob/main/python/ja4h.py (retrieved 2026-08-07)

**No reference value moves.** Every capture under `tests/foxio_vectors/` ran through
`JA4HFingerprinter` before the change and after it. The vector set produces 82 JA4H
values, and the two runs are equal, fingerprint and raw form alike. The count of
reference values that moved is **0**.

## How it is tested

`tests/test_ja4h_cookie_list.py` holds ten tests. The tests landed first. Six failed
against the code as it stood:

```
FAILED tests/test_ja4h_cookie_list.py::test_a_repeated_cookie_name_reaches_both_cookie_hashes
FAILED tests/test_ja4h_cookie_list.py::test_a_repeated_cookie_name_reaches_both_hashes_of_the_packet_path
FAILED tests/test_ja4h_cookie_list.py::test_the_cookie_value_hash_reads_the_same_cookie_list_as_the_raw_form
FAILED tests/test_ja4h_cookie_list.py::test_a_request_line_without_a_minor_version_produces_the_version_code_20
FAILED tests/test_ja4h_cookie_list.py::test_the_packet_path_reads_a_request_line_without_a_minor_version
FAILED tests/test_ja4h_cookie_list.py::test_the_parsed_request_path_keeps_a_repeated_cookie_name
6 failed, 3 passed, 1 warning in 0.29s
```

The first failure reads:

```
        assert part_c == _hash("a,a")
>       assert part_d == _hash("a=1,a=2")
E       AssertionError: assert 'd3043f41a038' == '2449bc002712'
```

The tenth test came from the self-review. `HTTP/\d+(?:\.\d+)?` read `HTTP/11` as
`HTTP/1`, so `GET / HTTP/11` and `GET / HTTP/1.1` carried one fingerprint. The version
token now holds a minor version, or it names major version 2 or 3, and a space, a tab or
a line terminator ends it. `ja4plus/utils/http_utils.py` holds it as
`REQUEST_LINE_PATTERN`, and both parse paths read that one name.

The three that passed are the regression guards: a unique cookie name keeps its
fingerprint, `HTTP/1.1` keeps the version code `11`, and the packet path and the stream
path report one cookie list.

## The five gates

```
.venv/bin/pytest tests/ -m "not spec_validation"   1036 passed, 8 xfailed, 93 subtests passed
.venv/bin/pytest tests/ -m spec_validation         1375 passed, 146 skipped, 120 xfailed
.venv/bin/ruff format --check ja4plus/ tests/      102 files already formatted
```

`120 xfailed` against the 120 keys of `tests/foxio_deviations.json`. The baseline held
`1026 passed` on the unit suite, and the ten new tests account for the difference.

Coverage rises: `ja4h.py` 87% to 89%, `http_utils.py` 90% to 91%, total 81%.

## Out of scope

The self-review found that the cookie parser caps neither the pair count nor the length
of a name or a value. The growth predates this pull request, so #175 records it.

## Docs

`docs/implementation_notes.md` records both readings. `docs/specs/features/02-correctness-audit.md`
marks rows 7 and 8 of the audit register as built. The `docs/specs/spec.md` Changelog
round belongs to the project manager, so this pull request writes none.


Crank-Git:
## Self-review

I reviewed the diff by two lenses, correctness and robustness, and ran both against the
code. The review found one real defect that I introduced, and one pre-existing defect
outside the scope of this issue.

### The widened pattern collided two versions — fixed in 37d6d72

`HTTP/\d+(?:\.\d+)?` read `HTTP/11` as `HTTP/1` and reported the version code `11`, so
`GET / HTTP/11` and `GET / HTTP/1.1` carried one fingerprint:

```
'GET / HTTP/11'  ge11nn010000_4a823118b9ba_000000000000_000000000000
'GET / HTTP/1.1' ge11nn010000_4a823118b9ba_000000000000_000000000000
```

The pattern before this pull request produced nothing for `HTTP/11`, so the change
turned a missing fingerprint into a wrong one. A fingerprint that describes malformed
traffic as ordinary traffic is worse than no fingerprint.

The version token now holds a minor version, or it names major version 2 or 3, and a
space, a tab or a line terminator ends it.
`test_a_request_line_that_names_no_version_produces_no_fingerprint` covers `HTTP/11`,
`HTTP/1`, `HTTP/23` and `HTTP/`. The test failed before the fix:

```
>           assert _extract_http_info_from_bytes(data) is None, request_line
E           AssertionError: GET / HTTP/11
E           assert {'method': 'GET', 'path': '/', 'version': 'HTTP/11', ...} is None
```

### The sort matches the FoxIO reference — no change needed

The review asked whether `sorted(name for name, _ in pairs)` for part c can differ from
the FoxIO form, which sorts the pairs and then reads the names. It cannot: a tie holds
one identical name, so either tie-break yields one list. A 200000-trial comparison over
inputs that repeat cookie names found no difference. Part d uses
`sorted(cookie_pairs, key=lambda pair: pair[0])`, which is the FoxIO expression.

### The cookie list has no maximum entry count — out of scope

A Cookie header that repeats one pair 100000 times builds a list of 100000 pairs. The
project rule reads "A state table has a maximum entry count and a maximum age. Nothing
grows without a limit." The review measured the same growth on the code before this
pull request, with a header of 100000 unique cookie names, so the defect predates the
change. I filed it rather than widening this issue.

### Rejected: no fallback to the cookies dictionary

`_ja4h_cookie_pairs` reads `cookie_fields` and `cookie_values` alone. A fallback to the
`cookies` dictionary would disagree with the cookie flag of part a, which already reads
`cookie_fields` alone. Every producer in the repository reports all three keys.

Crank-Git:
The cookie-list growth finding is filed as #175, untriaged. It predates this pull request and this pull request does not change it.

Issue #37, opened 2026-08-06
Add the malformed-input suite and run it on every pull request


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/02-correctness-audit.md`](docs/specs/features/02-correctness-audit.md):

> FR-correctness-audit-8 — No parser trusts a length field it read from the packet.

> FR-correctness-audit-9 — No fingerprinter raises when given a truncated packet.

> FR-correctness-audit-10 — No fingerprinter raises when given a packet whose declared length exceeds its actual length.

> FR-correctness-audit-13 — The malformed-input suite runs on every pull request.

## What to build

Treat every packet as hostile input. No parser trusts a length field it reads from the packet.

Four FoxIO captures exist for this purpose and are already committed by Epic 0:

| Capture | What it holds |
|---|---|
| `ssh2-malformed.pcap` | An SSH session with malformed records. |
| `ssh2-moloch-crash.pcap` | An SSH capture that crashed another implementation. |
| `badcurveball.pcap` | A TLS handshake with a malformed curve list. |
| `CVE-2018-6794.pcap` | A capture that exercises a reassembly bypass. |

Generate the truncated copies at test time from the committed captures. Do not commit them.

This is the last issue in the epic and proves the whole register.

## Acceptance criteria

- [ ] `pytest tests/fuzz/` passes.
- [ ] A TLS record that declares a length larger than the packet produces no exception and no fingerprint.
- [ ] An extension list that declares more extensions than the packet carries produces no exception.
- [ ] Every FoxIO malformed capture produces no exception.
- [ ] A truncated copy of every FoxIO capture produces no exception.

## Links

- Feature: [`docs/specs/features/02-correctness-audit.md`](docs/specs/features/02-correctness-audit.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #13
Depends on #36

Crank-Git:
## Plan

Batch 13, on `epic/13-batch13`, off `dev` at `845065c`. This issue runs in parallel with
the JA4H chain, #35 then #170. It is additive, so it should not meet them.

**`tests/fuzz/` does not exist.** The project manager confirmed that. This issue builds it.

**Approach.**

[...]

PR #176, opened 2026-08-07
Add the malformed-input suite and run it on every pull request

Crank-Git:
Part of #13. Builds #37.

## What

Adds `tests/fuzz/`, the malformed-input suite, and a `fuzz` job that runs it on every
pull request.

The suite runs every fingerprinter over the four committed FoxIO malformed captures,
over a truncated copy of each, and over a ClientHello whose length fields declare more
bytes than the packet carries. It generates each malformed copy at test time and
commits no capture.

## Why

`FR-correctness-audit-8` to `FR-correctness-audit-13`. Every packet is hostile input.
No parser trusts a length field it read from the packet. A parser that cannot read a
packet returns nothing, and it does not raise.

## The risk this suite had to avoid

A suite that asserts "no exception" passes when the parser is never reached. This
project has already found eleven assertions that cannot fail. Three guards keep this
suite from adding the twelfth:

1. **The harness observes a raise.** `Processor.process_packet` catches every
   exception, so a suite that drove the processor could never fail.
   `run_fingerprinters` reads each fingerprinter directly.
   `test_the_harness_reports_a_fingerprinter_that_raises` proves it.
2. **Every case names the parser it reaches.** `spy_on` counts the calls of that
   parser and fails on a count of zero. The counter patches the defining module and
   every module that bound the name, because a consumer that imports a name holds its
   own reference.
3. **Every case states what it produced.** `RECORDED_FINGERPRINTS` pins the value each
   capture produces. `SILENCED_METHOD` names a method that produces a value on the
   committed capture and none on the cut copy.

**Mutation check.** A parser patched to raise when a record declares more bytes than
the packet carries turns **34 cases red** across all four captures and all three
truncation fractions. The suite detects the defect it exists to detect.

## Measurements

| Capture | JA4 | JA4H | Parser calls, committed | Parser calls, cut |
|---|---|---|---|---|
| `badcurveball.pcap` | `t13d1615h2_46e7e9700bed_45f260be83e2` | — | `parse_tls_handshake` 8 | 8 |
| `CVE-2018-6794.pcap` | — | 12 values | `is_http_request` 18 | 18 |
| `ssh2-malformed.pcap` | — | — | `parse_ssh_packet` 15 | 13 |
| `ssh2-moloch-crash.pcap` | — | — | `parse_ssh_packet` 15 | 13 |

Under payload truncation JA4 drops to none on `badcurveball.pcap` and JA4H drops from
12 to none on `CVE-2018-6794.pcap`, while the parser call count stays above zero. The
parser is reached and returns nothing, which is the contract.

## No fingerprinter raises

The suite found no fingerprinter that raises on any of the four captures, on any
truncated copy, or on any mutated length field. No file under `ja4plus/` changed.

One finding sits in the harness, not in the library: `truncate_frames` never cuts below
the link-layer header, because scapy raises `struct.error` when it dissects a shorter
frame. Such a frame builds no packet object and reaches no fingerprinter, so a case
built on it would measure scapy instead of this project. `tests/fuzz/README.md` records
the reading.

## How tested

```
pytest tests/fuzz/                          106 passed
pytest tests/ -m "not spec_validation"      1132 passed, 8 xfailed
pytest tests/ -m spec_validation            1375 passed, 146 skipped, 120 xfailed
ruff check ja4plus/ tests/                  All checks passed
ruff format --check ja4plus/ tests/         108 files already formatted
mypy ja4plus/                               no issues in 27 source files
```

`120 xfailed` is unchanged, so the suite moves no fingerprint.
Line coverage rises from 80.08% to 80.14%.

Verified against: https://github.com/FoxIO-LLC/ja4/tree/main/pcap (retrieved 2026-08-07)
Verified against: https://www.rfc-editor.org/rfc/rfc8446 (TLS 1.3, sections 4 and 5.1)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -96,9 +96,12 @@ jobs:
           path: test-results.xml
 
   fuzz:
-    # `FR-correctness-audit-13`. The malformed-input suite runs on every pull request.
-    # The `test` job reads `tests/` and covers this suite too. This job names it
-    # separately, so a red malformed-input case is visible without a log search.
+    # `FR-correctness-audit-13`. The malformed-input suite runs on each pull request
+    # the `on:` block above accepts. This job names the suite, so a red
+    # malformed-input case is visible without a log search.
+    #
+    # This job runs one Python version. The `test` job reads `tests/`, so it runs this
+    # suite on every version of the matrix and on macOS.
     #
     # The suite cuts each committed capture at test time. Never commit a generated
     # capture: a committed copy stops following the capture it came from.
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -95,6 +95,40 @@ jobs:
           name: test-results-${{ matrix.os }}-py${{ matrix.python-version }}
           path: test-results.xml
 
+  fuzz:
+    # `FR-correctness-audit-13`. The malformed-input suite runs on every pull request.
+    # The `test` job reads `tests/` and covers this suite too. This job names it
+    # separately, so a red malformed-input case is visible without a log search.
+    #
+    # The suite cuts each committed capture at test time. Never commit a generated
+    # capture: a committed copy stops following the capture it came from.
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
+
+      - name: Set up Python
+        uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97  # v7.0.0
+        with:
+          python-version: "3.13"
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -e ".[dev]"
+
+      - name: Run the malformed-input suite
+        run: |
+          python -m pytest tests/fuzz/ -v --tb=short \
+            --junitxml=fuzz-results.xml
+
+      - name: Upload the malformed-input results
+        if: always()
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a  # v7.0.1
+        with:
+          name: fuzz-results
+          path: fuzz-results.xml
+
   conformance:
     # `FR-foundation-4`. The vectors are committed under `tests/foxio_vectors`, so this
     # job reads no network. Never add a step that runs `tests/download_test_vectors.py`:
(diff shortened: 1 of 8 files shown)


Crank-Git:
## Self-review

Two specialist lenses read the diff: correctness and robustness/CI.

### Falsifiability, checked twice by independent mutation

The one risk this suite had to avoid is an assertion that cannot fail. Two separate
mutations confirm the assertions are live:

1. **A parser that trusts a declared length.** `parse_tls_handshake` patched to raise
   when a record declares more bytes than the packet carries turns **34 cases red**,
   across all four captures and all three truncation fractions.
2. **The bound removed.** The correctness review removed the `end > len(raw_data)`
   guard in `ja4plus/utils/tls_utils.py` and reran the suite. Five cases fail:
   `test_a_record_that_declares_more_bytes_than_the_packet_produces_no_fingerprint`
   at each of `[12, 64, 256, 516]`, and
   `test_a_handshake_length_larger_than_the_packet_produces_no_fingerprint`.

The review also confirmed that the length-field offsets match the parser:
`RECORD_LENGTH_OFFSET = 3` matches the read at `tls_utils.py:67`,
`HANDSHAKE_LENGTH_OFFSET = 6` matches the read at `tls_utils.py:78`, and
`extensions_length_offset` reproduces the walk at `tls_utils.py:113-154`.

The spy covers every consumer of the three parsers it counts. `ja4s.py:11`,
`ja4ssh.py:14` and `ja4h.py:15` each bind their own reference, and `quic_utils.py`
imports inside a function body. The identity scan reaches all of them.

### One finding, fixed

The README claimed the `fuzz` job runs on "every pull request". The `on:` block accepts
a pull request that targets `master` or `dev`, so a pull request into an epic
integration branch starts no run. That is the batch model working as designed, and
changing the filter is not mine to do. Commit 45635cf corrects the wording and records
that the `test` job reads `tests/`, so the matrix runs this suite on Python 3.9 to 3.13
and on macOS while the `fuzz` job runs one version.

### Checked and clean

- **Python 3.9.** Every file in `tests/fuzz/` parses with `feature_version=(3, 9)`. No
  walrus, no `match`, no `X | Y` union, no new stdlib call. The `|` uses are bitwise OR
  on integers.
- **Isolation.** `packet.copy()` is a true deep copy, and every `spy_on` patch goes
  through `monkeypatch.setattr`, so pytest unwinds each one. The session-scoped
  fixtures are read-only for every consumer.
- **Determinism.** Every assertion compares sets or checks `is None`. No pinned value
  derives from a wall clock. The suite pins no JA4L value, because register row 5 stays
  open.
- **Runtime.** 106 cases in 0.57 s, slowest case 0.02 s.

Issue #177, opened 2026-08-07
Repair the 48 JA4L tests that fail under coverage


Crank-Git:
Found while measuring coverage for #170. It is not a defect that #170 introduced, and it is outside the scope of that issue.

## What is wrong

48 tests pass under `pytest tests/ -m "not spec_validation"` and fail under the same
command with `--cov`. The failure is present on `origin/epic/13-batch13` at `6915bad`,
before the #170 change.

```
.venv/bin/pytest tests/ -m "not spec_validation" -q
1048 passed, 1641 deselected, 8 xfailed, 93 subtests passed

.venv/bin/pytest tests/ -m "not spec_validation" -q --cov=ja4plus.utils.tcp_stream
48 failed, 988 passed, 1641 deselected, 8 xfailed, 93 subtests passed
```

Every failing test names JA4L or QUIC:

```
tests/test_ja4l_foxio.py
tests/test_ja4l_quic_repeated_connection.py
tests/test_ja4l_quic_server_hello.py
tests/test_ja4l_return_path.py
tests/test_ja4l_udp_direction.py
tests/test_ja4s_quic_direction.py
tests/test_ja4s_quic_split_server_hello.py
tests/test_quic_multipacket.py
tests/test_quic_server_initial.py
```

A sample failure:

```
FAILED tests/test_ja4l_foxio.py::TestJA4LAgainstTheFoxIOVectors::test_the_quic_form_reads_the_initial_and_the_handshake_packets
```

## Why it matters

A coverage measurement is the gate for the `coverage` practice. A suite that fails under
`--cov` gives no coverage reading for JA4L, so nothing measures the coverage of those
modules.

`ja4plus/fingerprinters/ja4l.py:320` reads `float(packet.time)` when the packet carries
one, and `time.time()` when it does not. A test that builds a packet rather than reading
a capture takes the wall clock, and coverage instrumentation moves the wall clock.
`.claude/rules/conformance.md` states that eviction and latency read the packet
timestamp, so this is worth reading as more than a test defect.

## What to build

1. Find why the 48 tests depend on wall-clock timing.
2. Make each one read a stated packet timestamp.
3. Confirm the suite reports the same counts with `--cov` and without it.

## Acceptance criteria

- [ ] `pytest tests/ -m "not spec_validation" --cov=ja4plus` reports 0 failed.
- [ ] The pass count with `--cov` equals the pass count without it.
- [ ] No test reads the wall clock to decide a JA4L value.

[...]

PR #178, opened 2026-08-07
Give a TCP reassembler stream a maximum age

Crank-Git:
Builds #170.

## What

`CLAUDE.md` states that a state table has a maximum entry count and a maximum age. #33
built the entry count and the two per-stream caps. No field held a timestamp, so a
connection that sent one segment and then stopped held its slot until 100 further
connections pushed it out.

A stream entry now carries `last_seen`, the packet timestamp of its most recent segment.
`add_segment` removes every stream that receives no segment for `max_stream_age` seconds.

## Why the removal drops the whole entry

#32 added a `seen` set of `(seq, length)` pairs, and #33 added the stored byte count. An
eviction that removed a segment and left either one behind would make `add_segment` drop
a re-sent segment as a duplicate, and would leave the byte count high. The age eviction
deletes the whole entry with `del self.streams[key]`, so all three leave together.
`test_an_age_eviction_accepts_the_same_segment_again` measures it.

## The clock

The age reads the packet timestamp, never the wall clock.
`.claude/rules/conformance.md`: "A capture file replays faster than real time, and a wall
clock would evict state the capture still needs."

`ja4plus/utils/tcp_stream.py` imports no clock. The caller states the timestamp through
the new `packet_seconds` helper in `ja4plus/utils/packet_utils.py`, which reads
`packet.time` and returns None when the packet carries none. A stream whose `last_seen`
is None never ages out, so a missing packet time evicts nothing rather than falling back
to a wall clock.

## The third caller

The issue body names two callers. A sweep of `ja4plus/` finds a third `add_segment` call
site at `ja4plus/fingerprinters/ja4ssh.py:164`, but its receiver is `SSHMessageTracker`
in `ja4plus/utils/ssh_utils.py:99`, a different class with a different signature. The
signature change therefore reaches exactly two call sites,
`ja4plus/fingerprinters/ja4h.py:92` and `ja4plus/fingerprinters/ja4x.py:130`.

## The measurement that bounds the risk

**The default is 600 seconds.** The longest gap between two segments of one stream across
`tests/foxio_vectors/` is **320.714503 seconds**, on `ssh-r.pcap`, stream
`192.168.1.169:64980-192.168.1.197:22`. The second reading is `http1.pcapng` at
60.042404 seconds. The measurement reads every capture with `rdpcap`, keys each TCP
segment the way `ja4h.py` keys a stream, and records the widest gap between two packet
timestamps of one key.

**No reference value moves. The count is zero.** The conformance suite reports 1375
passed, 146 skipped and 120 xfailed before the change and after it, against 120 keys in
`tests/foxio_deviations.json`.

**The bound is safe because of what today fingerprinters read, not because the bound is
inherently safe.** JA4H and JA4X read the head of a stream, so a stream that leaves late
in a capture changes no value. This has the same shape as the #33 byte cap, which
truncates `http2-with-cookies.pcapng` from 1853328 bytes to 1048554 and moves nothing. A
later issue that reads further into a stream, or that adds a vector with a wider gap,
must measure this age again.

## One divergence to note

`docs/specs/features/03-concurrency-safety.md` states "The default maximum age is 300
seconds." That is below the 320.714503-second measurement. That feature set is not built,
and its other defaults do not describe this reassembler: it states 10000 entries where
`max_streams` is 100 for JA4H and 50 for JA4X.

A 300-second age gives the same conformance counts, because the one stream above 300
seconds is SSH traffic that JA4H and JA4X read nothing from. This pull request follows
the measurement and takes 600, and `docs/implementation_notes.md` records the divergence.
The feature file stays for the issue that builds it.

## How tested

Every new test failed against the code as it stood. 11 of the 12 failed:

```
E  TypeError: TCPStreamReassembler.__init__() got an unexpected keyword argument
   max_stream_age. Did you mean max_streams?
E  AttributeError: TCPStreamReassembler object has no attribute max_stream_age.
   Did you mean: max_stream_bytes?
E  KeyError: last_seen
E  AssertionError: 10.0.0.1:12345-10.0.0.2:80 unexpectedly found in OrderedDict(...)
```

The twelfth, `test_the_module_reads_no_wall_clock`, passes against the code as it stands,
because the module imports no clock today. **It is not evidence for the packet-clock
criterion.** It is a guard against a later wall-clock import. The two behaviour tests
carry that criterion, and both failed before the change:
`test_removes_a_stream_that_receives_no_segment_for_the_maximum_age` and
`test_the_packet_clock_alone_ages_a_stream`.

The five gates:

```
pytest tests/ -m "not spec_validation"   1050 passed, 8 xfailed, 93 subtests passed
ruff format --check ja4plus/ tests/      102 files already formatted
```

Coverage of the two changed modules holds at 99%. `ja4plus/utils/packet_utils.py` is
100% and `ja4plus/utils/tcp_stream.py` is 99%, which are the readings before the change.

## Out of scope

48 JA4L and QUIC tests pass without `--cov` and fail with it, on this base commit before
this change. #177 records it. This pull request changes none of them.

Verified against: https://scapy.readthedocs.io/en/latest/api/scapy.packet.html (scapy 2.6, retrieved 2026-08-07)

### DIFF
--- a/docs/api_reference.md
+++ b/docs/api_reference.md
@@ -185,8 +185,8 @@ result = generate_ja4(packet)
 
 | Class/Function | Description |
 |----------------|-------------|
-| `TCPStreamReassembler(max_streams, max_stream_bytes)` | Sequence-aware TCP stream reassembly |
-| `.add_segment(key, seq, data)` | Add a TCP segment |
+| `TCPStreamReassembler(max_streams, max_stream_bytes, max_stream_segments, max_stream_age)` | Sequence-aware TCP stream reassembly |
+| `.add_segment(key, seq, data, timestamp)` | Add a TCP segment. `timestamp` is the packet time in seconds, and it ages the stream |
 | `.get_stream(key)` | Get reassembled contiguous bytes |
 | `.remove_stream(key)` | Remove a tracked stream |
 
@@ -196,6 +196,7 @@ result = generate_ja4(packet)
 |----------|-------------|
 | `get_ip_layer(packet)` | Return IP or IPv6 layer, or None |
 | `get_ttl(packet)` | Return TTL (IPv4) or Hop Limit (IPv6), or None |
+| `packet_seconds(packet)` | Return the capture timestamp in seconds, or None |
 
 ## CLI Module
 
--- a/docs/implementation_notes.md
+++ b/docs/implementation_notes.md
@@ -1058,6 +1058,56 @@ the keys the reassembler no longer holds once it passes `max_streams` entries, t
 **Location:** `ja4plus/utils/tcp_stream.py`, `ja4plus/utils/http_utils.py`,
 `ja4plus/fingerprinters/ja4h.py`. #33 built it, and it absorbed #103.
 
+### One stream holds a maximum age
+
+`CLAUDE.md` states that a state table has a maximum entry count and a maximum age. #33
+built the entry count and the two per-stream caps. No field held a timestamp, so a
+connection that sent one segment and then stopped held its slot until 100 further
+connections pushed it out.
+
+A stream entry now carries `last_seen`, the packet timestamp of its most recent segment.
+`add_segment` removes every stream that receives no segment for `max_stream_age` seconds.
+The removal drops the whole entry, so the segment list, the set of seen segments and the
+stored byte count leave together. A partial removal would leave a `(seq, length)` pair in
+the set, and `add_segment` would then drop that segment as a duplicate when the sender
+re-sends it.
+
+The age reads the packet timestamp, never the wall clock. `ja4plus/utils/tcp_stream.py`
+imports no clock, and the caller states the timestamp. `packet_seconds` in
+`ja4plus/utils/packet_utils.py` reads `packet.time` and returns None when the packet
+carries none. A stream whose `last_seen` is None never ages out, so a mixed clock evicts
+nothing.
+
+The age follows every segment the stream receives, not the segments the reassembler
+stores. A stream at its byte cap still sends, and an age that ignored a refused segment
+would evict a stream that is busy.
+
+**The default is 600 seconds, and the vectors measure the bound.** The longest gap
+between two segments of one stream across `tests/foxio_vectors/` is 320.714503 seconds,
+on `ssh-r.pcap`, stream `192.168.1.169:64980-192.168.1.197:22`. The measurement reads
+every capture with `rdpcap`, keys each TCP segment the way `ja4h.py` keys a stream, and
+records the widest gap between two packet timestamps of one key. The second reading is
+`http1.pcapng` at 60.042404 seconds. The default sits above the widest gap.
+
+No reference value moves. The conformance suite reports 1375 passed, 146 skipped and 120
+xfailed before the change and after it. A 300-second age gives the same three counts,
+because the one stream above 300 seconds is SSH traffic that JA4H and JA4X read nothing
+from.
+
+**The bound is safe because of what today's fingerprinters read, not because the bound is
+inherently safe.** JA4H and JA4X read the head of a stream, the way the #33 byte cap is
+safe. A later issue that reads further into a stream, or that adds a vector with a wider
+gap, must measure this age again.
+
+`docs/specs/features/03-concurrency-safety.md` states a 300-second default maximum age
+for a state table. That feature set is not built, and its other defaults do not describe
+this reassembler: it states 10000 entries where `max_streams` is 100 for JA4H and 50 for
+JA4X. #170 records the divergence and leaves the feature file for the issue that builds
+it.
+
+**Location:** `ja4plus/utils/tcp_stream.py`, `ja4plus/utils/packet_utils.py`,
+`ja4plus/fingerprinters/ja4h.py`, `ja4plus/fingerprinters/ja4x.py`. #170 built it.
+
 ---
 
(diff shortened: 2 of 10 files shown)


Crank-Git:
Self-review, correctness lens.

## The companion tables after an age eviction

`ja4h.py` keeps `self.unusable_base` and `ja4x.py` keeps `self.scan_offsets` and
`self.processed_certs`, all keyed by the stream key. An age eviction removes a
reassembler stream and leaves those entries. I traced the worst case: a stream leaves on
its age, and the same connection then sends again at the identical sequence number, so
the saved scan offset applies to a stream that starts empty.

```
scan_offsets after first: {"10.0.0.1:443-10.0.0.2:12345": (100, 30)}
streams after eviction:   ["10.0.0.1:444-10.0.0.2:12345"]
scan_offsets still held:  {"10.0.0.1:443-10.0.0.2:12345": (100, 30), ...}
resumed ok, result: None
stream bytes: 30
seen set: {(100, 30)}
```

The rebuilt stream holds a correct byte count and a correct set of seen segments.
`ja4x.py:168` guards the saved offset with `saved[0] == base_seq`, so a resumed
connection with a different base sequence number restarts the scan at zero.
`processed_certs` blocks a repeated certificate on its own.

This is the same path the `max_streams` eviction already takes, at `tcp_stream.py:76`.
The age eviction makes an existing removal fire more often. It opens no new case.

## Vacuous tests

The project has a history of tests that assert nothing, so I checked each new test
against the code before the change. 12 of the 13 fail. The failures are in the pull
request body.

`test_the_module_reads_no_wall_clock` is the exception. It passes against the code as it
stands, because the module imports no clock today. **I do not count it as evidence for
the packet-clock criterion.** It stays as a guard against a later wall-clock import. Two
behaviour tests carry that criterion, and both failed before the change.

`test_a_refused_segment_holds_the_stream_against_the_maximum_age` pins the one design
decision that a reader could reverse: `add_segment` records `last_seen` before the
duplicate check and before the two caps, so a busy stream at its byte cap keeps its age.
Without that line the stream ages out while it still sends.

## One gap I am leaving

The comparison at `tcp_stream.py:83` is `now - last_seen > self.max_stream_age`, so an
age exactly equal to `max_stream_age` keeps the stream. No test pins that boundary. The
behaviour is deliberate and the value is 600 seconds, so a one-second boundary changes
nothing a fingerprinter reads.

## Not fixed here

48 JA4L and QUIC tests fail under `--cov` on this base commit, before this change. #177
records it.

Crank-Git:
Second review lens, correctness. An independent pass over the diff found no blocker and two vacuous tests. Both findings hold, and `c8cd136` repairs them.

## Accepted: two tests asserted nothing

This is the failure mode the project keeps meeting, and I wrote two more of it.

**`test_the_default_maximum_age_passes_the_longest_gap_of_the_vectors`** compared 600
against the literal `320.714503`. Two literals. It passes whatever the vectors hold, so
it protects nothing when a later vector holds a wider gap. It now reads every capture
under `tests/foxio_vectors/` and measures the gap itself. It asserts the widest gap is
positive, names `ssh-r.pcap`, and equals 320.714503 to five places. The scan costs 1.1
seconds. A default of 300 now fails it:

```
E       AssertionError: 300 not greater than 320.7145028114319
```

**`test_the_packet_clock_alone_ages_a_stream`** compared a 1-second packet gap against a
100-second age. A reassembler that reads the wall clock also sees a near-zero real gap,
so the test passed either implementation. Its comment claimed otherwise, which is worse
than the weak test. It is now `test_the_stream_holds_the_stated_packet_time`, which
compares the stored value against the time the caller states. A reassembler that stores
the wall clock fails it:

```
E       AssertionError: 1786142217.540756 != 1000000001.0
```

I mutated the implementation to confirm both. With `stream["last_seen"] = _t.time()` in
place of the stated timestamp, two tests fail:

```
FAILED test_removes_a_stream_that_receives_no_segment_for_the_maximum_age
FAILED test_the_stream_holds_the_stated_packet_time
```

## Accepted with no code change: the trade-off note

The reviewer asks that the "a stream idle beyond 600 seconds restarts" limit is stated
somewhere durable. `docs/implementation_notes.md` carries it, under **The bound is safe
because of what today fingerprinters read, not because the bound is inherently safe.**
The `docs/specs/spec.md` Changelog round belongs to the project manager, so this pull
request adds no entry there.

## Rejected: two nits

**`ja4x.py` `scan_offsets` after a sequence-space wrap.** Reachable only when a resumed
stream recomputes the identical `base_seq`. `max_stream_bytes` is 1048576, so a wrap
needs 4 GiB on one connection. It is a property of the `scan_offsets` design that #33
shipped, and the age eviction reaches it no differently from the `max_streams` eviction
at `tcp_stream.py:76`. I measured the worst case and the rebuilt stream is correct.
Changing it is out of scope for #170.

**A guard around `float(packet.time)` in `packet_seconds`.** `.time` is a scapy
attribute, not a field read from the packet body, so the "every packet is hostile input"
rule does not reach it. `ja4l.py:320` already does the same bare conversion. A guard here
and none there is worse than neither.

## Categories the review cleared

Whole-entry deletion with no partial-eviction path. No stale companion state for a
realistic resumed stream. No crash and no over-eviction under non-monotonic timestamps,
because a negative difference never passes the age. Correct `EDecimal` conversion.

The five gates stay green: 1050 passed unit, 1375 passed and 120 xfailed conformance,
ruff clean, ruff format clean, mypy clean.

Issue #182, opened 2026-08-07
Reject a JA4SSH field that carries no side prefix


Crank-Git:
Found while building #36. Out of scope there, and this issue records it instead.

## What is wrong

`JA4SSHFingerprinter.interpret_fingerprint` reads a field by removing the first
character, and it never checks that the character is the side prefix. A field that
carries no prefix therefore reports a number that the fingerprint does not hold.

`ja4plus/fingerprinters/ja4ssh.py:352`:

```python
c_size = int(packet_sizes.split("s")[0][1:])
```

## How to reproduce

```
>>> JA4SSHFingerprinter().interpret_fingerprint("36s36_c50s50_c70s30")
{"session_type": "Unknown", ..., "details": {"packet_sizes": {"client": 6, "server": 36}, ...}}
```

The client packet size reads 6. The field holds 36, and the method removed the digit
`3` as though it were the prefix `c`.

## Rule for the fix

A field that carries no `c` prefix is a malformed fingerprint. Return the error
dictionary, as the method does for every other malformed form.

## Scope

`interpret_fingerprint` is an advisory helper. It reads a fingerprint and it produces
none, so no FoxIO vector covers it and no fingerprint moves.

Crank-Git:
## Plan

**Batch 15 (#185), member 2 of 4. Base: `batch/185-correctness-measurements`.**

### One correction to the issue body

The body names `ja4plus/fingerprinters/ja4ssh.py:352`. On `dev` at `9dfda2a` the two
lines are **479 and 480**:

```python
c_size = int(packet_sizes.split("s")[0][1:])
s_size = int(packet_sizes.split("s")[1])
```

Read the method before you accept the rest of the body. Line 480 takes the server field
with no prefix check of its own, and the body names only the client field.

### Approach

1. Write the failing case first: `interpret_fingerprint("36s36_c50s50_c70s30")` reports a
   client packet size of 6, and the field holds 36.
2. Read what `interpret_fingerprint` already returns for every other malformed form, and
   return that same error dictionary. Do not invent a second error shape.
3. Check every other field the method splits, not only `packet_sizes`. A field that
   removes a fixed character without checking it holds the same defect.
4. Prove each repair fails against the unrepaired method.

### Files

`ja4plus/fingerprinters/ja4ssh.py`, `tests/test_ja4ssh*.py`.

### Out of scope

`interpret_fingerprint` reads a fingerprint and produces none, so no FoxIO vector covers
[...]

Issue #36, opened 2026-08-06
Stop retaining packet objects and narrow the exception handlers


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/02-correctness-audit.md`](docs/specs/features/02-correctness-audit.md):

> FR-correctness-audit-12 — `get_raw_fingerprint` has one branch for the original order and one branch for the sorted order, and the two produce different values, or the branch is removed.

## What to build

Four register rows:

- `ja4plus/fingerprinters/base.py:38` — `add_fingerprint` stores the packet object, so a monitor holds every packet it ever fingerprinted. `ja4.py:325`, `ja4l.py:80` and `ja4s.py:96` do the same directly. Store what the result needs and never the packet.
- `ja4plus/fingerprinters/ja4.py:268` — the `if original_order:` branch and its `else:` branch build the same string. Remove the branch or make the two differ.
- `ja4plus/fingerprinters/ja4ssh.py:354` — `interpret_fingerprint` catches every exception and returns an error dictionary.
- `ja4plus/fingerprinters/ja4ssh.py:80` — on a non-standard port the lower port number decides which side is the server, which is arbitrary when both ports are ephemeral. Decide from the first SSH banner and fall back to the port.

The conformance suite must still pass after this slice.

## Acceptance criteria

- [ ] No fingerprinter stores a packet object after `process_packet` returns.
- [ ] `ja4plus/fingerprinters/ja4.py` has no branch whose two arms build the same string.
- [ ] `interpret_fingerprint` catches only the parse errors it expects.

## Links

- Feature: [`docs/specs/features/02-correctness-audit.md`](docs/specs/features/02-correctness-audit.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #13

Crank-Git:
## Plan

Batch 14, on `epic/13-batch14`, off `dev` at `ebf2ebc`. **This issue closes Epic 2.**
#179 runs beside it and touches documentation only, so the two do not meet.

**The line numbers are stale, and one defect is not where the body says.** The project
manager located all four before scheduling. Three issue bodies this session did not match
the file, so verify each one yourself before you build it.

**1. The packet objects.** `ja4plus/fingerprinters/base.py:38` stores the packet in
[...]

PR #183, opened 2026-08-07
Store the endpoints of a packet and never the packet

Crank-Git:
Builds #36. Part of #13.

## What this changes

**1. No fingerprinter holds a packet object.** Nine fingerprinters stored the packet
that produced each value. A monitor runs for weeks, so the process held every packet it
ever fingerprinted. `CLAUDE.md` states the rule, and row 11 of
`docs/specs/features/02-correctness-audit.md` records the defect.

The issue body named `base.py`, `ja4.py`, `ja4l.py` and `ja4s.py`. A sweep of the package
found a fifth writer the body did not name: `ja4h.py:170`. A run-time walk over each
fingerprinter confirmed the whole set and confirmed that no other state holds a packet:

```
JA4: produced=12 retained=12      JA4T:  produced=11 retained=11
JA4S: produced=11 retained=11     JA4TS: produced=11 retained=11
JA4H: produced=57 retained=40     JA4D:  produced=4  retained=4
JA4X: produced=15 retained=7      JA4D6: produced=0  retained=0
JA4L: produced=24 retained=24     JA4SSH: produced=1 retained=0
```

Every path ran through `fingerprints[i][packet]`. JA4SSH already held none.

An entry now holds `src`, `dst`, `srcport` and `dstport`. The values name the innermost
address layer and the innermost port layer, which are the layers
`tests/conformance_index.py` read from the packet before this change. A tunnelled
capture therefore groups under the same stream it grouped under before.

**2. The dead branch.** `get_raw_fingerprint` built the signature-algorithm form in both
arms of one `if`. The branch is gone. The cipher branch and the extension branch stay:
the `if` arm builds the wire order and the `else` arm builds the sorted order, which is
the `JA4_o` reading #132 settled. A test guards both against removal.

**3. The exception handler.** `interpret_fingerprint` caught every exception and
returned an error dictionary, so a defect inside this project reached a caller as a
malformed fingerprint. The handler now names `AttributeError`, `IndexError` and
`ValueError`.

## What this does NOT change, and why

**Row 12 — the SSH server side — is not built here.** The register says to decide the
server from the first SSH banner and to fall back to the port. A measurement of every
SSH stream in `tests/foxio_vectors/` disproves the premise of that rule: the first
`SSH-` banner comes from the client in 5 of 12 streams.

```
ssh-r.pcap:    192.168.1.169:64980 -> 192.168.1.197:22   [peer is on port 22]
ssh-r.pcap:    192.168.1.197:46394 -> 44.212.59.210:22   [peer is on port 22]
ssh-r.pcap:    192.168.1.197:46396 -> 44.212.59.210:22   [peer is on port 22]
ssh.pcapng:    172.16.225.48:57377 -> 54.160.114.75:22   [peer is on port 22]
ssh2.pcapng:   172.16.225.48:57377 -> 54.160.114.75:22   [peer is on port 22]
```

RFC 4253 has both sides send an identification string, and a client that does not wait
for the server sends first. The first banner is therefore a race, and it identifies no
side. The rule as written would replace one arbitrary answer with another arbitrary
answer.

Every stream above uses port 22, so the port branch decides all twelve and the vector
set proves no alternative either. That is a question for the user, not a choice to make
here.

## How this was tested

Each test failed against the code as it stands before it passed. Every test that asserts
"no packet is retained" first asserts that the fingerprinter produced a value, so no
assertion reports success on an empty list.

- `tests/test_no_retained_packets.py` — walks each fingerprinter for a packet object at
  any depth and under any key name. 21 cases. Before the change: 19 failed.
- `tests/test_ja4_raw_branches.py` — reads `ja4.py` with `ast` and reports every `if`
  whose two arms hold identical code. A dead branch produces the same value as the live
  branch, so no call can detect it. Before the change: 1 failed.
- `tests/test_ja4ssh_interpret.py` — an error the parser expects from no input reaches
  the caller, and every malformed form still returns the error dictionary. Before the
  change: 1 failed.

## Measurements

| Gate | Before | After |
|---|---|---|
| `pytest tests/ -m "not spec_validation"` | 1156 passed, 8 xfailed | 1188 passed, 8 xfailed |
| `pytest tests/ -m spec_validation` | 1375 passed, 146 skipped, 120 xfailed | 1375 passed, 146 skipped, 120 xfailed |
| `pytest tests/fuzz/` | 106 passed | 106 passed |
| `ruff check` / `ruff format --check` / `mypy` | clean | clean |
| Coverage | 81% (639 missed of 3385) | 81% (637 missed of 3404) |

`120 xfailed` against 120 keys in `tests/foxio_deviations.json`.

**Reference values that moved: 0.** The conformance suite reports the same counts.

**JA4SSH values that moved: 0.** Every JA4SSH value across `tests/foxio_vectors/` was
recorded before and after — 17 values across 7 captures — and the two files are
identical.

## Out of scope

#182 records a separate defect found here: `interpret_fingerprint` reads `36s36` as a
client size of 6, because it removes the first character without checking that it is the
side prefix.

Verified against: https://github.com/FoxIO-LLC/ja4/tree/main/pcap (retrieved 2026-08-07)


Crank-Git:
Self-review, decision needed on register row 12 (the SSH server side).

## The measurement

The rule the register gives — decide the server from the first SSH banner, fall back to
the port — rests on a premise the FoxIO material disproves. Across every SSH stream in
`tests/foxio_vectors/`, the sender of the first `SSH-` banner is the client in 5 streams
of 12.

| Capture | Sender of the first banner | The endpoint on port 22 |
|---|---|---|
| `gre-sample.pcap` | `66.59.109.137:22` | the sender |
| `ssh-r.pcap` | `192.168.1.169:64980` | the receiver |
| `ssh-r.pcap` | `192.168.1.197:46394` | the receiver |
| `ssh-r.pcap` | `192.168.1.197:46396` | the receiver |
| `ssh-scp-1050.pcap` | `192.168.1.197:22` | the sender |
| `ssh.pcapng` | `172.16.225.48:57377` | the receiver |
| `ssh2-malformed.pcap` | `10.0.0.2:22` | the sender |
| `ssh2-moloch-crash.pcap` | `10.0.0.2:22` | the sender |
| `ssh2.pcapng` | `172.16.225.48:57377` | the receiver |
| `sshv1.pcap` | `[3ffe:501:410:0:2c0:dfff:fe47:33e]:22` | the sender |
| `tcpdump-geneve.pcap` | `20.0.0.1:22` | the sender |
| `v6.pcap` | `[3ffe:501:410:0:2c0:dfff:fe47:33e]:22` | the sender |

RFC 4253 section 4.2 has both endpoints send an identification string. A client that
does not wait for the server sends first, and OpenSSH and PuTTY both do. The banner
carries no field that names a client or a server, and the two banners hold the same
form. The first banner is therefore a race.

## Why this stops the work rather than reshapes it

The rule as written replaces one arbitrary answer with another arbitrary answer, and it
would move a JA4SSH value on any capture where the client banner arrives first.

Every stream above holds an endpoint on port 22, so the port branch decides all twelve
and no vector exercises the non-standard-port branch. The vector set can therefore
neither prove the banner rule nor prove an alternative. `CLAUDE.md` rule 1 makes that a
question, not a judgment call.

## The alternative I did not build

The TCP handshake names the two endpoints exactly: the SYN sender is the client, and the
SYN+ACK sender is the server. It is a fact of the connection rather than a race, and
`ja4l.py` already reads the handshake for the same purpose through `_opens_a_connection`.
The order would read: an endpoint on port 22 decides, else the SYN decides, else the
lower port decides and the fingerprinter records the guess.

Adopting it changes `docs/specs/features/02-correctness-audit.md` row 12 and the
edge-case row that reads "Both endpoints use an ephemeral port for SSH. | The first
banner decides which side is the server." That is a spec change, so it needs the user.

Rows 10, 11 and 14, which carry the three acceptance criteria of #36, are built and
measured in this pull request.

Crank-Git:
Self-review by specialist lens. Two children reviewed the diff: one for correctness, one for whether the new tests can fail. Both ran read-only in the worktree.

## Finding accepted and fixed in a68aa09

**`packet_endpoints` read the wrong address layer.** The first version read the
innermost address layer, and the docstring claimed the reference describes the inner
stream. The reference does the opposite. It reports the outer address pair and the
inner port pair.

| Capture | Reference | Outer address pair | Inner address pair | Inner port pair |
|---|---|---|---|---|
| `gre-sample.pcap` | `172.27.1.66` `66.59.109.137` 40264 22 | `172.27.1.66` `66.59.109.137` | `66.59.111.190` `172.28.2.3` | 40264 22 |
| `gre-erspan-vxlan.pcap` | `100.20.9.2` `100.20.9.1` 65174 80 | `100.20.9.2` `100.20.9.1` | `10.16.27.12` `10.16.27.131` | 65174 80 |
| `tcpdump-geneve.pcap` | `20.0.0.2` `20.0.0.1` 51225 22 | `20.0.0.2` `20.0.0.1` | `30.0.0.1` `30.0.0.2` | 51225 22 |

The helper now reads `get_ip_layer` for the address pair and `innermost_layer` for the
port pair, which are the two layers `_reported_key` in `ja4l.py` already reads. No
conformance result moved on the correction, because the reference holds only JA4L
values on all three captures and JA4L names its stream with its connection key.

`test_a_tunnelled_packet_reports_the_endpoints_the_reference_reports` now reads each of
the three vectors against its expected-output file. Against the first version it fails
on all three:

```
AssertionError: gre-erspan-vxlan.pcap reports [(("10.16.27.12", "65174"), ("10.16.27.131", "80"))],
not (("100.20.9.1", "80"), ("100.20.9.2", "65174"))
```

## Second finding accepted and fixed in a68aa09

`test_an_entry_carries_the_endpoints_of_the_packet_that_produced_it` reads the presence
of the four fields and not their values, so it passes when the source and the
destination swap. `test_an_entry_names_the_endpoints_the_reference_names` now reads the
four values of a JA4 entry against `tls12.pcap.json`. Against a swap it reports:

```
AssertionError: assert "34.117.237.239" == "192.168.133.129"
```

## The mutation report

Every test of the three new files was mutation-tested. Each one went red under a
mutation of the source it covers. No test passes when the feature is absent.

One structural note: the JA4SSH case of the retention test has no single-line mutation
that reintroduces a stored packet, because `_close_window` reads aggregated connection
state and holds no packet variable. The case still guards against the vacuous form,
because it asserts that the fingerprinter produced a value before it reads the walk.

## Findings rejected

**A `TypeError` reaches the caller of `interpret_fingerprint` when a caller passes
`bytes`.** Correct, and it is the intended result. The method reads a fingerprint that
this library produced, and every caller in the repository passes a string or None. A
`bytes` argument is a defect in the caller, and the narrowed handler is what reports it.

**The deleted `_innermost` in `tests/conformance_index.py` held its own defect.** True:
it iterated the layer classes in tuple order, so `UDP` always won over `TCP` and it read
the outer VXLAN port 4789 rather than the inner port 65174. The deletion removes it, and
no behaviour depended on it. Recorded here rather than acted on.

## Gates after the correction

```
pytest tests/ -m "not spec_validation"   1191 passed, 8 xfailed
pytest tests/fuzz/                       106 passed
ruff check / ruff format --check / mypy  clean
coverage                                 81 percent, 637 missed of 3403
```

Reference values that moved: 0. JA4SSH values that moved: 0, measured across 17 values
in 7 captures.

Crank-Git:
Register row 12 is built, and the specification is corrected. The user decided on 2026-08-07 to adopt the TCP handshake rule.

## What this adds

`_decide_endpoints` in `ja4plus/fingerprinters/ja4ssh.py` reads three rules in order:

1. An endpoint on port 22 is the server.
2. Else the TCP handshake decides. The SYN sender is the client, and the SYN+ACK sender is the server.
3. Else the lower port is the server, and the fingerprinter records the guess.

The handshake arrives before any SSH data, and a SYN on a non-standard port creates no
connection, so `_record_handshake` runs before the guard that returns. It writes the
client endpoint into `_handshake_clients`, which holds at most
`MAX_HANDSHAKE_CONNECTIONS` entries and drops the oldest half when it fills. A monitor
reads a SYN for every TCP connection on the wire, so that table needs a limit.

`opens_a_connection` moved from `ja4l.py` to `ja4plus/utils/packet_utils.py`, beside a
new `accepts_a_connection`. `ja4l.py` now calls the moved function. Two readers of one
handshake that disagree would name two different clients for one connection, and
`test_one_reader_of_the_tcp_handshake_serves_both_fingerprinters` holds them together.

## The guess reaches the result

Every JA4SSH entry carries `server_decided_by`, which reads `port`, `handshake` or
`guess`. A JA4SSH value names a client packet size and a server packet size, so a
consumer that cannot tell a measured endpoint from a guessed one cannot read the value
safely. `docs/api_reference.md` names the field and its three values.

## The specification change

`docs/specs/features/02-correctness-audit.md` row 12 now reads the handshake. Its
edge-case row reads the handshake. A new section, **The banner decides nothing**,
records the measurement: RFC 4253 section 4.2, the 4 of 11 attributable streams, and
`ssh-r.pcap` and `ssh2.pcapng` by name.

## Tests, each proven to fail first

| Test | Mutation | Result |
|---|---|---|
| `test_the_syn_sender_is_the_client_when_no_endpoint_uses_port_22` | disable rule 2 | `AssertionError: the SYN sender must be the client / assert "10.0.0.2" == "10.0.0.1"` |
| `test_the_syn_ack_sender_is_the_server_when_the_capture_holds_no_syn` | disable rule 2 | `AssertionError: assert "10.0.0.2" == "10.0.0.1"` |
| `test_the_lower_port_decides_when_the_capture_holds_no_handshake` | record `port` in place of `guess` | `AssertionError: assert "port" == "guess"` |
| `test_the_fingerprint_entry_records_how_the_server_was_decided` | both mutations | `assert "guess" == "handshake"` and `assert "port" == "guess"` |
| `test_a_packet_that_carries_the_rst_flag_names_no_endpoint` | remove the RST guard | `AssertionError: a SYN and a RST named an endpoint` |

Every test that reads the handshake places the client on the LOWER port, so the
lower-port fallback names the client as the server. A test cannot pass through the
fallback.

## Second self-review, one finding accepted

A Sonnet child reviewed the commit. It found that `opens_a_connection` and
`accepts_a_connection` read only the SYN flag and the ACK flag, so a packet carrying SYN
and RST together named the client of a connection it never opened. Every packet is
hostile input, so both now report False for any packet that carries the RST flag. No
vector holds such a packet, measured across all 79 files of `tests/foxio_vectors/`, so
the fix moves nothing.

The same review could not reproduce a mid-connection orientation flip. It drove a late
contradicting SYN, and it evicted the handshake entry with 1010 unrelated SYNs mid-connection.
Both stayed one connection, because the `reversed_key` guard remaps a swapped decision
onto the existing entry and `is_client_to_server` reads the connection rather than a
fresh decision. It also confirmed the table never exceeds 1000 entries across 5000
unique pairs.

## The gate

**JA4SSH values that moved: 0.** 17 values across 7 captures, recorded before the first
commit of this pull request and again now. The two files are identical.

Every stream of `tests/foxio_vectors/` holds an endpoint on port 22, so the port rule
decides all of them and no vector exercises the handshake branch. That is why nothing
moves, and it is also why the vector set could not settle the question on its own.

```
pytest tests/ -m "not spec_validation"   1202 passed, 8 xfailed
pytest tests/fuzz/                       106 passed
ruff check / ruff format --check / mypy  clean
coverage                                 81 percent, 638 missed of 3447
```

`120 xfailed` against 120 keys in `tests/foxio_deviations.json`. Reference values that
moved: 0.

Base: merged `origin/epic/13-batch14` at `81134b9` before pushing. `docs/specs/spec.md`
is untouched.

Verified against: https://www.rfc-editor.org/rfc/rfc4253#section-4.2 (retrieved 2026-08-07)

Issue #186, opened 2026-08-08
Reject a JA4SSH part that carries a second separator


Crank-Git:
Found while building #182. Out of scope there, because the part carries the `c` prefix
that #182 checks.

## What is wrong

`JA4SSHFingerprinter.interpret_fingerprint` splits a part on `s` and reads element 1 as
the server value. It never checks that the split produced exactly two elements. A part
that carries a second `s` separator therefore passes, and the method reads the first
server number and drops the rest.

`ja4plus/fingerprinters/ja4ssh.py:486`:

```python
s_size = int(packet_sizes.split("s")[1])
```

## How to reproduce

```
>>> JA4SSHFingerprinter().interpret_fingerprint("c36s36s99_c50s50_c70s30")
{'session_type': 'Interactive SSH Session', ..., 'details': {'packet_sizes': {'client': 36, 'server': 36}, ...}}
```

The part holds `36s36s99`, which names no JA4SSH value. The method reports a client
packet size of 36 and a server packet size of 36, and it reports no error.

## Rule for the fix

A part that splits into other than two elements on `s` is a malformed fingerprint.
Return `{"error": "Invalid JA4SSH format"}`, which is the error the method already
returns for a wrong part count and for a missing client prefix.

## Scope

`interpret_fingerprint` reads a fingerprint and it produces none, so no FoxIO vector
covers it and no fingerprint moves.

Crank-Git:
Three more readings from the same class, measured while #182 was built. `int()` accepts
more than the JA4SSH form names, so a part that carries the `c` prefix and one `s`
separator still passes with content the form does not hold.

```
>>> f = JA4SSHFingerprinter()
>>> f.interpret_fingerprint("c 36s36_c50s50_c70s30")["details"]["packet_sizes"]
{'client': 36, 'server': 36}
>>> f.interpret_fingerprint("c-5s36_c50s50_c70s30")["details"]["packet_sizes"]
{'client': -5, 'server': 36}
>>> f.interpret_fingerprint("c٦s36_c50s50_c70s30")["details"]["packet_sizes"]
{'client': 6, 'server': 36}
```

The first holds a space, the second holds a negative number, and the third holds the
Arabic-Indic digit six. A JA4SSH part holds ASCII digits alone.

One repair covers this issue and all three: match each part against
`^c(\d+)s(\d+)$` with `re.ASCII`, and return `{"error": "Invalid JA4SSH format"}` for a
[...]

Issue #185, opened 2026-08-08
Batch 15: correctness measurements


Crank-Git:
Batch 15 groups the four correctness issues that no epic holds. Each one repairs
a measurement rather than a fingerprinter, so no fingerprint moves in this batch.

The members run in sequence on one integration branch. #172 runs last, so its
mutation sweep reads the tests the other three members add.

## Members

- [x] #177 Repair the 48 JA4L tests that fail under coverage — **closed as no change required on 2026-08-08.** Only `--cov=<dotted.module>` fails, and the cause is a `coverage` module-purge interaction, not a `ja4plus` defect. The project manager records the measurement and the workaround at this batch's gate.
- [x] #182 Reject a JA4SSH field that carries no side prefix
- [ ] ~~#175 Cap the cookie list the HTTP parser builds~~ — **merged as `be3604f`, reverted as `042e8c3` on 2026-08-08.** It produced no JA4H value past a bound, which discarded five readings that read no cookie, and it diverged from the reference with no `Divergence register` row. #175 reopens and leaves this batch.
- [ ] #172 Sweep the suite for assertions that cannot fail

## Why this order

#177 comes first. The `coverage must not fall` practice reads a coverage
measurement, and 48 tests fail under `--cov`, so the practice measures nothing
until #177 lands. #172 comes last, because a sweep that runs before the other
members reads a suite that the batch then changes.

## Gate

The batch pull request into `dev` runs the full continuous-integration suite
once. The register must hold the same key count as the `xfailed` count, and no
fingerprint may move.

Crank-Git:
## Batch 15 merged into `dev` as `6fc248f`

Nine of nine continuous-integration jobs green: `lint`, `conformance`, `fuzz`, and six
`test` jobs across Python 3.9 to 3.13 on Ubuntu and Python 3.12 on macOS.

| Member | Outcome |
|---|---|
| #182 | Merged. `interpret_fingerprint` rejects a JA4SSH part that carries no client prefix, in all three parts |
| #172 | Merged. The project holds `tests/mutation_sweep.py` and `docs/mutation_sweep.md`, and two vacuous `tests/test_ja4l.py` cases are repaired |
| #177 | Closed, no change required. The stated cause does not survive measurement |
| #175 | Merged as `be3604f` and reverted as `042e8c3`. Reopened, and it keeps its measurement |

**The register is unchanged: 120 keys against 120 `xfailed`, checksum
[...]

PR #187, opened 2026-08-08
Reject a JA4SSH fingerprint part that carries no client prefix

Crank-Git:
Builds #182. Batch 15 (#185), member 2 of 4.

## What is wrong

`JA4SSHFingerprinter.interpret_fingerprint` read a client value with `part[1:]`. It
never checked that the removed character is the prefix `c`. A part that carries no
prefix therefore reported the number that follows its first digit.

Measured on the base commit `9dfda2a`:

```
'36s36_c50s50_c70s30'    -> packet_sizes {'client': 6,  'server': 36}
'c36s36_50s50_c70s30'    -> ssh_ratio    {'client': 0,  'server': 50}
'c36s36_c50s50_70s30'    -> ack_ratio    {'client': 0,  'server': 30}
'x36s36_c50s50_c70s30'   -> packet_sizes {'client': 36, 'server': 36}
```

**The issue body named the packet size part alone. All three parts hold the defect.**
A wrong prefix passed as well, so the check tests for the `c` prefix and not for a
digit.

## Why the change is what it is

The method returns `{"error": "Invalid JA4SSH format"}`. That is the error the method
already returns for a wrong part count, so this change adds no second error shape.

## Two corrections to the written premises

**The issue body's line number is stale.** It names `ja4ssh.py:352`. The two lines are
479 and 480 on the base commit. The plan comment already recorded that.

**The plan expected a defect on line 480 and there is none.** A JA4SSH part names its
server value after the `s` separator and carries no server prefix, so
`packet_sizes.split(\"s\")[1]` removes no character. A part that holds no `s` raises
`IndexError`, which the method already catches and reports.

## A note on words

The issue body calls the underscore-separated unit a field. This change calls it a
part, which is the name `interpret_fingerprint` gives it at `ja4plus/fingerprinters/ja4ssh.py:470`.

## How it is tested

`tests/test_ja4ssh_interpret.py` gains six cases. Each one was run against the
unrepaired method first and failed:

```
FAILED test_a_part_that_carries_no_client_prefix_returns_an_error_dictionary[a packet size part that carries no side prefix]
FAILED test_a_part_that_carries_no_client_prefix_returns_an_error_dictionary[an SSH ratio part that carries no side prefix]
FAILED test_a_part_that_carries_no_client_prefix_returns_an_error_dictionary[an ACK ratio part that carries no side prefix]
FAILED test_a_part_that_carries_no_client_prefix_returns_an_error_dictionary[a packet size part that carries a wrong side prefix]
FAILED test_a_part_that_carries_no_client_prefix_returns_an_error_dictionary[an SSH ratio part that carries a wrong side prefix]
FAILED test_a_part_that_carries_no_client_prefix_reports_no_client_number
6 failed, 10 passed
```

## The five gates

| Gate | Before | After |
|---|---|---|
| `pytest tests/ -m \"not spec_validation\"` | 1202 passed, 8 xfailed | 1208 passed, 8 xfailed |
| `pytest tests/ -m spec_validation` | 1375 passed, 120 xfailed | 1375 passed, 120 xfailed |
| `tests/foxio_deviations.json` keys | 120 | 120 |
| `ruff check` | clean | clean |
| `ruff format --check` | clean | clean |
| `mypy ja4plus/` | clean | clean |
| Coverage, total | 81% | 81% |
| Coverage, `ja4ssh.py` | 98%, 4 lines missed | 98%, 4 lines missed |

No fingerprint moved. `interpret_fingerprint` reads a fingerprint and produces none, so
no FoxIO vector covers it.

## Out of scope, filed rather than fixed

#186. A part that carries a second `s` separator, such as `c36s36s99`, passes the
method, because nothing checks that the split produced two elements. That part carries
the `c` prefix, so #182 does not cover it.

## Docs

`docs/specs/spec.md` gains one Changelog entry with the round number `TBD`. The project
manager assigns the round number at the batch gate.


Crank-Git:

No finding. Four questions, each measured rather than read.

**1. Does the guard change the result for a well-formed fingerprint?** No.
`_close_window` is the only producer of a JA4SSH string in this repository, and it
builds all three parts with a literal `c`:

```
ja4plus/fingerprinters/ja4ssh.py:380:        part_a = f"c{client_mode}s{server_mode}"
ja4plus/fingerprinters/ja4ssh.py:385:        part_b = f"c{client_ssh_count}s{server_ssh_count}"
ja4plus/fingerprinters/ja4ssh.py:390:        part_c = f"c{client_ack_count}s{server_ack_count}"
```

**2. Can the guard raise?** No. `parts = fingerprint.split("_")` runs first, so a value
that carries no `split` raises before the guard, as it did on the base. Every element of
`parts` is a `str`, because `str.split` returns strings, so `startswith` raises nothing.
`bytes`, `None`, `int`, `""` and `"__"` produce the same result on the base and on this
branch.

**3. Would the six new tests fail if the code were wrong?** Yes. Each of the five
parametrized values returns a full details dictionary and no `error` key against the base
commit, so `assert "error" in result` fails there. The sixth compares the whole
dictionary and reads `{'client': 6, ...}` against the base. The failures are recorded in
the pull-request body.

**4. Does an existing test assert the old behaviour?** No.
`pytest tests/ -m "not spec_validation"` reports `1208 passed, 8 xfailed`, against
`1202 passed, 8 xfailed` on the base. The six are the six added here.

Crank-Git:
## Self-review, robustness and writing-standard lens

**Robustness.** The guard rejects a part that carries no `c` prefix. It rejects an
uppercase `C36s36` and a part with a leading space. Four malformed forms still pass, and
none of them carries a missing prefix, so none is in scope here:

| Input | Result |
|---|---|
| `c36s36s99_c50s50_c70s30` | `{'client': 36, 'server': 36}` — the third number is dropped |
| `c 36s36_c50s50_c70s30` | `{'client': 36, 'server': 36}` — `int()` strips the space |
| `c-5s36_c50s50_c70s30` | `{'client': -5, 'server': 36}` — a packet size below zero |
| `c٦s36_c50s50_c70s30` | `{'client': 6, 'server': 36}` — `int()` accepts a Unicode digit |

All four belong to one class: nothing checks the body of a part against the JA4SSH form.
#186 owns the class, and a comment there records the last three with the repair that
covers all four. Filed rather than fixed, because #182 states one rule and this is
another.

**Writing standard.** Three findings, all fixed in `c41a635`.

1. The code comment read "the number that follows its first digit". A wrong prefix such
   as `x36s36` carries no digit, so the comment now reads "its first character".
2. The parametrize ids read "side prefix" where the test names read "client prefix".
   That is synonym rotation. The ids now read "client prefix".
3. Three sentences of the Changelog entry ran past the 25-word description limit. Each
   is now split.

Rejected: no change to the code comment's length. It runs four sentences and states the
reason a reader cannot see from the code, which is the pattern `.claude/rules/ste.md`
names.

Issue #188, opened 2026-08-08
Cap the header list the HTTP parser builds


Crank-Git:
## Context

`CLAUDE.md` holds the rule:

> A state table has a maximum entry count and a maximum age. Nothing grows without a
> limit.

The HTTP header parser holds no maximum entry count. #175 covers the cookie list of the
same parsers. It does not cover the header list.

## The defect

`ja4plus/utils/http_utils.py` and `ja4plus/fingerprinters/ja4h.py` append one entry per
header line. Three sites hold the growth:

- `ja4plus/utils/http_utils.py:74` — `parse_http_request` fills the `headers`
  dictionary.
- `ja4plus/utils/http_utils.py:204` — `extract_http_info` fills `headers` and
  `header_names`.
- `ja4plus/fingerprinters/ja4h.py:216` — `_extract_http_info_from_bytes` fills the same
  two.

No site caps the entry count, and no site caps the length of a header name or of a
header value.

A measurement on the base commit `b6468dc`, with a request that carries 100000 header
lines:

```
_extract_http_info_from_bytes header names 100000 in 38.2 ms
parse_http_request headers 100000 in 17.4 ms
```

One request builds one list, so the memory returns when the request ends.

## What to build

- Cap the number of header lines one request contributes.
- Cap the length of one header name and of one header value.
- Decide what a request that exceeds a cap produces. #175 holds the same decision for
  the cookie list, and the two answers must agree.

## Measurement that bounds the cap

`tests/foxio_vectors/` produces 82 JA4H values, and the 82 requests behind them reach at
most 11 header lines and a header name of 25 bytes. A cap above those two readings leaves
every one of the 82 values unchanged.

## Acceptance criteria

- [ ] A request that carries more header lines than the cap produces a bounded list.
- [ ] The 82 JA4H values that `tests/foxio_vectors/` produces do not change.

## Links

- Feature: [`docs/specs/features/02-correctness-audit.md`](docs/specs/features/02-correctness-audit.md)
- Same parsers as: #175

Crank-Git:
## The policy is not inherited. Decided 2026-08-08

This issue was filed with the note that it takes the same over-cap policy as #175. **That
link is removed.** #175's bound merged and was reverted the same day, because producing no
JA4H value past a bound discards the method, the HTTP version, the header count, the
`Accept-Language` value and the header names — five readings that read no cookie.

**A header bound carries the same cost and a wider exposure.** A large header list is more
[...]

Issue #175, opened 2026-08-07
Name the boundary of the state rule and record that the cookie list carries no bound


Crank-Git:
## Context

`CLAUDE.md` holds the rule:

> A state table has a maximum entry count and a maximum age. Nothing grows without a
> limit.

The HTTP cookie parser holds no maximum entry count.

## The defect

`ja4plus/utils/http_utils.py` and `ja4plus/fingerprinters/ja4h.py` split the Cookie
header on `;` and append one entry per pair. Neither parser caps the entry count, and
neither caps the length of a name or of a value. One request builds one list, so the
memory returns when the request ends. A sender that emits a large Cookie header still
makes the parser allocate a list proportional to that header.

A measurement on a header that repeats `a=1; ` 100000 times builds a list of 100000
pairs in about 12 milliseconds.

The defect predates #35. #35 found it and did not widen its scope to cover it. A
measurement on the code before #35, with 100000 unique cookie names, built a dictionary
of the same size, so the growth is not a result of #35.

## What to build

- Cap the number of cookie pairs one request contributes.
- Cap the length of one cookie name and of one cookie value.
- Decide what a request that exceeds a cap produces. A truncated cookie list produces a
  fingerprint that describes traffic the sender did not send, so returning nothing may
  be the correct answer. That decision needs the vectors.

## Acceptance criteria

- [ ] A Cookie header that exceeds any cap produces no cookie parse result, and the request therefore produces no JA4H value. Decided 2026-08-08; see the decision comment.
- [ ] The caps are 512 cookie pairs, 256 bytes of cookie name and 4096 bytes of cookie value.
- [ ] All three parser sites take the same policy, including `ja4plus/fingerprinters/ja4h.py:229`.
- [ ] The 82 JA4H values that `tests/foxio_vectors/` produces do not change.

## Links

- Feature: [`docs/specs/features/02-correctness-audit.md`](docs/specs/features/02-correctness-audit.md)
- Found by: #35

Crank-Git:
## Plan

**Batch 15 (#185), member 3 of 4. Base: `batch/185-correctness-measurements`**, which
already carries #182.

### Read the premise before you build against it

Two premises in this batch have already failed contact with the code: #182's line numbers
were stale and its defect was wider than the body stated, and #177's stated cause was
wrong. Measure the growth this body describes on the base commit before you cap anything.

### Approach

1. Reproduce the measurement. The body reports 100000 pairs built in about 12
[...]

Issue #172, opened 2026-08-07
Sweep the suite for assertions that cannot fail


Crank-Git:
## The shape

**A comparison that is never made reads as a comparison that passes.** This project has
found eleven instances. Every one was found by accident, while an issue worked on
something else.

| Where | What |
|---|---|
| #80, #109, #121, #138, #141, #151 | Six found across batches 1 to 10 |
| Session 6, two workers | Found inside their own new tests, before the gate saw them |
| #32 | The **acceptance criterion itself**: "adding 10000 segments completes in under one second" already passed against the unfixed code, at 0.4511 s |
| #33 | **Three tests already committed and green**, `test_max_stream_size` among them. It read `assertLessEqual(len(result), 20)` against a cap of 10, so it passed whatever the reassembler did |

The last three matter most. #32's instance sat in an acceptance criterion, so the gate
would have reported it met with no change made. #33's three were **already in the suite,
green, for as long as the project has existed**. Nothing was looking for them.

#141 is the instructive case. The only vector reaching the JA4 ALPN branch held
`0xba 0xad`, and all three candidate rules fire on it, so the suite was green whichever
rule the code held.

## Why an issue

Every instance so far was found by a worker who happened to touch the code. The suite
holds about 2300 cases, and nobody has ever asked of them the question that matters:
**would this fail if the code were wrong?**

## What to build

1. **Sweep the suite with mutation.** For each module under `ja4plus/`, change the
   behaviour the tests claim to measure, and record which cases fail. A case that fails
   for no mutation of the code it names is a candidate.
2. **Report the candidates with evidence**: the test, the mutation attempted, and the
   result. Do not rewrite a test from the report alone; a case may be correct and the
   mutation wrong.
3. **Repair the cases that are genuinely vacuous**, and prove each repaired case fails
   against the mutation that exposed it.
4. **Give the project a way to keep asking.** A documented command, a marked subset, or a
   note in `.claude/rules/conformance.md` that states the question and how to run it.

## What not to build

- No fingerprinter change. A mutation is a measurement, and it is reverted.
- No new coverage target. Coverage measures which lines run, and this issue is about
  whether a line that runs is measured. They are different questions, and the confusion
  between them is part of why these eleven survived.

## Acceptance criteria

- [ ] Every module under `ja4plus/` is swept, and the sweep names which mutation it
      applied to each.
- [ ] The report lists every candidate with the test, the mutation and the result.
[...]

PR #189, opened 2026-08-08
Bound the cookie list the HTTP parser builds (#175)

Crank-Git:
Batch member of #185. Base: `batch/185-correctness-measurements`. Draft, `[skip ci]` on
every head commit.

## What

The HTTP cookie parser holds a maximum entry count and two maximum lengths.

| Bound | Value | Largest vector reading |
|---|---|---|
| `MAX_COOKIE_PAIRS` | 512 | 14 pairs |
| `MAX_COOKIE_NAME_BYTES` | 256 | 41 bytes |
| `MAX_COOKIE_VALUE_BYTES` | 4096 | 264 bytes |

`ja4plus/utils/http_utils.py:parse_cookie_header` holds the three bounds, and all three
parsers call it:

- `ja4plus/utils/http_utils.py:parse_http_request`
- `ja4plus/utils/http_utils.py:extract_http_info`
- `ja4plus/fingerprinters/ja4h.py:_extract_http_info_from_bytes`

**The issue body named two files, and the defect sat in three call sites.** The third
site is the one the JA4H reassembled-stream path reads, so a fix that covered the two
named files would have left the stream path uncapped. The issue body now names it.

## Why absent output, and not truncation

Past any bound the cookie parse produces no result, and the request therefore produces
no JA4H value. The user decided that on 2026-08-08, on #175.

The JA4H cookie hash reads the content and the order of the list. A truncated list
therefore produces a value that compares **equal** to a different sender's. A tool whose
purpose is to compare one output against another must not emit a value that describes
traffic the sender did not send. Absent output is safer than silently wrong output.

`ja4plus/utils/ssh_utils.py:242` already held that shape past `MAX_SSH_BANNER_BYTES`.
The truncating style at `ja4plus/fingerprinters/ja4x.py:334` stays where it is, because
a stopped certificate read fails to parse rather than producing a comparable value.

The accepted cost is stated and is not a defect: a sender suppresses its own JA4H value
with a large Cookie header. The SSH banner bound already permits the same.

## Two readings that shaped the code

**The bound covers the list the parser builds, not the value it returns.**
`parse_cookie_header` reads the header through `_cookie_segments`, a generator, and stops
one segment past the cap. A header of 100000 pairs allocates 513 segments. `str.split`
would build one entry for every pair before refusing the result, which is the growth the
bounds exist to stop.

**The length bounds read the UTF-8 byte count.** The parsers decode with
`errors="ignore"`, so a name of 256 two-byte characters is 512 bytes. A character count
accepts it.

## The measurement on the base commit `b6468dc`

The premise of the issue holds, and it is wider than the body states:

```
--- repeated a=1 x 100000, header bytes 499998
parse_http_request                       12.1 ms  100000
_extract_http_info_from_bytes            11.4 ms  100000
extract_http_info                        11.0 ms  100000
--- unique names x 100000, header bytes 988888
parse_http_request                       17.3 ms  100000
_extract_http_info_from_bytes            16.6 ms  100000
extract_http_info                        16.4 ms  100000
single pair: name bytes 1000000 value bytes 1000000
```

`TCPStreamReassembler` already refused a segment past `max_stream_bytes=1048576`
(`ja4plus/utils/tcp_stream.py:130`), which held the list near 260000 pairs. That is an
outer bound, not the entry count the `CLAUDE.md` state rule states.

## How tested

Every bound is proven by removing it and counting the failures. A cap that no test
detects is a cap that is not proven.

| Removal | Cases that fail |
|---|---|
| The pair cap | 4 |
| The name cap | 2 |
| The value cap | 2 |
| The UTF-8 byte count, replaced by the character count | 1 |
| The lazy segment reader, replaced by `str.split` | 1 |
| The third call site, left uncapped | 5 |

`tests/test_http_cookie_bounds.py` holds 14 cases. The counting case asserts
`0 < len(read) <= MAX_COOKIE_PAIRS + 1`, because the upper bound alone passes when the
parser never calls the reader.

The five gates, run in the worktree:

```
pytest tests/ -m "not spec_validation"   1222 passed, 1641 deselected, 8 xfailed, 93 subtests passed
pytest tests/ -m spec_validation         1375 passed, 146 skipped, 1230 deselected, 120 xfailed
ruff format --check ja4plus/ tests/      114 files already formatted
```

**No JA4H value moved, measured rather than asserted.** Every capture under
`tests/foxio_vectors/` was replayed through `JA4HFingerprinter` on `b6468dc` and on this
branch:

```
base values 82, branch values 82, captures that differ 0
```

The comparison covers the hashed form and the `JA4H_ro` raw original-order form.
`tests/foxio_deviations.json` holds 120 keys, against 120 `xfailed` cases, before and
after.

Coverage, `--cov=ja4plus`:

| Reading | Base | Branch |
|---|---|---|
| Total | 81% | 81% |
| Missed lines | 660 | 659 |
| `ja4plus/utils/http_utils.py` | 91% | 92% |
| `ja4plus/fingerprinters/ja4h.py` | 89% | 89% |

## Docs

- `docs/specs/features/03-concurrency-safety.md` gains a `Per-request bounds the code
  holds today` section that states the three bounds and the policy.
- `docs/specs/spec.md` gains a Changelog entry with the round number `TBD`.

## Out of scope, filed

#188 — the same three parsers hold no bound on the header list. A request of 100000
header lines builds 100000 entries in 38.2 ms. It takes the same policy when it is
scheduled.

Refs #175


Crank-Git:
## Self-review

### Correctness

`_cookie_segments` produces the same segments as `cookie_header.split(";")` on every
edge input I could name. Measured:

```
'' match
';' match
'a=1;' match
';a=1' match
'a=1' match
'a=1;;b=2' match
'  a = 1  ;b=2' match
```

`segment.partition("=")` accepts the same segments as the old
`if "=" in pair: pair.split("=", 1)`, and it produces the same name and the same value.
`str.partition` returns the whole string and two empty strings when it finds no
separator, so the `if not separator` test rejects exactly the segments the old `in` test
rejected.

No request that stays under every bound behaves differently. The 82 JA4H values that
`tests/foxio_vectors/` produces are unchanged, on the hashed form and on the raw
original-order form, and `tests/test_ja4h_cookie_list.py` passes unchanged.

`parse_cookie_header` raises nothing. `str.find`, `str.partition`, `str.strip` and
`str.encode` with `errors="ignore"` raise on no `str` input, and every caller reaches it
with a `str` the parser decoded with `errors="ignore"`.

### Time and memory on hostile input

The parse is linear in the header length, and it holds at most 512 entries. Measured:

```
huge name (20 MB, no separator)     None    1.1 ms
huge value (20 MB)                  None    0.7 ms
million pairs                       None    0.2 ms
```

The million-pair case is the one the issue names, and it costs 0.2 ms rather than the
12.1 ms the base commit measured, because the parser stops at 513 segments.

**One case costs more than the rest.** A segment that holds no equals sign contributes
no entry, and it counts against no bound, so a header of separators alone reads to the
end:

```
1 MiB of separators:  97.5 ms
2 MiB of separators: 198.7 ms
4 MiB of separators: 370.8 ms
```

That is linear, and the base commit costs more on the same input, because
`str.split(";")` builds one entry for every separator before anything reads them. It is
an improvement, not a regression. Counting an empty segment against the pair cap would
refuse a header the acceptance criteria do not name, so this change leaves it.

### One finding, filed rather than fixed: #190

`ja4plus/fingerprinters/ja4h.py:112` reads a complete header block, produces nothing, and
returns without removing the stream. Every later packet on that connection then re-reads
the whole buffer, which grows to `max_stream_bytes`, 1048576.

Measured with one request and 300 further packets of 1400 bytes:

```
refused by the new cookie bound                  123.6 ms  stream bytes held   424735  values 0
refused by the request line (base behaviour)      46.9 ms  stream bytes held   420036  values 0
accepted, one cookie                              37.8 ms  stream bytes held        0  values 1
```

**The path predates this change**, and `GET / HTTP/11` already reached it. This change
gives a sender a second entrance to it. The fix belongs in `process_packet`, it changes
the behaviour of a request this issue does not name, and it needs its own vector
measurement. #190 owns it.

This change widens no buffer and changes no read depth, so the bounds #33 and #170 added
need no re-measurement.

### Tests that would pass against wrong code

I looked for the vacuous shape the project has found twelve times. One case carried it
and now does not: `test_the_parser_reads_no_more_segments_than_one_past_the_pair_cap`
asserts `0 < len(read) <= MAX_COOKIE_PAIRS + 1`. The upper bound alone passes when the
parser never calls the reader at all, which is the state a `str.split` implementation
leaves.

Every bound is proven by removal:

| Removal | Cases that fail |
|---|---|
| The pair cap | 4 |
| The name cap | 2 |
| The value cap | 2 |
| The UTF-8 byte count, replaced by the character count | 1 |
| The lazy segment reader, replaced by `str.split` | 1 |
| The third call site, left uncapped | 5 |

Crank-Git:
## Review findings, and what each one produced

A second reviewer read the diff for vacuous assertions, for the writing standard, for
dead code, and for a documented number that contradicts the code. Four findings, three
accepted and one rejected. `5fe1d6d` carries the fixes.

**Accepted — one vacuous assertion.**
`test_the_parsed_request_path_reports_no_cookie_above_a_cap` ended with
`assert _convert_parsed_to_extract_format({}) is not None`. That assertion reads a
hard-coded empty dictionary, not the parse under test, and it passes against any
production code at all. The case is now
`test_the_parsed_request_path_converts_a_header_at_the_cap_and_nothing_above_it`: it
converts a header of 512 pairs, asserts the converted pair count is 512, and asserts the
513-pair header parses to nothing. A parse that truncated would reach the converter with
512 of the 513 pairs and fail the second assertion.

That is the thirteenth instance of this shape the project has found.

**Accepted — two sentences of 27 words**, above the 25-word limit `.claude/rules/ste.md`
states. Both are split. One comment used `encoding` as a noun and stated the guarantee
of the branch it does not take; it now names `str.encode` and reads as a bound.

**Accepted — a line number the change itself moves.** The Changelog cited
`ja4plus/fingerprinters/ja4h.py:229`, which held `cookie_values = []` after the change.
It now names `_extract_http_info_from_bytes` and no line. Stale line numbers have cost
this batch twice already.

**Rejected — the comment repeated at three call sites.** `# A header past a bound
produces no request. #175 states the reason.` appears at all three sites deliberately. A
reader of any one site needs the reason there, and the writing standard asks for one
concept in one word rather than three phrasings of it.

The reviewer also confirmed the three bound values and the three vector readings agree
across `ja4plus/utils/http_utils.py`, `docs/specs/features/03-concurrency-safety.md` and
the Changelog entry, and that no other case in the file is vacuous.

## The gates after the fixes

```
pytest tests/                            2597 passed, 146 skipped, 128 xfailed, 93 subtests passed
pytest tests/ -m "not spec_validation"   1222 passed, 8 xfailed
ruff format --check ja4plus/ tests/      114 files already formatted
tests/foxio_deviations.json              120 keys
```

The removal proof holds unchanged after the test rewrite: the pair cap fails 4 cases,
the name cap 2, the value cap 2, the byte count 1, the generator 1, and the third call
site 5.

The 82 JA4H values were replayed again against the base commit `b6468dc`:
`base values 82, branch values 82, captures that differ 0`.

Coverage, `--cov=ja4plus`: total 81% on the base and 81% on the branch, 660 missed lines
against 659.

Crank-Git:
## Gate finding accepted: the enforcement was measured and the three numbers were not

The finding is correct, and my verdict overstated what the boundary cases prove.

Every boundary case builds its input from the constant, as `MAX_COOKIE_PAIRS + 1`. Each
proves that the refusal happens **at** the constant. None reads the constant. Nothing in
the suite asserted 512, 256 or 4096, so a later `MAX_COOKIE_PAIRS = 5` would suppress the
JA4H value of ordinary traffic against a green suite. The user decided those three
numbers on 2026-08-08, and a decided value that no comparison reads is the shape #172
hunts.

`62a1cf9` adds `test_the_three_bounds_hold_the_values_the_user_decided_on_2026_08_08`. It
reads the three numbers, and its docstring cites
https://github.com/Crank-Git/ja4plus/issues/175#issuecomment-5223660889 and the vector
readings each number sits above. The boundary cases are unchanged, because building the
input from the constant is the right way to write them.

## The proof, run the way the other bounds were proven

One decided value changed at a time. The new case fails alone, and all 14 boundary cases
stay green:

```
MAX_COOKIE_PAIRS = 512  ->  MAX_COOKIE_PAIRS = 5
    FAILED test_the_three_bounds_hold_the_values_the_user_decided_on_2026_08_08
    1 failed, 14 passed
MAX_COOKIE_NAME_BYTES = 256  ->  MAX_COOKIE_NAME_BYTES = 8
    FAILED test_the_three_bounds_hold_the_values_the_user_decided_on_2026_08_08
    1 failed, 14 passed
MAX_COOKIE_VALUE_BYTES = 4096  ->  MAX_COOKIE_VALUE_BYTES = 16
    FAILED test_the_three_bounds_hold_the_values_the_user_decided_on_2026_08_08
    1 failed, 14 passed
```

The gate's own reading reproduces, and the new case closes it. All three raised to
1000000:

```
    FAILED test_the_three_bounds_hold_the_values_the_user_decided_on_2026_08_08
    FAILED test_the_reassembled_stream_path_still_produces_a_value_at_the_pair_cap
    FAILED test_the_parser_reads_no_more_segments_than_one_past_the_pair_cap
    3 failed, 12 passed
```

Before this commit that reading was 12 of 14 green.

The enforcement proof is unchanged and now reports against 15 cases:

| Removal | Cases that fail |
|---|---|
| The pair cap | 4 |
| The name cap | 2 |
| The value cap | 2 |
| All three guards at once | 8 |
| The UTF-8 byte count, replaced by the character count | 1 |
| The lazy segment reader, replaced by `str.split` | 1 |
| The third call site, left uncapped | 5 |

The new case still passes when the guards are removed, which is correct: it pins the
values, and the boundary cases pin the enforcement.

## Gates

```
pytest tests/                            2598 passed, 146 skipped, 128 xfailed, 93 subtests passed
pytest tests/ -m "not spec_validation"   1223 passed, 8 xfailed
ruff format --check ja4plus/ tests/      114 files already formatted
coverage --cov=ja4plus                   81% total, 659 missed lines
```

The `docs/specs/spec.md` Changelog entry is extended rather than duplicated. It records
the gap, the reading that found it, and why the boundary cases stay as they are.

Issue #191, opened 2026-08-08
Remove the deprecated ja4plus/collector.py


Crank-Git:
## What the sweep measured

The mutation sweep of #172 applied 12 mutations to `ja4plus/collector.py`. **No mutation
failed one case.** No test imports the module, and no code under `ja4plus/` imports it.

## Why it can go

`ja4plus/collector.py` opens with its own removal notice:

```
DEPRECATED: Use the `ja4plus` CLI command instead.
This module is kept for backward compatibility and will be removed in v0.4.0.
```

The project publishes 0.6.0, so the stated removal release passed two minor versions
ago. The module holds 217 lines that no case measures, and `ja4plus/cli.py` carries the
command-line program.

## What to build

1. Confirm that no caller outside the repository needs the module. The Go port at
   `Crank-Git/ja4plus-go` names the decision if it ships an equivalent.
2. Remove `ja4plus/collector.py`.
3. Record the removal in the `docs/specs/spec.md` Changelog.

## Acceptance criteria

- [ ] `ja4plus/collector.py` no longer exists.
- [ ] The unit suite and the conformance suite report the counts they report today.
- [ ] The `docs/specs/spec.md` Changelog holds the removal.

Found by the mutation sweep on 2026-08-07. Out of scope for #172, which repairs cases
that cannot fail and adds no coverage target.

Crank-Git:
## `collector.py` now diverges from the command-line program on JA4SSH

#214 landed the rule that JA4SSH emits the window a connection holds open when the
packet source ends. `ja4plus/cli.py` runs `close_open_windows()` after the file reader
of `analyze` and after `live` stops. `tests/conformance_index.py` runs it after the last
packet of a capture.

**`ja4plus/collector.py:195-217` holds its own `sniff` loop and its own fingerprinter
map, and it runs no such call.** A live SSH connection read through `collector.py`
therefore emits no trailing window, and the same traffic read through `ja4plus live`
emits one.

**#214 did not wire it, for two reasons.**

1. The output of `collector.py` is packet-shaped. `collector.py:140-181` builds each
   record from the packet that produced the value. A trailing window has no packet, so
   a record needs a shape that this module does not hold. That is a design decision,
   and #214 carried no license to make it.
2. This issue removes the module. A new record shape in a module that this issue deletes
   is work that the delete throws away.

**This comment records the divergence so that the removal is the answer to it.** If the
removal does not land, the module needs the call and the record shape.

[...]

PR #205, opened 2026-08-08
Sweep the suite for assertions that cannot fail

Crank-Git:
Builds #172. Batch 15 (#185), member 4 of 4.

## What this adds

`tests/mutation_sweep.py` asks of every case: **would this fail if the code were
wrong?** It changes one expression in one module under `ja4plus/`, runs the suite, and
records which cases fail. The change is a measurement, and the sweep reverts it.

Coverage answers a different question. Coverage reports which line runs. The sweep
reports whether a line that runs is measured.

## The first sweep

12 mutations for each of the 26 modules, 275 in all. Each mutation ran the whole suite
of 2857 cases. 976 cases failed for no mutation, and 248 of those are cases the baseline
reports as skipped or as xfailed, which cannot fail.

`docs/mutation_sweep.md` is the generated report. It lists every module with the
mutations the sweep applied to it, each mutation with its result, and every candidate
with its file.

| Module | Mutations | Killed |
|---|---|---|
| `ja4plus/fingerprinters/ja4t.py` | 12 | 12 |
| `ja4plus/utils/tcp_stream.py` | 12 | 11 |
| `ja4plus/utils/x509_utils.py` | 12 | 1 |
| `ja4plus/collector.py` | 12 | 0 |
| `ja4plus/fingerprinters/__init__.py` | 9 | 0 |

## The two cases repaired, and the proof of each

**`tests/test_ja4l.py` decided no JA4L value.** The handshake assigned no packet time,
so scapy stamped the wall clock, and the assertions matched `JA4L-S=\d+_\d+`, which
every latency meets.

```
LATENCY_DIVISOR = 2  ->  3    survives all 224 mutations of ja4l.py against tests/test_ja4l.py
                              fails 138 cases of the whole suite
```

The FoxIO vectors held the divisor. The unit suite did not. The handshake states its own
times now, and the cases compare against `JA4L-S=10000_64` and `JA4L-C=15000_128`.
Against the same mutation:

```
FAILED tests/test_ja4l.py::TestJA4L::test_ja4l_direct_generation
FAILED tests/test_ja4l.py::TestJA4L::test_ja4l_fingerprinter_class
```

**`test_ja4l_os_estimation` failed for none of the 224 mutations.** It read one word of
one answer at one TTL of each range, and `assertIn("Windows", ...)` also accepts the
mutated string `Windows_mutated`. It names the edges 64, 65, 128 and 129 now. Against
`if ttl <= 64` -> `if ttl <= 65`:

```
FAILED tests/test_ja4l.py::TestJA4L::test_ja4l_os_estimation
```

Swept again after the repair, `tests/test_ja4l.py` reports **0 candidates** where it
reported 1, and 96 of 224 mutations killed where it reported 92.

## Candidates the reader rejected

A case may be correct and the mutation wrong. Every mutation of `tcp_stream.py` against
`tests/test_tcp_stream.py` named three candidates, and each is a correct case the
operators cannot reach: one reads the module attributes, one reads an early
`return None`, and one compares a default against a gap the vectors hold.

**The sweep also found a defect in itself.** `pytest` prints a failing subtest as
`SUBFAILED(hop_count=23)` and prints no `FAILED` line for the case that holds it, so
the first form counted a case with subtests as measured by nothing. That produced one
false candidate, `test_reads_each_row_of_the_foxio_table`. The pattern reads both forms
now.

## Out of scope, filed

- `ja4plus/utils/x509_utils.py` loses 173 of its 181 mutations against the four JA4X
  test files, and `--cov=ja4plus` reports 17 percent for it. That is a coverage
  question, and #172 excludes coverage.
- `ja4plus/collector.py` loses all 12, no module and no test imports it, and it carries
  its own removal notice for v0.4.0 against a published 0.6.0. **#191** owns the removal.

## How this was tested

```
pytest tests/ -m "not spec_validation"   1208 passed, 8 xfailed, 93 subtests passed
pytest tests/ -m spec_validation          1375 passed, 146 skipped, 120 xfailed
ruff check ja4plus/ tests/                All checks passed!
ruff format --check ja4plus/ tests/       114 files already formatted
mypy ja4plus/                             Success: no issues found in 27 source files
pytest tests/ --cov=ja4plus               TOTAL 3449 statements, 638 missed, 82%
```

Both counts equal the counts on the base commit `b6468dc`. `tests/foxio_deviations.json`
holds 120 keys and its checksum `0ed08f5a6c3efa71ede56f4eacc012c9b64a0627` is unchanged.
No fingerprinter changed, and no fingerprint moved. Coverage is 82 percent before and
after.

Verified against: https://docs.pytest.org/en/stable/how-to/output.html (pytest 9.1.1, retrieved 2026-08-07)


Crank-Git:
## Self-review

Three lenses: correctness of the sweep tool, safety of a tool that writes to source
files, and the writing standard. `af5c580` carries the fixes.

### Fixed

1. **`tests/mutation_sweep.py`: `signal.SIGHUP` does not exist on Windows.**
   `signal.signal(signal.SIGHUP, stop)` raises `AttributeError` there, and
   `pyproject.toml` classifies the project `Operating System :: OS Independent`. The
   handler reads each name with `getattr` now.
2. **A `SIGKILL` never reaches the restore.** This happened during the work: a killed
   sweep left `'JA4TSFingerprinter_mutated'` in `ja4plus/fingerprinters/__init__.py`.
   The startup check caught it, because the sweep refuses to start on a dirty
   `ja4plus/`. The rule and the module docstring name the recovery command.
3. **A checkpoint belongs to one commit.** It keys each result on the byte offset of the
   expression, so a code change moves the key. The rule states it.

### Read and kept

- **Restore safety.** Every module write sits inside a `try`/`finally` that writes the
  original text back. The handler raises `SystemExit`, so a stop signal reaches the
  `finally`. A `KeyboardInterrupt` reaches it as well.
- **The splice.** A mutation replaces the span between two operand offsets, so a string
  literal that holds an operator word cannot be reached. A chained comparison advances
  the left operand to each comparator, so `a < b < c` gives two mutations and each one
  is bounded.
- **The constant guard.** `source[start:start + 1]` rejects a prefixed literal such as
  `r"..."`, and an f-string is a `JoinedStr` and not a `Constant`, so the sweep splices
  neither.
- **Python 3.9.** The module opens with `from __future__ import annotations` and uses
  `Dict`, `List` and `Optional`. `ruff` passes at `target-version = "py39"`.
- **The two JA4L values.** `scapy` stores an assigned `packet.time` as a `float`, and
  `round(1000.020 * 1000000)` gives `1000020000` exactly. The measured latency is 10000
  and 15000 on every run.

### Not fixed, and why

The mutation operators change no return value and remove no clause. Three cases of
`tests/test_tcp_stream.py` and two of `tests/test_ja4t_deep.py` reach this limit, and
each is a correct case. `docs/mutation_sweep.md` states the limit above the candidate
list, so a reader meets it before the names.

### Gates after the fixes

```
pytest tests/ -m "not spec_validation"   1208 passed, 8 xfailed, 93 subtests passed
ruff format --check ja4plus/ tests/      114 files already formatted
```

`tests/foxio_deviations.json` holds 120 keys, checksum
`0ed08f5a6c3efa71ede56f4eacc012c9b64a0627`, unchanged.

Issue #34, opened 2026-08-06
Emit one JA4L client fingerprint per connection


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/02-correctness-audit.md`](docs/specs/features/02-correctness-audit.md):

> FR-correctness-audit-4 — JA4L emits one client fingerprint per TCP connection.

#89 withdrew `FR-correctness-audit-5`, which this issue also quoted. The vectors contradicted it. `FR-spec-conformance-17` now gives the client measurement point, and #88 built it. This issue owns multiplicity alone.

> FR-correctness-audit-11 — The project removes `_src_is_client` from `ja4plus/fingerprinters/ja4l.py`.

## What to build

`ja4plus/fingerprinters/ja4l.py:223` matches every ACK without a SYN flag. Each later ACK overwrites timestamp `C` and emits another client fingerprint with a larger latency. A connection with 1000 post-handshake ACKs produces 1000 JA4L-C values, each one wrong.

`_src_is_client` at line 279 is never called. Remove it.

Do not change the QUIC path in the same slice; it was fixed in `1ed28b3` and a vector covers it.

## Acceptance criteria

- [ ] A TCP connection with 100 ACKs after the handshake produces exactly one `JA4L-C` value.
- [ ] ~~The client fingerprint comes from the first ACK after the SYN-ACK.~~ Withdrawn by #89. `browsers-x509.pcapng` stream 0 contradicts it: the reference `JA4L-C` is `278_128`, and the first ACK after the SYN-ACK gives 56.
- [ ] `rtk git grep -n "_src_is_client" ja4plus/` reports nothing.

## Links

- Feature: [`docs/specs/features/02-correctness-audit.md`](docs/specs/features/02-correctness-audit.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #13

Crank-Git:
## This issue widens to the server side, and its acceptance criterion is wrong

#80 ran the JA4L fingerprinter in the conformance harness for the first time. The
measurement is on #80 and #12.

### The acceptance criterion is contradicted by a vector

This issue says:

> The client fingerprint comes from the first ACK after the SYN-ACK.

**A vector says otherwise.** `browsers-x509.pcapng` stream 0:

[...]

Issue #209, opened 2026-08-08
Batch 16: the register and the state rule


Crank-Git:
Batch 16, loose members. Integration branch `batch/193-register-and-state-rule`, off `dev`
at `6fc248f`.

The two members share one subject: **the register and the rules a reader meets before it.**
Neither moves a fingerprint.

## Members

- [ ] #193 — Six streams where the reference holds a JA4H value and this project holds
      none, and 49 register entries that carry a decision the register does not mark.
      `priority:high`.
- [ ] #175 — Name the boundary of the state rule, and record that the cookie list carries
      no bound. **Sequenced after #195 sub-merges on `epic/194-read-the-specification-images`,
      because both edit `CLAUDE.md`.**

## The gate

- `pytest tests/ -m spec_validation` reports the same `xfailed` count as the key count of
  `tests/foxio_deviations.json`, before and after. The count is **120** on `dev` at
  `6fc248f`.
- No file under `ja4plus/` changes unless a measurement requires it, and #193 part 1 is the
  only member that may require one.

Opened 2026-08-08. Nothing merged yet.

Crank-Git:
**#193 sub-merged as `9780b91`**, Changelog round 51. Both defects repaired and proven by reverting; the register moved from 120 keys to 114, and 114 equals the `xfailed` count.

**#175 is in flight**, branch `issue/175-name-the-state-rule-boundary`. It was sequenced behind #195, which merged on the epic branch.

Register on this branch: **114 keys, 108 decided, 6 undecided.** The six that stay all name #34, and the audit's claim that Changelog round 38 settled them is **not confirmed** — round 38 closed #34 on the stored-list reading and round 39 settled the return path, and neither reaches this shape.

Crank-Git:
**Batch 16 merged into `dev` as `e8c0252`.** All nine continuous integration jobs pass.

| Member | Round | State |
|---|---|---|
| #193 | 51 | Closed |
| #175 | 55 | Closed |

**The register on `dev` now holds 114 keys and 6 undecided entries**, against 120 and 55 at the start of the session. The key count equals the `xfailed` count.

[...]

Issue #193, opened 2026-08-08
Read the six JA4H streams the reference holds and this project omits, and mark the settled register entries


Crank-Git:
## Context

Found by a register audit on 2026-08-08, after #175 was reverted. The audit read all 120
entries of `tests/foxio_deviations.json` and grouped them by direction: does `ja4plus`
produce a value the reference omits, or does the reference produce a value `ja4plus`
omits?

**Six entries read the second way, and no decision covers them.** Their owning issue #35
is closed, so nothing in the tracker holds them.

## Part 1. Six streams where the reference holds a JA4H value and this project holds none

| Entry | Cause the register states |
|---|---|
| `CVE-2018-6794.pcap/JA4H` | `ja4plus produces no JA4H fingerprint the reference holds.` |
| `CVE-2018-6794.pcap/JA4H_ro` | The same |
| `http-empty-useragent.pcap/0:57722/JA4H.1` | `ja4plus produces no JA4H value on this stream.` |
| `http-empty-useragent.pcap/0:57722/JA4H_ro.1` | The same |
| `http-empty-useragent.pcap/JA4H` | The same |
| `http-empty-useragent.pcap/JA4H_ro` | The same |

None carries a `decided` marker. **An entry with no marker states that nobody has read
it**, and these six have sat in the register while the count around them was measured
eleven times.

**The name of one capture states the likely cause.** `http-empty-useragent.pcap` holds a
request whose `User-Agent` header carries no value. A parser that rejects that request
rejects a request the reference reads. #151 was the same shape: a parser rejected seven
streams the reference reads, and repairing it raised the register from 99 keys to 104.

### What to build

1. **Measure why each of the six produces nothing.** Read the parser path each stream
   takes. Do not accept the register's cause text as the reason; it records the symptom.
2. **Repair what is a defect.** A request the reference reads and this project rejects is
   a defect unless a reading of the reference proves otherwise.
3. **Decide what is not a defect, and record it.** An entry that stays needs a `decided`
   marker, a cause that states the measured reason, and a `Divergence register` row if the
   two implementations disagree on behaviour rather than on a defect.

### Acceptance criteria

- [ ] Each of the six entries is measured, and the measurement names the parser path and
      the reason the stream produces nothing.
- [ ] Every one that is a defect is repaired, and a test proves the repair by reverting it.
- [ ] Every one that stays carries a `decided` marker and a cause that states the measured
      reason.
[...]

Issue #190, opened 2026-08-08
Drop a JA4H stream whose complete header block produces nothing


Crank-Git:
## Context

`JA4HFingerprinter.process_packet` reads the reassembled stream on every packet. It
removes the stream after it records a fingerprint. It removes no stream when the parse
produces nothing.

## The defect

`ja4plus/fingerprinters/ja4h.py:112` reads a complete header block and produces nothing
when the parse fails. The method returns at line 114 and leaves the stream in the
reassembler. `_drop_an_unusable_stream` does not cover the case, because the buffer
starts with an HTTP method token and `is_http_request` reports True.

Every later packet on that connection therefore re-reads the whole buffer. The buffer
grows to `max_stream_bytes`, which is 1048576, so the cost per packet grows with the
buffer and the packet count carries no bound.

A complete header block admits no later byte that changes its parse. Nothing reads the
buffer again.

## The measurement

Base commit `b6468dc`, one request followed by 300 further packets of 1400 bytes:

```
refused by the request line (base behaviour)      46.9 ms  stream bytes held   420036  values 0
accepted, one cookie                              37.8 ms  stream bytes held        0  values 1
```

The accepted request leaves no stream. The refused request holds 420036 bytes and
re-reads them 300 times.

The cost of one re-read grows with the buffer. A Cookie header of semicolons measures
97.5 ms for one mebibyte, 198.7 ms for two, and 370.8 ms for four.

## What #175 changes

#175 gives a sender a second way to reach this path: a Cookie header past a bound
produces no parse result. The path itself predates #175, and the request line
`GET / HTTP/11` already reached it. #175 does not widen the buffer, and it does not
change how deep a parser reads.

## What to build

- Remove the stream when the header block is complete and the parse produces nothing.
- Prove the removal with a test that counts the reassembler entries after the refused
  request.

## Acceptance criteria

- [ ] A request whose header block is complete and whose parse produces nothing leaves
      no entry in the reassembler.
- [ ] The 82 JA4H values that `tests/foxio_vectors/` produces do not change.

## Links

- Feature: [`docs/specs/features/02-correctness-audit.md`](docs/specs/features/02-correctness-audit.md)
- Found by: #175

Crank-Git:
## Plan — batch 20, integration branch `batch/259-interpret-fingerprint`

**Changelog round 74.** It is assigned. Do not choose one, and never write `TBD`.

[...]

PR #212, opened 2026-08-08
Read the six JA4H streams the reference holds and mark the settled register entries

Crank-Git:
Part of batch 16, tracked on #209. Refers to #193.

## What this changes

**Part 1 repairs two JA4H defects. Part 2 marks 43 register entries and leaves 6 open.**

### The measurement contradicts the register on one of the two captures

The register stated one cause for all six entries: `ja4plus` produces no JA4H value the
reference holds. **That is true of one capture and false of the other.**

| Entry | The register said | The measurement says |
|---|---|---|
| `http-empty-useragent.pcap` (4 entries) | `ja4plus` produces no value | Correct. The parser reads `\r\n` as the only line ending. |
| `CVE-2018-6794.pcap` (2 entries) | `ja4plus` produces no value | **Wrong direction.** Both values match, and 10 extra occurrence keys fail the case. |

`test_the_produced_fingerprint_equals_the_reference[CVE-2018-6794.pcap-stream0:8089-JA4H.1]`
passes on the base commit. The failing case is
`test_the_produced_occurrence_keys_equal_the_reference[CVE-2018-6794.pcap-JA4H]`, and its
message reads `10 extra occurrence key(s)`.

### The specification decides intent, and the vector decides the bytes

`technical_details/JA4H.md` holds two facts. It states that JA4H fingerprints the HTTP
client "based on each HTTP request". It states that part a carries the "2 digit number of
headers, not counting Cookie and Referer". It names no line ending, and it names no rule
that reads the value of a header.

Both defects follow from that one sentence.

1. A request whose lines end with one line feed is an HTTP request. `ja4plus` read the two
   bytes `\r\n` as the only line ending, so it read
   `GET / HTTP/1.0\nUser-Agent:\n\n` as one line and produced nothing.
2. A retransmitted TCP segment carries no new request. `ja4plus` removed the stream from
   the reassembler after a value, so the retransmission rebuilt the same request.

**The name of the capture does not state the cause.** `http-empty-useragent.pcap` is named
for the header that carries no value, and the empty value is legal on both readings.
`b8bcd45ac095` is the first 12 characters of the SHA-256 hash of the 10 bytes
`User-Agent`, so the reference counts a header that carries no value. The line ending is
the whole cause.

Neither reading is a reference defect, so nothing is declined and no `Divergence register`
row is needed.

### The code

- `ja4plus/utils/http_utils.py` gains `split_http_lines` and `header_block_end`. All three
  parse paths read them, which is the shape Changelog round 40 records.
- `ja4plus/fingerprinters/ja4h.py` gains `consumed_seq`. It holds the sequence range of
  the request each stream produced a value for, and a segment that lies inside that range
  produces nothing. The comparison holds across a wrap of the 32-bit sequence number. A
  pipelined request sits above the range, so it still produces a second value.
- `ja4plus/utils/tcp_stream.py` renames `_seq_before` to `sequence_before` and `_SEQ_MASK`
  to `SEQUENCE_MASK`, because a caller outside the module now reads both.
- `consumed_seq` is bounded at 100 entries and at 600 seconds, which the `CLAUDE.md` state
  rule requires. The age reads the packet timestamp and never the wall clock.

## Every repair is proven by reverting it

| Revert | Result |
|---|---|
| `split_http_lines` and `header_block_end` removed | 6 cases of `tests/test_ja4h_line_endings.py` fail, and the conformance suite reports `4 failed`. |
| The `consumed_seq` guard removed | 2 cases of `tests/test_ja4h_retransmission.py` fail, and the conformance suite reports `2 failed`. |
| The start comparison removed | 3 cases of `test_a_new_connection_on_one_address_pair_still_produces_a_value` fail. |
| The range ended at the header block again | `test_a_retransmitted_request_that_carries_a_body_produces_one_value` fails. |
| `consumed_seq.popitem` removed | `test_the_consumed_request_table_stays_inside_the_stream_cap` fails. |
| The age eviction removed | `test_a_consumed_request_leaves_the_table_after_the_maximum_stream_age` fails. |

The tests came before the repairs. The first run of the two new files reported
`8 failed, 4 passed`, and the 4 that passed are the cases that guard the unchanged
behaviour.

## The self-review found three defects this branch introduced

Three lenses ran over the diff: correctness, hostile input, and documentation. The
hostile-input lens returned no finding. The other two are answered in the PR comments,
and all three code defects are repaired on this branch.

1. **The guard lost the first request of a reused connection.** It read the end of a
   segment alone, and a second connection on one address pair starts at its own initial
   sequence number. `consumed_seq` now holds the start of the request as well.
2. **The consumed range ended after the header block.** A request that carries a body
   reaches the fingerprinter in a segment that ends past the header block, so the
   retransmission produced a second value. The range now ends after the whole buffer.
3. **The new line rule moves a CRLF request whose header value holds a bare line feed.**
   `tshark` settles it in favour of the new rule, and the FoxIO Python implementation
   reads the `tshark` fields:

   ```
   $ tshark -r lf_in_value.pcap -T fields -e http.request.line
   X-Data: line1\n
   ```

   The request is
   `GET / HTTP/1.1\r\nX-Data: line1\n\nHost: example.com\r\nUser-Agent: probe\r\n\r\n`.
   This branch reports `['X-Data']`, the base commit reported `['User-Agent']`, and
   `tshark` reports one line. **The new rule matches the reference and the old one did
   not.** Each of the three readings is now a case the suite runs.

## The regression reading is zero

Every capture under `tests/foxio_vectors/` was replayed on `6fc248f` and on this branch,
through all six fingerprinters, and every produced value was compared.

- 38 captures, 647 values on the base and 638 on the branch.
- **36 of 38 captures are identical.**
- `CVE-2018-6794.pcap` JA4H falls from 12 values to 2. The 10 that leave are exact copies
  of the 2 that stay, and the 2 that stay equal the reference.
- `http-empty-useragent.pcap` JA4H rises from 0 values to 1, and it equals the reference.
- No other method moves on any capture. **No value that matched before stopped matching.**

## The invariant holds

| Reading | Base `6fc248f` | This branch |
|---|---|---|
| `pytest tests/ -m spec_validation` | `1375 passed, 146 skipped, 120 xfailed` | `1381 passed, 146 skipped, 114 xfailed` |
| `tests/foxio_deviations.json` keys | 120 | 114 |
| Entries with no `decided` marker | 55 | 6 |
| Entries with a `decided` marker | 65 | 108 |

The key count equals the `xfailed` count, before and after. Part 1 moved both together,
and part 2 moved neither.

## Part 2: what is marked, and what stays open

**43 entries gain a marker.** All 43 name #129 and state that the reference decrypts a
capture that carries key material. Changelog round 26 settled that on 2026-08-07:
`ja4plus` reads no key material, because a program that reads a key log file reads the
plain text of the traffic. Each cause now names the round.

**6 entries stay unmarked, and no round settled them.** They all name #34 and they all
report a JA4L value on a connection the FoxIO Python expected-output file holds no JA4L
value for.

| Entry | Why it is open |
|---|---|
| `CVE-2018-6794.pcap/JA4L-C` | The reference file holds no JA4L record. `ja4plus` reports 3 connections. |
| `CVE-2018-6794.pcap/JA4L-S` | The same. |
| `https-connect.pcap/JA4L-C` | The reference file holds stream 0 and no JA4L key. |
| `https-connect.pcap/JA4L-S` | The same. |
| `ssh2.pcapng/JA4L-S` | One extra occurrence key, `stream=15 .../JA4L-S.2`. |
| `tls-handshake.pcapng/JA4L-S` | 20 extra occurrence keys, all on streams the reference file omits. |

Round 38 closed #34 on the stored-list reading and moved the return path to #156. Round 39
settled the return path. **Neither reaches an entry of this shape**, so the audit's claim
that #34's six were settled is not confirmed, and the plan told this issue to leave them.

**One reading may settle two of the six, and it is a decision this issue did not take.**
`tests/foxio_vectors/rust_expected/ja4__insta@https-connect.pcap.snap` holds
`ja4l_c: 45_64` and `ja4l_s: 13532_57` for stream 0. `ja4plus` produces
`JA4L-C=45_64` and `JA4L-S=13532_57` on that stream. Changelog round 30 states that the
Rust snapshot decides for a stream the Python file omits, and the unit is the stream.
Whether "omits a stream" covers "holds the stream and omits the method" is a question for
the project manager.

## Two records outside the issue's own subject

- `docs/specs/features/03-concurrency-safety.md` gained a row for
  `JA4HFingerprinter.unusable_base`, which #33 built and the #179 audit omitted. The table
  states that it records what the code holds, and the new `consumed_seq` row sits beside
  it.
- Changelog round 48 records #190, which owns `ja4h.py:112`. This branch changed that line
  and kept its behaviour. #190 is untouched.

## How this was tested

```
pytest tests/ -m spec_validation      1381 passed, 146 skipped, 114 xfailed
pytest tests/ -m "not spec_validation"  1227 passed, 8 xfailed, 93 subtests passed
ruff check ja4plus/ tests/            All checks passed
ruff format --check ja4plus/ tests/   116 files already formatted
mypy ja4plus/                         Success: no issues found in 27 source files
```

Coverage rose. `pytest tests/ -m "not spec_validation" --cov=ja4plus` reports
`3449 statements, 660 missed` on the base and `3493 statements, 661 missed` on the branch.

`tests/generate_foxio_baseline.py` did not run. Every register edit is by hand.

Verified against: https://github.com/FoxIO-LLC/ja4/blob/main/technical_details/JA4H.md
(FoxIO JA4H specification, retrieved 2026-08-08)


Crank-Git:
## Self-review finding 1: the guard lost the first request of a reused connection

**I found this in my own change and repaired it in `644a028`.**

`_segment_carries_no_new_request` read the end of a segment alone. A second connection on
one address pair and one port pair starts at its own initial sequence number, and that
number sits below the stored one about half the time. The guard then read the first
request of the second connection as a retransmission and produced no value.

The probe, on the first form of the guard:

```
2000000000 -> 1000000000 values: 1
1000 -> 500 values: 1
3000000000 -> 2000000000 values: 1
```

Each line sends one complete request on one 4-tuple, then a second complete request at a
lower initial sequence number. Each should produce 2 values.

`consumed_seq` now holds the start of the consumed request beside its end. A
retransmission repeats bytes the request carried, so it starts at or after the request. A
segment that starts before the request names a new connection, and the stored entry leaves.

A first repair used a distance bound of `max_stream_bytes` instead. **It is rejected**,
because it still lost the `1000 -> 500` case: 500 bytes below the watermark reads as a
retransmission on that rule. The start comparison has no such gap.

`test_a_new_connection_on_one_address_pair_still_produces_a_value` proves the repair.
Three of its cases fail when the start comparison is removed:

```
FAILED tests/test_ja4h_retransmission.py::test_a_new_connection_on_one_address_pair_still_produces_a_value[2000000000-1000000000]
FAILED tests/test_ja4h_retransmission.py::test_a_new_connection_on_one_address_pair_still_produces_a_value[1000-500]
FAILED tests/test_ja4h_retransmission.py::test_a_new_connection_on_one_address_pair_still_produces_a_value[3000000000-2000000000]
```

The replay reading is unchanged: 36 of 38 captures identical, and the two that differ are
the two the issue names.

Crank-Git:
## Self-review findings 2 and 3, and one finding rejected

Three review lenses ran over the diff: correctness, hostile input, and documentation.
The hostile-input lens returned no finding. Two correctness findings and five
documentation findings are below.

### Finding 2, accepted: the consumed range ended after the header block

**Repaired in `d6f6c10`.** A request that carries a body reaches the fingerprinter in a
segment that ends past the header block. A range that ends at the header block leaves
that segment outside itself, so the retransmission produced a second value. That is the
same defect this issue repairs, with a body present.

```
POST /one HTTP/1.1\r\nHost: example.com\r\nContent-Length: 200\r\n\r\n<200 bytes>
sent twice on one 4-tuple, at one sequence number
-> 2 values before the repair, 1 after
```

The range now ends after the whole contiguous buffer the stream held.
`test_a_retransmitted_request_that_carries_a_body_produces_one_value` fails when the
range ends at the header block again.

### Finding 3, rejected: one line feed inside a header value

The review reported that the new line rule moves a CRLF request whose header value holds
a bare line feed, and that no vector requires the move. **The first half is correct and
the second is not.** `tshark` settles it, and the FoxIO Python implementation reads the
`tshark` fields.

```
$ tshark -r lf_in_value.pcap -T fields -e http.request.line
X-Data: line1\n
```

The capture holds one request,
`GET / HTTP/1.1\r\nX-Data: line1\n\nHost: example.com\r\nUser-Agent: probe\r\n\r\n`.
`tshark` reports one header line and stops at the line feed pair. It drops `Host` and
`User-Agent`.

| Rule | Header names |
|---|---|
| `tshark`, the reference | one line, `X-Data: line1\n` |
| This branch | `['X-Data']` |
| The base commit | `['User-Agent']` |

**The new rule matches the reference and the old rule did not.** The two behaviours are
one rule, not two. Once a line feed ends a line, no byte sits inside a value past it, and
that same rule is what gives `http-empty-useragent.pcap` its reference value. The reading
is now `test_one_line_feed_ends_a_line_inside_a_header_value`, and
`docs/implementation_notes.md` records the command and its output.

### Documentation findings, all accepted

Four sentences over the 25-word cap of `.claude/rules/ste.md` are split. The stale
docstring of `_remember_the_consumed_request` now names a range rather than one sequence
number. The review confirmed every number in the Changelog row against the register, and
it confirmed the state-bounds table against the code.

### The replay reading is unchanged

36 of 38 captures identical. The two that differ are the two the issue names.

Issue #195, opened 2026-08-08
Record the specification inventory and the reading method


Crank-Git:
Part of #194. **This lands before any other sub-issue of Epic 10 starts.**

## Context

An audit on 2026-08-08 read `technical_details/` at the pinned commit and measured what it
holds. **Two premises this project holds about the specification are false.**

| File | Size | What it holds |
|---|---|---|
| `README.md` | 1567 B | The table of twelve methods. Text |
| `JA4.md` | 9153 B | The JA4 algorithm. Text, and the only complete one |
| `JA4H.md` | **278 B** | **A stub.** One sentence of purpose and one rule, `2 digit number of headers, not counting Cookie and Referer`. It builds no fingerprint |
| 9 PNG files | — | `JA4`, `JA4D`, `JA4D6`, `JA4H`, `JA4L`, `JA4S`, `JA4SSH`, `JA4T`, `JA4X` |

**`docs/specs/spec.md` states that `technical_details/` holds `JA4.md` and `JA4H.md` as
text.** It does, but `JA4H.md` specifies no algorithm. **One method of twelve holds a
complete text specification.**

**Three methods hold no diagram at all: `JA4LS`, `JA4TS` and `JA4TScan`.** `JA4L.png` and
`JA4T.png` presumably carry their server forms, and that presumption is unconfirmed.

## What to build

1. **Record the inventory** in `docs/specs/foxio/README.md`: every file in
   `technical_details/` at the pinned commit, its size, its **SHA-256**, and the retrieval
   date. This is the base every other sub-issue reads.
2. **State the reading method** that the six method sub-issues follow:
   - Read the image directly. Transcribe it as this project's own prose into
     `docs/specs/foxio/<METHOD>.md`, with the source address, the pinned commit, the
     retrieval date and the SHA-256 of the image read.
   - **No image file is copied into this repository.** JA4+ carries the FoxIO license and
     this project publishes to PyPI.
   - **Corroborate every transcribed rule against two FoxIO-authored sources that are not
     the image**: the `ja4` README, the Wireshark plugin documentation, the doc comments
     under `rust/ja4/src/`, the Zeek scripts, or a FoxIO blog post. **A rule with fewer
     than two corroborations is marked uncertain and keeps the vector fallback.**
   - **Cite the source of every statement.** A transcription with an uncited rule is not
     accepted.
3. **Correct the two false premises** in `docs/specs/spec.md` and in
[...]

Issue #179, opened 2026-08-07
Correct the maximum-age default in the concurrency-safety feature file


Crank-Git:
## The defect

`docs/specs/features/03-concurrency-safety.md` states:

> The default maximum age is 300 seconds.

**The committed vectors hold a longer gap than that.** #170 measured the longest gap
between segments of one TCP stream across `tests/foxio_vectors/` and found **320.71
seconds**, in `ssh-r.pcap` on the connection between `192.168.1.169:64980` and
`192.168.1.197:22`. The project manager re-measured independently and read **320.34
seconds** on the same capture and the same connection, keying each direction separately
rather than the whole stream. Both readings are above 300.

A 300-second default therefore evicts a stream that a committed vector still holds.

## Why this is a specification issue and not a defect in #170

#170 shipped 600 seconds, following the measure-and-exceed rule its plan set. That value is
built and it is correct. The feature file is the artifact that is wrong.

**Nothing breaks today at either value.** The conformance counts are identical at 300 and
at 600, because the one stream above 300 seconds is SSH traffic that JA4H and JA4X read
nothing from. That is why the documented number survived until a measurement looked at it.

`docs/specs/features/03-concurrency-safety.md` is unbuilt, and its other numbers do not
describe this reassembler either. It names 10000 entries, and `max_streams` is 100 in JA4H
and 50 in JA4X.

## What to build

1. Correct the maximum-age default in `docs/specs/features/03-concurrency-safety.md` to
   600 seconds, and name the measurement that sets it.
2. Reconcile the other state numbers in that file against what the code holds, or state
   which of them are targets for Epic 3 rather than descriptions of today.
3. Add a line to `docs/specs/spec.md` § Changelog.

**Do not change a default in `ja4plus/`.** This issue corrects the document.

## Acceptance criteria

- [ ] `docs/specs/features/03-concurrency-safety.md` states 600 seconds and names the
      measurement.
- [ ] Every state bound the file states is either matched to the code or marked as an Epic
      3 target.
- [ ] No file under `ja4plus/` changes.

Found by #170 on 2026-08-07. Part of #14.

Crank-Git:
## Plan

Batch 14, on `epic/13-batch14`, off `dev` at `ebf2ebc`. #36 runs beside it and touches
`ja4plus/` only, so the two do not meet.

**This is a documentation change. No file under `ja4plus/` changes.**

1. Correct the maximum-age default in `docs/specs/features/03-concurrency-safety.md` to
[...]

PR #218, opened 2026-08-08
Name the boundary of the state rule (#175)

Crank-Git:
Builds #175. Batch 16 member, tracked on #209. Base is `batch/193-register-and-state-rule`.

## What

The state rule stated a bound and never named the object it binds. This branch names the
boundary. A state table survives across packets. A structure that one packet or one
request builds and releases is not a state table, and it holds neither bound.

One wording lands in four files:

- `CLAUDE.md`, in the `Conventions` list. **Rule 1 is untouched**, because #195 landed the
  authority rule there on another branch and this base forked `dev` before it.
- `.claude/rules/conformance.md`, in `## State rules`.
- `docs/specs/spec.md`, in `## Cross-cutting concerns`.
- `docs/specs/spec.html`, in the same table.

## Why

The user decided on 2026-08-08 that the cookie list carries no bound. The FoxIO reference
caps nothing, the Go port caps nothing, and no FoxIO vector reaches any candidate bound.
`ja4plus` matches all three. A bound merged as `be3604f` and the user reverted it as
`042e8c3`, so **no file under `ja4plus/` changes here**.

`docs/specs/features/02-correctness-audit.md` records the decision and the measurement in
one section. That file held no cookie bound paragraph to drop: the paragraph the revert
removed sat in `docs/specs/features/03-concurrency-safety.md`, and `042e8c3` already took
it out.

## This does not weaken Epic 3

All four files state it. The six unbounded state tables that #179 records survive across
packets, so the rule reaches them and they keep their bound.

## The divergence register gains no row

The register records a reading that diverges from the FoxIO reference. The reference holds
no cookie bound and `ja4plus` holds none, so the two agree.
`docs/specs/features/02-correctness-audit.md` states that, so a reader does not look for a
row that is not missing.

## The test, and how it is proven to fail

`tests/test_ja4h_large_cookie_header.py` holds 5 cases. Each builds a Cookie header past
all three reverted bounds: 5000 pairs, a 1024-byte name and a 8192-byte value. Each
asserts a complete JA4H value, part a through part d.

The proof of failure runs against both candidate policies:

1. **The refusing parser.** `git apply -3` of `be3604f` onto
   `ja4plus/utils/http_utils.py` and `ja4plus/fingerprinters/ja4h.py` restores
   `parse_cookie_header` and its three call sites. All 5 cases fail: 4 with
   `assert None is not None`, and `test_every_parser_site_reads_all_5000_pairs_of_a_large_cookie_header`
   with `AttributeError: 'NoneType' object has no attribute 'get'` at
   `ja4plus/fingerprinters/ja4h.py:342`.
2. **The truncating parser.** Changing that same `parse_cookie_header` to truncate rather
   than to refuse fails 4 of the 5. The case that passes is
   `test_a_cookie_header_of_5000_pairs_keeps_the_parts_that_read_no_cookie`, which reads
   part a and part b alone, and truncation keeps both. That is the correct result.

Both parser files are restored, and the diff touches nothing under `ja4plus/`.

## Verification

```
pytest tests/ -m "not spec_validation" -q    1232 passed, 8 xfailed, 93 subtests passed
pytest tests/ -m spec_validation -q          1381 passed, 146 skipped, 114 xfailed
tests/foxio_deviations.json                  114 keys
ruff check ja4plus/ tests/                   All checks passed!
ruff format --check ja4plus/ tests/          117 files already formatted
mypy ja4plus/                                Success: no issues found in 27 source files
pytest --cov=ja4plus                         81%, 661 missed of 3495
```

Coverage does not fall. The same run without `tests/test_ja4h_large_cookie_header.py`
reports 81% and 661 missed, so the new cases exercise lines the suite already covered.

## Changelog

The row carries the round `TBD`. Rounds 49, 50 and 51 are taken.

Verified against: https://github.com/FoxIO-LLC/ja4/tree/main/python (FoxIO `python/ja4h.py`, as quoted in `docs/implementation_notes.md`)


Crank-Git:
## Self-review

I ran the review sequentially. Two Sonnet children were spawned for a parallel lens pass
and neither reported inside 20 minutes, so this is my own pass. Four findings, all
repaired in `0b908d5`.

**1. Four assertions could not fail.** `tests/test_ja4h_large_cookie_header.py` held
`assert part_c != "000000000000"` and `assert part_d != "000000000000"` in two cases. Each
sat after an assertion that compares the same part against a sha256 prefix, which never
holds twelve zeros. Once the first assertion passes, the second cannot fail. `096f94f`
swept the suite for that exact shape. The four assertions are gone, and the docstring of
each case now names the parser change that fails it.

**2. `CLAUDE.md` did not name #179.** The other three files stated that the six unbounded
state tables keep their bound, and `CLAUDE.md` stated the weaker "The boundary removes no
bound from a state table." All four now name #179.

**3. `docs/specs/features/02-correctness-audit.md` said only that it held no cookie bound
paragraph.** A reader cannot see from that where the paragraph was. The text now states
that the paragraph sat in `docs/specs/features/03-concurrency-safety.md` and that `042e8c3`
removed it.

**4. The same section used `cap` and `bound` for one concept.** `.claude/rules/ste.md`
rule 7 refuses synonym rotation. The section now uses `bound` throughout. The phrase
`over-bound path` is also gone, in favour of `the path past a bound`.

### Two things I checked and left alone

**`produce` appears where the `## Terms` table lists it under "Do not use" for `emit`.**
The table forbids it as a synonym for `emit`, which is "to return a fingerprint from a
fingerprinter". Every use here describes a parse result rather than a fingerprinter
return, the existing cases in `tests/test_ja4h_cookie_list.py` read the same way, and the
user's own decision comment of 2026-08-08 reads "produces no JA4H value". Rewording would
diverge from the specification text this issue implements.

**`_expected_cookie_hashes` duplicates the sort rule of the implementation.** It sorts the
names for part c and sorts the pairs by name for part d, which is what
`_generate_ja4h_from_info` does. A case built this way follows the implementation if the
sort rule itself is wrong. That is deliberate: `tests/test_ja4h_cookie_list.py` already
asserts the sort rule against literal expected hashes, and this file asserts completeness
and pair count, which is the property the reverted work broke.

### The failure proof, re-run after the repair

The proof runs from the worktree and restores both parser files afterwards.

1. `git apply -3` of `be3604f` onto `ja4plus/utils/http_utils.py` and
   `ja4plus/fingerprinters/ja4h.py` restores `parse_cookie_header` and its three call
   sites. **All 5 cases fail.**
2. Changing that same `parse_cookie_header` to truncate rather than to refuse **fails 4 of
   the 5**. The case that passes is
   `test_a_cookie_header_of_5000_pairs_keeps_the_parts_that_read_no_cookie`, which reads
   part a and part b alone, and truncation keeps both. That is the correct result.

Issue #200, opened 2026-08-08
Transcribe JA4L and settle the server form, its part count and the delta fields


Crank-Git:
Part of #194. **Depends on #195**, which sets the reading method.

## What to build

1. **Transcribe `technical_details/JA4L.png`** by the method #195 states: read the image,
   write this project's own prose into `docs/specs/foxio/JA4L.md`, and record the
   source address, the pinned commit, the retrieval date and the **SHA-256 of the image**.
   **No image file is copied into this repository.**
2. **Corroborate every transcribed rule against two FoxIO-authored sources that are not the
   image.** The candidates are the `ja4` README, the Wireshark dissector, the doc comments
   under `rust/ja4/src/`, **the Zeek implementation under `zeek/`** and a FoxIO blog post.
   A rule with fewer than two corroborations is marked uncertain and keeps the vector
   fallback.
3. **Compare every transcribed rule against `ja4plus/fingerprinters/ja4l.py`**, field by field, and report each
   disagreement with a `file:line`.
4. **Compare the transcribed rules against the register.** An entry of
   `tests/foxio_deviations.json` that names this method is either explained by the
   specification or it is not, and the report states which.

## Three open questions this transcription must answer

1. **Does `JA4L.png` specify `JA4LS`?** `technical_details/` holds no `JA4LS.png`, and
   `technical_details/README.md` names `JA4LatencyServer` as a separate method.
2. **How many parts does the server form carry?** `zeek/tests/Traces/Scripts.ja4-conn/conn.log`
   holds `ja4ls 18862_59_14792`, which is three parts. `python/test/testdata/` holds a
   `JA4L-S` of two parts, and `ja4plus` emits the two-part form.
3. **What are `ja4l_delta` and `ja4ls_delta`?** The Zeek baseline holds both. No other
   reference this project reads mentions them, and `ja4plus` produces nothing of that name.

The register holds six `JA4L-C` and `JA4L-S` entries owned by #34, and the
`Divergence register` holds the return-path row that #156 settled.

## What not to build

**No fingerprint moves in this issue.** A disagreement produces a decision issue, and a
person decides it. `CLAUDE.md` rule 1 binds every fingerprinter change.

## Acceptance criteria

- [ ] `docs/specs/foxio/JA4L.md` holds the transcription, with the image SHA-256, the
      pinned commit, the retrieval date, and two cited corroborations per rule.
- [ ] Every rule that could not reach two corroborations is marked uncertain and listed.
[...]

Issue #198, opened 2026-08-08
Read the Zeek implementation and its baselines as a fourth reference


Crank-Git:
Part of #194.

## Context

An audit on 2026-08-08 enumerated every FoxIO-authored implementation of JA4+. **This
project uses three of six, and it uses the fourth for one method only.**

| Implementation | Where | This project uses it |
|---|---|---|
| Python | `python/` | Primary reference for every method |
| Rust | `rust/` | Ten snapshots, for the streams the Python file omits |
| Wireshark | `wireshark/` | JA4D and JA4D6 only |
| **Zeek** | `zeek/` | **JA4D only, through one log file** |
| **`FoxIO-LLC/ja4tscan`** | Separate repository, C | **Not read** |
| **`FoxIO-LLC/ja4-nginx-module`** | Separate repository, C | **Not read** |

**The Zeek package implements eight methods**: `ja4`, `ja4d`, `ja4h`, `ja4l`, `ja4s`,
`ja4ssh`, `ja4t` and `ja4x`. It carries its own test baselines under
`zeek/tests/Traces/`, and those baselines are an independent reading of the same captures
this project already holds.

## The finding that makes this urgent

`zeek/tests/Traces/Scripts.ja4-conn/conn.log` holds these fields:

```
ja4l   ja4ls   ja4l_delta   ja4ls_delta   ja4t   ja4ts
```

and this row of values:

```
ja4l 35_64   ja4ls 18862_59_14792   ja4l_delta (empty)   ja4ls_delta 0.8
ja4t 65535_00_00_00   ja4ts 65535_00_00_00
```

Three readings follow, and each one is a gap in this project.

1. **`JA4T` and `JA4TS` hold reference values after all.** #196 records that no
   `python/test/testdata/` file, no Wireshark file and no Rust snapshot holds one, so two
   shipped fingerprinters have no conformance evidence. **The Zeek baseline holds one.**
2. **`ja4ls` carries three parts, and `JA4L-S` in `python/test/testdata/` carries two.**
   This project emits the two-part form. Whether the two implementations name the same
   measurement, and which one the specification states, is unmeasured.
3. **`ja4l_delta` and `ja4ls_delta` appear in no other reference this project reads**, and
   `ja4plus` produces nothing of that name.

## What to build

1. **Enumerate every Zeek baseline** under `zeek/tests/Traces/` and record which methods
   and which captures each covers.
2. **Read the Zeek implementation of each method** under `zeek/ja4*/`. It is
[...]

Issue #196, opened 2026-08-08
Transcribe JA4T and measure the two methods that hold no vector coverage


Crank-Git:
Part of #194. **Depends on the inventory sub-issue.**

## Context

An audit on 2026-08-08 measured the FoxIO vector coverage per method. **`JA4T` and `JA4TS`
hold none.**

`ja4plus/fingerprinters/ja4t.py` and `ja4plus/fingerprinters/ja4ts.py` ship in the package.
A search for `JA4T` across every expected-output file under `tests/foxio_vectors/` matches
**only the `NOTICE` file**. No `python/test/testdata/` file, no Wireshark file and no Rust
snapshot holds a `JA4T` or a `JA4TS` value.

**Two shipped fingerprinters therefore have no conformance evidence of any kind.** The
conformance suite reports them neither as matching nor as deviating, and
`tests/foxio_deviations.json` holds no entry for either, because nothing compares them.

**This is the shape #172 sweeps for, at the scale of a whole method.** A comparison that is
never made reads as a comparison that passes. The suite has been green on JA4T and JA4TS for
as long as the project has existed, and that greenness measures nothing.

`technical_details/` holds `JA4T.png` and **no `JA4TS.png`**, so one diagram presumably
carries both forms. Confirm that rather than assume it.

## What to build

1. **Transcribe `JA4T.png`** by the method the inventory sub-issue states, and establish
   whether it specifies `JA4TS` as well as `JA4T`.
2. **Compare the transcribed rules against `ja4t.py` and `ja4ts.py`**, field by field, and
   report every disagreement. Do not change a fingerprinter in this issue.
3. **Find whether any reference value exists anywhere** — the Wireshark dissector, the Zeek
   scripts, the Rust snapshots, or a capture in the FoxIO `pcap/` directory that carries TCP
   traffic the method reads. Report what you find, including finding nothing.
4. **State what conformance evidence is achievable**, so a later issue can build it.

## Acceptance criteria

- [ ] `docs/specs/foxio/JA4T.md` holds the transcription with two corroborations per rule.
- [ ] The transcription states whether the diagram specifies `JA4TS`.
- [ ] Every disagreement between the transcription and `ja4t.py` or `ja4ts.py` is reported
      with a `file:line`.
- [ ] The search for a reference value is reported, with the sources searched named.
- [ ] No file under `ja4plus/` changes, and no fingerprint moves. A disagreement produces a
      decision issue.

Crank-Git:
[...]

Issue #194, opened 2026-08-08
Epic 10: Read the specification images


Crank-Git:
## Goal

**Read the JA4+ specification for the seven methods FoxIO publishes as images, and hold
the reading in this repository as text.**

## Why

`docs/specs/spec.md` records the risk: FoxIO publishes `JA4.md` and `JA4H.md` as text, and
it publishes JA4L, JA4S, JA4SSH, JA4T, JA4X, JA4D and JA4D6 as **PNG images**.
`.claude/rules/external-apis.md` then states the fallback:

> Where an image is **ambiguous**, the expected-output files under
> `python/test/testdata/` decide.

**That rule presumes a person read the image and found it ambiguous. Nobody read them.**
No file from `technical_details/` is in this repository, and every one of the seven methods
took the fallback by default. The project therefore holds no statement of intent for seven
of its ten fingerprinters, and it inferred their behaviour from the reference
implementations.

The user restated the project's authority rule on 2026-08-08:

> **The specification decides intent and schema. The vectors decide the exact bytes where
> intent runs out. A provable reference defect is declined and recorded.**

An unread image is not intent that ran out. It is intent that nobody looked at.

## What this epic produces

1. **A transcription of each image, held as text in this repository**, with the source
   address, the pinned commit, the retrieval date and the **SHA-256 of the image it was
   read from**. The transcription is this project's own prose. **No image file is copied
   into this repository**, because JA4+ carries the FoxIO license and this project
   publishes to PyPI.
2. **A corroboration for every transcribed rule**, from FoxIO-authored prose that is not
   the image: the `ja4` README, the Wireshark plugin documentation, the doc comments under
   `rust/ja4/src/`, the Zeek scripts, or a FoxIO blog post. **A rule with fewer than two
   independent corroborations is marked uncertain and keeps the vector fallback.**
3. **A comparison of each transcribed rule against what `ja4plus` does today and against
   what the vectors hold**, and a report of every disagreement.

## What this epic does not do

**No fingerprint moves in this epic.** A disagreement between the transcribed
specification and the behaviour of `ja4plus` produces a **decision issue**, and a person
decides it. `CLAUDE.md` rule 1 still binds every fingerprinter change.

This is expected to find disagreements. `JA4.md` is one of the two methods published as
text, and its ALPN prose already contradicts both FoxIO implementations. Seven more
methods will probably hold more of that shape. **A disagreement found is the value of this
epic, not a failure of it.**

## Exit criteria

[...]

PR #223, opened 2026-08-08
Transcribe JA4L and settle the server form, its part count and the delta fields

Crank-Git:
Part of #194. Builds #200. Base is `epic/194-read-the-specification-images`.

## What this changes

- `docs/specs/foxio/JA4L.md` is new. It is this project's own prose form of `technical_details/JA4L.png`, written under the #195 method.
- `docs/specs/foxio/README.md` closes the open question and adds the transcription row.
- `docs/specs/spec.md` gains one Changelog row. The round is `TBD`, because rounds 49 to 53 are taken.

**No file under `ja4plus/` changes and no fingerprint moves.** `git diff --name-only origin/dev...HEAD -- ja4plus/` returns nothing.

## The open question, answered

**`JA4L.png` specifies the client form alone.** It titles itself `JA4L: Light Distance/Location Fingerprint`, it labels its one example `JA4L=`, and it states no server rule. `technical_details/README.md:9` and `README.md:50` name `JA4LatencyServer` / `JA4LS` as a separate method, and `technical_details/` holds no `JA4LS.png`.

The two halves of the round 49 presumption have opposite answers. #196 found `JA4T/S` in the title of `JA4T.png`. JA4LS reaches no image at all.

## The finding that raises a decision

**The image labels three parts, and this project publishes two.**

```
JA4L=5191_42_45014
     \__/ \_/ \___/
      a    b    c

a = One-way TCP latency in µs (1ms = 1,000µs)
b = Observed TTL
c = One-way application handshake latency
```

- `wireshark/source/packet-ja4.c:1371-1374` and `:1381-1386` write the same three-part form. `wireshark/test/testdata/` holds 88 values in 15 files, and every one holds three parts.
- `zeek/ja4l/main.zeek:191-192` and `:132-133` write a third part. #198 owns that reading.
- `python/ja4.py:157` and `rust/ja4/src/time.rs:18-22` write two parts. **Every one of the 114 JA4L values in `tests/foxio_vectors/*.json`, across 24 files, holds two parts.**

R4 is therefore marked uncertain and keeps the vector fallback, so the two-part form stands and no fingerprint moves.

## The two questions #198 already answered

This page cites #198 and re-derives neither.

- The three-part `ja4ls` of the Zeek baseline. Zeek is the outlier against the reference values, and the image agrees with Zeek on the part count.
- `ja4l_delta` and `ja4ls_delta` are ratios and neither is a fingerprint part. The image corroborates that from the other side: the delta it names is a subtraction, `45014-5191 = 39823µs`, and it feeds the distance formula. `wireshark/source/packet-ja4.c:1377` and `zeek/ja4l/main.zeek:269` both write a ratio in a field separate from the fingerprint.

## The six register entries

**The specification explains none of the six.** The image states the schema of one value and the distance a reader derives from it. It states no emission rule.

Round 38 settled the stored list and the return path of `process_packet`. It did not settle these six.

**Five of the six compare against an expected-output file that publishes no JA4L key at all.** `CVE-2018-6794.pcap` and `https-connect.pcap` hold `JA4H` keys alone, and `tls-handshake.pcapng` holds `JA4`, `JA4S` and `JA4X` keys alone. `python/ja4.py:339` deletes `JA4L-S` and `JA4L-C` when the run names another method. `grep -c '"JA4L-'` reports 0 for each of the three upstream files and 19 for `python/test/testdata/ssh2.pcapng.json`.

`ssh2.pcapng/JA4L-S` is the one entry that compares two measured counts: 11 against 10.

## The comparison against `ja4plus/fingerprinters/ja4l.py`

Seventeen named fields agree, and the page names every one. Four disagreements are reported.

- **D1** — two parts against the image's three. `https-connect.pcap` gives `45_64` against the Wireshark `45_64_66`.
- **D2** — `calculate_distance` cannot compute the VPN case the image works, because there is no part c to read.
- **D3** — no `q` part and no `_quic` part on a QUIC connection. This matches the Python reference.
- **D4** — `ja4l.py:416-425` returns a second server value for a retransmitted SYN-ACK. `ssh2.pcapng` gives 11 `JA4L-S` values against the reference 10; the extra is `JA4L-S=6252_58` on `tcp_172.16.225.48:57380_184.150.157.177:80`.

The worked example of the image reproduces exactly against this project: `calculate_distance(5191, ttl=42)` returns 415.28 miles, because TTL 42 gives 22 hops and the table row for 22 gives 1.6.

## Uncertain rules

Four of the eleven rules rest on the image alone, and none of the four builds a fingerprint: the initial TTL table of 64, 128 and 255, the hop count formula, the distance formula `D = jc/p`, and the propagation factor table. Measured at the pinned commit outside `zeek/`, `grep -rn "propagation"` and `grep -rn "Hop Count"` report no match. The one FoxIO function that subtracts a TTL, `python/ja4.py:140-147`, reads `initial_ttl = 54` for the low band, and nothing calls it.

## How this was tested

`ci: skip`, so the full local suite replaces provider CI.

```
pytest tests/ -m spec_validation -q      : 1375 passed, 146 skipped, 120 xfailed
pytest tests/ -m "not spec_validation" -q: 1208 passed, 8 xfailed, 93 subtests passed
ruff check ja4plus/ tests/               : All checks passed!
ruff format --check ja4plus/ tests/      : 115 files already formatted
mypy ja4plus/                            : Success: no issues found in 27 source files
git diff --name-only origin/dev...HEAD -- ja4plus/ : (empty)
```

**`tdd` does not apply to this issue.** The change is a documentation transcription and it adds no behaviour, so no unit test can fail before it and pass after it. Every claim of the page is instead a command the page reproduces. Coverage cannot fall, because no file under `ja4plus/` changes.

Verified against: https://github.com/FoxIO-LLC/ja4/blob/main/technical_details/JA4L.png (retrieved 2026-08-08, commit `27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8`, SHA-256 `04036284edfcf0d8e94f3ca6660b1ab688813df7bfb82b0f001b65d7daa07354`)


Crank-Git:

A Sonnet child verified about 60 `file:line` citations of `docs/specs/foxio/JA4L.md` against the FoxIO checkout at the pinned commit and against `ja4plus/fingerprinters/ja4l.py`. It found two errors, and `f8d4d63` fixes both.

1. **`python/common.py:181` does not hold the truncation.** The page cited it twice, in R2 and in the rounding row of the comparison table. Line 181 reads `dt2 = parse_timestamp(t2)`, and line 182 reads `return int((dt2-dt1).microseconds/2)`. Both citations now read 182.
2. **`tests/foxio_vectors/ssh2.pcapng.json` holds twelve method keys, not eleven.** The register table said eleven. The twelve are `JA4`, `JA4H`, `JA4H_ro`, `JA4L-C`, `JA4L-S`, `JA4S`, `JA4SSH`, `JA4S_r`, `JA4X`, `JA4_o`, `JA4_r` and `JA4_ro`.

Every other citation checked out: the 22 `ja4l.py` line numbers, the Wireshark `packet-ja4.c` citations, the Zeek `main.zeek` citations that #198 owns, the Rust citations, the `python/ja4.py` citations, the two README rows, the image SHA-256, the 114 local vector values across 24 files, the 88 Wireshark values across 15 files, the 27 Rust snapshot files, the four register-entry counts, and the worked example `5191 x 0.128 / 1.6 = 415.28`.

No finding was rejected.

## Scope lens

`git diff --name-only origin/dev...HEAD -- ja4plus/` returns nothing. Three files change and all three are documentation.

The page raises four disagreements and five register readings, and it changes no fingerprinter and no register entry. A decision issue belongs to the user, as #215, #214 and #219 do for #196, #198 and #203.

Crank-Git:
## The three coordinator requests are addressed in `a03e1f7`

**1. The decision issue is #225**, filed untriaged with no status label. It covers D1 (the part count), D3 (the protocol part) and D4 (the duplicate server value), and it carries D2 as the consequence of D1. It states the five-source table, states that no vector this project reads can measure a third part because `python/test/testdata/` publishes two, and lists the four things the project needs before any change: a Wireshark expected-output file for a capture this repository holds, a precedence rule that overrides `.claude/rules/external-apis.md` for this method, a new measurement point in `ja4l.py`, and **a choice between the two definitions of part c**, because `wireshark/source/packet-ja4.c:1371-1374` measures from the first client application packet to the first server application packet and `zeek/ja4l/main.zeek:191-192` measures from the ClientHello to the ServerHello. The two references that write three parts do not agree on what the third part holds. **The issue decides nothing.**

**2. `docs/specs/foxio/zeek.md` is corrected in three places.**

- The section heading `the Zeek script adds a third part that no other reference holds` is replaced. The section now opens by naming its own former error, carries the measured five-source table, and states that **Zeek is not the outlier on the part count**. It cites `wireshark/test/testdata/https-connect.pcap.json:58-59` holding `45_64_66`, the 88 Wireshark values across 15 files, and the image reading.
- Reading 2 of `The three readings of the audit` now records the correction and points at the section.
- **The rounding claim is scoped, not deleted.** `The FoxIO Python reference agrees with ja4plus, and the Zeek script is the outlier` now reads `On the rounding rule, ...`, with a line stating that it does not hold for the part count. That claim is correct in its own scope: on the half-to-even rounding, the Python reference does agree with this project.

The correction says plainly that **#198's measurement stands and that the conclusion it drew from two sources of five does not.** #198 compared the Python reference against the Zeek script and read neither the image nor the Wireshark expected output.

`docs/specs/foxio/JA4L.md` carries the same correction in R4, so a reader who opens either page finds it.

**3. The Changelog round is 56.** The row also records the correction to round 52 and names #225.

## Verification on the new head

```
pytest tests/ -m spec_validation -q      : 1375 passed, 146 skipped, 120 xfailed
pytest tests/ -m "not spec_validation" -q: 1208 passed, 8 xfailed, 93 subtests passed
ruff check ja4plus/ tests/               : All checks passed!
ruff format --check ja4plus/ tests/      : 115 files already formatted
git diff --name-only origin/dev...HEAD -- ja4plus/ : (empty)
gh pr checks 223                         : no checks reported
```

All three commits end with `[skip ci]`, verified byte-exactly, and the PR is still a draft into `epic/194-read-the-specification-images`.

**On the wrong premise:** the brief was right that #198 measured Python against Zeek correctly. It was the framing of Zeek as the outlier that the image disproved, and that only surfaced because reading the image was the one part of this issue nobody had done yet.

Crank-Git:
## The count is corrected in `3908cd5`, and the cause was not a line-based artifact

I re-measured before changing anything, with a walk that recorded the JSON path of every hit so a duplicate would be visible. **There is no duplication.** Every one of the 88 hits sits at a distinct path of the form `/<packet index>/_source/layers/ja4.ja4l` or `.../ja4.ja4ls`.

Counting each key separately explains the disagreement:

```
ja4.ja4l : 44 values in 15 files | part counts {3: 44} | quic third part 9
ja4.ja4ls: 44 values in 15 files | part counts {3: 44} | quic third part 9
both keys: 88 in 15 files
```

**Both numbers were right about different key sets.** 44 is `ja4.ja4l` alone, which is the key you named. 88 is `ja4.ja4l` and `ja4.ja4ls` together, which is what my line read. 88 being exactly twice 44 is not a doubling bug — it is that the dissector writes a client value and a server value for the same connection, so the two keys appear in equal number.

I did not silently accept the correction and I did not silently keep my number. The fix is to remove the ambiguity that produced the disagreement.

## What changed

**Every place that carried the bare number now names the key beside it.**

- `docs/specs/foxio/JA4L.md`, the reference-value table: `44 ja4.ja4l values and 44 ja4.ja4ls values, 88 in all, in 15 files. Every one of the 88 holds three parts.`
- `docs/specs/foxio/JA4L.md`, the R4 part-count table: the Wireshark row now reads `3 on all 44 ja4.ja4l values and all 44 ja4.ja4ls values`.
- `docs/specs/foxio/zeek.md`: `Every ja4.ja4l value and every ja4.ja4ls value under wireshark/test/testdata/ holds three parts: 44 of each key, 88 in all, in 15 files.`
- The round 56 Changelog row, in both places it appeared.
- Issue #225, prerequisite 1.

**The reproduce command is now the JSON walk, not a grep.** The old `grep -rhA1` line is replaced, and the page states why the method matters: the two keys sit side by side, so a line pattern reports a number whose key set the reader cannot see. The command is verified to print exactly `ja4.ja4l 44 values in 15 files {3: 44}` and `ja4.ja4ls 44 values in 15 files {3: 44}` against a checkout at the pinned commit.

**The stronger claim is stated, as you asked.** Not most, not a majority: every Wireshark JA4L value carries three parts.

## Verification on the new head

```
pytest tests/ -m spec_validation -q      : 1375 passed, 146 skipped, 120 xfailed
pytest tests/ -m "not spec_validation" -q: 1208 passed, 8 xfailed, 93 subtests passed
ruff check ja4plus/ tests/               : All checks passed!
ruff format --check ja4plus/ tests/      : 115 files already formatted
git diff --name-only origin/dev...HEAD -- ja4plus/ : (empty)
```

Nothing else changed. Four commits, each ending `[skip ci]`, still a draft into `epic/194-read-the-specification-images`.

Issue #227, opened 2026-08-08
Write s2 for the TLS version value 0x0002


Crank-Git:
Found while #221 read the deleted FoxIO text specifications. This issue is untriaged.

## The defect

**`ja4plus` writes the SSL 2.0 version value that FoxIO retracted.**

`technical_details/JA4.md` at the pinned commit `27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8`
states `0x0002 = SSL 2.0 = “s2”`. It states no SSL 1.0 row.

The form FoxIO deleted on 2024-02-22 states `0x0200 = SSL 2.0 = “s2”` and
`0x0100 = SSL 1.0 = “s1”`. **FoxIO corrected both statements** in commit `3e02a27`, dated
2024-08-23, titled `Fix SSL version fields: SSL 2.0 is 0x0002, SSL 1.0 never existed`.

**Three lines of this project follow the retracted value.**

```
ja4plus/fingerprinters/ja4.py:127   elif version == 0x0200:  # SSL 2.0
ja4plus/fingerprinters/ja4.py:250   elif version == 0x0200:  # SSL 2.0
ja4plus/fingerprinters/ja4s.py:393  0x0200: "s2",  # SSL 2.0
```

A hello whose protocol version is `0x0002` therefore writes `00`, and the pinned
specification requires `s2`. A hello whose protocol version is `0x0200` writes `s2`, and
the pinned specification states no meaning for that value.

## What is already correct

**This is one row of one table, and it is the only stale row.** `ja4plus` writes `d1`, `d2`
and `d3` for the three DTLS versions, which the deleted form does not carry, and it writes
no `s1`, which the deleted form does carry. The rest of the version table follows the
pinned form.

## Why no test fails today

**No capture in `tests/foxio_vectors/` carries an SSL 2.0 hello.** No conformance case
fails, and `tests/foxio_deviations.json` holds no entry for this. The reading is a code
reading against the pinned specification.

An SSL 2.0 ClientHello also uses a record format that this project's TLS parser does not
read, so a repair may reach no packet. **Measure that before you change a line.**

## What to build

1. **Measure whether any parser path can present a version of `0x0002`.** State the result,
   including a negative result.
2. **Write `s2` for `0x0002` in the three places named above**, and state what happens to
   `0x0200`.
3. **Add one unit test that names the behaviour**, in the form
   `writes s2 for the SSL 2.0 version value 0x0002`.
[...]

Issue #221, opened 2026-08-08
Read the seven deleted FoxIO text specifications


Crank-Git:
Found while #199 transcribed `technical_details/JA4SSH.png`. This issue is untriaged.

## What the transcription found

**FoxIO published a text specification for seven methods, and one commit deleted all
seven.** Commit `b6f3ff4`, titled `Update README.md`, removed these files.

- `technical_details/JA4.md`
- `technical_details/JA4H.md`
- `technical_details/JA4L.md`
- `technical_details/JA4S.md`
- `technical_details/JA4SSH.md`
- `technical_details/JA4T.md`
- `technical_details/JA4X.md`

`JA4.md` and `JA4H.md` exist again at the pinned commit
`27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8`. **The other five do not, and no transcription
has read them.**

Reproduce the list from a checkout at the pinned commit. The command reaches no network,
because the checkout holds the whole history.

```bash
git log --oneline --diff-filter=D --name-only -- 'technical_details/*.md'
```

## Why this material counts

**The pinned commit points at one of the deleted files.** `rust/ja4/src/ssh.rs:283`, at
the pinned commit, cites
`https://github.com/FoxIO-LLC/ja4/blob/16850cc2c8bcb8328c1a43a851a3a9a6eaa56103/technical_details/JA4SSH.md#how-to-measure-the-mode-for-tcp-payload-lengths-across-200-packets-in-the-session`
for its mode rule. FoxIO's own implementation therefore treats the deleted file as the
statement of intent for that rule.

**The deleted `JA4SSH.md` states rules the image only draws.** It states that the mode
reads the TCP payload length and not the packet length, that the count excludes a
retransmission, and that the window is configurable. `docs/specs/foxio/JA4SSH.md` records
each one with its corroboration.

## What to build

1. **Read the deleted file for each of the five methods** at the commit that holds it:
   `JA4L.md`, `JA4S.md`, `JA4T.md`, `JA4X.md`, and the earlier form of `JA4.md`.
2. **Record the provenance of each one**: the path, the commit that holds it, the blob
   SHA-1, the byte count, the commit that created it, and the commit that deleted it.
   `docs/specs/foxio/JA4SSH.md` holds the shape.
3. **Fold each statement into the transcription of that method** as a corroboration, and
   state where it agrees with the image and where it does not.
4. **State the standing of a deleted file in `docs/specs/foxio/README.md`.** A deleted
[...]

Issue #199, opened 2026-08-08
Transcribe JA4SSH and test the three declined reference behaviours against it


Crank-Git:
Part of #194. **Depends on #195**, which sets the reading method.

## What to build

1. **Transcribe `technical_details/JA4SSH.png`** by the method #195 states: read the image,
   write this project's own prose into `docs/specs/foxio/JA4SSH.md`, and record the
   source address, the pinned commit, the retrieval date and the **SHA-256 of the image**.
   **No image file is copied into this repository.**
2. **Corroborate every transcribed rule against two FoxIO-authored sources that are not the
   image.** The candidates are the `ja4` README, the Wireshark dissector, the doc comments
   under `rust/ja4/src/`, **the Zeek implementation under `zeek/`** and a FoxIO blog post.
   A rule with fewer than two corroborations is marked uncertain and keeps the vector
   fallback.
3. **Compare every transcribed rule against `ja4plus/fingerprinters/ja4ssh.py`**, field by field, and report each
   disagreement with a `file:line`.
4. **Compare the transcribed rules against the register.** An entry of
   `tests/foxio_deviations.json` that names this method is either explained by the
   specification or it is not, and the report states which.

## Read this method first among the six

The register holds **ten JA4SSH entries** that a person decided without a specification to
read: #96, #97 and #105 each decline a FoxIO reference behaviour that the audit calls a
defect. **Those three declines rest on a reading of the reference implementation, not on a
statement of intent.** The transcription either confirms them or it reopens them, and both
outcomes are worth having.

`zeek/tests/Traces/Scripts.ja4-ssh2/ja4ssh.log` is an independent baseline for this
method, and #198 owns reading it.

## What not to build

**No fingerprint moves in this issue.** A disagreement produces a decision issue, and a
person decides it. `CLAUDE.md` rule 1 binds every fingerprinter change.

## Acceptance criteria

- [ ] `docs/specs/foxio/JA4SSH.md` holds the transcription, with the image SHA-256, the
      pinned commit, the retrieval date, and two cited corroborations per rule.
- [ ] Every rule that could not reach two corroborations is marked uncertain and listed.
- [ ] Every disagreement with `ja4plus/fingerprinters/ja4ssh.py` is reported with a `file:line` and both readings.
- [ ] Every register entry that names this method is stated as explained or unexplained.
- [ ] No file under `ja4plus/` changes, and no fingerprint moves.

Crank-Git:
[...]

PR #230, opened 2026-08-08
Transcribe JA4S and compare it against the fingerprinter

Crank-Git:
Closes nothing on its own. Part of #194, and #201 owns it.

## What this changes

`docs/specs/foxio/JA4S.md` holds the fourth transcription written under the #195 method.
`docs/specs/foxio/README.md` gains the row for it. `docs/specs/spec.md` gains a Changelog
row, and the round stays `TBD` because the project manager assigns it.

`git diff --name-only origin/dev...HEAD -- ja4plus/` returns nothing.

## Why

JA4S reaches more published reference values than any other method of this epic, so this
page measures where #196, #199 and #200 reasoned from source code.

## How this is tested

The page is prose, so no new unit test applies. Every quantity in it is a measurement, and
each one names the command that reproduces it. The suites run unchanged.

```
pytest tests/ -m spec_validation -q      : 1375 passed, 146 skipped, 1216 deselected, 120 xfailed
pytest tests/ -m "not spec_validation" -q: 1208 passed, 1641 deselected, 8 xfailed, 93 subtests passed
ruff check ja4plus/ tests/               : All checks passed!
ruff format --check ja4plus/ tests/      : unchanged
```

## The findings

**The image's own worked example reproduces byte for byte.** `JA4S.png` draws
`JA4S=t120400_c030_4e8089b08790`. That value belongs to `tls-handshake.pcapng`, which this
repository holds, and `ja4plus` produces it with the raw form
`t120400_c030_0005,0017,ff01,0000` that the deleted `technical_details/JA4S.md` states.

**Fourteen rules carry two FoxIO-authored corroborations each. Three are uncertain and keep
the vector fallback.** The DTLS protocol character, an ALPN byte that is not ASCII
alphanumeric, and GREASE in the server extension list.

**R13 is two against two.** `python/ja4.py:182` includes a GREASE extension and
`rust/ja4/src/tls.rs:604-613` applies no GREASE test, where
`wireshark/source/packet-ja4.c:757` and `zeek/ja4s/main.zeek:91` drop it from the list and
from the count. No vector measures it: 0 ServerHello messages across the 38 captures of
`tests/foxio_vectors/` carry a GREASE extension type.

**Five disagreements are reported, and no vector measures any of them.** `ja4s.py:276`
writes a `d` protocol character that the image does not name, and `ja4plus` produces 126
JA4S values over the 38 captures of which 97 open with `t`, 29 open with `q` and 0 open
with `d`. `ja4s.py:281-283` reads the first supported version where the references read the
largest, and it falls back to the legacy version where the Rust reference emits nothing.
`ja4s.py:300-302` writes no value for a ServerHello that names no cipher.
`ja4s.py:393` maps the retracted `0x0200`, which **#227 owns and this page only cites**.

**The plan's premise about the register is false, and the page says so.** The plan stated
that 16 entries under #162 concern JA4S and ALPN. Those 16 name `JA4.1`, `JA4_o.1`,
`JA4_r.1` and `JA4_ro.1` on four streams of `alpn-condition.pcap`, and none names JA4S. The
capture holds 7 packets and 0 ServerHello messages, its expected-output file holds 0 JA4S
keys, and `ja4plus` produces 0 JA4S values for it. **Fourteen entries do name JA4S**, nine
under #138 and five under #151, and the specification explains none of them and contradicts
none of them.

**#221 gains one correction.** It read the deleted text's `(q or t)` as a file that predates
the DTLS addition. The image at the pinned commit states the same two values, so that
reading now rests on `technical_details/JA4.md` and the Wireshark dissector rather than on
the age of the deleted file.

Verified against: https://github.com/FoxIO-LLC/ja4/blob/main/technical_details/JA4S.png (retrieved 2026-08-08, commit `27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8`, SHA-256 `a4d303c3c51c2862d86abd69d6dfe6d28a43e86556a91d2c1c8261fa4de15458`)


Crank-Git:
## Self-review

I reviewed the diff myself. A Sonnet child was spawned for a second pass and it had not
reported when I wrote this, so the findings below are mine and every one is reproduced by a
command.

**Citation accuracy.** I checked all 85 `file:line` citations into the FoxIO checkout at
the pinned commit, by reading each cited line. **Five were off by one to three lines and
each is now corrected**: the Rust ALPN read (`tls.rs:423-424`, not `:436-437`), the Rust
`is_quic` read (`tls.rs:434`, not `:441`), the Wireshark empty-hash ternary
(`packet-ja4.c:573`, not `:572`), the Wireshark GREASE guard (`packet-ja4.c:757`, not
`:756`), and the Zeek cipher format (`zeek/ja4s/main.zeek:165`, not `:162`). The 15
citations into `ja4plus/` were correct, except the end of `compute_alpn_value`, which is
`ja4.py:37-84`.

**Arithmetic.** The first draft read "329 published values across four FoxIO
implementations". 84 + 120 + 125 = 329 covers three, and the Zeek baseline adds 20 more.
Corrected in the page and in the Changelog row.

**An inference replaced by a measurement.** D4 read "every ServerHello in
`tests/foxio_vectors/` names a cipher", which the produced values cannot prove, because a
ServerHello that names no cipher produces nothing. The page now reports the parse: the
reader parses 97 TCP ServerHello messages across the 38 captures and 0 of them name no
cipher.

**One sentence stated the register pairing backwards.** Twelve entries form six pairs of a
`JA4S` entry and a `JA4S_r` entry. `https-connect.pcap` and `socks4-https.pcap` hold a
`JA4S` entry and no `JA4S_r` entry.

**The writing standard.** Four sentences ran past the 25-word limit of
`.claude/rules/ste.md`. Each now splits into two facts.

**Internal consistency.** R1 to R14 are all defined and no rule number is referenced that
does not exist. D1 to D5 likewise. The register table holds 14 rows, and they equal the 14
entries of `tests/foxio_deviations.json` whose method is `JA4S` or `JA4S_r`, owner for
owner.

**No finding was rejected.** Every one above is fixed in the branch.

Crank-Git:
## The second review reported, and it found three more defects

My earlier comment said the Sonnet review child had not reported. **It reported after
that comment, and it found three defects my own pass missed.** All three are real, all
three are now fixed in `6935129`, and I reject none of them.

1. **`docs/specs/foxio/JA4S.md:120` cited `python/ja4.py:195` for the TLS version map.**
   Line 195 reads `x['version'] = get_supported_version(x['supported_versions'])`. The
   `TLS_MAPPER[x['version']]` expression the page quotes sits at line 196. Corrected.
2. **`docs/specs/foxio/JA4S.md:116` cited `packet-ja4.c:550` for the `val_to_str_const`
   call.** Line 550 opens the format string. The call sits at line 551. Corrected. The
   parallel citation `packet-ja4.c:567` was already right.
3. **`docs/specs/foxio/JA4S.md:413` claimed the 110 Rust values sit in 8 files, and they
   sit in 7.** `grep -lE 'ja4s: ' tests/foxio_vectors/rust_expected/*.snap` reports 7, and
   the table directly below the sentence already listed those 7 captures. The sentence
   contradicted its own table. Corrected in the page and in the Changelog row.

The child re-measured every quantity in the page independently and reproduced all of them
exactly: 84, 120 in 15 files with 29 distinct, 125, 194, 126 values with 97 `t` / 29 `q` /
0 `d`, 12 with an ALPN, 0 ServerHello messages with a GREASE extension, and the four
`alpn-condition.pcap` zeros. It found no STE violation, and it confirmed that `reading` is
a defined term in the `## Terms` table of `docs/specs/spec.md` rather than an `-ing` noun.
It confirmed R1 to R14 and D1 to D5 are all defined and referenced, and that the 14
register rows match `tests/foxio_deviations.json` key for key and owner for owner.

Suites re-run after the fix.

```
pytest tests/ -m spec_validation -q      : 1375 passed, 146 skipped, 1216 deselected, 120 xfailed
pytest tests/ -m "not spec_validation" -q: 1208 passed, 1641 deselected, 8 xfailed, 93 subtests passed
ruff check ja4plus/ tests/               : All checks passed!
git diff --name-only origin/dev...HEAD -- ja4plus/ : empty
```

Issue #216, opened 2026-08-08
Compare JA4T against the Rust snapshot values the harness already parses


Crank-Git:
Raised by #196. See `docs/specs/foxio/JA4T.md`.

## The gap

`tests/test_foxio_rust_parity.py:72` reads two methods from a snapshot.

```python
SNAPSHOT_METHODS = (("JA4", "ja4"), ("JA4S", "ja4s"))
```

**The six snapshots in `tests/foxio_vectors/rust_expected/` carry 38 `ja4t` values, and the
harness never reads the field.** No other test compares a JA4T value or a JA4TS value
against a FoxIO value, so `tests/foxio_deviations.json` holds no entry for either method.

The JA4T cases in `tests/test_comprehensive.py` build a packet with scapy and assert this
project's own format, so no reference decides them. **The suite has been green on both
methods for the life of the project, and that greenness measures nothing.** This is the
thirteenth instance of the shape #172 sweeps for, and the first at the scale of a whole
method.

## What exists already

| Capture in `tests/foxio_vectors/` | `ja4t` values in the local snapshot |
|---|---|
| `browsers-x509.pcapng` | 3 |
| `chrome-cloudflare-quic-with-secrets.pcapng` | 1 |
| `https-connect.pcap` | 1 |
| `latest.pcapng` | 6 |
| `ssh2.pcapng` | 19 |
| `tls3.pcapng` | 8 |

`rust/ja4/src/snapshots/` holds 64 `ja4t` values in 26 files, 19 of them distinct. No
snapshot holds a `ja4ts` value.

## What to build

1. **Add `("JA4T", "ja4t")` to `SNAPSHOT_METHODS`.** This needs no new file. Two cases fail
   at once, and #215 owns the decision they raise.
2. **Copy the snapshot for `gre-erspan-vxlan.pcap`.** The capture is already local, and it
   is the one case that measures the empty option list.
3. **Record every remaining mismatch in `tests/foxio_deviations.json`** rather than change a
   fingerprinter. #215 decides the repairs.
4. **State that JA4TS reaches no reference value except the Zeek baseline.** #198 owns that
   reading.

**Warning: do not change a fingerprinter in this issue.** #215 holds the decisions.

## Acceptance criteria

- [ ] The parity harness compares JA4T against every local snapshot value.
- [ ] Reverting the harness change to `SNAPSHOT_METHODS` makes the new cases fail.
[...]

Issue #238, opened 2026-08-08
Batch 17: act on what Epic 10 measured


Crank-Git:
Integration branch `batch/238-act-on-epic-10`, off `dev` at `dea4e32`.

Epic 10 read the specification and moved no fingerprinter. **This batch acts on what it measured.**

## Members

- [ ] #216 — Compare JA4T against the 38 Rust snapshot values the harness already parses.
- [ ] #229 — Extend the harness to compare JA4X, and correct the register cause that cites a test which does not test it. **Sequenced after #216**, because both edit `tests/test_foxio_rust_parity.py:72`.
- [ ] #219 — The four JA4H readings the user decided on 2026-08-08.
- [ ] #227 — Write `s2` for the version value `0x0002`, which FoxIO corrected in `3e02a27`.

#211 waits for the next batch.

## The invariant

`dev` holds **114 register keys against 114 `xfailed` cases**, and 6 undecided entries. #216 and #229 add comparisons, so the count may rise; the two numbers must stay equal. #219 and #227 must move **no** vector value.

## The shape this batch must not repeat

Three members touch a comparison that has never run. **A comparison that is never made reads as a comparison that passes** — #196 found it for `ja4t`, #199 for `ja4ssh`, #202 for `ja4x`. Every member proves its work by reverting it and states the count of cases that then fail.

Opened 2026-08-08. Nothing merged yet.

Crank-Git:
**#227 and #216 sub-merged.** Rounds 62 and 63.

| Member | Round | State |
|---|---|---|
| #227 | 62 | Sub-merged `784fcb5` |
| #216 | 63 | Sub-merged `1170a4d` |
| #229 | — | **Unblocked**, starting now |
| #219 | — | In flight |

Register on the branch: **116 keys against 116 `xfailed`**, up from 114 because #216 made the JA4T comparison run for the first time.

**A concurrency hazard was found and it affected no repository content.** The session scratchpad is shared between workers, and two of them wrote unprefixed filenames. Pull request #239 briefly carried another issue's body, and pull request #240 carried #219's body until the project manager read it at the gate. **The title stayed correct while the body was wrong**, so nothing automatic would have caught it. Both are corrected. Every scratchpad filename is now namespaced by issue number.

Crank-Git:
**Batch 17 merged into `dev` as `fe50713`.** All nine continuous integration jobs pass. Four members, Changelog rounds 62 to 65.

| Round | Issue | Result |
[...]

Issue #215, opened 2026-08-08
Decide the JA4T and JA4TS form where this project disagrees with FoxIO


Crank-Git:
Raised by #196. **This issue needs a decision from the user. It changes fingerprints that
`ja4plus` 0.6.0 already publishes.**

`docs/specs/foxio/JA4T.md` holds the transcription and the full comparison. #196 changed no
fingerprinter.

### D1 — the empty option list, and a zero part

Three FoxIO-authored forms exist and `ja4plus` matches none of them.

| Producer | Value for a SYN that carries no option |
|---|---|
| FoxIO Rust, `ja4__insta@gre-erspan-vxlan.pcap.snap` | `8192__0_0` |
| Wireshark dissector, `packet-ja4.c:664` | `8192_00_00_00` |
| Zeek baseline, `conn.log` (see #198) | `65535_00_00_00` |
| `ja4plus`, `ja4t.py:85` | `8192_0_0_0` |

`tests/foxio_vectors/gre-erspan-vxlan.pcap` reproduces it.

**Question: which form does `ja4plus` write?**

### D2 — the End of Option List pad bytes

`ja4t.py:67` reads scapy's parsed `tcp.options`, which reports one `EOL` entry for several
`0x00` pad bytes. Both FoxIO implementations count one entry per byte.

The SYN of `tests/foxio_vectors/chrome-cloudflare-quic-with-secrets.pcapng` carries the
option bytes `020405a0010303060101080a88d73b2c0000000004020000`.

```
FoxIO rust : 65535_2-1-3-1-1-8-4-0-0_1440_6
ja4plus    : 65535_2-1-3-1-1-8-4-0_1440_6
```

A repair reads the raw option bytes rather than the parsed list. **That changes a published
fingerprint.**

### D4 — one value per connection, or one per SYN

`rust/ja4/src/tcp.rs` reads only the first SYN of a connection. `ja4t.py` holds no
connection state, so `tests/foxio_vectors/ssh2.pcapng` emits 44 values over the 19
connections the reference reads once each. The distinct values agree.

`.claude/rules/conformance.md` states that a method that emits more fingerprints than the
reference is a defect.

### D6 and D7 — does JA4TS carry the retransmission part?

**The image and the Wireshark dissector disagree.** `JA4T.png` states that the
retransmission part appears only on JA4TScan. `wireshark/source/packet-ja4.c:1595` writes
the `JA4TS` field through `ja4t()` with a connection, which appends that part. Line 1608
[...]

PR #240, opened 2026-08-08
Compare JA4T against the Rust snapshot values the harness already parses

Crank-Git:
Builds #216. #238 tracks batch 17. Do not close #216 from here; the batch pull request closes it.

## What this changes

`tests/test_foxio_rust_parity.py` read `ja4` and `ja4s` from a FoxIO Rust snapshot and
never read `ja4t`. The six local snapshots carry 38 JA4T values, so **no case compared a
JA4T value against any FoxIO value, and the suite reported green on a comparison it never
made.** #196 measured the gap and this pull request closes it.

- `SNAPSHOT_METHODS` holds `("JA4T", "ja4t")`.
- `tests/conformance_index.py` runs the JA4T fingerprinter, so the produced index holds
  the method.
- `TestTheJa4tValuesTheRustSnapshotHolds` collects 38 value cases and 6 occurrence-key
  cases. Each case carries a register key, so a disagreement another issue owns reports
  as `xfailed` and a disagreement the register does not name fails the suite.
- `TestTheLocalSnapshotsHoldNoJa4tsValue` states the JA4TS negative result as a check.
- `tests/test_spec_validation.py` reads the parity keys too, because it rejects a register
  entry that matches no collected case.

**No file under `ja4plus/` changes, and no fingerprint moves.**
`git diff --name-only origin/dev...HEAD -- ja4plus/` returns nothing. #215 owns every
repair, and it is blocked on this measurement.

## The 38 values, and the 37 that reproduce

| Capture | Values | Result |
|---|---|---|
| `browsers-x509.pcapng` | 3 | Each value matches. |
| `chrome-cloudflare-quic-with-secrets.pcapng` | 1 | D2. The value differs by one `0`. |
| `https-connect.pcap` | 1 | The value matches. |
| `latest.pcapng` | 6 | Each value matches. |
| `ssh2.pcapng` | 19 | Each value matches, and D4 makes 10 streams carry more than one. |
| `tls3.pcapng` | 8 | Each value matches. |

**37 of the 38 values reproduce exactly, and one differs.** No value the register does not
name disagrees.

## The two register entries

| Key | Owner | The reading it records |
|---|---|---|
| `chrome-cloudflare-quic-with-secrets.pcapng/0:57098/JA4T.1` | #215 | D2. The SYN carries two End of Option List bytes, and scapy reports one `EOL` entry for the pair, so part b holds one `0` where the reference holds two. `rust=65535_2-1-3-1-1-8-4-0-0_1440_6 ja4plus=65535_2-1-3-1-1-8-4-0_1440_6` |
| `ssh2.pcapng/JA4T` | #215 | D4. `ja4plus` holds no connection state and fingerprints every SYN, so 10 of the 19 streams carry more than one value against the reference count of one. The distinct values agree, so the count differs and the value does not. |

**Neither entry is `decided`**, because #215 decides the repair and a fix removes the
entry. `tests/foxio_deviation_owners.json` gains #215 as an open, non-epic owner.

## The revert proof

A comparison that passes with the new method entry removed measures nothing.

Remove `("JA4T", "ja4t")` from `SNAPSHOT_METHODS` and **44 cases stop running**: 38 value
cases and 6 occurrence-key cases. The module collects 83 cases with the entry and 41
without it. The full run then reports `3 failed, 1383 passed, 148 skipped, 114 xfailed`,
and three checks name the loss.

| Check that fails on the revert | What it reports |
|---|---|
| `test_the_local_snapshots_hold_the_thirty_eight_values_the_reading_counts` | The snapshot reader finds no JA4T value. |
| `test_the_suite_collects_one_case_for_every_value_the_snapshots_hold` | The parameter list is empty. |
| `test_every_register_entry_matches_a_collected_case` | The two JA4T register entries match no case. |

## JA4TS reaches no reference value here

**No local Rust snapshot writes a `ja4ts` field.** The check reads all ten files under
`tests/foxio_vectors/rust_expected/` and finds none, which confirms the count #196
reported. No FoxIO Python expected-output file holds a `JA4T` key or a `JA4TS` key either.

**JA4TS therefore stays uncovered.** Its one reference value is the Zeek baseline
`ja4ts 65535_00_00_00`, and #198 owns that reading. `docs/specs/foxio/JA4T.md` says so.

## The one snapshot this pull request declines to add

Item 2 of the issue body asks for `ja4__insta@gre-erspan-vxlan.pcap.snap`, which holds
`ja4t: 8192__0_0` and is the one local case that measures D1. **It is not added, and the
reason is a second disagreement.**

| Reference | Addresses it names |
|---|---|
| `python/test/testdata/gre-erspan-vxlan.pcap.json` | `100.20.9.2` and `100.20.9.1` |
| `rust/ja4/src/snapshots/ja4__insta@gre-erspan-vxlan.pcap.snap` | `10.16.27.12` and `10.16.27.131` |

The capture carries ERSPAN and VXLAN, so one packet holds two address pairs. `ja4plus`
reads the outer pair, which is the pair the FoxIO Python file names. The harness matches a
stream by its address pair, so a case built from that snapshot would compare against a
stream identity `ja4plus` never produces, and **the failure would report the address layer
rather than D1.** #242 holds that question, untriaged.

**The D1 reading stands on its own** and needs no case: `ja4plus` produces `8192_0_0_0`
for the SYN of `tests/foxio_vectors/gre-erspan-vxlan.pcap`, and the snapshot holds
`8192__0_0`. #215 item 1 owns the decision.

## One footgun recorded, not repaired

`tests/generate_foxio_baseline.py` runs `tests/test_spec_validation.py` alone, so it
collects no case of the parity module and a regeneration drops the two JA4T entries. Its
docstring now carries the warning. The loss is loud rather than silent: each dropped entry
makes its case fail the conformance suite by name.

## How tested

The full local suite ran in the worktree. `ci: skip`, so the pull request is a draft into
`batch/238-act-on-epic-10` and every head commit ends with `[skip ci]`.

| Command | Result |
|---|---|
| `pytest tests/ -m spec_validation -q` | `1428 passed, 146 skipped, 116 xfailed` |
| `python -c "len(json.load(open('tests/foxio_deviations.json')))"` | `116` |
| `pytest tests/ -m "not spec_validation" -q` | `1232 passed, 8 xfailed, 93 subtests passed` |
| `ruff check ja4plus/ tests/` | `All checks passed!` |
| `ruff format --check ja4plus/ tests/` | `118 files already formatted` |
| `mypy ja4plus/` | `Success: no issues found in 27 source files` |
| `pytest tests/ --cov=ja4plus` | `82%`, 639 missed of 3495 statements, the same as the base |

**The register invariant holds.** The base carries 114 keys against 114 `xfailed` cases.
This branch carries 116 against 116.

**The test came first.** `112423c` added the comparison and left two cases failing with no
register entry. `eee223e` then registered them.

`docs/specs/spec.md` records the work as Changelog round 63.

Verified against: https://github.com/FoxIO-LLC/ja4/blob/main/rust/ja4/src/snapshots (retrieved 2026-08-08, commit `27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8`)


Crank-Git:
## Self-review

A correctness lens ran on the diff, and I read the diff myself for scope and for the
writing standard. Both returned.

### Findings, and what each one produced

**F1 — The value count was wrong. Fixed.** `docs/specs/foxio/JA4T.md` and the Changelog
row read `35 of the 38 values`. The measurement is 37 of 38 reproduced exactly and one
differed. D4 changes the count of values on 10 streams of `ssh2.pcapng`, and it changes no
value. Corrected in `fe37889`.

**F2 — `tests/generate_foxio_baseline.py` drops the two JA4T entries. Documented, not
fixed.** The generator runs `tests/test_spec_validation.py` alone, so it collects no case
of this module and a run removes `chrome-cloudflare-quic-with-secrets.pcapng/0:57098/JA4T.1`
and `ssh2.pcapng/JA4T`. The loss is loud rather than silent: each dropped entry makes its
case fail the conformance suite by name. The docstring of the generator now carries the
warning. A repair changes the generator, which #216 does not own.

**F3 — The address layer of a tunneled capture. Filed as #242.** The Rust snapshot of
`gre-erspan-vxlan.pcap` names the stream by the inner addresses and `ja4plus` names it by
the outer pair, so the one snapshot that measures D1 reaches no case. The issue is
untriaged.

### What the lens confirmed

1. **No vacuous comparison.** `test_the_produced_ja4t_equals_the_rust_snapshot_value`
   calls `pytest.fail` on both branches, on an absent value and on a differing value. Run
   alone it reports 37 passed and 1 xfailed, so the 38 comparisons are made.
2. **The `read_rust_snapshot` return type changed and every caller followed.**
   `omitted_cases` is the one caller. Two blocks that share one stream identity still
   merge their values, and the index and the source port of the first block win.
3. **`register_keys()` returns 44 unique keys**, and the register holds an entry for two
   of them.
4. **The import in `tests/test_spec_validation.py:539` is no cycle.**
   `tests/test_foxio_rust_parity.py` imports nothing from it.
5. **`handshake_cases()` weakens no #138 check.** It moves the JA4T cases out of a path
   that lets an absent value pass, into cases that fail on an absent value.

### One nit I left

`tcp_cases` sorts by the stream index as a string, so `10` sorts before `2`. That orders
the test identifiers and it reaches no comparison.

Issue #219, opened 2026-08-08
Decide six JA4H readings the transcription found


Crank-Git:
#203 transcribed `technical_details/JA4H.png` into `docs/specs/foxio/JA4H.md`. The
comparison against `ja4plus/fingerprinters/ja4h.py` and `ja4plus/utils/http_utils.py`
found six disagreements. **Each one needs a person, because each changes a fingerprint
this project publishes, or it changes a recorded cause.** #203 changed no fingerprinter.

`CLAUDE.md` rule 1 binds every fingerprinter change. `.claude/rules/conformance.md` states
the two shapes that decline a reference defect.

Read `docs/specs/foxio/JA4H.md` first. It holds every rule, its two corroborations, and
the measurement behind each item below.

## D2 — Does field a3 read the Cookie header, or the parsed cookie list?

`ja4h.py:288` writes `has_cookie = "c" if http_info.get("cookie_fields", []) else "n"`.

R5 of the transcription states that all three FoxIO implementations read the header alone.
`rust/ja4/src/http.rs:83` sets the flag from the header name. `wireshark/source/packet-ja4.c:1165`
sets it from the `http.cookie` field. `python/ja4h.py:18` sets it from the presence of the
cookies key.

A Cookie header whose value carries no `=` therefore writes `n` here and `c` in the
reference.

```
request : GET / HTTP/1.1\r\nHost: a\r\nCookie: sessionid\r\n\r\n
ja4plus : ge11nn010000_4a823118b9ba_000000000000_000000000000
```

**Decision needed.** Does field a3 read the header?

## D3 — Does field a4 read the Referer header, or its value?

`ja4h.py:289` writes `has_referer = "r" if http_info.get("referer", "") else "n"`.

R6 states that all three references read the header name. A Referer header with an empty
value writes `n` here and `r` in the reference.

```
request : GET / HTTP/1.1\r\nHost: a\r\nReferer: \r\n\r\n
ja4plus : ge11nn010000_4a823118b9ba_000000000000_000000000000
```

**Decision needed.** Does field a4 read the header?

## D4 — Which list does field a5 count?

`ja4h.py:291` to `ja4h.py:294` count one list, and `ja4h.py:321` to `ja4h.py:325` hash
another. A header whose name is empty reaches the count and not the hash.

R9 states that all three references count the list they hash.

```
request        : GET / HTTP/1.1\r\nHost: a\r\n : b\r\n\r\n
[...]

PR #241, opened 2026-08-08
Read the header for JA4H fields a3 and a4, and open the method list

Crank-Git:
Closes nothing on its own. #219 holds the four decisions, and #238 tracks batch 17.

## What this changes

The user settled D1, D2, D3 and D4 on 2026-08-08. `docs/specs/foxio/JA4H.md` measured all
four, and #193 had already repaired D5 and D6.

| Ruling | Repair | Register row |
|---|---|---|
| **D2.** Field a3 reads the Cookie header, not the parsed cookie list. R5. | `ja4h.py:_ja4h_part_a` reads the header name list. | None. All three references read the header, so this is a defect. |
| **D3.** Field a4 reads the Referer header, not the header value. R6. | `ja4h.py:_ja4h_part_a` reads the header name list. | None. Same reason. |
| **D4.** Field a5 counts the list part b hashes. R9. | `ja4h.py:_ja4h_part_a` counts `_ja4h_header_names`. | None. |
| **D1.** Field a1 reads the first two characters of any method. | `http_utils.py:REQUEST_LINE_PATTERN` reads a method token and names no method. | **One row in `docs/specs/spec.md`.** `python/ja4h.py:9` reads any method, `rust/ja4/src/http.rs:342` rejects a method outside the nine, and `wireshark/source/packet-ja4.c:1087` holds a 40-entry map. |

The three parse paths now read one request-line pattern, so `parse_http_request` reads the
same method list and the same version list as the other two.

**The reassembly gate keeps a narrow test.** `is_http_request` admits a method the nine
tokens omit only when the buffer holds a whole request line. `ja4l.py:365` reads the same
gate, and a wider test of the same shape would admit an SSH banner, which starts with
method characters and one space. A JA4L measurement point would move with it.

## Why the vectors do not move

**No FoxIO vector reaches any of the four readings.** The largest JA4H vector carries no
valueless Cookie header, no empty Referer value, no header named with a space and no method
outside the nine.

The 38 captures under `tests/foxio_vectors/` replay on `origin/dev` and on this branch.

```
origin/dev  : 38 captures, 791 fingerprint values, 73 of them JA4H
this branch : 38 captures, 791 fingerprint values, 73 of them JA4H
diff        : 0 values differ
```

`origin/dev` and the base of this branch hold identical `ja4plus/` and identical
`tests/foxio_vectors/`, so the base replay is the `origin/dev` replay.

```
$ git diff --stat origin/dev origin/batch/238-act-on-epic-10 -- ja4plus/ tests/foxio_vectors/
(no output)
```

The register and the conformance suite hold too.

```
before: 114 xfailed, tests/foxio_deviations.json holds 114 keys
after : 114 xfailed, tests/foxio_deviations.json holds 114 keys
```

## Why the cases can fail

A test built only from constructed cases proves the new behaviour and nothing about the
old. `tests/test_ja4h_part_a_readings.py` holds 42 cases, and 28 of them fail on the base
commit. Each repair is also reverted alone and measured.

| Repair reverted | Cases that fail |
|---|---|
| Field a3 reads `cookie_fields` again | 3 |
| Field a4 reads the `referer` value again | 5 |
| Field a5 counts the whole header list again | 5 |
| `http_utils.py` restored from `origin/dev` | 15 |
| `ja4h.py` restored from `origin/dev` | 13 |

## How this was tested

```
pytest tests/ -m "not spec_validation" -q   1274 passed, 8 xfailed, 93 subtests passed
pytest tests/ -m spec_validation -q         1381 passed, 146 skipped, 114 xfailed
ruff check ja4plus/ tests/                  All checks passed
ruff format --check ja4plus/ tests/         119 files already formatted
mypy ja4plus/                               no issues found in 27 source files
coverage (--cov=ja4plus), base 81%          after 81%
```

## Notes

- The Changelog round is `TBD`. Rounds 49 to 61 are taken.
- The self-review found two more gaps and both are repaired. See the review comment.
- `tests/test_ja4h_raw.py:_info` now appends `Cookie` to the header list when the request
  carries a cookie. The fixture modelled a request that no parse path produces, because a
  request that carries a cookie carries a Cookie header.

Verified against: https://github.com/FoxIO-LLC/ja4/blob/main/technical_details/JA4H.png (retrieved 2026-08-08, commit `27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8`, transcribed in `docs/specs/foxio/JA4H.md`)
Verified against: https://www.rfc-editor.org/rfc/rfc9110#name-method (section 5.6.2 names the token characters of a method)


Crank-Git:
## Self-review

**I ran this review myself.** Two Sonnet review children were spawned for a correctness
lens and a test-quality lens, and neither returned. Nothing below comes from them.

### Findings I fixed

**1. `is_http_request` decoded the whole stream buffer.** The new request-line test ran on
the whole buffer, and `ja4l.py:365` calls the gate for every TCP payload. A stream buffer
grows to 1048576 bytes, so every packet of every stream that carries no HTTP paid a 1 MB
decode.

```
before: 200 calls on a buffer of 1000003 bytes: 9.2 ms
after : 200 calls on a buffer of 1000003 bytes: 0.3 ms
```

Fixed in `221e8c2`. The test reads 8192 leading bytes, and
`test_the_reassembly_gate_reads_no_more_than_the_request_line_limit` holds the bound.

**2. R8 sat on a line I rewrote, and no case held it.** The D4 repair rewrote the
expression that applies `min(..., 99)`. Removing the cap failed 0 cases in the whole
suite. Fixed in `1c07728`; removing the cap now fails 2 cases.

**3. The one request-line pattern narrows `parse_http_request`, and nothing recorded it.**
Measured on `origin/dev`:

```
b'GET / HTTP/11\r\nHost: a'  -> ('GET', '/', 'HTTP/11')
b'GET  /  HTTP/1.1\r\nHost'  -> ('GET', '', '/')
```

The path split the request line on one space and read no version. Both are repairs, not
regressions, and `ef96433` records them as cases.

### Findings I checked and cleared

**No catastrophic backtracking.** `[tchar]+`, `\s+` and `\S+` are disjoint character sets,
so the pattern holds no ambiguous alternation. Measured on adversarial input:

```
50 matches on 20000 bytes of 'A'                      9.42 ms
50 matches on 'A'*10000 + ' ' + '/'*10000 + ' HTTP/'  9.83 ms
50 matches on 'GET' + ' '*10000 + '/' + ' '*10000     8.06 ms
```

That is about 0.2 ms per match, and it grows linearly. The gate also caps its input.

**The pattern matches no response line and no other protocol.** `tchar` omits `/`, so
`HTTP/1.1 200 OK` never reads as a request line. Measured on all three parse paths and on
the gate:

```
b'HTTP/1.1 200 OK\r\nHost: a' -> from_bytes None | parse None | gate False
b'\x16\x03\x03\x00E'          -> from_bytes None | parse None | gate False
b'SSH-2.0-OpenSSH_9.6\r\n'    -> from_bytes None | parse None | gate False
b'random data here\r\n\r\n'   -> from_bytes None | parse None | gate False
```

`test_the_reassembly_gate_refuses_a_payload_that_holds_no_request_line` holds all four.
This matters because `ja4l.py:365` reads the same gate, and a false positive there moves a
JA4L measurement point. The vector replay confirms it: 791 values on ten methods, 0 differ.

**Header name case is consistent.** `_ja4h_part_a` lowercases the header list before it
tests for `cookie` and `referer`, and `_ja4h_header_names` lowercases each name for the
same test. `test_reads_the_referer_header_name_without_regard_to_its_case` holds it.

**Field a5 moves for no request an ordinary parser produces.** `_ja4h_header_names` drops a
name that is empty, a name that starts with `:`, `cookie` and `referer`. The old count
dropped only the last two. The three parse paths produce an empty name only for a header
line whose name is whitespace, and they produce no name that starts with `:`, because
`^([^:]+):` needs one non-colon character first. The 73 JA4H values of the vector set are
unchanged, which measures this.

### The character set

`METHOD_TOKEN_CHARACTERS` is the `tchar` production of RFC 9110, character for character.

Verified against: https://www.rfc-editor.org/rfc/rfc9110.html#name-tokens (section 5.6.2,
retrieved 2026-08-08)

Crank-Git:
## Final verification, measured on `e4dc315`

The base moved twice while this branch was open. #227 landed as Changelog round 62 and
#216 landed as round 63, and both are merged in. **The #219 row stays `TBD`.**

**The register figure changed under this branch, and #216 changed it, not #219.** The
brief named 114 keys against 114 `xfailed`. #216 added two JA4T entries, so the base now
holds 116. The measurement that matters is the delta against the base, and it is zero.

```
$ git show origin/batch/238-act-on-epic-10:tests/foxio_deviations.json  ->  116 keys
$ tests/foxio_deviations.json on this branch                           ->  116 keys
keys this branch adds:    []
keys this branch removes: []
```

### The vectors do not move

```
origin/batch/238-act-on-epic-10 : 38 captures, 791 values, 73 of them JA4H
this branch                     : 38 captures, 791 values, 73 of them JA4H
values that differ              : 0
JA4H values that differ         : 0
```

The replay reads the stored list of all ten fingerprinters, so the count covers JA4, JA4S,
JA4H, JA4T, JA4TS, JA4L, JA4X, JA4SSH, JA4D and JA4D6.

**Note on the count.** The brief named 82 JA4H values. This branch and its base both
produce 73. The 82 predates #193, which cut `CVE-2018-6794.pcap` from 12 values to 2. The
figure that gates this work is the difference, and it is 0.

### The gates

```
pytest tests/ -m "not spec_validation" -q   1290 passed, 8 xfailed, 93 subtests passed
pytest tests/ -m spec_validation -q         1428 passed, 146 skipped, 116 xfailed
ruff check ja4plus/ tests/                  All checks passed
ruff format --check ja4plus/ tests/         120 files already formatted
mypy ja4plus/                               no issues found in 27 source files
coverage (--cov=ja4plus)                    81%, and the base is 81%
```

### The cases can fail

`tests/test_ja4h_part_a_readings.py` holds 42 cases. Measured on the merged tree:

| What is reverted | Cases that fail |
|---|---|
| `ja4plus/` restored from the base | 28 |
| `ja4h.py` alone restored from the base | 13 |
| `http_utils.py` alone restored from the base | 15 |
| Field a3 reads `cookie_fields` again | 3 |
| Field a4 reads the `referer` value again | 5 |
| Field a5 counts the whole header list again | 5 |
| `min(..., 99)` removed from the count | 2 |

Issue #229, opened 2026-08-08
Compare the JA4X values of the FoxIO Rust snapshots


Crank-Git:
#202 transcribed `technical_details/JA4X.png` and measured the JA4X values of two
captures. The measurement found that **no test in this repository compares a JA4X value
against a FoxIO Rust snapshot.**

## What the register claims

`tests/foxio_deviations.json` holds the key `https-connect.pcap/JA4X`. Its cause reads
`tests/test_foxio_rust_parity.py measures the match.`

## What the harness does

`tests/test_foxio_rust_parity.py:72` reads:

```python
SNAPSHOT_METHODS = (("JA4", "ja4"), ("JA4S", "ja4s"))
```

The harness holds no JA4X entry. `read_rust_snapshot` also takes the two-space level of a
snapshot alone, and its docstring states that it ignores `tls_certs` and `http`. Every
`ja4x` value sits under `tls_certs:` at a deeper indent, so the reader skips it by design.

## The value the comparison would hold

`tests/foxio_vectors/rust_expected/ja4__insta@https-connect.pcap.snap` holds two values,
and `ja4plus` produces both, in the same order. Measured on 2026-08-08.

```
FoxIO rust snapshot : 7d5dbb3783b4_2bab15409345_5e17a2514980
                      7d5dbb3783b4_7d5dbb3783b4_9c5875a5c227
ja4plus JA4XFingerprinter : 7d5dbb3783b4_2bab15409345_5e17a2514980
                            7d5dbb3783b4_7d5dbb3783b4_9c5875a5c227
```

The other nine snapshots under `tests/foxio_vectors/rust_expected/` hold further values.

## What to build

1. Read the nested `tls_certs` block in `read_rust_snapshot`, and key each `ja4x` value on
   its stream and its position.
2. Add `JA4X` to the comparison, and record the result.
3. Correct the cause of `https-connect.pcap/JA4X` so that it names the test that measures
   the match.

## What not to build

**No fingerprint moves in this issue.** `CLAUDE.md` rule 1 binds every fingerprinter
change. A mismatch produces a register entry or a decision issue.

## Acceptance criteria

- [ ] `tests/test_foxio_rust_parity.py` compares the JA4X values of every local snapshot.
- [ ] The `https-connect.pcap` comparison passes, and the test names both values.
- [ ] The cause of `https-connect.pcap/JA4X` names a test that measures the match.
- [ ] The conformance suite reports the same count of `xfailed` cases, or a lower one.
[...]

PR #243, opened 2026-08-08
Compare JA4X against the Rust snapshot values the reader skipped (#229)

Crank-Git:
Refs #229. Batch 17, tracked on #238. This pull request targets `batch/238-act-on-epic-10`.

## What the harness did not do

`read_rust_snapshot` took the two-space level of a FoxIO Rust snapshot alone. Every
`ja4x` value sits under `tls_certs:` at a deeper indent, so the reader skipped it by
design. Five of the ten local snapshots hold a certificate, 19 streams hold at least one
value, and those streams carry **43 JA4X values that no case compared**.

`tests/foxio_deviations.json` stated for `https-connect.pcap/JA4X` that
`tests/test_foxio_rust_parity.py measures the match.`, and it marked the entry `decided`.
**That sentence was false when it was written.** This is the third comparison the project
has found that never ran, after `ja4t` (#196) and `ja4ssh` (#199).

## What this pull request builds

1. The reader enters the `tls_certs` block and keeps each `ja4x` value in snapshot order.
2. `TestTheJa4xValuesTheRustSnapshotHolds` collects 43 value cases and 5 count cases.
3. The cause of `https-connect.pcap/JA4X` keeps its sentence, and it names the class that
   holds the comparison. The sentence is now true rather than corrected.

## The result

| Measurement | Count |
|---|---|
| Local Rust snapshots | 10 |
| Snapshots that hold a JA4X value | 5 |
| Streams that hold at least one | 19 |
| JA4X values in those streams | 43 |
| Values `ja4plus` reproduces exactly, in snapshot order | 43 |
| Values that differ | 0 |
| Streams whose value count differs | 0 |

**All 43 values agree, so the register gains no entry.** The register holds 116 keys
before and 116 after, against 116 `xfailed` cases.

## The revert proof

Remove the `tls_certs` branch of the reader, and **48 cases stop running**: 43 value
cases and 5 count cases. Three checks fail and name the loss.

```
FAILED TestTheJa4xValuesTheRustSnapshotHolds::test_the_local_snapshots_hold_the_forty_three_values_the_reading_counts
FAILED TestTheJa4xValuesTheRustSnapshotHolds::test_the_suite_collects_one_case_for_every_value_the_snapshots_hold
FAILED TestTheJa4xValuesTheRustSnapshotHolds::test_the_https_connect_stream_is_the_one_stream_the_python_file_omits
```

## One stream carries a register key of this module

`tests/test_spec_validation.py` builds the key form `<capture>/<stream>:<port>/JA4X.<n>`
from the FoxIO Python file, and that file holds a JA4X value for 18 of the 19 streams. A
measurement found **41 of the 43 proposed keys already collected by that module**. One
entry that matches a case of each module marks both, and neither fix reports itself, so
`certificate_key` keys `https-connect.pcap` stream 0 on port 54723 alone.
`test_no_register_key_of_this_module_belongs_to_the_spec_validation_suite` holds the two
key sets apart.

## What stays unmeasured

- **D1, D2 and R8 of `docs/specs/foxio/JA4X.md`.** No snapshot field holds a raw form, no
  capture carries a certificate with a duplicate attribute, and no capture carries a
  certificate with no extension. None is a mismatch.
- **The `ja4ssh` values of the snapshots.**
  `tests/foxio_vectors/rust_expected/ja4__insta@ssh2.pcapng.snap:215` holds a `ja4ssh:`
  block with two values, and no other local snapshot holds the field. #199 reports the
  gap and owns the reading. This pull request changes nothing there and widens no scope.

## No fingerprint moves

`git diff --name-only origin/batch/238-act-on-epic-10...HEAD -- ja4plus/` returns nothing.

## How this was tested

```
pytest tests/ -m spec_validation -q      1481 passed, 146 skipped, 116 xfailed
pytest tests/ -m "not spec_validation" -q 1248 passed, 8 xfailed, 93 subtests passed
ruff format --check ja4plus/ tests/      119 files already formatted
```

The conformance suite reported 1428 passed and 116 xfailed on the base. The 53 added
cases are the whole difference.

Verified against: https://github.com/FoxIO-LLC/ja4/tree/main/rust/ja4/src/snapshots (pinned commit `27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8`, retrieved 2026-08-08)

The Changelog round is **65**, assigned by the coordinator. Round 64 belongs to #219, which landed on the integration branch after this branch left it, so this branch shows 63 and then 65.


Crank-Git:
Self-review, correctness lens. One Sonnet child reviewed the parser state machine, the effect on the pre-existing JA4/JA4S/JA4T cases, the vacuous-case risk and the occurrence numbering. It reported no defect and it returned its evidence, so no review is reported here that did not run.

Two findings of my own reading are now fixed in 98f12ca.

1. **The count expression read case order.** `test_the_produced_ja4x_count_equals_the_rust_snapshot_count` built `expected[case.identity] = case.occurrence`, which is the count only while `certificate_cases` returns a sorted list. A later sort change would have made the check read a smaller count and pass on a stream that differs. A `Counter` states the count directly.
2. **Three reads held the literal `JA4X` beside `SNAPSHOT_CERT_METHOD`.** One concept, one word. Each one now reads the constant.

The conformance suite reports 1481 passed and 116 xfailed after the fix, which is the same as before it.

Issue #246, opened 2026-08-08
Write the JA4TS value that a RST produces


Crank-Git:
Split from #226, which built part e of JA4TS and measured that the RST value separates from it.

## What the FoxIO sources state

The deleted `technical_details/JA4T.md` states three rules that #226 did not build.

1. A RST appends `R` and its delay to part e. Its example ends
   `65535_2-1-3-1-1-4_65495_8_1-2-4-8-R6`.
2. The delay of the RST is the interval since the last SYN-ACK, in whole seconds.
3. "Note that RST packets do not contain TCP options or window sizes, as such the
   program will need to be aware of the previous JA4TS."

`wireshark/source/packet-ja4.c:1608` writes a second `JA4TS` field on the RST packet, and
it rebuilds part a through part d from the stored connection.
`zeek/ja4t/main.zeek:232-234` appends `-R<seconds>` to the delay list.

## Why this separates from part e

Both implementations nest the RST branch inside the delay branch.
`wireshark/source/packet-ja4.c:693` reads `rst_time` only when `syn_ack_count > 1`, and
`zeek/ja4t/main.zeek:233` reads `rst_ts` only when the delay list holds a value. A delay
list is complete with no RST, and the deleted file's second example shows one:
`62727_2_8961_00_1-2-4-8-16`.

The RST value needs two things part e does not: a reader of RST packets, and stored part
a through part d for a packet that carries neither.

`docs/specs/foxio/JA4T.md` records the reading as R13.

## Acceptance criteria

- `ja4plus/fingerprinters/ja4ts.py` appends `R<seconds>` to part e when the server sends
  a RST on a connection that already holds a delay.
- A RST on a connection with no delay produces no part e, which both implementations do.
- The RST value reads part a through part d from the stored connection, because the RST
  packet carries no window size and no option.
- A test builds the example of the deleted file and asserts
  `65535_2-1-3-1-1-4_65495_8_1-2-4-8-R6`.
- The replay of every capture in `tests/foxio_vectors/` names every value that moves, and
  every value that moves is a JA4TS value.
- `tests/foxio_deviations.json` key count equals the `xfailed` count of the conformance
  suite.

## What blocks it

No vector reaches this rule. `tests/foxio_vectors/` holds one connection with more than
one SYN-ACK, in `ssh2.pcapng`, and the server sent no RST on it. A capture that reaches
the rule must be built, the way `tests/build_ssh_retransmission.py` builds one.

[...]

Issue #245, opened 2026-08-08
Batch 18: build the five fingerprint decisions


Crank-Git:
Integration branch `batch/245-fingerprint-decisions`, off `dev` at `fe50713`.

**Every member changes a fingerprint `ja4plus` 0.6.0 publishes.** The user decided all five on 2026-08-08, each recorded as a decision comment on its own issue.

## Members

- [ ] #225 — JA4L carries three parts, Wireshark's definition. **114 values change. The largest change of the session.**
- [ ] #214 — JA4SSH emits the trailing window. Four captures, one rule.
- [ ] #226 — JA4TS carries part e. **Reverses the #215 D6/D7 ruling of earlier today.**
- [ ] #228 — JA4X writes the zero sentinel for an empty extension list.

#231, #211 and #242 go in the next batch.

## The four touch four different fingerprinters

`ja4l.py`, `ja4ssh.py`, `ja4ts.py`, `ja4x.py`. They run in parallel and the sub-merge conflicts should be confined to `docs/specs/spec.md`.

## The invariant, and it will move

`dev` holds **116 register keys against 116 `xfailed`**, 8 undecided. **These changes are expected to move both numbers**, because #225 diverges from the FoxIO Python reference on every JA4L comparison. **The two numbers must stay equal**, and every member reports its own movement.

## What every member must do

**Report the projected register movement before building**, then the measured movement after. **Prove every repair by reverting it** and state the count of cases that fail. **Replay every capture and name every value that moves** — this batch is the first in two sessions where values are supposed to move, so a silent extra change is the risk.

**#226 carries a known cost the user accepted:** the only JA4TS reference value in existence, Zeek's `65535_00_00_00`, has four parts. #198 measured JA4TS at 9 of 10 matching. Adding part e diverges from it. Measure the new number and record it; do not leave `docs/specs/foxio/zeek.md` reading as though 9 of 10 still holds.

Opened 2026-08-08. Nothing merged yet.

Crank-Git:
**Batch 18 merged into `dev` as `07f5a7a`.** All nine continuous integration jobs pass. Three members, Changelog rounds 66 to 68.

| Round | Issue | Result |
|---|---|---|
| 66 | #225 | JA4L keeps two timing parts; a QUIC connection gains the `quic` marker. **36 values move, all QUIC. 0 TCP values move** |
| 67 | #226 | JA4TS carries part e. **1 of 803 values moves** |
[...]

Issue #226, opened 2026-08-08
Decide whether JA4TS carries the retransmission part


Crank-Git:
Found while #221 read the deleted FoxIO text specifications. This issue is untriaged.

**Related to #215 item 4, which asks the same question from the image alone.** #221 found
a FoxIO-authored source that #215 did not have. **A triager may fold this issue into #215.**

## The contradiction

**The deleted `technical_details/JA4T.md` states that JA4TS carries part e. The image at
the pinned commit states that part e appears only on JA4TScan.**

| Source | What it states |
|---|---|
| `technical_details/JA4T.png`, the pinned specification | The caption of part e reads `TCP Retransmission Timings (only on JA4TScan)` |
| The deleted `technical_details/JA4T.md` | The heading reads `__JA4TS and JA4TScan Fingerprint formats:__` and the form under it reads `WindowSize_TCPOptions_MSSValue_WindowScale_TimeSinceLastSYNACK` |
| `wireshark/source/packet-ja4.c:1595` | It writes the `JA4TS` field through `ja4t()` with a connection, which appends part e |
| `rust/ja4/src/tcp.rs` | It writes four parts and no timing part |
| `ja4plus/fingerprinters/ja4ts.py` | It writes four parts and no timing part |

The deleted file states the behaviour in two more sentences.

> JA4TS takes into account the number of SYNACK TCP Retransmissions, or RST, as well as the
> time delay between each retransmission or RST.

> If no retransmissions are seen, as there shouldn't be in normal network communications,
> the fingerprint will omit section e. If retransmissions are seen, the fingerprint will
> fill out section e.

## The provenance

| Item | Value |
|---|---|
| Path | `technical_details/JA4T.md` |
| Commit that holds it | `b6f3ff4^` |
| Blob SHA-1 | `54f207980273c77d5df928d592c694e4cae52cd7` |
| SHA-256 | `d8ec985535f5bbf2cf411989f6ab19b17191585137e4ee448041e9ee91191d76` |
| Bytes | 6423 |
| Commit that deleted it | `b6f3ff4`, `Update README.md`, 2024-02-22 |

```bash
git cat-file -p b6f3ff4^:technical_details/JA4T.md
```

## Why this needs the user

**The image at the pinned commit outranks a deleted file, so the image decides.** The
deleted text is nonetheless FoxIO-authored, and it agrees with one of the three FoxIO
[...]

PR #247, opened 2026-08-08
Write part e of JA4TS, the time since the last SYN-ACK

Crank-Git:
Builds #226. Part of batch 18, tracked on #245.

**The user reversed the D6/D7 ruling of #215 on 2026-08-08.** JA4TS carries part e.

## What changed

`ja4plus/fingerprinters/ja4ts.py` writes part e, the delay between each SYN-ACK of one
connection, in whole seconds. The fingerprinter is now stateful, so it overrides `reset`
and `cleanup_connection`.

## Why — three FoxIO sources, and nothing ever corrected them

The image is the one source that keeps part e off JA4TS. Its caption reads
`TCP Retransmission Timings (only on JA4TScan)`, and **its own example value `1-2-4-8-R6`
carries the `R` suffix that the caption denies.**

- The deleted `technical_details/JA4T.md` heads its form
  `__JA4TS and JA4TScan Fingerprint formats:__` and writes
  `WindowSize_TCPOptions_MSSValue_WindowScale_TimeSinceLastSYNACK`. `b6f3ff4` deleted the
  file on 2024-02-22 and no commit restored it.
- `wireshark/source/packet-ja4.c:1595` writes the `JA4TS` field through `ja4t()` with a
  connection, which appends part e at line 684.
- `zeek/ja4t/main.zeek:227-236` appends the delay list to `c$conn$ja4ts`. **This third
  source is one the earlier reading did not name.**

## The rules, read from the blob rather than inferred

The issue required reading the deleted file for what part e measures and what it writes
when no retransmission occurs. It states both.

> If no retransmissions are seen, as there shouldn't be in normal network communications,
> the fingerprint will omit section e. If retransmissions are seen, the fingerprint will
> fill out section e.

**Part e is absent, and it is not `00`.** Both implementations corroborate it, because
each appends part e only when a delay exists.

> we start with the timestamp of the first SYNACK and subtract it from the next SYNACK,
> rounding the result to the nearest whole number in seconds

`timediff` in the dissector calls the C `round`, which carries a half away from zero, so
`ja4ts.py` calls `math.floor(delta + 0.5)`. The Python built-in `round` carries a half to
the even number and would write `0` where the dissector writes `1`.
`zeek/ja4t/main.zeek:180` truncates instead, and the prose and the dissector outvote it.

**The state bound is ten retransmissions and a timeout of two minutes.**
`zeek/ja4t/main.zeek:185` corroborates the count. `MAX_SYN_ACK_TIMES` in the dissector
stores ten timestamps, which holds nine delays, so the dissector reads one fewer.

## The Zeek cost the user accepted, measured

**The user accepted a fall in the Zeek comparison in advance. The measurement reports
none.**

| Reading | Before | After |
|---|---|---|
| JA4TS rows compared | 10 | 10 |
| JA4TS rows that match | 9 | **9** |
| `ja4plus` values carrying part e | 0 | **0** |

The three `conn.log` baselines hold 10 JA4TS values, and `Scripts.ja4-conn/conn.log`
holds `ja4ts 65535_00_00_00`. **Each of the ten baseline connections holds one SYN-ACK**,
so the omission rule keeps every value at four parts. The one row that differs is the
`DLT_NULL` defect of the Zeek script, which #198 already recorded and which #226 did not
change.

`docs/specs/foxio/zeek.md` gains the new reading beside #198's, in the shape #200 used.
#198's measurement is not rewritten.

## Every value that moves

The 38 captures under `tests/foxio_vectors/` were replayed through `Processor` before and
after.

| Reading | Before | After |
|---|---|---|
| Values produced | 803 | 803 |
| Values that differ | — | **1** |
| Keys in `tests/foxio_deviations.json` | 116 | 116 |
| Conformance `xfailed` | 116 | 116 |

**One value moved, and it is a JA4TS value.** `ssh2.pcapng` packet 372 moves from
`64240_2-1-1-4-1-3_1460_7` to `64240_2-1-1-4-1-3_1460_7_0`. **No value of another method
moved.**

It is the one connection in the whole vector set the server answered twice:
`184.150.157.177:80` to `172.16.225.48:57380`, answered at 1688672871.347960 and
1688672871.353789. The interval of 0.005829 s rounds to `0`. Packet 369 is the first
SYN-ACK and keeps four parts.

**The register invariant holds: 116 keys against 116 `xfailed`, before and after.** No
conformance case compares a JA4TS value, so none could move.

## The RST value separates, and #246 owns it

The issue asked whether the second `JA4TS` value on a RST with the `-R<interval>` suffix
is inseparable from part e. **It separates.** Both implementations nest the RST branch
inside the delay branch: `wireshark/source/packet-ja4.c:693` reads `rst_time` only when
`syn_ack_count > 1`, and `zeek/ja4t/main.zeek:233` reads `rst_ts` only when the delay list
holds a value. A delay list is complete with no RST, which the deleted file's own second
example shows, ending at `62727_2_8961_00_1-2-4-8-16`. The RST value also needs a reader
of RST packets and stored parts a to d that the RST packet does not carry.

Raised as #246. `docs/specs/foxio/JA4T.md` records the measurement as R13.

## How it is tested

`tests/test_ja4ts_part_e.py` holds 20 cases, written before the implementation. One case
replays the deleted file's own example timestamps and asserts its stated sequence, from
`62727_2_8961_0` through `62727_2_8961_0_1-2-4-8-16`.

**Seven mutations prove the cases bite.**

|---|---|
| Drop part e from the output | 10 |
| Remove the omission rule | 7 |
| Use the Python built-in `round` | 2 |
| Use `floor(delay + 0.5)` | 1 |
| Truncate the delay, as Zeek does | 3 |
| Cap the delay list at nine | 3 |
| Raise the timeout to one hour | 2 |
| Remove the connection table bound | 3 |
| Key part e on the server address alone | 3 |
| Drop the delay list bound | 2 |

Reverting the whole of `ja4plus/` to the base loses all 20 cases, which stop collecting.

**The self-review found a case that could not fail, and repaired it.** Three bound cases
read the constant they asserted, so the cap-at-nine mutation changed nothing. They now
state the literal values, and a separate case pins the three constants.

## Verification

- `pytest tests/ -m "not spec_validation"` — 1310 passed, 8 xfailed, 93 subtests passed.
- `pytest tests/ -m spec_validation` — 1481 passed, 146 skipped, **116 xfailed**.
- `tests/foxio_deviations.json` — **116 keys**. Equal, before and after.
- `ruff check` clean. `ruff format --check` clean. `mypy ja4plus/` clean, 27 files.
- Coverage 81% total, unchanged. `ja4plus/fingerprinters/ja4ts.py` at 93%.

Verified against: `git show b6f3ff4^:technical_details/JA4T.md` in FoxIO-LLC/ja4 at the
pinned commit `27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8` (retrieved 2026-08-08). The blob
was read directly rather than through the transcription in
`docs/specs/foxio/deleted-text-specifications.md`, and it confirms every quotation there.

**The self-review found three more items and repaired them**, including a rounding
expression that matched the C `round` only for a delay of zero or more. The review
comment on this pull request holds the measurements.

The Changelog round stays `TBD`.


Crank-Git:
## Self-review

Two Sonnet review children were spawned and **neither reported before this comment**, so
**I ran the review myself** and record it here. Three findings, all repaired in
`e33787c` and `5393a10`.

### REAL — the rounding matched the C `round` only for a delay of zero or more

`math.floor(delay + 0.5)` carries a negative half towards zero. It writes `0` for `-0.5`,
and `timediff` in `wireshark/source/packet-ja4.c` writes `-1`. A capture that holds a
SYN-ACK out of order produces a negative delay, and every packet is hostile input.

Measured against the C library over -50 to 50 in steps of 1/8: `math.floor(x + 0.5)`
differs on every negative half. `int(math.copysign(math.floor(abs(x) + 0.5), x))` matches
on all 801 inputs.

`tests/test_ja4ts_part_e.py` gains `test_part_e_rounds_a_negative_half_second_delay_away_from_zero`,
and reverting the expression fails that case alone.

### MINOR — a comment stated the opposite of what the code does

The comment read "A retransmission past the bound changes no fingerprint, so the packet
only keeps the entry alive." The entry stores no timestamp past the bound, so a later
SYN-ACK does **not** keep it alive; the entry ages from the tenth retransmission. The
comment now states that.

### REAL, and the design already held — the timeout alone does not bound the table

The timeout reads the capture timestamp of the packet, so a capture whose timestamps
**decrease** never ages an entry out: `now - stamps[-1]` is negative. Nothing in the
diff said the entry count is what holds the table in that case.

Measured, 50000 packets each:

| Hostile pattern | Entries | Timestamps |
|---|---|---|
| Distinct connections, one instant | 1000 | 1000 |
| **Ever-decreasing timestamps** | **1000** | 1000 |
| Spoofed source addresses | 1000 | 1000 |
| One connection, many SYN-ACKs | 1 | **11** |

The two bounds are independent and both hold. `tests/test_ja4ts_part_e.py` gains
`test_a_capture_with_decreasing_timestamps_bounds_the_connection_table` and
`test_one_connection_stores_eleven_timestamps_at_most`.

### Categories checked and clean

- **No unhandled exception escapes.** A packet with no address layer gives a null
  connection key, and a packet with no timestamp gives none; both return four parts
  rather than raising. `generate_ja4ts` catches the four parse errors it expects and no
  bare `Exception`.
- **No packet reference is retained.** The tracker stores float timestamps only.
  `tests/test_no_retained_packets.py` passes.
- **No cross-connection corruption.** The key is the outer address pair with the
  innermost port pair, which is the convention `packet_endpoints` already applies for
  every method. `test_two_connections_hold_separate_delay_lists` covers it, and keying on
  the server address alone fails 3 cases.
- **Cost.** One SYN-ACK against a full 1000-entry table takes 40.5 us, and only SYN-ACK
  packets reach the expiry scan.

## The mutation battery, re-run against the final 20 cases

|---|---|
| Drop part e from the output | 10 |
| Remove the omission rule | 7 |
| Use the Python built-in `round` | 2 |
| Use `floor(delay + 0.5)` | 1 |
| Truncate the delay, as Zeek does | 3 |
| Cap the delay list at nine | 3 |
| Raise the timeout to one hour | 2 |
| Remove the connection table bound | 3 |
| Key part e on the server address alone | 3 |
| Drop the delay list bound | 2 |

**Every mutation kills at least one case.** The earlier cap-at-nine mutation killed none,
which is the vacuous shape `.claude/rules/conformance.md` names; three bound cases read
the constant they asserted. They now state the literal values.

## Final verification

- `pytest tests/ -m "not spec_validation"` — 1310 passed, 8 xfailed, 93 subtests passed.
- `pytest tests/ -m spec_validation` — 1481 passed, 146 skipped, **116 xfailed**.
- `tests/foxio_deviations.json` — **116 keys**. Equal, before and after.
- `ruff format --check` — 121 files already formatted. `ruff check` clean.
- `mypy ja4plus/` — clean, 27 files.
- Coverage **81% total, unchanged**; `ja4plus/fingerprinters/ja4ts.py` at 93%.
- Replay of all 38 captures after the rounding repair: **803 values, still exactly 1
  differs**, and it is the same JA4TS value.

Crank-Git:
## The third acceptance criterion needed a vector case, and it now has one

> - [ ] If the decision changes a fingerprint, one vector measures the new value.

The decision does change a fingerprint, and every case up to `5393a10` was constructed
from `scapy` packets. That satisfied the first two criteria and not the third.

`8b2f564` adds four cases that read `tests/foxio_vectors/ssh2.pcapng` through
`Processor`. It holds the one connection of the whole vector set that the server answered
twice.

| Case | What it asserts |
|---|---|
| `test_the_retransmitted_syn_ack_carries_part_e` | Packet 372 is `64240_2-1-1-4-1-3_1460_7_0` |
| `test_the_first_syn_ack_of_that_connection_carries_no_part_e` | Packet 369 is `64240_2-1-1-4-1-3_1460_7` |
| `test_one_value_of_the_capture_carries_part_e` | Exactly one of the nine values carries part e |
| `test_the_capture_produces_nine_ja4ts_values` | A value that appears or disappears fails |

The mutation battery re-run against the final **24** cases:

|---|---|
| Drop part e from the output | **12** |
| Remove the omission rule | **9** |
| Use the Python built-in `round` | 2 |
| Use `floor(delay + 0.5)` | 1 |
| Truncate the delay, as Zeek does | 3 |
| Cap the delay list at nine | 3 |
| Raise the timeout to one hour | 2 |
| Remove the connection table bound | 3 |
| Key part e on the server address alone | **4** |
| Drop the delay list bound | 2 |

The two mutations that reach the vector value rose by two and by two, which is what the
four new cases should do.

Final: `pytest -m "not spec_validation"` 1314 passed, 8 xfailed; `pytest -m
spec_validation` 1481 passed, **116 xfailed** against **116 register keys**; `ruff` and
`mypy` clean.

Issue #225, opened 2026-08-08
Decide the JA4L form: the part count, the protocol part and the duplicate server value


Crank-Git:
Raised by #200, which transcribed `technical_details/JA4L.png` into `docs/specs/foxio/JA4L.md`. #196 raised #215 in this shape, #198 raised #214 and #203 raised #219.

**This issue decides nothing on its own. Each item changes a fingerprint that this project publishes, so the user decides it.** #200 changed no file under `ja4plus/` and moved no fingerprint.

## The reading that raises this

**The specification labels three parts, and this project publishes two.**

```
JA4L=5191_42_45014
     \__/ \_/ \___/
      a    b    c

a = One-way TCP latency in µs (1ms = 1,000µs)
b = Observed TTL
c = One-way application handshake latency
```

Measured on 2026-08-08 across every reference this project reads.

| Source | Parts |
|---|---|
| `technical_details/JA4L.png`, the specification | **three** |
| `wireshark/test/testdata/`, the `ja4.ja4l` and `ja4.ja4ls` keys | **three** |
| `zeek/tests/Traces/`, the `ja4l` and `ja4ls` columns | **three** |
| `python/test/testdata/` | two |
| `ja4plus` | two |

**The specification and two implementations write three parts. The Python reference writes two, and this project follows the Python reference.**

The image is at the pinned commit `27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8`, SHA-256 `04036284edfcf0d8e94f3ca6660b1ab688813df7bfb82b0f001b65d7daa07354`. `docs/specs/foxio/JA4L.md` holds the transcription and marks the rule uncertain, so the vector fallback keeps the two-part form today.

**This corrects a reading of #198.** `docs/specs/foxio/zeek.md` stated that the Zeek script adds a third part that no other reference holds. #198 compared the Python reference against the Zeek script and read neither the image nor the Wireshark expected output. #200 corrected that section on the same branch. **Zeek is not the outlier on the part count.**

## D1 — Does this project publish part c?

`ja4l.py:408`, `ja4l.py:423`, `ja4l.py:439`, `ja4l.py:504` and `ja4l.py:549` each write `"{}_{}"`.

```
foxio wireshark/test/testdata/https-connect.pcap.json:58-59 (ja4.ja4l) : 45_64_66
ja4plus JA4L-C on tests/foxio_vectors/https-connect.pcap               : 45_64
```

Part a and part b agree. This project publishes no part c.

[...]

PR #248, opened 2026-08-08
Keep two JA4L timing parts and mark a QUIC connection (#225)

Crank-Git:
Builds #225. Batch 18, tracked on #245.

## What this changes

**JA4L keeps two timing parts. A QUIC connection gains the `quic` marker.**

The three-part adoption was proposed, measured and reversed. The measurement is the
substance of this pull request, and it is recorded on #225 and in the Changelog.

## Why the three-part form is declined

`JA4L.png` labels three parts, `wireshark/test/testdata/` holds 44 `ja4.ja4l` values and 44
`ja4.ja4ls` values that all carry three, and the Zeek baselines carry three.
`python/test/testdata/` and `rust/ja4/src/snapshots/` carry two.

Two findings decline the adoption.

**1. No vector in this repository can verify a three-part form.** Every one of the 114 JA4L
values in `tests/foxio_vectors/*.json` holds two parts, and
`tests/foxio_vectors/wireshark_expected/` holds `dhcp.pcapng.json` and `dhcpv6.pcap.json`,
neither carrying a JA4L key. A three-part form deviates from all 114 comparisons, which
moves the register from 116 entries to **230** and leaves **no case that measures any JA4L
value**. 20 of the 114 sit on 10 captures the dissector publishes no JA4L value for.

**2. The Wireshark part c is not computable where `ja4plus` writes a value.**

| Point | Dissector line | What it marks |
|---|---|---|
| D | `packet-ja4.c:1328` | The first client application packet after the handshake. |
| E | `packet-ja4.c:1343` | The first server application packet after D. |
| F | `packet-ja4.c:1366` | The first client application packet after E. |

`packet-ja4.c:1370` reads the D to E interval for `ja4ls`, and `packet-ja4.c:1382` reads the
E to F interval for `ja4l`. `ja4l.py:423` emits the server value on the SYN-ACK and
`ja4l.py:439` emits the client value on the first packet that carries both relative
numbers. Both points lead D. **Adopting part c is an emission-model change and not a format
change.** The dissector publishes 44 values on each key where the Python reference
publishes 46 `JA4L-C` and 48 `JA4L-S`, so the two references emit on different connections.

## Why the QUIC marker survives

It is the one part of the three-part reading that a published reference value verifies.

**The spelling follows Wireshark, and the sources settle it.**

| Source | Marker | Published values |
|---|---|---|
| `packet-ja4.c:1441` and `:1447` | `quic` | 18 |
| `zeek/ja4l/main.zeek:233` and `:252` | `q` | 14 |
| `python/ja4.py:165` | none | two parts |
| `rust/ja4/src/time.rs` | none | two parts |

`.claude/rules/external-apis.md:95` reads "Read no JA4L or JA4LS value of a Zeek baseline as
a reference value", and `:101` names the `q` marker as one of three Zeek divergences. **The
rule declines the Zeek spelling**, so the Wireshark spelling is the one published reading
that remains. No guess was made.

**The `tcp` literal is not adopted.** `packet-ja4.c:1348` and `:1354` write it on an HTTP
connection, and **none of the 88 published Wireshark values carries it**.

## The replay

The change touches two format strings. Over the 38 captures under `tests/foxio_vectors/`:

| Reading | Count |
|---|---|
| JA4L values before | 143 |
| JA4L values after | 143 |
| Values that move | 36, every one on a QUIC connection |
| **TCP values that move** | **0** |
| TCP values that stay | 107 |

| Capture | Values that move |
|---|---|
| `chrome-cloudflare-quic-with-secrets.pcapng` | 2 |
| `ssh2.pcapng` | 3 |
| `tls3.pcapng` | 11 |

**16 of the 36 reach a conformance case.** `tls-handshake.pcapng` carries no JA4L key in its
expected-output file, so its existing occurrence entry already covers its 20. On all 16 the
two timing parts agree, so the marker is the whole difference.

## The register

| Reading | Before | After |
|---|---|---|
| `tests/foxio_deviations.json` keys | 116 | 132 |
| `xfailed` in the conformance suite | 116 | 132 |

`tests/foxio_deviation_owners.json` gains issue 225.

## The revert proof

Restore `ja4plus/` from `origin/batch/245-fingerprint-decisions` and **32 cases fail**: 16
in the conformance suite and 16 in the unit suite. Every one is a QUIC JA4L case, and none
is a TCP case.

## Out of scope

**D2 and D4 stay open.** #225 ruled on the part count and on the marker alone.
`ssh2.pcapng/JA4L-S` stays under #34 unchanged, and `docs/specs/foxio/JA4L.md` records that
both await a decision.

## How this was tested

- `pytest tests/ -m "not spec_validation"`: 1296 passed, 8 xfailed, 93 subtests passed.
- `pytest tests/ -m spec_validation`: 1465 passed, 146 skipped, 132 xfailed.
- `ruff check`, `ruff format --check`, `mypy ja4plus/`: all clean.
- Six new cases in `tests/test_ja4l_quic_marker.py`, written before the change. Four failed
  against the base and the two TCP guards passed.

Verified against: https://github.com/FoxIO-LLC/ja4 (commit `27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8`, read from the pinned local checkout, retrieved 2026-08-08)

The Changelog round is `TBD`.


Crank-Git:
## Self-review

**I ran this review myself rather than delegating it to child agents.** Four review
children stalled without reporting earlier in this session, and the production diff here is
14 added lines across one file, so a direct read is both cheaper and more reliable than a
fan-out. No child agent contributed to this review.

### Correctness

**The change reaches exactly the two QUIC emission sites, and no TCP site.**
`ja4plus/fingerprinters/ja4l.py` holds five sites that return a JA4L value.

| Line | Branch | Parts |
|---|---|---|
| 415 | TCP, a reordered capture whose SYN follows the SYN-ACK | 2 |
| 430 | TCP, the SYN-ACK | 2 |
| 446 | TCP, the client measurement point | 2 |
| 511 | QUIC, the server Initial packet | 3 |
| 558 | QUIC, the client Handshake packet | 3 |

`generate_ja4l` sends a TCP packet to `_tcp_ja4l` and a UDP packet to `_quic_ja4l`, so the
two branches share no format string. The replay confirms the reading: 107 TCP values do not
move.

**No downstream consumer parses a JA4L value positionally.** This is the failure mode a
part-count change usually causes, so I searched for it. `ja4plus/` holds one split of a
fingerprint on `_`, at `ja4plus/fingerprinters/ja4ssh.py:470`, and it reads a JA4SSH value
and rejects any value that does not hold three parts. `ja4plus/ja4db.py` names no JA4L
method. `cli.py` and `processor.py` carry the fingerprinter and never read the value. **No
finding.**

**The client value reads the stored TTL and falls back to the packet TTL.**
`ttls.get("client", ttl)` at line 559 is unchanged behaviour, and the marker is appended
after it. No finding.

### Security

The marker is a module constant. The change adds no parsing of packet bytes, no state, and
no growth in any table. **No finding.**

### Scope

The production diff is one constant and two format strings. The register gains 16 entries,
each naming a measured cause. Twelve unit cases that asserted the two-part QUIC form now
assert the marked form, and no TCP case changes. **No finding.**

### One observation, no action

`tests/compare_zeek_baselines.py` is a manual script that no test imports, and it needs a
FoxIO checkout path to run. Its `ZEEK_COLUMNS` list reads `ja4l` and `ja4ls`. The `ja4plus`
QUIC values now carry a marker as the Zeek columns do, and the two spellings still differ,
`quic` against `q`. The script compares values as text, so a reader who runs it will see
that difference. **This is the recorded divergence and not a defect**, and
`.claude/rules/external-apis.md:101` already names the `q` marker as a Zeek divergence.

### Verdict

No finding that needs a change. The gates are green and the register count equals the
`xfailed` count.

Issue #214, opened 2026-08-08
Decide whether JA4SSH emits the trailing window on ssh2.pcapng


Crank-Git:
Found by #198, which read the FoxIO Zeek package as a fourth reference.

## The decision

`ssh2.pcapng` holds one SSH connection, `172.16.225.48:57377` to `54.160.114.75:22`.
Three FoxIO-authored readings of it exist, and they do not agree.

| Source | First value | Second value |
|---|---|---|
| `python/test/testdata/ssh2.pcapng.json` | `c36s36_c76s124_c74s5` | `c36s36_c0s0_c2s0` |
| `zeek/tests/Traces/Scripts.ja4-ssh2/ja4ssh.log` | `c36s36_c76s124_c74s5` | `c36s52_c42s76_c51s2` |
| `ja4plus` | `c36s36_c76s124_c74s5` | none |

The first value matches across all three. **The second value is the decision.**

## Why #97 does not settle it

`tests/foxio_deviations.json` declines the Python second value under #97. The reason it
records is that `c0s0` states a window that holds no SSH packet, and
`.claude/rules/conformance.md` declines a fingerprint that describes no traffic.

**The Zeek second value describes real traffic.** `c42s76` states 42 client packets and
76 server packets. It therefore falls outside the shape #97 declined, and the #97
decision does not reach it.

`zeek/ja4ssh/main.zeek:160-164` emits the value from `connection_state_remove`, so the
Zeek script always writes the window a connection holds open when it closes.

## The question for the user

Does `ja4plus` emit a trailing JA4SSH value for the window a connection holds open when
it closes, on a connection where that window holds SSH packets?

- If yes, which counts does it write? The Zeek counts and the Python counts differ, so
  the two references disagree about the window boundary as well as about the value.
- If no, the decline needs a stated reason that #97 does not already carry, and the
  register entry needs to name it.

Round 15 of the Changelog records that `gre-sample.pcap`, `sshv1.pcap` and `v6.pcap`
already hold a trailing JA4SSH value from `ja4plus`, so the behaviour is not uniform
across the vector set. That is part of the same question.

## Evidence

Reproduce the comparison from the root of a checkout.

```bash
python tests/compare_zeek_baselines.py <path-to-FoxIO-ja4-checkout>
```

`docs/specs/foxio/zeek.md` records the whole reading.

## What this issue does not ask

It asks for no change to the first value, and for no change to any other method. #198
found no other open decision.

[...]

PR #249, opened 2026-08-08
Emit the JA4SSH window a connection holds open at the end of a capture

Crank-Git:
## What

JA4SSH emits the window a connection holds open when the capture ends, not only when a
FIN+ACK packet closes the connection. One rule for every capture.

`close_open_windows()` reads every connection of the state table and emits its open
window. The caller runs it when the packet source ends. `tests/conformance_index.py`
runs it after the last packet of a capture. `ja4plus/cli.py` runs it after the file
reader of `analyze`, and after `live` stops.

Refs #214.

## Why

**The specification cannot settle this, and the reference values can.**
`docs/specs/foxio/JA4SSH.md` R11 records the gap. #199 read the deleted
`technical_details/JA4SSH.md`, the exact blob `rust/ja4/src/ssh.rs:283` cites, and it
names no FIN packet, no connection that closes, no end of a capture and no trailing
window.

**The references split two against two.**

| Reference | What closes the last window |
|---|---|
| FoxIO Python | A FIN+ACK packet, on a stream whose index is not `0` (`python/ja4.py:554-556`) |
| FoxIO Wireshark | A FIN+ACK packet on port 22 (`wireshark/source/packet-ja4.c:1399-1404`) |
| FoxIO Rust | The end of the capture (`rust/ja4/src/ssh.rs:45-55`) |
| FoxIO Zeek | The removal of the connection state (`zeek/ja4ssh/main.zeek:160-164`) |

`python/ja4.py:610` runs no end-of-capture step at all. The line
`#finalize_ja4ssh() if 'ja4ssh' in output_types else None` is commented out.

**Two FoxIO references agree on the value, and this repository already held the proof.**
`ssh2.pcapng` carries 452 TCP packets on port 22 and zero FIN+ACK, so no FIN+ACK rule
fires on it. `tests/foxio_vectors/rust_expected/ja4__insta@ssh2.pcapng.snap:215-217` and
the Zeek baseline both hold `c36s52_c42s76_c51s2`, and the `ja4plus` counters already
held that value.

The user decided it on 2026-08-08. The decision is reversible.

## The measurable target

**`ssh2.pcapng` now produces `c36s52_c42s76_c51s2` as its second value.** Held by
`test_the_long_capture_produces_the_rust_and_zeek_trailing_window`.

## Every value that moved

The measurement replayed all 38 committed captures through
`tests/conformance_index.py::index_produced` before and after.

| Vector | Method | Value the end of the capture adds |
|---|---|---|
| `ssh2.pcapng` | JA4SSH | `c36s52_c42s76_c51s2` |
| `ssh.pcapng` | JA4SSH | `c36s52_c42s76_c0s0` |
| `ssh-scp-1050.pcap` | JA4SSH | `c0s1460_c0s53_c6s0` |
| `ssh2-malformed.pcap` | JA4SSH | `c16s23_c7s6_c3s4` |
| `ssh2-moloch-crash.pcap` | JA4SSH | `c16s23_c7s6_c3s4` |
| `tcpdump-geneve.pcap` | JA4SSH | `c144s48_c10s11_c6s4` |

**Six comparisons moved. Every one is a JA4SSH comparison. No value of another method
moved.**

**`gre-sample.pcap`, `sshv1.pcap` and `v6.pcap` did not move.** Each one carries a
FIN+ACK packet, so each already produced a trailing value. Changelog round 15 recorded
that, and the behaviour is now uniform.

## #97 stays declined

A window that holds no SSH packet describes no traffic. The new rule reaches the guard
at `ja4plus/fingerprinters/ja4ssh.py:368-369`, so the value `c0s0` does not return.
Three cases hold it:

- `test_the_end_of_the_capture_emits_nothing_for_a_connection_of_bare_acks_alone`
- `test_the_end_of_the_capture_emits_nothing_when_the_open_window_holds_no_ssh_packet`
- `test_the_long_capture_produces_no_declined_empty_window`

The FoxIO Python second value on `ssh2.pcapng` is `c36s36_c0s0_c2s0`. It is not a
trailing window. `.claude/rules/external-apis.md` gives `python/test/testdata/`
precedence over a Rust snapshot, and here the Python value is the value #97 already
declined, so the decline outranks the precedence rule and the Rust snapshot decides.

## The register

**116 keys before, 120 after. 116 `xfailed` before, 120 `xfailed` after. The two are
equal on both sides.**

- Removed `ssh2.pcapng/JA4SSH`. Its occurrence keys now equal the reference, so the
  strict marker reported `XPASS`.
- Added five entries under #214, one per vector, for the trailing window the FoxIO
  Python reference does not emit.
- `tests/foxio_deviation_owners.json` gains #214.

## How tested

- Unit suite: `1305 passed, 8 xfailed`.
- Conformance suite: `1480 passed, 143 skipped, 120 xfailed`.
- `ruff check` clean, `ruff format --check` clean, `mypy ja4plus/` clean.
- Coverage 82%, unchanged from the base.

**Proof by revert.** Restore the body of `close_open_windows` to `return []` and **13
cases fail**: 10 of the unit suite and 3 of the conformance suite. A revert of the whole
method instead breaks `tests/conformance_index.py`, which calls it on every
fingerprinter, so that measurement reports 1556 failures and states nothing about the
behaviour.

## The state table

`close_open_windows` emits the open window and evicts nothing, so
`JA4SSHFingerprinter.connections` holds the same keys after the call.
`test_the_end_of_the_capture_holds_no_connection_longer` proves it. That table is one of
the six unbounded state tables #179 records, and Epic 3 owns its bound. **This change
holds no entry longer.**

## Parity

The port exposes no equivalent method, so rule 2 names no interface to adopt.
`docs/specs/spec.md` gains two divergence register rows: one for the behaviour, one for
the `close_open_windows` name.

## Notes

- The Changelog round is **68**, assigned by the project manager at the merge gate. Round 66 is #225 and round 67 is #226, and both land ahead of this branch.
- `ja4plus/cli.py` lost a bare `except Exception` around the call. `CLAUDE.md` states
  that a fingerprinter catches the parse errors it expects and does not catch bare
  `Exception`.

Verified against: https://github.com/FoxIO-LLC/ja4/blob/main/rust/ja4/src/ssh.rs
(retrieved 2026-08-08, commit `27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8`)
Verified against: https://github.com/FoxIO-LLC/ja4/blob/main/technical_details/JA4SSH.png
(retrieved 2026-08-08, commit `27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8`)


Crank-Git:

A Sonnet child reviewed the diff of `ja4plus/` and `tests/` for correctness. It reported
two findings and cleared six checks. Both findings are addressed.

### Cleared

- `close_open_windows` appends and returns the same entry. `_close_window` has two exits:
  it returns `None` before any append, or it appends and returns a string.
  `_generate_ja4ssh` cannot return `None` there, because the caller already read
  `self.connections[conn_key]`.
- `list(self.connections)` is safe. `_close_window` resets counters on the existing entry
  and adds no key and removes none.
- The empty-window guard sits inside `_close_window`, so it applies to the new rule
  unchanged. #97 stays declined.
- The method is idempotent. `_close_window` clears the counters it emits.
- The `partition("-")` of the connection key is correct for IPv6. A JA4SSH key is
  `<address>:<port>-<address>:<port>`, and an IPv6 address holds a colon and no hyphen.
- `Processor.close_open_windows` matches the documented contract of the class, which
  swallows a per-fingerprinter error at `DEBUG`.

### Finding 1, major — `ja4plus/collector.py` holds its own capture loop

`collector.py:195-217` runs its own `sniff` loop with its own fingerprinter map, and it
runs no close. The same SSH traffic through `ja4plus live` emits a trailing window and
through `collector.py` emits none.

**Not fixed here, and recorded on #191 instead.** The output of that module is
packet-shaped: `collector.py:140-181` builds each record from the packet that produced
the value. A trailing window has no packet, so a record needs a shape the module does not
hold, and that is a design decision this issue carries no license to make. #191 removes
the module. https://github.com/Crank-Git/ja4plus/issues/191#issuecomment-5225423531

### Finding 2, minor — the trailing close is unreachable on the error path

A read error in `cmd_analyze`, or a capture error in `cmd_live`, exits with the status 1
before the close runs.

**This is the intended contract, and it is now written down.** `_close_open_windows`
states it: a partial window that follows an error describes a connection the command
failed to read.

### One gap the review found that the child did not name

`examples/pcap_analysis.py` and `examples/tcpdump_workflows.py` each replay a whole
capture and then report `get_fingerprints()`. Both reached the end of a capture and
emitted no trailing window, so both disagreed with `ja4plus analyze` on the same file.
Fixed in `3d8a9a7`.

### Not fixed, and not mine

`ruff check examples/` reports three `F541` errors in `examples/demo_fingerprints.py` and
`examples/monitoring_daemon.py`. They pre-date this branch, they sit in files this branch
does not touch, and `examples/` is outside the lint gate `CLAUDE.md` names
(`ruff check ja4plus/ tests/`).

Crank-Git:
## Self-review, test quality and documentation lens

A second Sonnet child reviewed test quality and documentation accuracy. It verified the
measurements independently and reported one nit. The nit is fixed.

### Verified independently by the reviewer

- **The revert measurement.** It applied the same `return []` edit, ran the suite, and
  measured **exactly 13 failures**: 10 of the unit suite and 3 of the conformance suite.
  It restored the file and confirmed the suite green.
- **The value.** It ran the fingerprinter on `tests/foxio_vectors/ssh2.pcapng` and read
  the trailing value `c36s52_c42s76_c51s2`.
- **The register.** 116 keys at `origin/batch/245-fingerprint-decisions`, 120 here. The
  diff moves exactly four keys, and no key of another method changed.
- **The line references.** `ja4plus/fingerprinters/ja4ssh.py:328-351` is the method, and
  `ja4plus/fingerprinters/ja4ssh.py:368-369` is the guard.

It could not verify the FoxIO `file:line` citations, because those repositories are not
vendored here. Those citations come from `docs/specs/foxio/JA4SSH.md`, which #199 wrote
against the pinned commit.

### Nit, fixed — a new assertion that could not fail

`test_processor_close_open_windows_returns_nothing_without_state` asserted that a
processor with no state returns an empty list. **That assertion passes whatever the
method does.** #205 swept the suite for assertions that cannot fail, and this one was new.

The case now builds a connection of bare ACKs alone and asserts that the processor
reports no window for it. Fixed in `3f2b2ac`.

### A stronger proof of the #97 decline

The revert of `close_open_windows` does not exercise the decline, because a negative
assertion passes when the feature is absent. **The guard is proven by its own removal
instead.** Replace the guard condition of `_close_window` with `False` and **11 cases
fail, 5 of them new cases of this issue**:

- `test_the_end_of_the_capture_emits_nothing_for_a_connection_of_bare_acks_alone`
- `test_the_end_of_the_capture_emits_nothing_when_the_open_window_holds_no_ssh_packet`
- `test_the_end_of_the_capture_emits_nothing_after_the_connection_closes`
- `test_a_second_call_at_the_end_of_the_capture_emits_nothing`
- `test_processor_close_open_windows_declines_a_window_with_no_ssh_packet`

The other six are pre-existing cases that hold the same guard, which shows the guard is
one guard and the new rule reaches it. `docs/specs/foxio/JA4SSH.md` and
`docs/specs/spec.md` now record this measurement.

### Rejected — the density of the `docs/specs/spec.md` decision-log row

The reviewer flagged the `| TBD |` row for sentence length and passive voice under
`.claude/rules/ste.md`. **The row keeps its format**, because all 65 rows of that table
carry one narrative cell and a single row in a different shape reads as a defect. I split
the one 47-word sentence the reviewer quoted. The table format is a question for the
person who owns the Changelog, not for this issue.

### Coverage

82% before and after. The exact figure moves from 81.75% to 81.88%, so it did not fall.

PR #257, opened 2026-08-08
Mark the sixteen QUIC marker entries Changelog round 66 settled

Crank-Git:
Builds #251. Batch 19, integration branch `batch/252-above-the-epics`. Changelog round 69.

## What

`tests/foxio_deviations.json` held 136 keys, and 24 carried no `decided` marker. Sixteen
of the 24 record the QUIC marker decision the user made on 2026-08-08, which Changelog
round 66 settled. This change marks those 16 and states why the other eight stay open.

## Why

An unmarked entry states that nobody read it. That is the register's only job. A reader of
`dev` counted 24 open entries where 8 are open. #193 repaired the same fault on 43 entries
under #129, and batch 18 put it back.

## How

- The 16 entries owned by #225 carry `"decided": true`. Each cause names Changelog round
  66, states that the decision is reversible, and cites the published reference value:
  `wireshark/source/packet-ja4.c:1441` and `:1447` write the marker, and
  `wireshark/test/testdata/` publishes it on 18 of 88 values.
- The six entries owned by #34 and the two owned by #215 stay unmarked. Each cause states
  why it is open. **No entry is marked to make a count come out even.**
- `docs/specs/spec.md` gains Changelog round 69.
- No file under `ja4plus/` changed and no fingerprint moved.

## Every premise re-measured

| Premise | Result |
|---|---|
| 136 keys, 24 undecided | Confirmed. |
| #225 owns 16, #34 owns 6, #215 owns 2 | Confirmed. All 16 of #225 were undecided. |
| `python/ja4.py:339` deletes both JA4L keys | **Corrected.** Line 339 is the guard `if 'ja4l' not in output_types:` and line 340 is the `delete_keys` call. Round 56 and round 66 cite 339 alone, and the five causes now cite `python/ja4.py:339-340`. |
| The three files publish no JA4L key | Confirmed. `grep -c '"JA4L-'` reports 0, 0, 0 and 19 for `ssh2.pcapng.json`. |
| `packet-ja4.c:1441` and `:1447` write the marker | Confirmed by `grep -n '%d_%d_quic'`. |
| 18 of 88 published Wireshark values carry the marker | Confirmed by a walk of every JSON file of `wireshark/test/testdata/`. |
| `ssh2.pcapng` gives 11 `JA4L-S` against the reference 10 | Confirmed. The extra value is `JA4L-S=6252_58`, and packets 369 and 372 carry the same SYN-ACK. |
[...]

PR #254, opened 2026-08-08
Remove the deprecated ja4plus/collector.py (#191)

Crank-Git:
Issue #191. This pull request removes `ja4plus/collector.py`.

This branch rebases onto `d147592`, the head of `batch/252-above-the-epics` after #256
landed. Every count below is measured against that commit.

## What changed

- `ja4plus/collector.py` is deleted. The module carried its own removal notice for
  v0.4.0, and the project publishes 0.6.0.
- `docs/specs/spec.md` gains Changelog round 71.
- `docs/specs/spec.md` moves the `FR-typed-api-11` traceability row from #48 to #191,
  because #48 closed as absorbed by #191 on 2026-08-08.

No file under `ja4plus/fingerprinters/` and no file under `ja4plus/utils/` changed.

## Why

The mutation sweep of #172 applied 12 mutations to the module and no mutation failed one
case. The module holds 119 statements and no case reaches one of them. `ja4plus/cli.py`
carries the command-line program.

The removal also answers the divergence #214 recorded on the issue. `collector.py:195-217`
held its own `sniff` loop and ran no `close_open_windows()` call, so a live SSH connection
read through the module emitted no trailing window.

## The measurement that decides whether the removal is safe

`git grep -i collector d147592` over every tracked file of the base commit reports 53
matching lines in 12 files outside the removed module. **No line imports it.** The
command names the base commit, because Changelog round 71 adds a match to the count it
reports.

- `ja4plus/__init__.py` re-exports it nowhere.
- `pyproject.toml` declares one console script, `ja4plus = "ja4plus.cli:main"`.
- The six files under `examples/` name it nowhere.
- The three files under `.github/workflows/` name it nowhere.
- `README.md` names it nowhere.

Every remaining match holds the word for another thing: the `_CaseCollector` class of
`tests/generate_foxio_baseline.py`, the QUIC fragment collector of
`tests/test_quic_server_initial.py`,
`tests/test_comprehensive.py::TestJA4TSComprehensive::test_collector_uses_base`, which
reads `JA4TSFingerprinter`, the Go garbage collector of
`docs/specs/features/03-concurrency-safety.md`, the remote collector of
`docs/specs/features/06-live-capture.md`, and the `processor` row of the `## Terms`
table, which rejects `collector` as a synonym. The rest sit in `docs/mutation_sweep.md`,
`docs/implementation_notes.md`, `docs/specs/features/02-correctness-audit.md`,
[...]

Issue #251, opened 2026-08-08
Mark the sixteen QUIC marker entries the user decided


Crank-Git:
Found by the project manager at the batch 18 gate on 2026-08-08, after the merge.

## The defect

**`tests/foxio_deviations.json` holds 136 keys and 24 carry no `decided` marker. Sixteen of those 24 record a decision the user made.**

| Owner | Entries | State |
|---|---|---|
| #225 | **16** | The user decided the QUIC marker on 2026-08-08. **No marker** |
| #34 | 6 | Genuinely open. #200 measured that five are artifacts of `python/ja4.py:339` |
| #215 | 2 | Genuinely open. D2 and D4 await the build |

The 16 causes are accurate. One reads:

> `ja4plus` writes the quic marker as the third part on a QUIC connection, and the FoxIO Python reference writes two parts. The two timing parts agree, so the marker is the whole difference. #225 holds the decision.

**The cause names the decision. The entry does not carry the marker that says a person made it.**

## Why this matters, and why it is the same fault twice

**This is exactly what #193 spent an issue repairing.** That issue found 43 entries carrying a real Changelog decision and no `decided` marker, so a reader counted 55 open when 12 were. The register fell to 6 undecided and the count became meaningful.

**Batch 18 put it straight back.** A reader of `dev` today counts 24 open entries, and 16 of them were settled by a user decision recorded in Changelog round 66.

**An unmarked entry states that nobody has read it.** That is the register's only job.

## What to build

- [ ] The 16 entries owned by #225 carry `"decided": true` and a cause naming **Changelog round 66**, in the form #193 used.
- [ ] The 6 entries owned by #34 and the 2 owned by #215 **stay unmarked**, and the reason each is open is stated. Do not mark an entry to make a count come out even; #193 refused to do that and its refusal was correct.
- [ ] The undecided count falls from 24 to **8**.
- [ ] The key count does not change: 136 keys against 136 `xfailed`.
- [ ] No file under `ja4plus/` changes and no fingerprint moves.

## What not to build

Do not run `tests/generate_foxio_baseline.py`. #161 made it merge rather than rewrite, and #216 recorded that it runs `test_spec_validation.py` alone and would drop the JA4T and JA4X entries. Edit by hand.

## A question for the next batch, not this issue

[...]

Issue #242, opened 2026-08-08
Settle which address pair names a stream of a tunneled capture


Crank-Git:
Raised by #216.

## The finding

`gre-erspan-vxlan.pcap` carries ERSPAN and VXLAN, so one packet holds two address pairs.
**The two FoxIO references name the stream by different pairs.**

| Reference | Addresses it names |
|---|---|
| `python/test/testdata/gre-erspan-vxlan.pcap.json` | `100.20.9.2` and `100.20.9.1` |
| `rust/ja4/src/snapshots/ja4__insta@gre-erspan-vxlan.pcap.snap` | `10.16.27.12` and `10.16.27.131` |

`ja4plus` reads the outer pair, which is the pair the FoxIO Python file names.

Measured on 2026-08-08. `ja4plus` produces one JA4T entry for the capture:

```
{'fingerprint': '8192_0_0_0', 'src': '100.20.9.2', 'dst': '100.20.9.1', 'srcport': 65174, 'dstport': 80}
```

## Why it blocks work

`tests/test_foxio_rust_parity.py` matches a stream by its address pair. A case built from
the Rust snapshot of this capture compares against a stream identity `ja4plus` never
produces, so the failure reports the address layer rather than the value.

**#216 therefore declined to commit `ja4__insta@gre-erspan-vxlan.pcap.snap`.** That
snapshot holds `ja4t: 8192__0_0`, and it is the one local case that measures D1 of
`docs/specs/foxio/JA4T.md`: `ja4plus` writes `8192_0_0_0` for an empty option list, and no
FoxIO reference writes that form.

## What to decide

1. Which address pair names a stream of a tunneled capture?
2. Does the parity harness match such a stream by port alone, or by a second identity?

`.claude/rules/conformance.md` states that a fingerprint exists so that one tool output
can be compared against another tool output. A stream identity that two references
disagree on makes that comparison unreachable for the whole capture.

`tests/generate_foxio_baseline.py` already records a related reading: #101 owns the
mirrored capture, where `ja4plus` groups the two directions as two connections.

Untriaged. #215 owns the D1 decision itself.

Crank-Git:
## Plan — batch 19, integration branch `batch/252-above-the-epics`

**Changelog round 72.** It is assigned. Do not choose one, and never write `TBD`.

### The project manager's decision on question 1

**`ja4plus` keeps the outer address pair, and no fingerprinter changes.**

[...]

PR #260, opened 2026-08-08
Name the stream of a tunneled capture by a recorded alias

Crank-Git:
Settles issue #242. Batch 19, integration branch `batch/252-above-the-epics`.

## What this changes

`gre-erspan-vxlan.pcap` carries GRE, ERSPAN and VXLAN, so one packet holds three address
layers. The two FoxIO references name its one stream by different address pairs, so #216
declined to commit the Rust snapshot of it. This pull request settles the identity and
commits the snapshot.

No file under `ja4plus/` changes. No fingerprint moves.

## Question 1 — which address pair names the stream

**The outer pair names the stream, and no fingerprinter changes.**

Measured on 2026-08-08 with `scapy` and `tests.conformance_index.index_produced`:

| Layer | Address pair |
|---|---|
| Outer, under GRE and ERSPAN | `100.20.9.2` and `100.20.9.1` |
| Middle, under VXLAN | `172.16.27.131` and `172.16.27.121` |
| Inner, the session | `10.16.27.12` and `10.16.27.131` |

The ports `65174` and `80` come from the inner layer. Both references name them.

```
{'fingerprint': '8192_0_0_0', 'src': '100.20.9.2', 'dst': '100.20.9.1', 'srcport': 65174, 'dstport': 80}
```

`ja4plus` reports the outer pair with the inner ports, which is the pair
`python/test/testdata/gre-erspan-vxlan.pcap.json` reports. The Rust snapshot reports the
inner pair. Rule 1 of `CLAUDE.md` states that the vectors decide the exact bytes where
intent runs out, and here the two references disagree with each other, so no vector
decides. The ALPN disputed region of #162 settled this shape: a move is forbidden where
the two implementations disagree. The present behaviour stands.

## Question 2 — how the harness names such a stream

**By a recorded address-pair alias, which is the narrowest of the three candidates.**

| Identity | What it stops measuring |
|---|---|
| The port pair alone | Every address. `ssh2.pcapng` holds 19 JA4T streams, and a port pair matches any stream that shares it. |
| The value alone, with no stream | The whole stream attribution. A value that moves to another stream still passes. |
| The recorded address pair | Nothing. A wrong address on either side finds no entry, so the case reports `ja4plus=<none>` and fails. |

`SNAPSHOT_ADDRESS_ALIASES` in `tests/test_foxio_rust_parity.py` holds the identity the
snapshot names, and it returns the identity `ja4plus` produces. `read_rust_snapshot`
reads the map for every stream of every snapshot.

## Proof that a wrong value still fails

Four mutations, each run against `tests/test_foxio_rust_parity.py -m spec_validation`.

| Mutation | Result |
|---|---|
| The snapshot value becomes `8192_0_0_0`, the `ja4plus` form | 2 failed. The strict `xfail` case reports `XPASS`. |
| The snapshot value becomes `65535__0_0` | 1 failed. `test_the_stream_measures_d1_and_the_two_forms_differ` names the value. |
| The alias outer address becomes `203.0.113.9` | 3 failed. |
| The alias entry is deleted | 3 failed. `test_the_alias_maps_the_snapshot_pair_to_the_pair_ja4plus_produces` and the count case report. |

`test_a_stream_the_alias_does_not_name_finds_no_entry` carries six parameter sets: a
wrong source address, a wrong destination address, two wrong addresses, a wrong source
port, a wrong destination port, and the outer pair, which the map holds as a value and
never as a key.

## What lands

- `tests/foxio_vectors/rust_expected/ja4__insta@gre-erspan-vxlan.pcap.snap`, copied
  without change. `sha256 a1a6ebfb95c3a63e98473492c3e71102b89414f065d306ee1c2f3a2c84716acd`.
- `tests/test_foxio_rust_parity.py`: `SNAPSHOT_ADDRESS_ALIASES`, `TCP_ONLY_CAPTURES`,
  `SNAPSHOT_CAPTURES`, and `TestTheStreamIdentityOfTheTunneledCapture` with seven checks.
  The reader guard moves from a check for the `JA4` name to a check for any value the
  module compares, because this snapshot writes `ja4t` alone.
- `tests/foxio_deviations.json`: **one new key**, `gre-erspan-vxlan.pcap/0:65174/JA4T.1`,
  which names #215. No other entry changes. #251 owns this file in the same batch.
- `tests/download_test_vectors.py` and `tests/foxio_vectors/NOTICE`, so a vector refresh
  reproduces the snapshot.
- `docs/specs/foxio/JA4T.md`, `docs/implementation_notes.md`, and `docs/specs/spec.md`
  round 72.

**D1 of `docs/specs/foxio/JA4T.md` now reaches a conformance case for the first time.**
`ja4plus` writes `8192_0_0_0` for an empty option list, and the snapshot holds
`8192__0_0`. #215 owns the decision, and it is scheduled for batch 20.

## How this was tested

The five gates ran in a worktree virtual environment.

```
.venv/bin/pytest tests/ -m "not spec_validation"   1358 passed, 8 xfailed, 93 subtests passed
.venv/bin/ruff format --check ja4plus/ tests/      122 files already formatted
.venv/bin/mypy ja4plus/                            Success: no issues found in 26 source files
```

The branch is rebased on `batch/252-above-the-epics` after #254 and #257 sub-merged.
#257 added `TestTheOpenRegisterEntries`, which holds the count of register entries no
Changelog round settled. The new entry is undecided, because #215 decides the D1 form,
so the open count rises from eight to nine and `TCP_OPTION_KEYS` gains the key.

| Measure | Before | After |
|---|---|---|
| Keys in `tests/foxio_deviations.json` | 136 | 137 |
| `xfailed` cases in the conformance suite | 136 | 137 |
| Passing cases in the conformance suite | 1464 | 1477 |
| Undecided register entries | 8 | 9 |
| JA4T values the local snapshots hold | 38 | 39 |
| Files under `tests/foxio_vectors/rust_expected/` | 10 | 11 |
| Coverage of `ja4plus/`, `--cov=ja4plus` | 85%, 524 of 3467 missed | 85%, 524 of 3467 missed |

Verified against: https://github.com/FoxIO-LLC/ja4/tree/27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8/rust/ja4/src/snapshots (retrieved 2026-08-08)
Verified against: https://github.com/FoxIO-LLC/ja4/tree/27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8/python/test/testdata (retrieved 2026-08-08)


Crank-Git:
## Self-review

Two Sonnet review children were spawned, for correctness and for documentation. Neither
returned. The review below is self-run, and this note records that.

### Correctness

1. **A negative check that reports a pass on nothing. Fixed in c6baedd.**
   `test_a_stream_the_alias_does_not_name_finds_no_entry` asserts that six wrong
   identities miss `SNAPSHOT_ADDRESS_ALIASES`. An empty map passes all six, so the check
   could not fail. It now asserts that the map holds `TUNNEL_INNER_IDENTITY` first.

2. **Alias collision with another snapshot. No finding.**
   `test_the_alias_moves_no_other_capture` reads the other ten snapshots and checks that
   none of them produces a stream identity the map returns. It passes.

3. **The reader guard. No finding.** The guard moved from `any("JA4" in stream.values)` to
   `any(stream.values or stream.certs)`. A snapshot that names no `ja4`, `ja4s`, `ja4t` or
   `ja4x` field still produces empty `values` and empty `certs` for every stream, so the
   assertion still fires. `gre-erspan-vxlan.pcap` writes `ja4t` alone, which the old guard
   rejected.

4. **`DIVERGENT_CAPTURES` against `SNAPSHOT_CAPTURES`. No finding.** Three uses of
   `DIVERGENT_CAPTURES` remain, at lines 538, 543 and 577. All three sit in
   `TestTheRustImplementationHoldsWhatJa4plusProduces`, which measures the #138 handshake
   gaps. The new capture holds no handshake value, so it belongs to neither.
   `tcp_captures` and `certificate_captures` moved to `SNAPSHOT_CAPTURES`.

5. **The hard-coded counts. No finding.** 39 values across seven captures, 11 files under
   `tests/foxio_vectors/rust_expected/`, 137 register keys, 137 `xfailed` cases. Each one
   is measured, and the suite holds each as a check.

### Documentation

6. **Stale claims in `docs/specs/foxio/JA4T.md`. Fixed in 53a14b1 and earlier.**
   The page said the repository does not hold the snapshot, that D1 reaches no local
   value, that six snapshots hold 38 values, and that ten files sit under
   `rust_expected/`. Each claim is corrected. The section is renamed from "The snapshot
   the comparison cannot reach" to "The snapshot the comparison now reaches".

7. **`tests/foxio_vectors/NOTICE`. No finding.** The committed file is the exact render of
   `NOTICE_TEMPLATE` in `tests/download_test_vectors.py` with the module's own constants.
   Verified by rendering and diffing.

8. **Controlled vocabulary.** The new prose uses one word the `## Terms` table did not
   hold, `alias`. The table gains one row. No word in the "Do not use" column appears in
   new prose that the repository does not already use.

### The shared file

`tests/foxio_deviations.json` gains exactly one key,
`gre-erspan-vxlan.pcap/0:65174/JA4T.1`. No other entry changes. #251 owns this file in
the same batch, and the project manager resolves the overlap at the sub-merge gate.

Issue #259, opened 2026-08-08
Batch 20: reject a malformed JA4SSH part


Crank-Git:
Batch 20 of the `issue-flow` loop. Session 10, 2026-08-08.

**This batch runs beside batch 19 (#252), which is still open.** Two live integration
branches cannot see each other. Session 9 recorded a real cross-batch collision, so this
batch takes only members that touch neither `tests/foxio_deviations.json` nor any
fingerprint. Everything that moves the register waits for batch 19 to reach `dev`.

## Members

- [x] #186 Reject a JA4SSH part that carries a second separator
- [x] #190 Drop a JA4H stream whose complete header block produces nothing
- [x] #262 Return an error dictionary for a bytes JA4SSH fingerprint

## Held back, and why

| Issue | Why it is not a member yet |
|---|---|
| #215 | **Depends on #242**, which is a member of batch 19. Without the snapshot #242 commits, D1 lands measured by nothing |
| #228 | Decides the JA4X empty-list form and the JA4X raw form. It can move a fingerprint, so it can move the register, which #251 is repairing in batch 19 |
| #231 | Eleven JA4D and JA4D6 readings. Same reason |

**These three schedule as batch 21, off `dev`, once batch 19 merges.**

## The Changelog round

Rounds 1 to 68 are taken on `dev`. Batch 19 holds 69 to 72.

| Member | Round |
|---|---|
| #186 | 73 |
| #190 | 74 |
| #262 | 75 |

**Batch 19 merges into `dev` first.** If this batch reaches the gate before batch 19, the
project manager renumbers rather than merging out of order.

## Integration branch

`batch/259-interpret-fingerprint`, forked from `dev` at `07f5a7a`. The member opens a draft
pull request into it, carrying `[skip ci]` in every head commit.

Crank-Git:
Gate note for batch 20, from the batch 19 cross-batch check.

Batch 19 moves `tests/foxio_deviations.json` from 136 keys to **137**, because #242 committed the snapshot that measures D1. Batch 20 was built on `dev` at `07f5a7a`, where the file holds 136.

**Changelog round 73 reads "136 keys before and after".** That statement is a measurement of #186's own change against its own base, so it is accurate, and it is not the present-tense claim that session 9 caught in five Epic 10 pages. **It still reads as stale once batch 19 reaches `dev`.**

[...]

PR #261, opened 2026-08-08
fix(ja4ssh): reject a JA4SSH part that carries a second separator

Crank-Git:
Closes nothing on its own. It builds issue #186, and batch #259 closes the issue.

## What is wrong

`JA4SSHFingerprinter.interpret_fingerprint` split a JA4SSH part on `s` and read element
1 as the server value. It never counted the elements. A part that carried a second
separator therefore passed, and the method reported a number the fingerprint does not
hold.

## What the measurement found

**The issue body named the packet size part, and all three parts held the same unchecked
split.** Measured on the base commit `07f5a7a`:

```
'c36s36s99_c50s50_c70s30' -> {'packet_sizes': {'client': 36, 'server': 36}, ...}
'c36s36_c50s50s1_c70s30' -> {'packet_sizes': {'client': 36, 'server': 36}, ...}
'c36s36_c50s50_c70s30s7' -> {'packet_sizes': {'client': 36, 'server': 36}, ...}
```

**The three readings a comment on #186 records all reproduce, and the measurement adds
four more.** `int()` accepts a space, a sign and a Unicode digit on the server side as
well as on the client side, so `c36s 36`, `c36s-5` and `c36s٦` pass. `c36s+36` passes on
a plus sign. `c36s36_c50s50_c70s30\n` and `c36s36_c50s50_c70s30 ` pass, because the last
part reaches `int()` whole.

**The issue body's line number is stale.** It names
`ja4plus/fingerprinters/ja4ssh.py:486`, and the line is 514. Changelog round 45 records
the same correction for #182.

## What this pull request changes

One guard replaces the split, the `int()` call and #182's prefix guard together.

```python
JA4SSH_PART = re.compile(r"c(\d+)s(\d+)", re.ASCII)
```

The method calls `fullmatch` on each part, and a part that does not match returns
`{"error": "Invalid JA4SSH format"}`. That is the error the method already returns for a
wrong part count, so no second error shape appears. The regular expression subsumes
#182's prefix guard, and #182's five cases pass unchanged.

**`fullmatch` is the comparison and an anchored `match` is not.** Python `$` matches
before a trailing line feed, so `c70s30\n` passes `^c(\d+)s(\d+)$`.

**One error message changes.** `c36_c50s50_c70s30` returned
`Failed to interpret: list index out of range`, and it now returns
`{"error": "Invalid JA4SSH format"}`.

**One reading stays open and this pull request changes nothing for it.** A leading zero
passes, so `c0036s36_c50s50_c70s30` reports a client packet size of 36. The regular
expression the issue comment states accepts it, no FoxIO reference states a rule for it,
and no emitter of this project writes one.

## How it is tested

Thirteen new cases in `tests/test_ja4ssh_interpret.py`. They were written first and they
failed first: `13 failed, 16 passed`. The self-review added a fourteenth.

**Four mutations prove the parts of the guard bite.**

|---|---|
| `match` in place of `fullmatch` | 5 |
| The pattern without `re.ASCII` | 2 |
| The pre-#186 parser restored | 13 |
| `ValueError` dropped from the handler | 1 |

## Local gates

| Gate | Result |
|---|---|
| `pytest tests/ -m "not spec_validation"` | 1349 passed, 8 xfailed, 93 subtests passed |
| `pytest tests/ -m spec_validation` | 1464 passed, 143 skipped, 136 xfailed |
| `ruff format --check ja4plus/ tests/` | 122 files already formatted |
| `--cov=ja4plus` | 82% total, 98% on `ja4plus/fingerprinters/ja4ssh.py` |

Coverage reads 82% and 98% on the base commit as well, so it does not fall.

## No fingerprint moves

`interpret_fingerprint` reads a fingerprint and it produces none, so no FoxIO vector
covers it. The conformance suite reports 136 `xfailed` before and after, and
`tests/foxio_deviations.json` holds 136 keys before and after. No register entry changes,
and batch 19 owns that file.

## Self-review

Two lenses ran in parallel on Sonnet, each read only this worktree, and each posted its
findings as a comment. The correctness lens reports no finding. The robustness lens
reports two.

1. **Repaired here.** The comment on the handler stated that the part guard rejects a
   malformed string before `IndexError` and `ValueError` can reach it. CPython converts
   4300 digits at most, so a part of 5000 digits matches the pattern and `int()` raises
   `ValueError`. The comment now states that reading, and the fourteenth case pins it.
2. **Filed as #262.** A `bytes` fingerprint raises `TypeError` from
   `fingerprint.split("_")` at `ja4ssh.py:503`. The same input raises the same error at
   the base commit `07f5a7a`, so it sits outside #186.

Changelog round 73 in `docs/specs/spec.md` records the whole measurement.


Crank-Git:

I ran the correctness lens and the robustness lens in parallel, each on Sonnet, and each read only this worktree. The correctness lens reports no finding.

- No string the old parser accepted and answered correctly is rejected by the new guard.
- Every value this module emits matches the pattern. `ja4ssh.py:413`, `:418` and `:423` build a part as `f"c{n}s{n}"`, and `n` is a mode or a count, so it is never negative and it carries no leading zero.
- The unpack order matches the loop order and the order the old code read.
- The 13 new cases each fail against the pre-#186 parser, 5 against `match`, and 2 against a pattern without `re.ASCII`.

Crank-Git:
## Self-review, robustness lens

Two findings. One is mine and I repair it. One is older than this branch and I file it.

**Finding 1, worth fixing, mine.** `ja4plus/fingerprinters/ja4ssh.py:560-566`. The comment I wrote states that the part guard rejects a malformed string before `IndexError` and `ValueError` can reach the handler. `ValueError` still reaches it. CPython limits an integer conversion to 4300 digits, so a part of 5000 digits matches the pattern and `int()` raises. The method still returns an error dictionary, so the behaviour is right and the comment is wrong. I repair the comment and I add the case.

**Finding 2, note, older than this branch.** `ja4plus/fingerprinters/ja4ssh.py:503`. A `bytes` fingerprint raises `TypeError` from `fingerprint.split("_")`, and the handler does not name `TypeError`. The same input raises the same error at the base commit `07f5a7a`, so this branch neither adds it nor removes it. It sits outside #186, so I file it rather than fix it.

**No finding on the regular expression.** It holds no nested quantifier and the literal `s` separates the two groups, so it cannot backtrack catastrophically. Measured: a 200000-digit match runs in 0.00045 s and a 100001-character failure runs in 0.00072 s.

PR #263, opened 2026-08-08
Drop a JA4H stream whose complete header block produces nothing

Crank-Git:
Refs #190. Batch 20, integration branch `batch/259-interpret-fingerprint`.

## What

`JA4HFingerprinter.process_packet` removed a stream after it recorded a value, and after
a buffer that starts no HTTP request. It removed no stream when the buffer held a
complete header block and the parse produced nothing. The request line `GET / HTTP/11`
reaches that path, because the version token names no HTTP version.

The fingerprinter now removes such a stream. `_drop_a_stream_whose_base_holds` holds the
removal rule, and both unusable-buffer paths read it.

## Why

Every later packet on the connection read the whole buffer again, and the buffer grows to
`max_stream_bytes`, which is 1048576.

Measured on one refused request followed by 300 packets of 1400 bytes:

```
before   refused       35.5 ms  stream bytes held   420055  values 0
before   accepted      21.3 ms  stream bytes held        0  values 1
after    refused       20.1 ms  stream bytes held     1400  values 0
after    accepted      20.4 ms  stream bytes held        0  values 1
```

## The premise of the issue is wrong for a byte that precedes the buffer

The issue states that a complete header block admits no later byte that changes its
parse. That holds for a byte the sender appends. It fails for a segment that arrives out
of order, which puts the middle of a request at the start of the buffer. The head of the
request lowers the base sequence number when it arrives.

The removal therefore waits for one further packet that leaves the base sequence number
in place. That is the rule #33 already built for a buffer that starts no HTTP request.

`test_a_reordered_request_whose_head_arrives_late_produces_a_fingerprint` measures it.
The request carries the head `POST /a HTTP/1.1\r\nX-Note: ` at sequence number 100, and
the tail `GET / HTTP/11\r\n\r\n` arrives first. An immediate removal loses the value.

## How tested

Three cases join `tests/test_ja4h_reassembly.py`.

- Against the base, two of the three fail:
  `test_a_complete_header_block_that_produces_nothing_leaves_no_entry` and
  `test_the_buffer_of_a_refused_request_stops_growing`, which reports
  `AssertionError: 420055 not less than 8192`. The third passes, because it guards a
  value the base already produces.
- Against an immediate removal, two fail:
  `test_a_complete_header_block_that_produces_nothing_leaves_no_entry` and
  `test_a_reordered_request_whose_head_arrives_late_produces_a_fingerprint`.

**No JA4H value moved.** All 38 captures under `tests/foxio_vectors/` were replayed
before and after, through `ja4plus.processor.Processor` and through
`tests/conformance_index.py`. Both replays give 73 JA4H values, 73 `JA4H_ro` values and
293 streams left in the reassembler, and the two dumps are byte-identical.

**The count of 82 in the second acceptance criterion is stale.** Round 40 of the
changelog measured 82 on 2026-08-07. The base commit of batch 20 produces 73.

Local gates, all green:

```
pytest tests/ -m "not spec_validation"     1338 passed, 8 xfailed, 93 subtests passed
pytest tests/ -m spec_validation           1464 passed, 143 skipped, 136 xfailed
ruff check ja4plus/ tests/                 All checks passed!
ruff format --check ja4plus/ tests/        122 files already formatted
mypy ja4plus/                              Success: no issues found in 27 source files
```

The register holds 136 keys and the conformance suite reports 136 xfailed, before and
after. Coverage of `ja4plus/fingerprinters/ja4h.py` rises from 90% to 91%, and the whole
project holds at 82%.

## Documentation

`docs/specs/spec.md` gains Changelog round 74.


Crank-Git:
Self-review, one correctness lens, run by a child agent on the diff. Two minor findings, both rejected, with the reason.

**Finding 1 — the redundant membership guard in `_drop_a_stream_whose_base_holds`.** Rejected. `_drop_an_unusable_stream` checks the membership before it calls the new method, and the new caller in `process_packet` does not. A payload of zero bytes reaches `add_segment`, which stores nothing, so the guard covers the second caller. It stays.

**Finding 2 — `test_a_reordered_request_whose_head_arrives_late_produces_a_fingerprint` passes against the base.** Correct, and it is the point of the case. It measures the premise the issue body states, that a complete header block admits no later byte that changes its parse. The premise fails for a byte that precedes the buffer. Replace the call to `_drop_a_stream_whose_base_holds` with `self.reassembler.remove_stream(stream_key)` and the case fails, with `test_a_complete_header_block_that_produces_nothing_leaves_no_entry`: `2 failed, 1336 passed`.

**The reasoning of the review on question 1 is wrong, and the answer is right.** The review states that the reassembly is append-only. `TCPStreamReassembler.get_stream` starts at the earliest sequence number the stream holds, so a segment that arrives late prepends bytes and moves every offset. `TCPStreamReassembler.base_seq` exists for that reason. The two-packet wait is what makes the removal safe, and the measurement above proves it.

Crank-Git:
The second acceptance criterion states 82 JA4H values. **The base commit of batch 20 produces 73.** Changelog round 40 measured 82 on 2026-08-07, and the number has moved since. The measurement of this pull request compares the values, not the stale count: all 38 captures under `tests/foxio_vectors/` give 73 JA4H values, 73 `JA4H_ro` values and 293 streams left in the reassembler, before the change and after it, and the two dumps are byte-identical. The project manager owns the correction of the criterion.

Crank-Git:
Rebased onto `batch/259-interpret-fingerprint` at `f0ef8fc`, after #186 sub-merged. The one conflict was `docs/specs/spec.md`, resolved by appending: round 73 for #186, then round 74 for this issue. Neither row was dropped and neither round number moved. #263 now reports `MERGEABLE` / `CLEAN`, and the head commit still ends with `[skip ci]`.

Changelog round 74 gains two clauses the project manager asked for.

- **It names the commit its counts were measured against**, `batch/259-interpret-fingerprint` at `f0ef8fc`, and it states that batch 19 raises the register to 137 keys because #242 commits the snapshot that measures D1.
- **It states the limit of the removal.** The stream leaves on the second packet, and not on the refused packet itself. A stream that receives one packet and nothing else leaves through `max_stream_age`, which is 600 seconds.

Re-measured on the new base. Every count holds.

```
pytest tests/ -m "not spec_validation"     1352 passed, 8 xfailed, 93 subtests passed
pytest tests/ -m spec_validation           1464 passed, 143 skipped, 136 xfailed
ruff check ja4plus/ tests/                 All checks passed!
ruff format --check ja4plus/ tests/        122 files already formatted
mypy ja4plus/                              Success: no issues found in 27 source files
tests/foxio_deviations.json                136 keys
coverage                                   TOTAL 82%, ja4plus/fingerprinters/ja4h.py 91%
```

**No JA4H value moved on the new base either, and the comparison was made against the new base and not the old one.** `f0ef8fc` with `ja4plus/fingerprinters/ja4h.py` restored to its base form gives 73 JA4H values, 73 raw forms and 293 streams left behind across the 38 captures. The branch gives a byte-identical dump. The cost measurement holds: the refused request costs 20.2 ms and holds 1400 stream bytes, where the base costs 35.5 ms and holds 420055.

Issue #262, opened 2026-08-08
Return an error dictionary for a bytes JA4SSH fingerprint


Crank-Git:
Found while building #186, and out of scope there. The same input raises the same error at `07f5a7a`, so it is older than that branch.

## What is wrong

`JA4SSHFingerprinter.interpret_fingerprint` calls `fingerprint.split("_")` at
`ja4plus/fingerprinters/ja4ssh.py:503`. A `bytes` value raises `TypeError`, and the
handler names `AttributeError`, `IndexError` and `ValueError` alone. The error
reaches the caller as a stack trace.

## How to reproduce

```
>>> JA4SSHFingerprinter().interpret_fingerprint(b"c36s36_c50s50_c70s30")
TypeError: a bytes-like object is required, not 'str'
```

## Why it is inconsistent

The method returns `{"error": ...}` for `None`, which is also not a string.
`tests/test_ja4ssh_interpret.py` holds the case, and `_close_window` returns `None`,
so a caller reaches the method with a value that is not a string.

## The decision this needs

A reader must decide whether a value of the wrong type is a malformed fingerprint or a
defect in the caller. #36 set the handler to the parse errors this method expects, so
adding `TypeError` to it is a decision and not a repair.

Crank-Git:
## Plan — batch 20, integration branch `batch/259-interpret-fingerprint`

**Changelog round 75.** It is assigned. Do not choose one, and never write `TBD`.

**#186 already sub-merged into this integration branch**, and it rewrote the part parser of
this same method. Your base holds that work. Read it before you write.

### The project manager's decision

**A value of the wrong type is a malformed fingerprint, and the method returns the error
dictionary.** The issue asks whether it is that, or a defect in the caller. Three readings
settle it:

1. **The method already answers this question for `None`**, which is also not a string, and
   it answers it with `{"error": ...}`. A `bytes` value is the same class of input, and two
   answers to one question is the inconsistency the issue names.
2. `_close_window` returns `None`, so a caller reaches this method with a value that is not
   a string **through this project's own code**.
3. `CLAUDE.md` states that a parser that cannot read its input returns nothing rather than
   raising, and that a fingerprinter catches the parse errors it expects rather than a bare
   `Exception`. **Naming `TypeError` in the handler follows both halves of that rule.** A
   bare `except Exception` does not, and it is forbidden.

**Measure the premise before you build on it.** If `interpret_fingerprint` does not in fact
[...]

PR #265, opened 2026-08-08
fix(ja4ssh): return an error dictionary for a bytes JA4SSH fingerprint

Crank-Git:
Issue #262. Batch 20, into `batch/259-interpret-fingerprint`. Changelog round 75.

The branch rebased onto `715d0f7`, which the other member of batch 20 added to the
integration branch. The two changes touch no common file, and the only conflict was the
adjacent changelog row.

## What is wrong

`JA4SSHFingerprinter.interpret_fingerprint` raised `TypeError` for a fingerprint of type
bytes. `bytes.split` denies a separator of type str, and the handler named
`AttributeError`, `IndexError` and `ValueError` alone. The error reached the caller as a
stack trace.

## Why the method returns the error dictionary

The method already answers `None` with the error dictionary, and `None` is also not a
string. Two answers to one question is the inconsistency the issue names. `_close_window`
returns `None`, so a caller reaches the method with a value that is not a string through
this project's own code. `TypeError` joins the handler beside the three errors it already
names. The handler names no bare `Exception`, which `CLAUDE.md` forbids.

## What the measurement adds

The issue body names `bytes`. Thirteen types reached the method at the base commit
`f0ef8fc`.

- `bytes` and `bytearray` raised `TypeError: a bytes-like object is required, not 'str'`.
- `None`, `memoryview`, `int`, `float`, `bool`, `list`, `tuple`, `dict` and `set` already
  returned the error dictionary through `AttributeError`. Each holds no `split`
  attribute, and `memoryview` is the one bytes-like type that answers this way.
- A string returns the session dictionary, before and after.

A second place raises `TypeError`. The part guard `JA4SSH_PART` holds a string pattern,
so `fullmatch` raises `TypeError` on a part of type bytes. One handler covers both
places, and a stand-in whose `split` returns `[b"c36s36", b"c50s50", b"c70s30"]` reaches
the second one.

## The issue body's line number is correct

The body names `ja4plus/fingerprinters/ja4ssh.py:503`, and that line held
`fingerprint.split("_")` at the base commit. Rounds 45 and 73 each record a stale line
number on this same method, so this round breaks that run of two.

## How it is tested

Test first, then the handler. The four new cases failed with `TypeError` before the
change and pass after it.

- `test_a_fingerprint_of_a_bytes_like_type_returns_an_error_dictionary` covers a bytes
  fingerprint, a bytearray fingerprint and an empty bytes fingerprint.
- `test_a_fingerprint_whose_parts_hold_bytes_returns_an_error_dictionary` covers the part
  guard.

**The guard is proven by its removal.** Drop `TypeError` from the handler and exactly 4
cases of the unit suite fail, all four of them new. No other case fails, so nothing
depended on the stack trace.
`test_an_error_the_parser_expects_from_no_input_reaches_the_caller` still passes, so a
`RuntimeError` still reaches the caller.

## Gates

```
.venv/bin/pytest tests/ -m "not spec_validation"   1356 passed, 8 xfailed, 93 subtests
.venv/bin/pytest tests/ -m spec_validation         1464 passed, 143 skipped, 136 xfailed
.venv/bin/ruff format --check ja4plus/ tests/      122 files already formatted
.venv/bin/mypy ja4plus/                            no issues found in 27 source files
```

No fingerprint moved. `interpret_fingerprint` reads a fingerprint and produces none, so
no FoxIO vector covers it. The conformance suite reports 136 `xfailed` before and after,
and `tests/foxio_deviations.json` holds 136 keys before and after. The file is untouched;
batch 19 owns it. Coverage holds at 82% with `--cov=ja4plus`, and
`ja4plus/fingerprinters/ja4ssh.py` holds 98% before and after.

## Documentation

- `docs/specs/spec.md` gains Changelog round 75.
- `docs/specs/features/02-correctness-audit.md` row 14 now records the four errors the
  handler names.
- The docstring of `interpret_fingerprint` now states the failure mode.

No external interface changed, so this pull request cites none.


Crank-Git:
## Self-review, four lenses

A Sonnet child ran the review inside this worktree and it returned. Four lenses:
correctness, robustness, test quality and the writing standard.

**Correctness — no finding.** The reviewer read every operation inside the `try` block
and named no operation that raises `TypeError` for a defect inside this project. Both
groups of `JA4SSH_PART` are mandatory, so `match.group(1)` and `match.group(2)` return no
`None` after a successful `fullmatch`, and `int()` raises no `TypeError` there. The loop
always fills `values` with three elements, so the tuple unpack raises no `TypeError`. The
two places that raise `TypeError` both name a value of the wrong type.

**Robustness — no finding.** A bytes-like fingerprint reaches no success branch, and the
change alters the answer for no string input.

**Test quality — no finding.** The reviewer reproduced the pre-change handler
independently and measured the four cases:

```
bytes           -> RAISED TypeError a bytes-like object is required, not 'str'
bytearray       -> RAISED TypeError a bytes-like object is required, not 'str'
empty bytes     -> RAISED TypeError a bytes-like object is required, not 'str'
SplitsIntoBytes -> RAISED TypeError cannot use a string pattern on a bytes-like object
```

The two sites raise two different messages, which confirms that `SplitsIntoBytes` reaches
the part guard and not the split.

**Writing standard — one finding, fixed in 7fbe8a2.** The handler comment held a sentence
of 36 words that carried two facts. STE rule 1 caps a description sentence at 25 words,
and rule 2 allows one fact per sentence. The comment now holds one sentence for the
bytes-like fingerprint and one for the part of type bytes.

All five gates pass after the fix.

Issue #266, opened 2026-08-08
Batch 21: the register gate and three decided fingerprint forms


Crank-Git:
Batch 21 of the `issue-flow` loop. Session 10, 2026-08-08.

**This is the last batch of the backlog above the original epics.** When it merges, Epic 3
(#14) is the next body of work and #38 to #43 are ready.

**Every member of this batch touches `tests/foxio_deviations.json`.** That is why none of
them could join batch 20, which forked from `dev` at `07f5a7a` before the register repair.
This branch forks from `dev` at `a5e9f8e`, where the register holds **137 keys and 9
undecided entries**.

## Members

- [x] #253 Gate the register marker rule so an unmarked decided entry fails
- [x] #228 Decide the JA4X empty-list form and the JA4X raw form
- [x] #231 Decide the eleven JA4D and JA4D6 readings the images leave open
- [x] #246 Write the JA4TS value that a RST produces
- [x] #255 Refresh the register owner list and re-home the six JA4L entries

## Sequencing

**#253 lands first, and the other three land against it.** It gates the rule that an entry
whose cause names a Changelog round carries the `decided` marker. Batch 18 added 21 entries
and 16 were born unmarked with no gate catching it, and #251 repaired that in batch 19 —
for the second time in one day. **A member of this batch that adds a register entry must
satisfy the gate, which is the whole point of landing the gate first.**

## Three members change a published fingerprint

#228, #231 and #246 each move a value that `ja4plus` 0.6.0 publishes. **The user decided
#228 and #231 on 2026-08-08**, and both decisions are recorded as comments on their issues.
**#246 builds a rule the deleted `technical_details/JA4T.md` states**, and #226 already
landed part e of JA4TS in Changelog round 67.

**#215 is not a member.** It needs a user ruling on D1, D2 and D4, and it is
`status:needs-feedback`.

## The Changelog rounds

Rounds 1 to 72 are on `dev`. **Batch 20 holds 73 to 75 and it merges first.**

| Member | Round |
|---|---|
| #253 | 76 |
| #228 | 77 |
| #231 | 78 |
| #246 | 79 |
| #255 | 80 |

## The invariant

`dev` at `a5e9f8e` reports `1477 passed, 143 skipped, 137 xfailed`, and
`tests/foxio_deviations.json` holds 137 keys. **The equality is the invariant.** A member
that moves the key count states the count before and after.

## Integration branch

[...]

Issue #253, opened 2026-08-08
Gate the register marker rule so an unmarked decided entry fails


Crank-Git:
Raised by the project manager at the batch 19 plan, from the closing section of #251.

## The gap

**Nothing enforces the register's own marker rule.** `tests/foxio_deviations.json` gained
21 entries in batch 18 and 16 of them were born with no `decided` marker while their cause
text names a decision the user made. No gate caught it. #193 had repaired the identical
fault earlier the same day, and #251 repairs it again.

**An unmarked entry states that nobody has read it.** A reader of `dev` between batch 18
and #251 counted 24 open entries when 8 were open.

## The rule to enforce

An entry whose cause text names a Changelog round, or names a decision issue, carries
`"decided": true`. An entry that carries no such reference stays unmarked, and that is the
correct state for a genuinely open entry.

## What to build

- [ ] A case in the conformance suite reads `tests/foxio_deviations.json` and fails when an
      entry whose cause names a Changelog round carries no `decided` marker.
- [ ] The case names every offending key, so the failure is actionable rather than a count.
- [ ] The case passes on `dev` after #251 merges, with 8 entries correctly unmarked.
- [ ] **The case can fail.** Prove it: add a marker-free entry whose cause names a round,
      show the case fails, and remove it. This project has recorded fourteen instances of a
      comparison that cannot fail, and two of them sat in an acceptance criterion.
- [ ] The key count still equals the `xfailed` count.

## What not to build

Do not mark any entry. #251 owns the repair; this issue owns the gate. Do not weaken the
rule so that the current file passes — if the file fails, that is the finding.

## Depends on

**Depends on #251.** The gate is written against the repaired file.

Crank-Git:
Unblocked. #251 merged into `dev` in batch 19, as `a5e9f8e`.

**The register on `dev` now holds 137 keys and 9 undecided entries.** The 9 are #34's six, #215's two, and the one `gre-erspan-vxlan.pcap/0:65174/JA4T.1` entry that #242 added, which #215 decides.

**Write the gate against those 9.** #251 also added `tests/test_foxio_deviations.py`, which already holds `TestTheOpenRegisterEntries` and asserts the exact set of undecided keys. **Read that file first — the gate this issue asks for may belong beside those cases rather than in a new file.**

Crank-Git:
## Plan — batch 21, integration branch `batch/266-register-gate-and-decisions`

[...]

PR #269, opened 2026-08-08
Gate the register marker rule so an unmarked decided entry fails

Crank-Git:
Builds issue #253. Batch member of #266, so this pull request is a draft and every head
commit carries `[skip ci]`.

## What changed

`unmarked_decisions` in `tests/test_foxio_deviations.py` reads the register and returns one
message for every entry that cites a decision and carries no `decided` marker.
`TestTheRegisterMarkerRule` holds twelve cases. No file under `ja4plus/` changed, no
fingerprint moved, and `tests/foxio_deviations.json` is untouched.

## Why

The register gained 21 entries in batch 18, and 16 of them named a decision the user had
made while carrying no marker. #193 repaired the same fault on 43 entries under #129, and
#251 repaired the 16. No case caught either fault. An unmarked entry states that nobody
read it, which is the register's only job. Changelog round 69 recorded that the invariant
is not enforced and that #253 owns the gate.

## The citation forms the gate reads as evidence

The gate reads four forms as evidence that a person decided an entry. Every form is past
tense, because a decision that was made is the only evidence that somebody read the entry.

| Form | Example |
|---|---|
| A Changelog round named by number | `Changelog round 66` |
| A decision recorded against an issue or a date | `decided on #105`, `decided on 2026-08-07` |
| An issue named as the place the decision was made | `#138 decided` |
| An issue named as the record of the decision | `#162 records the decision` |

Two near misses are not citations, and both appear in the register today.

- **A denied citation.** The six #34 entries close their cause with `no Changelog round
  settled it`. A rule on the words `Changelog round` alone marks six open entries. The rule
  requires a number, and a lookbehind drops a denied citation.
- **The present tense.** The three #215 entries read `#215 decides whether ja4plus reads
  the raw option bytes`. An entry that names the issue that will decide it names no
  decision.

**The rule runs in one direction.** It recognizes a citation in 90 of the 128 decided
entries. It reads no citation in 38 of them. The 38 are the 34 entries of #138 and 4 of the
5 entries of #151. A FoxIO Rust snapshot settled those by measurement, and no person
decided them. The fifth #151 entry cites `#138 decided`. A decided entry that cites nothing
is therefore not a defect.

## The gate can fail, and the proof was reverted

A marker-free entry keyed `253-proof.pcap/0:1234/JA4.1`, whose cause names Changelog round
66, makes `test_the_committed_register_hides_no_decision` fail:

```
E       AssertionError: assert ['253-proof.p...cided marker'] == []
E         Left contains one more item: '253-proof.pcap/0:1234/JA4.1 names Changelog round 66, and the entry carries no decided marker'
```

Adding the marker alone cleared the failure, which proves the gate reads the marker and not
the key. `git checkout` restored the register, and `git status` reports it unchanged at 137
keys.

The first ten cases were written before `unmarked_decisions` existed, and all ten failed.
Self-review added two more, and the comment on this pull request records every finding,
including the four it rejects and why.

## The register holds nine open entries and not eight

The issue body states eight. The measurement reports nine, because #242 added
`gre-erspan-vxlan.pcap/0:65174/JA4T.1` after that body was written. #34 owns six and #215
owns three. The plan comment on #253 states nine, and the measurement agrees with the plan.

## How it was tested

Measured in a worktree virtual environment against base `a5e9f8e`.

| Gate | Before | After |
|---|---|---|
| `pytest tests/ -m "not spec_validation"` | 1358 passed, 8 xfailed | 1370 passed, 8 xfailed |
| `pytest tests/ -m spec_validation` | 1477 passed, 143 skipped, 137 xfailed | 1477 passed, 143 skipped, 137 xfailed |
| `ruff format --check ja4plus/ tests/` | 122 files formatted | 122 files formatted |
| `mypy ja4plus/` | success, 26 files | success, 26 files |
| Coverage `--cov=ja4plus` | 3467 statements, 524 missed, 85% | 3467 statements, 524 missed, 85% |

Register keys: 137 before, 137 after, against 137 `xfailed` cases before and after.

## Documentation

`docs/specs/spec.md` gains Changelog round 76. The module docstring of
`tests/foxio_deviations.py` gains a `## The marker rule` section, next to the procedure that
adds an entry, because that is where an author decides whether to set the marker.

Verified against: https://github.com/FoxIO-LLC/ja4/tree/main/python/test/testdata (pinned
commit `27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8`, read on disk 2026-08-08). No external
interface is called by this change.


Crank-Git:
## Self-review

Two review lenses ran against the diff: an adversarial correctness review of the rule, and
a prose review against `.claude/rules/ste.md`. Both ran to completion and every claim below
was re-measured against the committed register.

### Accepted, and fixed in 9197f55

1. **The gate missed a citation form the register already holds.** `#162 records the
   decision` closes the cause of 16 entries. A future entry that copied the form and carried
   no marker passed the gate. Three other members of this batch add register entries, so the
   hole was reachable. The gate now reads a fourth form, and
   `test_the_check_reports_an_entry_that_names_the_issue_that_records_the_decision` holds it.
   The recognition counts do not move, because those 16 entries also name the date of the
   decision.
2. **A denied citation must hide no decision that the same cause states later.** The
   behaviour was correct and untested. `test_the_check_reads_a_citation_that_follows_a_denied_citation`
   now holds it.
3. **An `-ing` noun in two files.** "Naming the issue that will decide an entry is not naming
   a decision" breaks rule 14. Both copies now read "An entry that names the issue that will
   decide it names no decision."
4. **One British spelling.** `recognises` in the round 76 row is now `recognizes`.
5. **A passive test name.** `..._names_the_issue_a_decision_was_made_on` is now
   `..._names_the_issue_the_user_decided_on`.
6. **One sentence over the 25-word cap** in the round 76 row is split into three.

### Rejected, with the reason

7. **"The negation guard is too narrow."** The review showed that `Nobody has yet decided on
   #105` and `There was no formal Changelog round 76` defeat the guard, which reads the word
   before the citation alone. The finding is correct and the guard stays narrow. A wider
   guard was measured and rejected: the register's causes carry `no` constantly, as in
   `reads no QUIC handshake`, so a sentence-level negator suppresses real citations and turns
   a false positive into a false negative. **The two errors are not equal.** A false positive
   fails the gate, prints the key, and a person rewords the cause or marks the entry. A false
   negative is silent, and a silent miss is the fault that produced 16 unmarked entries in
   batch 18. The gate is biased toward asking. The comment on `DECISION_CITATION` and the
   `## The marker rule` section now state the bias and the repair.
8. **"The message can cite the denied reference rather than the real one."** This occurs only
   under the loose-negation case of finding 7, which is rejected. The acceptance criterion
   asks the case to name every offending **key**, and it does. The citation is context.
9. **Speculative citation forms** — `Resolved by #228`, `decided on issue #231`, and case
   variants such as `changelog round 66`. No entry holds any of them. The issue states that
   the rule must not be weakened to accommodate an entry that does not exist, and the mirror
   holds: the gate reads forms the register uses, measured, not forms invented for it.
10. **Future tense in one test name**, `..._names_the_issue_that_will_decide_it`. The subject
    of that case is a future-tense citation, so the name states the behaviour under test.

### The count the issue body carries is superseded

The issue body states eight open entries. The register holds **nine**. #242 added
`gre-erspan-vxlan.pcap/0:65174/JA4T.1` after the body was written, and the plan comment on
#253 states nine. #34 owns six and #215 owns three. No entry was marked and
`tests/foxio_deviations.json` is byte-identical to the base.

Crank-Git:

**The gate was proven independently rather than on the verdict's word.** Against the branch as published:

1. Unmodified: `132 passed`.
2. A marker-free entry whose cause reads `Changelog round 66 settled this entry, and it carries no marker.` inserted into the register: the check fails and names it — `pm-audit.pcap/0:9999/JA4.1 names Changelog round 66, and the entry carries no decided marker`.
3. `"decided": true` added to that same entry: **the 12 marker-rule cases pass.** So the gate reads the marker and not the key.

The register blob is byte-identical to the base, `06c09da1b4c0778020b6e841c3d1ea0b5d236b2d`, so no entry was touched.

**Two disclosures, both accepted.**

- **The gate runs in the unit suite and not in the marked `spec_validation` suite.** The issue body says "a case in the conformance suite" and the plan named `tests/test_foxio_deviations.py`, where #251 put `TestTheOpenRegisterEntries`. The worker followed the plan and said so. **Continuous integration runs both suites**, so the gate binds either way, and splitting the register cases across two files would cost more than the wording is worth.
- **The narrow negation guard is the right bias and the review finding is correctly rejected.** Register causes read `reads no QUIC handshake` constantly, so a sentence-level negator would suppress real citations. **A loud false positive is a wrong entry named; a silent false negative is the fault that produced 16 unmarked entries in batch 18.** The worker chose the loud failure. Keep it.

**Step 2 of the proof also fired `TestTheRegisterOwners` twice**, because the injected entry named issue #999. That is #255's subject, and it confirms independently that the owner check works and that only a stale snapshot keeps it quiet.

PR #270, opened 2026-08-08
Record the decided JA4X empty-list form

Crank-Git:
Builds issue #228, rescoped to Decision 1 alone. Changelog round 77. Batch 21, #266.

## What the user decided

An empty JA4X list writes the zero sentinel `000000000000`. The decision is reversible.

## What changed, and what did not

**No file under `ja4plus/` changed.** The project already wrote the decided form.
`ja4plus/fingerprinters/ja4x.py:65` to `:71` hold one guard for each of the three parts.
This branch measured that rather than assuming it, and it did not manufacture a change.

The product of this issue is the record and the gate.

- `docs/specs/foxio/JA4X.md` records the decision and the contradiction beside R8.
- `tests/test_ja4x_empty_ext.py` pins the sentinel with 7 cases.
- `tests/test_edge_cases.py` loses one docstring that stated the rejected form.
- `docs/specs/spec.md` gains Changelog round 77.

## Why the question needed the user

`technical_details/JA4X.png` contradicts its own example. Its two `Sliver, Havoc C2` rows
write `000000000000` in part a. Its `Qakbot` row writes
`2bab15409345_af684594efb4_e3b0c44298fc` in part c, for the same condition.

Reproduced in this worktree on 2026-08-08:

```
$ python3 -c "import hashlib; print(hashlib.sha256(b'').hexdigest()[:12])"
e3b0c44298fc
```

The Qakbot part c is therefore the truncated SHA-256 of the empty string, and not a
sentinel.

| Form | Sources at the pinned commit |
|---|---|
| **Decided.** `000000000000` | `rust/ja4x/src/lib.rs:166` to `:176`, whose test at `:181` asserts `assert_eq!(hash12(""), "000000000000");`. `wireshark/source/packet-ja4.c:590` to `:592`. `README.md` line 146, which reads ```JA4X=2bab15409345_af684594efb4_000000000000``` for the same certificate. |
| **Rejected.** `e3b0c44298fc` | `python/ja4x.py:87`, which hashes the join with no guard. The Qakbot row of the image. |

The first two parts of `README.md` line 146 and of the image row match exactly, and the
third part does not.

Verified against: https://github.com/FoxIO-LLC/ja4/tree/27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8 (FoxIO JA4+, pinned commit, read on disk 2026-08-08)

## How it is tested

**No fingerprint moved.** A replay of all 38 committed captures produces 60 JA4X values
[...]

PR #275, opened 2026-08-08
Write the JA4TS value that a RST produces

Crank-Git:
Builds issue #246. Batch member of #266, so this is a draft into
`batch/266-register-gate-and-decisions` and every head commit ends with `[skip ci]`.

## What this changes

`ja4plus/fingerprinters/ja4ts.py` reads the RST packet that the server sends. A RST on a
connection that already holds a delay appends `-R<seconds>` to part e. The delay is the
interval since the last SYN-ACK, in whole seconds. That value reads part a through part d
from the first SYN-ACK of the connection, because a RST packet carries neither a window
size nor an option.

A RST on a connection with no delay produces no value. A client RST produces no value.

## Why

The deleted `technical_details/JA4T.md` is the primary source. `b6f3ff4` removed it on
2024-02-22 and no commit restored it, so this branch read the blob out of the repository
history rather than inferring the text. It reads: "the final TCP packet, a RST packet,
should be appended to the last JA4TS denoted with 'R' and its delay." Its example ends
`65535_2-1-3-1-1-4_65495_8_1-2-4-8-R6`. It states the reason the value reads the stored
connection: "Note that RST packets do not contain TCP options or window sizes, as such the
program will need to be aware of the previous JA4TS."

Two implementations corroborate the suffix. `wireshark/source/packet-ja4.c:1608` writes a
second `JA4TS` value on the RST packet, and lines 1601 to 1607 restore `window_scale`,
`window_size`, `mss_val` and `tcp_options` from the connection. `zeek/ja4t/main.zeek:233`
appends `-R%d` to the delay list. Both nest the RST branch inside the delay branch:
`wireshark/source/packet-ja4.c:693` reads `rst_time` only when `syn_ack_count > 1`, and
`zeek/ja4t/main.zeek:232` reads `rst_ts` only when the delay list holds a value.

## Three citations of the issue body and the plan were stale

Every `file:line` was re-measured against the pinned commit
`27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8`.

| Citation | Stated | Measured |
|---|---|---|
| The dissector tests the RST flag combination | `packet-ja4.c:1297` | `packet-ja4.c:1296` |
| The Zeek script stores part a through part d | `main.zeek:181-182` | `main.zeek:177-178` |
| The Zeek script reads the timeout | `main.zeek:161` | `main.zeek:162` |

`packet-ja4.c:693`, `packet-ja4.c:694`, `packet-ja4.c:1608` and `main.zeek:233` all
measured correct.

## Two reference disagreements, and the reading each takes

**The flag combination.** `zeek/ja4t/main.zeek:167` tests `rph$tcp$flags & TH_RST`, so any
packet that carries the RST bit reaches the rule. `wireshark/source/packet-ja4.c:1296`
tests `tcp_flags == 0x004` for equality, so the dissector reads no RST that also carries
ACK. The deleted file writes "a RST packet" and names no flag combination, so the prose
and the Zeek script agree and the dissector reads one case fewer. A RST that carries ACK
is common on real traffic. This branch reads the bit.

**The RST that carries no delay.** `wireshark/source/packet-ja4.c:1599` writes the `JA4TS`
field whenever `syn == 3`, so the dissector emits the four parts a second time on such a
RST. That value repeats the value the SYN-ACK already produced and describes no packet of
its own, which `.claude/rules/conformance.md` declines under its second shape.
`zeek/ja4t/main.zeek` writes one JA4TS value for one connection and never a second, so the
Zeek package corroborates the decline. This branch emits nothing.

Both readings are recorded in `docs/specs/foxio/JA4T.md` R13 and in the
`Divergence register` row, and both are reversible.

## The capture is built, because none exists

`tests/foxio_vectors/` holds 10 RST packets across its 38 captures, and none of the 10
sits on a connection with more than one SYN-ACK. `ssh2.pcapng` holds the one connection of
the set that the server answered twice, and the server sent no RST on it.

`tests/build_ja4ts_rst.py` writes the capture from the six capture times the deleted file
lists. `.gitignore` commits no capture other than the FoxIO vectors, so
`tests/test_ja4ts_reset.py` writes the file to a temporary directory and reads it back,
the way `tests/test_ja4ssh_message_count.py` reads `tests/build_ssh_retransmission.py`.

## The replay of every vector

A replay of the 38 captures of `tests/foxio_vectors/` produces **797 values before the
change and 797 after, and 0 differ**. No value of any method moves, so no value of a
method other than JA4TS moves.

`tests/foxio_deviations.json` holds **137 keys before and 137 after**, against **137
`xfailed` cases before and 137 after**. The register gains no entry, so the marker gate of
#253 stays satisfied.

## The cases, and the proof that each can fail

`tests/test_ja4ts_reset.py` holds 22 cases. They were written first, and **16 of them
failed on the unchanged base**. The 6 that passed assert that a RST produces no value, and
the unchanged code produced no value at all.

**A revert of `ja4plus/fingerprinters/ja4ts.py` alone fails 17 cases** (the 16 above plus
`test_a_reset_on_a_connection_with_no_syn_ack_produces_no_value`, which self-review
strengthened to read the tracker directly).

**13 mutations prove that all 22 can fail**, and no case survives all 13.

| Mutation | Cases it fails |
|---|---|
| Drop the delay gate | 1 |
| The RST delay reads the first SYN-ACK | 10 |
| Store the parts of the latest SYN-ACK | 14 |
| Read the RST flag combination for equality | 1 |
| Drop the timeout from the RST path | 1 |
| Key the connection on the server address alone | 2 |
| Normalize the connection key across both directions | 3 |
| Crash on a connection the tracker never saw | 1 |
| The empty-string guard of part e reaches the RST path | 1 |
| The RST branch never runs | 13 |
| The RST value omits its suffix | 10 |
| `clear` leaves the stored parts behind | 1 |
| `drop` leaves the stored parts behind | 2 |

Self-review found one case that no mutation killed,
`test_a_reset_on_a_connection_with_no_syn_ack_produces_no_value`, because `generate_ja4ts`
catches `TypeError` and hides a crash. The case now reads
`SynAckTracker.reset_value` directly, and the crash mutation fails it. The review ran
here, in this worktree, and no review child produced it.

## No bound moved

The stored parts live in the connection table that Changelog round 67 bounded at 1000
entries and two minutes. `prefixes` holds the same keys as `times`, and `clear`, `drop`,
`_expire` and `_evict` all release both. No table was added and no bound changed.

## How this was tested

All five gates, run in the worktree with `.venv/bin/<tool>`:

```
.venv/bin/pytest tests/ -m "not spec_validation"   1392 passed, 8 xfailed, 93 subtests
.venv/bin/ruff format --check ja4plus/ tests/      124 files already formatted
.venv/bin/mypy ja4plus/                            Success: no issues found in 26 source files
```

Coverage with `--cov=ja4plus`: TOTAL **85% before and 85% after**.
`ja4plus/fingerprinters/ja4ts.py` moves from 93% to 94%. Coverage does not fall.

## Documentation

- `docs/specs/foxio/JA4T.md` — R13 states the rule, the four adopted readings, the two
  reference disagreements and the reachability measurement.
- `docs/specs/spec.md` — `Divergence register` gains one row, and the Changelog gains
  round 79.
- `docs/api_reference.md` — `JA4TSFingerprinter` and `generate_ja4ts` name the RST value.

## One finding outside this issue's scope

R12 of `docs/specs/foxio/JA4T.md` cites `zeek/ja4t/main.zeek:227` for the part e gate, and
Changelog round 67 cites `zeek/ja4t/main.zeek:227-236`. The measured line is **229**. R12
and round 67 belong to #226, so this branch left both unchanged.

Verified against: https://github.com/FoxIO-LLC/ja4 at commit
`27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8` (`technical_details/JA4T.md` at `b6f3ff4^`,
`wireshark/source/packet-ja4.c`, `zeek/ja4t/main.zeek`, retrieved 2026-08-08).


Crank-Git:
## Self-review

Two review children were spawned and neither returned, and this session holds no channel
to query a running child. **This review ran here, in the worktree, and no part of it is
reported from a child.**

### Findings that were fixed

1. **A case that could not fail.** `test_a_reset_on_a_connection_with_no_syn_ack_produces_no_value`
   read `process_packet` alone. `generate_ja4ts` catches `TypeError`, so a tracker that
   crashes on an unknown connection still returns None and the case still passed. The case
   now also calls `SynAckTracker.reset_value` directly, where no `except` hides the crash.
   The mutation `if stamps is not None and len(stamps) < 2` now fails it.
2. **Three stale `file:line` citations**, all of them written into this branch from the
   issue body and the plan. Commit 1 corrects them. The measured lines are
   `packet-ja4.c:1296`, `main.zeek:177-178` and `main.zeek:162`.
3. **Three docstrings that describe the code before the RST value existed.** Commit 2
   corrects them. `_connection_key` reads a RST packet too, and `reset` and
   `cleanup_connection` release the stored parts as well as the times.

### Findings that were measured and accepted

1. **The RST branch now runs before the SYN-ACK mask check.** A packet that carries SYN,
   ACK and RST together therefore reaches the RST branch, where the old code produced a
   SYN-ACK value. The 38 captures hold **0** such packets, and the replay reports 0 values
   moved. `zeek/ja4t/main.zeek:167` reads the RST bit before it reads the SYN-ACK flags,
   so the order follows the Zeek script.
2. **`reset_value` calls `_expire`, so a RST now ages the connection table.** A RST that
   arrives more than two minutes after the last SYN-ACK drops the connection, where the
   old code left it for the next SYN-ACK to drop. `zeek/ja4t/main.zeek:162` reads the
   timeout on the RST path too. A capture whose timestamps decrease is the only case that
   separates the two, and every packet is hostile input, so the entry-count bound still
   holds the table. `test_a_capture_with_decreasing_timestamps_bounds_the_connection_table`
   covers it.
3. **`times` and `prefixes` cannot drift.** Only `record` inserts, and it inserts into both
   under one condition. `clear`, `drop`, `_expire` and `_evict` all release both, because
   `_expire` and `_evict` now call `drop`.
   `test_the_stored_parts_hold_one_entry_for_each_tracked_connection` asserts
   `sorted(prefixes) == sorted(times)`, and the mutations `M12` and `M13` fail the release
   cases.
4. **No new state table and no bound moved.** The stored parts live in the connection table
   that Changelog round 67 bounded at 1000 entries and two minutes.

### The replay was re-measured after two rebases

The batch branch advanced twice while this branch was open, and #270 changed a JA4X value.
The replay was therefore measured again against the **current** base: **797 values before,
797 after, 0 moved.** No value of any method moves.

`tests/foxio_deviations.json` holds **137 keys**, against **137 `xfailed`** cases.

Crank-Git:

**Measured independently rather than on the verdict's word.** Reverting `ja4plus/fingerprinters/ja4ts.py` alone to the base, keeping every case, fails **17** — the count the verdict reports. The conformance suite on the branch reports `1477 passed, 143 skipped, 137 xfailed`, and the register blob is byte-identical to the base, so **no vector value moved and the register did not move.**

**This issue built a rule no vector reaches, and it did the thing that makes such work trustworthy.** It constructed the capture, it built the deleted file's own example `65535_2-1-3-1-1-4_65495_8_1-2-4-8-R6` as a case, and it ran 13 mutations with **0 survivors.** A rule with no reference value is exactly where a comparison that cannot fail hides, and 22 cases that all bite is the answer to that.

**One measurement in the verdict is worth keeping in the record.** The 38 captures hold **10 RST packets and none of them sits on a connection with more than one SYN-ACK**, and **no packet carries SYN, ACK and RST together.** So the rule is reachable and unreached, and the branch-ordering risk the new code introduces is measured rather than assumed.

Issue #42, opened 2026-08-06
Make the module-level lookup client thread-safe and bound its cache


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/03-concurrency-safety.md`](docs/specs/features/03-concurrency-safety.md):

> FR-concurrency-safety-13 — The module-level `lookup` function builds its client once, and two threads that call it at the same time receive one client.

> FR-concurrency-safety-14 — The lookup cache holds no more than a maximum entry count.

## What to build

`ja4plus/ja4db.py:132` holds `_default_client = None` and builds it lazily at line 138 with no guard, so two threads can build two clients. `JA4DBClient._cache` at line 68 is a plain dictionary that grows without limit and stores a result for every fingerprint ever looked up.

The port solves the same problem with `sync.Once` over a map that is read-only after load. Use the equivalent guard here.

Cache a miss as well as a hit, so a repeated miss costs nothing.

## Acceptance criteria

- [ ] Two threads that call `ja4plus.ja4db.lookup` at the same time receive results from one client instance.
- [ ] The lookup cache holds no more than its maximum entry count.

## Links

- Feature: [`docs/specs/features/03-concurrency-safety.md`](docs/specs/features/03-concurrency-safety.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #14

Crank-Git:
## Plan — Epic 3, integration branch `epic/14-concurrency-and-resource-safety`

**Changelog round 83.** It is assigned. Do not choose one, and never write `TBD`.

**#38 sub-merged as `849f58d`.** `ja4plus/utils/state_table.py` holds `BoundedStateTable`.
**Use it for the cache bound rather than writing a second bounding mechanism** — removing
exactly that duplication is why Epic 3 exists.

### The two things to build

1. **`ja4plus/ja4db.py:132` holds `_default_client = None` and builds it lazily at line 138
   with no guard**, so two threads can build two clients. **Re-measure those line numbers**;
   three issues this session carried a stale one.
2. **`JA4DBClient._cache` at line 68 is a plain dictionary that grows without limit.** It is
   one of the six tables #179 records.

**The port solves the first with `sync.Once` over a map that is read-only after load.**
[...]

Issue #41, opened 2026-08-06
Report processor statistics


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/03-concurrency-safety.md`](docs/specs/features/03-concurrency-safety.md):

> FR-concurrency-safety-11 — The processor reports how many entries each state table holds.

> FR-concurrency-safety-12 — The processor reports how many entries it evicted.

## What to build

`ProcessorStats` reports the entry count, the eviction count and the packet count for one method. Epic 4 makes it a typed dataclass; a plain object is acceptable here.

A connection that is evicted and then sends another packet creates a new entry, and the fingerprint for that connection may be incomplete. Record that in the statistics so an operator can see it happening.

The port has no equivalent. This is an addition, not a divergence: parity rule 2 applies only where the port has already shipped a choice.

## Acceptance criteria

- [ ] `Processor().stats()` returns one entry per method.
- [ ] Each entry reports the state-table entry count, the eviction count and the packet count.

## Links

- Feature: [`docs/specs/features/03-concurrency-safety.md`](docs/specs/features/03-concurrency-safety.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #14
Depends on #40

Crank-Git:
## Plan — Epic 3, integration branch `epic/14-concurrency-and-resource-safety`

**Changelog round 85.** It is assigned. Do not choose one, and never write `TBD`.

**Four members of Epic 3 have sub-merged:** `849f58d` #38 `BoundedStateTable`, `ac4ee62` #42
the lookup client and its bounded cache, `631790a` #39 **thirteen** tables moved, `0d3480e`
#40 `thread_safe` and one `RLock` per fingerprinter. **Read all four.**

### What you report, and where the numbers already exist

`BoundedStateTable` already counts evictions — #39's flood run read
`ja4l.connections: entries=10000 max=10000 evictions=390000`. **Read the structure rather
than adding a second counter.**

**The set is thirteen tables, not six.** #179 recorded six and #39 measured thirteen;
#179 now carries the correction. **Re-measure the set yourself** — #39's own finding is that
a list written in an issue is a starting point and not the measurement.

### The thing the issue asks for that is easy to miss

[...]

PR #278, opened 2026-08-08
Add a bounded state table with count and age limits

Crank-Git:
Part of #38. Part of Epic 3, #14. Changelog round 81.

`ja4plus/utils/state_table.py` holds `BoundedStateTable`. #39 moves the six unbounded
state tables of #179, and this class is the shape four of them take. The self-review
below records what the other two need. **This pull request moves no fingerprinter and
changes no caller.**

## The maximum age default is 600 seconds, and the issue body's 300 is stale

**The issue body contradicts its own feature file, and the feature file governs.**
`docs/specs/features/03-concurrency-safety.md` line 117 reads "The default maximum age
is 600 seconds. A measurement sets it." Line 205 publishes
`max_connection_age: float = 600.0` on `Processor`, and line 211 states that the value
matches `DEFAULT_MAX_STREAM_AGE`. #170 is that measurement and #179 corrected the file.
The issue body was written before both, and nobody updated its number.

The worker re-measured the longest gap between two packets of one connection over all
38 captures under `tests/foxio_vectors/`, in four readings.

| Reading | Longest gap | Capture and connection |
|---|---|---|
| Directional, packets that carry payload | **320.714503 s** | `ssh-r.pcap`, `192.168.1.169:64980` to `192.168.1.197:22` |
| Directional, every packet | 320.336760 s | `ssh-r.pcap`, `192.168.1.197:22` to `192.168.1.169:64980` |
| Bidirectional, packets that carry payload | 320.329879 s | `ssh-r.pcap`, the same connection |
| Bidirectional, every packet | 320.283589 s | `ssh-r.pcap`, the same connection |

Three flow keys hold a gap above 300 seconds. None holds a gap above 600 seconds.

Every reading reproduces #170. The first reading matches `TCPStreamReassembler`, which
keys a stream by direction and refuses a segment that carries no payload, so it is the
reading that governs the default. **A 300 second default evicts state that a committed
capture still needs.**

`DEFAULT_MAX_CONNECTION_AGE` therefore equals `DEFAULT_MAX_STREAM_AGE`, and one case
asserts that the two numbers match. One project holds one maximum age.

**The measurement is a case, not a note.** A replay of `ssh-r.pcap` at 600 seconds
evicts nothing, and the same replay at 300 seconds evicts.

The correction is a comment on #38, so the issue body's number misleads no later
reader.

## What the table does, and why

- **The count bound removes the least recently used entry on the insert that reaches
  `max_connections`.** A count bound that waits for the next pass overshoots the count
  it states.
- **The age bound runs one pass for every 1000 packets.** A pass reads every entry, so
  a pass on each packet costs the entry count on each packet.
- **`on_packet` carries the packet timestamp.** The table reads the wall clock only
  when a packet carries no timestamp. Both clocks read seconds since the epoch, so the
  two compare. A capture file replays faster than real time, and a wall clock evicts
  state the capture still needs.
- **The library starts no thread.** Eviction runs on packet arrival.
- **A read of one key holds that entry.** `__getitem__`, `get` and the `in` operator
  each refresh the entry against both bounds. A pass over the whole table holds no
  entry, so `keys`, `values`, `items` and iteration change no eviction order.
- **`__iter__` returns a snapshot.** A read inside a loop moves an entry, and an
  `OrderedDict` refuses that move during its own iteration.
- **`pop`, `del` and `clear` raise no eviction count.** Each is the caller's removal.
  `cleanup_connection` is the caller's way to evict early, and it stays.
- The argument names are `max_connections` and `max_connection_age`, which match the
  `Processor` interface that `features/03-concurrency-safety.md` publishes, so #39 and
  #40 pass them through unchanged.

## The memory baseline is measured, and the ceiling stays open

`features/03-concurrency-safety.md` asks this issue to measure the baseline and to
propose the ceiling. **This pull request publishes the measurement and proposes no
ceiling.** The measurement forces a floor near 239 MiB, and nothing forces the distance
above that floor. That distance is a product judgement, and a ceiling is a number this
package states to its users. **#279 carries the question to the user.**

Measured on this branch, on macOS, with
`resource.getrusage(resource.RUSAGE_SELF).ru_maxrss`.

| Reading | Resident memory | Bytes for each entry |
|---|---|---|
| Python alone | 15.80 MiB | — |
| `ja4plus` plus scapy plus an idle `Processor()` | 91.03 MiB | — |
| 10000 entries shaped like `JA4LFingerprinter.connections` | 10.69 MiB | 1121 |
| 100000 entries of the same shape | 102.36 MiB | 1073 |
| 10000 entries shaped like `JA4SSHFingerprinter.connections`, at a 200 packet window | 51.72 MiB | 5423 |

The JA4SSH shape is the largest of the twelve tables. Ten tables at the default bound
of 10000 entries project 91.03 + 51.72 + 96.21 = **238.96 MiB**. That number is the
floor for the one-million-packet case of #41, and #279 decides the ceiling above it.

The projection holds whatever the connection count is. `BoundedStateTable` never passes
10000 entries, so 100000 distinct connections reach no more than 10000 entries in one
table.

**No constant in `ja4plus/` names a ceiling, and no case asserts one.**

## How it is tested

**The cases were written first.** Against an empty stub, 27 of 28 failed. The one that
passed reads the capture and not the module, and it guards the replay.

**Twenty mutations of the module were then applied by hand, one at a time, and every
one is reverted.** The first round left two behaviours with no comparison against them.

1. `the contains operator holds no entry` killed no case. No case asserted that the
   `in` operator refreshes an entry.
2. `a second write inserts a new entry` killed no case. No case asserted that a rewrite
   of one key holds that entry against the count bound.

Three cases now cover both, and **every one of the twenty mutations kills at least one
case**. The mutations cover each default, the count guard, the eviction order, the age
comparison, the clock choice, the iteration snapshot, the `keys` snapshot, the two
constructor guards, and the three removals that count no eviction.

## The self-review found one defect and five readings

The full write-up is a comment on this pull request. In short:

- **Repaired.** `keys` returned the live `OrderedDict` view, so a read inside a loop over
  it raised `RuntimeError: OrderedDict mutated during iteration`. It now returns a list.
- **Recorded.** `dict(table)` holds every entry, because it reads each key through
  `__getitem__`. `items` is the pass that holds none.
- **Recorded.** `popitem` removes the least recently used entry, and a dictionary removes
  the entry it received last. No call site reaches it.
- **Recorded for #39.** `BaseFingerprinter.fingerprints` is a list that `ja4l.py:169`
  indexes and `ja4ssh.py:357` reads at index -1. This class cannot hold it, and the
  module docstring now says so rather than claiming all six tables of #179 fit.
- **Recorded for #42.** `JA4DBClient._cache` reads no packet, so nothing calls
  `on_packet` and its age pass never runs on its own. Its count bound holds.
- **Recorded, and not repaired.** One `on_packet()` call that states no timestamp moves
  the table from the capture clock to the wall clock, and a replay of a capture recorded
  in the past then ages out whole. That is the rule
  `features/03-concurrency-safety.md` states, so `on_packet` carries a warning and a case
  pins the consequence.

## Local checks

`ci: skip`, so the full local suite replaces provider CI. Run with `.venv/bin/<tool>`,
because the system Python holds no scapy and no `cryptography`.

| Gate | Result |
|---|---|
| `pytest tests/ -m "not spec_validation"` | 1416 passed, 8 xfailed, 93 subtests passed. The base reads 1379 passed and 8 xfailed. |
| `pytest tests/ -m spec_validation` | 1477 passed, 143 skipped, **137 xfailed**. The base reads the same. |
| `ruff format --check ja4plus/ tests/` | 124 files already formatted |
| `pytest --cov=ja4plus` | TOTAL 85%, and the base reads 85%. `ja4plus/utils/state_table.py` reads 100%. |

`tests/foxio_deviations.json` holds **137 keys** before and after, and the conformance
suite reports **137 xfailed** before and after. This pull request touches
`tests/foxio_deviations.json` not at all.

Practices: the cases came first and the stub run records it; coverage holds at 85% and
the new module reads 100%; the commits use `type(scope): summary`; the documentation
lands here in `docs/specs/spec.md` round 81 and `docs/api_reference.md`.

## Files

- `ja4plus/utils/state_table.py`, new.
- `tests/test_state_table.py`, new. 37 cases.
- `docs/specs/spec.md`, Changelog round 81.
- `docs/specs/features/03-concurrency-safety.md`, the open question that asks #38 for a
  ceiling. It now records that #38 measured and #279 decides.
- `docs/api_reference.md`, the `ja4plus.utils.state_table` section.


Crank-Git:
## Self-review

Two review passes ran over the diff at `65064ab`, on the Sonnet tier: one correctness
lens, and one lens that reads the class against the six tables of #179. Every finding
below is reproduced, and each states what happened to it. `7475eba` carries the fixes.

### Accepted, and repaired

**1. `keys` returned the live `OrderedDict` view.** A caller that reads a value inside a
loop over that view raises an error.

```
Looping over `table` (uses overridden __iter__, a snapshot) and reading each key:
  OK: [0, 1, 2, 3, 4]
Looping over `table.keys()` directly (a live OrderedDict view) and reading each key:
  RuntimeError: OrderedDict mutated during iteration
```

`__iter__` already returned a snapshot and `keys` did not, and every case that touched
`keys` wrapped it in `sorted` or `list` first, so nothing measured it. `keys` now returns
a list. `test_a_read_inside_a_loop_over_the_keys_method_raises_no_error` covers the loop,
and it fails when `keys` returns the view again.

**2. `dict(table)` holds every entry against both bounds.** The constructor reads each
key through `__getitem__`, so it bypasses `items`. The class docstring promised that a
pass over the whole table holds no entry, and that promise did not reach `dict(table)`.
The docstring now names the difference, and `items` is the pass that holds none.
`test_the_dict_constructor_holds_every_entry_against_the_age_bound` and
`test_a_pass_over_the_items_holds_no_entry_against_the_age_bound` pin both readings.

**3. `popitem` removes the least recently used entry.** A dictionary removes the entry it
received last. The inherited `MutableMapping.popitem` pops `next(iter(self))`, and this
table iterates least recently used first. No call site reaches the operation, so the
behaviour stays and the docstring states it.
`test_popitem_removes_the_least_recently_used_entry` pins it.

**4. `BaseFingerprinter.fingerprints` is a list, and this class cannot hold it.**
`ja4l.py:169` overwrites an entry at a stored index, and `ja4ssh.py:357` reads the entry
at index -1. The module docstring said the six tables of #179 all move onto this shape,
and that is wrong for two of them. The docstring now names four, and it records what
`fingerprints` and `JA4DBClient._cache` each need instead.

**5. `JA4DBClient._cache` reads no packet, so its age pass never runs on its own.**
Nothing calls `on_packet` for it. Its count bound still holds, because `__setitem__`
applies that bound at once, and a caller with no packet calls `evict_aged`. #42 owns the
lookup cache, so this issue records the reading rather than repairing it.

**6. One `on_packet()` call without a timestamp ages the whole table out.** The table
moves from the capture clock to the wall clock for that packet, and a capture recorded in
the past is then older than any age bound.

```
conn-a survives normal packet-clock cadence: True evictions: 0
evictions after a single on_packet(None) call amid an otherwise timestamped replay: 1
conn-a survived? False
```

**This behaviour is not repaired, because it is the rule the specification states.**
`features/03-concurrency-safety.md` reads: "Eviction by age uses the packet timestamp
when the packet carries one, and the wall clock otherwise." `on_packet` now carries a
warning, and `test_one_packet_without_a_timestamp_moves_the_table_to_the_wall_clock`
pins the consequence so that #39 meets it in a case and not in production.

### Rejected, with the reason

**A guard that warns when a consumer forgets to call `on_packet`.** The review asked for
a heuristic that reports many operations with no packet announced. Two reasons decline
it. `JA4DBClient._cache` announces no packet by design, so the heuristic fires on a
correct caller. The library also adds no handler and sets no level, so a warning reaches
an operator who configured none. The module docstring names the requirement for #39
instead.

**`in` and `get` refresh the entry.** The review flagged that a plain dictionary reads
`in` as free. That difference is the requirement, not a defect. FR-concurrency-safety-8
evicts an entry "that has not been read", and `ja4ssh.py:127` asks `if conn_key not in
self.connections` to decide that a connection is alive. Two cases assert the refresh
against each bound.

### The cases were written first, and every mutation kills one

Against an empty stub, 27 of 28 cases failed. The one that passed reads `ssh-r.pcap` and
not the module, and it guards the replay.

**Twenty mutations of the module were then applied by hand, one at a time, and every one
is reverted.** The first round left two behaviours with no comparison against them: no
case asserted that the `in` operator holds an entry, and no case asserted that a second
write to one key holds that entry against the count bound. Three cases now cover both,
and **every one of the twenty mutations kills at least one case**.

Crank-Git:

**This worker returned `needs-feedback` rather than shipping a default the measurement disproved. That is the mechanism working, and it is the reason this issue is right.**

**Measured independently rather than on the verdict's word.** On the branch, with `__pycache__` cleared before each run:

- Clean: `37 passed`.
- `popitem(last=False)` to `popitem(last=True)`, so the table evicts the most recently used entry: **4 failed.**
- `DEFAULT_MAX_CONNECTION_AGE` from 600 to 300: **2 failed.** The default is genuinely pinned, and it is not a case built from the constant it asserts — which is the fault #175's boundary cases carried.

**The 300-second default in the issue body is stale, and the specification already said so.** `docs/specs/features/03-concurrency-safety.md` states 600 at line 117, records at lines 126-127 that 300 stood in the file until #170 measured the vectors, and publishes `max_connection_age: float = 600.0` at line 205. The worker's four readings reproduce #170 exactly — the longest gap is 320.714503 s in `ssh-r.pcap`, three flow keys exceed 300 s, and none exceeds 600 s. **Shipping 300 would have evicted state a committed capture still needs.** Changelog round 81 records the correction and names #170.

**The 512 MiB ceiling is withdrawn from the whole branch.** The one remaining `512` in `docs/specs/spec.md` is Changelog round 48, the #175 cookie bound, and it is pre-existing on the base. **The branch publishes the 238.96 MiB measurement and no ceiling**, and #279 holds the number for the user.

The register is untouched at 137 keys against `137 xfailed`, and `ja4plus/utils/state_table.py` reports 100% coverage.

Issue #60, opened 2026-08-06
Bound the lookup cache


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/07-db-enrichment.md`](docs/specs/features/07-db-enrichment.md):

> FR-db-enrichment-9 — The lookup cache holds no more than a maximum entry count.

> FR-db-enrichment-10 — The lookup cache evicts the least recently used entry when it reaches its maximum.

## What to build

`JA4DBClient._cache` at `ja4plus/ja4db.py:68` is a plain dictionary that grows without limit. A monitor that looks up every fingerprint it sees keeps every one of them forever.

The maximum is 100000 entries. Cache a miss as well as a hit, so a repeated miss costs nothing.

Epic 3 made the module-level client thread-safe. Keep that guarantee.

## Acceptance criteria

- [ ] The cache holds no more than `cache_size` entries after 200000 distinct lookups.
- [ ] The least recently used entry is evicted at the maximum.

## Links

- Feature: [`docs/specs/features/07-db-enrichment.md`](docs/specs/features/07-db-enrichment.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #18
Depends on #57

Crank-Git:
Related work landed, and this issue stays open on purpose.

#42 merged into `epic/14-concurrency-and-resource-safety` and bounded `JA4DBClient._cache` with `BoundedStateTable` at **100000 entries**, the number `docs/specs/features/07-db-enrichment.md:134` publishes as `cache_size`. **Both acceptance criteria of this issue now hold, and each carries a case.**

**The project manager did not close it**, because this issue also depends on #57 for the `source` value, and #57 has not landed. #42's worker raised the question rather than deciding it, which was right.

**What is left here is the `source` value alone.** When #57 lands, re-read this issue against the code before scheduling a worker — most of it may already be built.

Crank-Git:
## The project manager answered, 2026-08-08, and #358 is sub-merged

**Question 1 — yes, #60 closes on this, and the project manager owns the mistake.**

**The bound already existed.** #42 and pull request #280 replaced the plain dictionary with a
100000-entry least-recently-used `BoundedStateTable`, cached misses, and documented it.
[...]

Issue #57, opened 2026-08-06
Make the remote lookup opt-in at the client


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/07-db-enrichment.md`](docs/specs/features/07-db-enrichment.md):

> FR-db-enrichment-1 — `JA4DBClient` performs no network request by default.

> FR-db-enrichment-2 — `JA4DBClient(allow_remote=True)` permits the remote lookup.

> FR-db-enrichment-14 — The remote lookup has a timeout.

> FR-db-enrichment-15 — A remote lookup failure returns nothing and does not raise.

## What to build

`ja4plus/ja4db.py:97` calls `self._remote_lookup(fingerprint)` on every miss with no opt-in at that layer. The command-line flag `--lookup` reads as a request for a local lookup, and the operator is not told that a request to `https://ja4db.com/api/read/<fingerprint>` follows.

A fingerprint describes traffic the operator observed. Sending it to a third party discloses that traffic. A default that reaches the network is a defect, not a convenience.

The port already separates the two, in `feat(lookup): opt-in ja4db.com remote fallback` dated 2026-05-09. Parity rule 2 applies: adopt the port's shape.

The timeout is 5 seconds and is not configurable before version 1.0.0. The service publishes no versioned API document, so treat any unexpected shape as a miss.

## Acceptance criteria

- [ ] A test that fails on any outbound socket passes while `JA4DBClient()` looks up 1000 fingerprints.
- [ ] `JA4DBClient(allow_remote=True)` performs the remote lookup on a miss.
- [ ] A remote lookup that times out returns `None` and does not raise.
- [ ] A remote response with an unexpected shape returns `None`.

## Links

- Feature: [`docs/specs/features/07-db-enrichment.md`](docs/specs/features/07-db-enrichment.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #18

Crank-Git:
Evidence for this issue, measured while building #25 in Epic 0.

`tests/test_ja4db.py::TestCLILookupFlag` makes live network requests to `ja4db.com` during an ordinary test run. Measured on `epic/11-foundation`:

[...]

Issue #14, opened 2026-08-06
Epic 3: Concurrency and resource safety


Crank-Git:
## Goal

State the concurrency contract, prove it, and bound every state table.

## Exit criteria

The thread-safety suite passes. Every state table has a maximum count and a maximum age. A one-hour packet stream holds flat memory.

## Spec

- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Feature: [`docs/specs/features/03-concurrency-safety.md`](docs/specs/features/03-concurrency-safety.md)

Depends on: #13

## Sub-issues

- [x] #38
- [x] #39
- [x] #40
- [x] #41
- [x] #42
- [x] #43

Crank-Git:
**Epic 3 merged into `dev` as `ed7d83f`. Nine of nine continuous-integration jobs passed.**

| Member | Round | What it landed |
|---|---|---|
| #38 | 81 | `BoundedStateTable` — count bound, age bound, least-recently-used eviction |
| #39 | 82 | Thirteen state tables moved onto it, across six fingerprinters |
| #42 | 83 | One lookup client under a lock, and the lookup cache bounded at 100000 |
| #40 | 84 | `thread_safe`, defaulting to `True`, and one `RLock` for each fingerprinter |
| #41 | 85 | `Processor.stats()` — entries, evictions, packets, returned connections |
| #43 | 86 | The memory-bound suite and the documented concurrency contract |
| #285 | 87 | The sixteenth table, which the merge with `dev` orphaned |

**Every state table in this package now holds a maximum entry count and a maximum age**, which
`CLAUDE.md` has required since before #179 recorded that six of them did not.

## The count of state tables read four different numbers

| Reading | Count | Why it changed |
|---|---|---|
| #179, session 7 | 6 | The list written into an issue |
| #39 | 13 | A sweep for every container assigned to `self` that survives a packet |
| #41 | 15 | A nested walk — a flat `vars()` walk reports 14 and misses `SynAckTracker.times` |
| #285 | 16 | #246 added one on a branch Epic 3 could not see |

**Only the last was caused by new code.** `STATE_TABLE_COUNT` is pinned at 16, and a case
compares an independent walk against `Processor.stats()` so a change hiding a table from
either one fails.

## The invariant

**137 register keys against 137 `xfailed`, unchanged through all seven members. No fingerprint
moved** — the 38-capture replay is byte-identical and `tests/foxio_deviations.json` never
changed.

## What is left open, deliberately

[...]

PR #280, opened 2026-08-08
Build one lookup client and bound the lookup cache

Crank-Git:
Part of #14. Builds #42.

## What this changes

`ja4plus.ja4db.lookup` builds one client, and two threads that call it at the same
time receive results from that one client. The lookup cache is a `BoundedStateTable`
of 100000 entries, so it holds no more than that count.

## Why

`ja4db.py:132` held `_default_client = None` and `:138` built the client with no
guard, so two threads built two clients and each one read the mapping file again.
`ja4db.py:68` held a plain dictionary that stored a result for every fingerprint a
caller ever looked up. All three line numbers were re-measured against the base
commit `849f58d`, and all three hold.

## How

- A module-level `threading.Lock` guards the construction, and `lookup` reads the
  client again inside the lock. The port guards the same construction with
  `sync.Once`, at `lookup.go:29-37`. FoxIO specifies nothing here, so parity rule 2
  decides.
- The cache is `BoundedStateTable`, the class #38 built. The module holds no second
  bounding mechanism.
- One lookup announces itself as one arrival, so the lookup cache runs its own age
  pass. The lookup cache reads no packet, and the module docstring of
  `ja4plus/utils/state_table.py` now records the call.
- The client holds a lock over the lookup cache read and over the lookup cache write.
  No thread holds that lock across `_do_lookup`, because a remote lookup takes up to
  5 seconds.

## The bound, and where the number comes from

The bound is **100000 entries**. `docs/specs/features/07-db-enrichment.md` line 112
states it, and line 134 publishes it as the `cache_size` default.

The 10000 of `docs/specs/features/03-concurrency-safety.md` describes a different
object. The `## Terms` table names a state table "The dictionary a fingerprinter uses
to hold per-connection data", and `JA4DBClient` is no fingerprinter.
FR-concurrency-safety-14 and FR-db-enrichment-9 state one requirement, and one number
satisfies both. The `## Terms` table gains a `lookup cache` row, because `cache`
alone is a rejected synonym of `state table`.

The maximum age is 600 seconds, the `DEFAULT_MAX_CONNECTION_AGE` of #38. One project
holds one maximum age.

A measurement sets the eviction interval. A pass reads every entry, so 200000 lookups
on a full cache of 100000 entries cost 1.91 s at an interval of 1000, 0.36 s at 10000,
and 0.19 s at 100000. The client reads the smaller of 100000 and its own `cache_size`,
so a client that holds a smaller lookup cache runs its age pass too.

## The race is forced, and the guard is proven by its removal

The case holds the first caller inside the constructor until the second caller runs,
so the second caller is guaranteed to arrive while the first still builds.

| Tree | Runs | Failures |
|---|---|---|
| With the guard | 100 | 0 |
| With the guard removed | 100 | 100 |

Each failure reads `AssertionError: the module built 2 clients`. Each run is one
process, because pytest deduplicates a repeated test id and reports 20 identical ids
as one run.

The lock over the lookup cache is measured the same way. Eight threads share one
client whose lookup cache holds two entries and ages every entry out, so an eviction
runs beside every read.

| Tree | Runs | Failures |
|---|---|---|
| With the lock | 20 | 0 |
| With a no-op in place of the lock | 20 | 17 |

Each failure is a `KeyError` that a concurrent eviction raises.

## What a cached miss does when the mapping file is reloaded

`JA4DBClient` reads the mapping file once, in `__init__`, and no code path reloads
it, so the data cannot change under a live cache today. The `db update` command of
#57 writes a file that the next client reads at construction, so a caller that
replaces the mapping file builds a new client and receives an empty cache. The age
bound also removes a miss that receives no read for 600 seconds.

## No fingerprint moves

A replay of all 38 committed captures produces a byte-identical dump, at sha256
`368fe3d831e6640017e52a56e7af306475b744353fb63a6d38be80b9eb2900ee` before and after.
`tests/foxio_deviations.json` holds 137 keys before and after, against 137 `xfailed`.
No fingerprinter imports `ja4plus/ja4db.py`, and `ja4plus/cli.py` is its one importer
under `ja4plus/`.

## How this was tested

Twelve cases join `tests/test_ja4db.py`. The cases ran first, and 11 of the 12 failed
against the base commit. The twelfth reads the lazy build, which the base already
held.

```
.venv/bin/pytest tests/ -m "not spec_validation"   1428 passed, 8 xfailed  (base: 1416 passed, 8 xfailed)
.venv/bin/pytest tests/ -m spec_validation         1477 passed, 143 skipped, 137 xfailed (unchanged)
.venv/bin/ruff format --check ja4plus/ tests/      124 files already formatted
coverage --cov=ja4plus                             ja4plus/ja4db.py 82% -> 85%, project 85% -> 85%
```

## #60

The two acceptance criteria of #60 hold after this change: one case looks 200000
distinct fingerprints up and the cache holds `cache_size`, and one case proves that
the cache evicts the least recently used entry. #60 keeps its dependency on #57 for
the `source` value, so this pull request does not close it.

Verified against: https://github.com/Crank-Git/ja4plus-go/blob/master/lookup.go
(commit `d06dfd9f8b68dd2cb637bc8c2514dc0795cf50cc`, retrieved 2026-08-08)

## The self-review

Two Sonnet children read the diff, one for concurrency and one for the writing
standard. Three findings changed the code, and they are recorded in Changelog round 83.

- The eviction interval was fixed at 100000, so `JA4DBClient(cache_size=50)` never ran
  its age pass. The interval now reads the smaller of 100000 and the entry count.
- The `in` operator holds an entry and the read that follows holds it again. `lookup`
  now reads the entry once, through `get` against a sentinel value.
- The module docstring of `ja4plus/utils/state_table.py` recorded that nothing calls
  `on_packet` for this table. This change makes that false, and the docstring now
  records the call.

One further finding is recorded and not repaired: a bounded lookup cache can evict a
result that an unbounded dictionary kept, so `result1 is result2` no longer holds for
an unlimited interval. That is the change #42 asks for, and `docs/api_reference.md`
states it.


Crank-Git:
## Self-review by lens

Two child agents read the diff on the Sonnet tier, one for concurrency and one for the writing standard. This comment records every finding and its disposition.

### Concurrency lens

| Finding | Disposition |
|---|---|
| The double-checked read of `_default_client` is correct, and no thread reads a half-built client. | Accepted, no change. |
| Every mutation of the lookup cache runs under `_cache_lock`, and `__contains__` and `__getitem__` both mutate. | Accepted, no change. |
| No deadlock and no reentrancy. `_do_lookup` runs outside both locks. | Accepted, no change. |
| A hit held the entry twice, once through `in` and once through the read that follows. | **Repaired.** `lookup` reads the entry through `get` against a sentinel value. |
| `eviction_interval` stayed at 100000 whatever `cache_size` states, so a small lookup cache never ran its age pass. | **Repaired.** The client reads the smaller of 100000 and `cache_size`. |
| A bounded lookup cache evicts a result an unbounded dictionary kept, so `result1 is result2` holds for no unlimited interval. | Recorded, not repaired. That is the change #42 asks for, and `docs/api_reference.md` states it. |
| `test_an_entry_that_receives_no_read_ages_out` depends on the clock advancing. | Rejected. The case sleeps 10 ms and `time.time()` reads microseconds. |

### Writing standard lens

| Finding | Disposition |
|---|---|
| The new `## Terms` row rejected the bare word `cache`, and the diff then used it. | **Repaired.** The row rejects `state table`, `lookup store` and `result map`, and the new prose spells `lookup cache`. |
| One comment rotated `the cache` and `the table` for one object. | **Repaired.** |
| The module docstring of `ja4plus/utils/state_table.py` states that nothing calls `on_packet` for this table, which this change makes false. | **Repaired.** The docstring records the call. |
| Sentence length, voice, tense, headings, docstring openings and test names are clean. | Accepted, no change. |

### One case measured nothing, and the review of the cases found it

`test_two_threads_that_look_up_at_the_same_time_hold_the_cache_bound` gave each of eight threads 2000 distinct fingerprints. It reported 0 failures out of 30 runs both with the lock and with a no-op in its place, so it measured no lock. `test_eight_threads_that_share_one_client_raise_nothing` replaces that measurement: eight threads share a lookup cache of two entries that ages every entry out, and the case reports 0 failures out of 20 runs with the lock and 17 failures out of 20 runs without it.

Crank-Git:

**The project manager's brief was wrong and this worker measured it.** The plan said the cache bound comes from `docs/specs/features/03-concurrency-safety.md`. **That file names no bound for this table** — its only number is the 10000-entry default per state table, and the `## Terms` table defines a state table as the dictionary a fingerprinter uses for per-connection data, which `JA4DBClient` is not. **The bound is `docs/specs/features/07-db-enrichment.md:134`, which publishes `cache_size: int = 100_000`.** Confirmed on `dev`. That is why FR-concurrency-safety-14 exists beside FR-concurrency-safety-7, and 100000 satisfies both it and FR-db-enrichment-9.

**This is the eighth premise a worker has disproven this session, and the first the project manager wrote itself.** It is exactly why no brief in this project says "do not re-measure".

**Measured independently rather than on the verdict's word.** Clean: `25 passed`. Replacing the double-checked `with _client_lock:` with `if True:` and running the race case **ten times: 10 failures out of 10.** The race is forced rather than sampled, which is the whole point.

**The worker found a vacuous case in its own work, and this is the sixteenth-instance shape.** Its first concurrent case — 8 threads over 2000 distinct fingerprints each — reported **0 failures out of 30 runs both with and without the cache lock.** It measured nothing. The replacement, a two-entry cache that ages every entry out, reports 0 failures of 20 with the lock and **17 of 20 without it**, each a `KeyError` from a concurrent eviction. **A concurrency case that passes because the race never happened also passes in continuous integration**, so catching it here is the only place it gets caught.

**The change to `ja4plus/utils/state_table.py` is a docstring alone**, recording that `JA4DBClient.lookup` calls `on_packet` so the age pass runs on the lookup schedule. No behaviour moved, so #39 cannot be broken by it.

**#60 is left open and untouched, correctly.** Both of its acceptance criteria now hold with cases, but it also depends on #57 for the `source` value. **The project manager agrees: #60 stays open**, and it is not closed by this issue.

The register is untouched at 137 keys against `137 xfailed`, and the replay of all 38 captures is byte-identical at sha256 `368fe3d8…`.

Issue #40, opened 2026-08-06
Add the thread_safe argument and per-fingerprinter locks


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/03-concurrency-safety.md`](docs/specs/features/03-concurrency-safety.md):

> FR-concurrency-safety-1 — `Processor.__init__` accepts a `thread_safe` argument.

> FR-concurrency-safety-2 — The `thread_safe` argument defaults to `True`.

> FR-concurrency-safety-3 — When `thread_safe` is `True`, two threads may call `process_packet` on one processor at the same time.

> FR-concurrency-safety-4 — When `thread_safe` is `True`, a call to `reset` does not corrupt a state table that another thread is reading.

> FR-concurrency-safety-5 — When `thread_safe` is `False`, the processor acquires no lock.

> FR-concurrency-safety-6 — The processor passes its `thread_safe` value to every fingerprinter it builds.

## What to build

The lock is one lock per fingerprinter, not one global lock. Ten fingerprinters holding ten locks let ten threads work at once on different methods.

The lock is a `threading.RLock`, because a fingerprinter may call its own method that also locks.

`thread_safe=False` is a promise the caller makes, not a mode the library checks. The documentation states that a caller who breaks the promise gets undefined results.

There is no lock anywhere in the package today, and `Processor.get_shard_key` exists so a caller can spread traffic across workers. Concurrency is expected; nothing states what is safe.

## Acceptance criteria

- [ ] Eight threads feeding one `Processor()` for 60 seconds produce the same fingerprint set as one thread feeding the same packets.
- [ ] Eight threads feeding one `Processor()` raise no exception.
- [ ] A test that calls `reset` from one thread while eight threads process packets raises no exception.
- [ ] `Processor(thread_safe=False)` acquires no lock, proven by a test that replaces the lock with an object that fails when acquired.

## Links

- Feature: [`docs/specs/features/03-concurrency-safety.md`](docs/specs/features/03-concurrency-safety.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #14
Depends on #39

Crank-Git:
## Plan — Epic 3, integration branch `epic/14-concurrency-and-resource-safety`

[...]

Issue #43, opened 2026-08-06
Add the thread-safety and memory-bound suites and document the contract


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/03-concurrency-safety.md`](docs/specs/features/03-concurrency-safety.md):

> FR-concurrency-safety-15 — The documentation states the concurrency contract.

## What to build

New files `tests/test_thread_safety.py` and `tests/test_memory_bounds.py`.

The contract to state: a processor is safe to share between threads by default; a caller who shards traffic with `get_shard_key` and wants the fastest path constructs `Processor(thread_safe=False)` and must not then share it.

This is the last issue in the epic and proves the whole contract. Epic 8 expands the README text into a documentation page.

## Acceptance criteria

- [ ] `pytest tests/test_thread_safety.py` passes.
- [ ] `pytest tests/test_memory_bounds.py` passes.
- [ ] The README states the concurrency contract in one paragraph.

## Links

- Feature: [`docs/specs/features/03-concurrency-safety.md`](docs/specs/features/03-concurrency-safety.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #14
Depends on #40

Crank-Git:
## Plan — Epic 3, integration branch `epic/14-concurrency-and-resource-safety`

**Changelog round 86.** It is assigned. Do not choose one, and never write `TBD`. **This
issue closes Epic 3.**

**Four members have sub-merged:** `849f58d` #38 `BoundedStateTable`, `ac4ee62` #42 the lookup
client and its bounded cache, `631790a` #39 **thirteen** tables moved, `0d3480e` #40
`thread_safe` and one `RLock` per fingerprinter. **#41 runs beside you** and adds the
statistics; if it sub-merges first, rebase and document what it built.

### The contract must state this, and it is the reason the project manager wrote this plan

**#40 measured that eight threads over five concatenated captures differ from one thread on
4 of 5 runs.** The cause is measured and it is **not a race**: it is the eviction order of
the bounded tables. Per single capture the match is exact on 3 of 3 runs.

**So the contract is narrower than "thread safe" and it must say so:**

- A caller who gives each thread **whole connections**, which is what `Processor.get_shard_key`
[...]

PR #282, opened 2026-08-08
Add the thread_safe argument and one lock for each fingerprinter

Crank-Git:
Closes nothing on its own. Part of #14, and it satisfies issue #40.

## What this changes

`Processor.__init__` accepts `thread_safe`, and it defaults to True. Every
fingerprinter holds one `threading.RLock`. Ten fingerprinters hold ten locks, so ten
threads work at once on ten methods. A processor that reads `thread_safe=False` gives
every fingerprinter the shared `NULL_LOCK`, whose `__enter__` acquires nothing.

## Why the lock is reentrant

`Processor.process_packet` holds the lock of a fingerprinter and then calls
`process_packet` on that fingerprinter, which holds it again. A `threading.Lock` in its
place deadlocks on the first packet, and a run with one times out on 3 of 3 runs.

The processor holds that lock across the read of `last_raw` too. The raw form describes
the value the call produced, so a second thread that runs between the two pairs one
fingerprint with the raw form of another packet.

## Three methods hold no lock, and the measurement decided it

`BaseFingerprinter.add_fingerprint`, `get_fingerprints` and `reset` each run one
operation over the value list. A lock on the three failed 0 of 20 runs with it removed,
and this project ships no lock that nothing measures. `ja4l.py` holds the one composite
over that list, and its `process_packet` holds the lock.

`SynAckTracker` holds no lock either. Every caller of it runs inside a method of
`JA4TSFingerprinter`, and that method holds the lock of that fingerprinter.

## Every lock is proven by its removal

Each row removes one lock, clears `__pycache__`, and runs
`tests/test_thread_safety.py` twenty times.

| Lock removed | Runs that failed |
|---|---|
| Every fingerprinter lock at once | 20 of 20 |
| The processor lock around `last_raw` | 20 of 20 |
| A plain `Lock` in place of the `RLock` | 3 of 3, each a deadlock |
| `JA4Fingerprinter.process_packet` | 20 of 20 |
| `JA4LFingerprinter.process_packet` | 20 of 20 |
| `JA4LFingerprinter.reset` | 20 of 20 |
| `JA4HFingerprinter.process_packet` | 20 of 20 |
| `JA4SSHFingerprinter.process_packet` | 20 of 20 |
| `JA4XFingerprinter.process_packet` | 20 of 20 |
| `JA4SFingerprinter.process_packet` | 20 of 20 |
| `JA4TSFingerprinter.process_packet` | 20 of 20 |

**The first sweep found six locks that nothing measured, and each reported 0 of 20.**
Every case then reached the fingerprinter through `Processor.process_packet`, which
holds the lock of that fingerprinter across every call, so the case measured the
processor and not the fingerprinter. Seven cases now call `process_packet` on the
fingerprinter itself and count the threads inside its guarded region.

**The forced reset case parked on the wrong packet and reported 9 of 20.** It patched
`packet_endpoints`, and the first call came from the SYN-ACK, which reports `JA4L-S` and
reads no stored index. The case now runs the SYN and the SYN-ACK before the patch and
parks the writer on the ACK.

## The first acceptance criterion reads a partition by connection

Eight threads over five concatenated captures produce a fingerprint set that differs
from one thread on 4 of 5 runs. **The cause is the eviction order of the bounded tables,
not a race.** A concatenation replays several independent capture sessions into one
processor, and the interleaving moves which stream a bounded table evicts.

One capture at a time, with `get_shard_key` sending every packet of one connection to
one thread, matches the single-thread list exactly: `latest.pcapng`, `ssh2.pcapng`,
`browsers-x509.pcapng` and `http1.pcapng` each match on 3 of 3 runs. The committed soak
runs 2112 rounds in 60 seconds and every round matches.

## The criterion that no lock is acquired is measured by a pair

A case replaces `threading` inside `ja4plus/fingerprinters/base.py` with a namespace
whose `RLock` returns an object that raises when a caller acquires it.
`Processor(thread_safe=False)` replays 209 packets and raises nothing.
`Processor(thread_safe=True)` raises on the first packet, which is the control that
proves the first case can fail.

**A criterion that reads an exception must read the log.** `Processor.process_packet`
catches every exception a fingerprinter raises and logs it at `DEBUG`, so the two cases
that feed eight threads beside a `reset` read `caplog`. Neither can pass on a swallowed
failure.

## What the lock costs

A replay of 1963 packets over four captures reports a median of 263.9 ms with
`thread_safe=True` and 260.7 ms with `thread_safe=False`. The lock costs **1.22% of the
run, and 1.63 microseconds for each packet**. The default is True, because 1.63
microseconds buys a contract and a caller who shards pays nothing for it.

## How these locks meet the module lock of #42

They never meet. No fingerprinter calls `ja4plus.ja4db.lookup`. `ja4ssh.py` holds
`lookup_hassh`, which reads a local dictionary, and `ja4plus/cli.py` calls the lookup
client in `_output_results`, after `process_packet` returns. No thread holds one lock
while it waits for the other, so the two cannot deadlock.

## No fingerprint moved

A replay of all 38 committed captures dumps **809 values before the change and 809
after, and the two dumps are byte-identical.** The dump holds no wall-clock field, and
two runs on the base commit produced identical bytes, which is the reading that proves
it.

`tests/foxio_deviations.json` holds **137 keys before and after**, and the conformance
suite reports **137 `xfailed` before and after**. This branch changes neither file.

## Re-measured premises

- **The thirteen tables hold.** A recursive walk of `Processor()` reports 13
  `BoundedStateTable` instances over 7 fingerprinters, plus 2
  `TCPStreamReassembler.streams` and 10 `BaseFingerprinter.fingerprints` lists. A first
  walk reported 12, because it read the attribute name `tracker` and JA4TS names its
  tracker `syn_ack_times`.
- **The brief states the two suite results in the wrong order.** The unit suite reports
  1472 passed and 8 `xfailed` on the base commit `631790a`, and the conformance suite
  reports 1477 passed, 143 skipped and 137 `xfailed`.

## How this was tested

| Gate | Before | After |
|---|---|---|
| `pytest tests/ -m "not spec_validation"` | 1472 passed, 8 xfailed | 1494 passed, 8 xfailed |
| `pytest tests/ -m spec_validation` | 1477 passed, 143 skipped, 137 xfailed | 1477 passed, 143 skipped, 137 xfailed |
| `ruff format --check ja4plus/ tests/` | clean | 126 files already formatted |
| `mypy ja4plus/` | clean | no issues in 27 source files |
| Coverage `--cov=ja4plus` | 85%, 523 missed of 3539 | 85%, 522 missed of 3581 |

Verified against: https://docs.python.org/3/library/threading.html#rlock-objects
(CPython 3.14, retrieved 2026-08-08), for the reentrant acquisition of `threading.RLock`
by one thread.

Changelog round 84 records the whole reading in `docs/specs/spec.md`.


Crank-Git:
## Self-review

I ran the review myself. Two review children were not spawned, because the review is a read of my own diff by lens and every finding below is one I fixed or declined in this branch.

**Correctness.** The lock-removal sweep is the review that mattered, and it found two real defects in the first pass.

1. **Six per-fingerprinter locks that nothing measured**, each reporting 0 of 20 runs failed. Every case reached the fingerprinter through `Processor.process_packet`, which holds the lock of that fingerprinter across every call, so each case measured the processor. Fixed by seven cases that call `process_packet` on the fingerprinter itself and count the threads inside its guarded region. Each of the six now fails 20 of 20.
2. **A lock on `BaseFingerprinter.add_fingerprint`, `get_fingerprints` and `reset` that nothing measured**, at 0 of 20. `list.append` and the rebind each run as one operation. Removed rather than kept, because this project ships no lock that nothing measures.
3. **The forced reset case parked on the wrong packet**, at 9 of 20. It patched `packet_endpoints` and caught the SYN-ACK, which reports `JA4L-S` and reads no stored index. Fixed, and it now reports 20 of 20.

**Correctness, declined.** `Processor.close_open_windows` reads `entry["fingerprint"]` after the lock releases. The entries come from JA4SSH alone, and no other fingerprinter rewrites them, so the read is safe. `ja4l.py` rewrites an entry at a stored index, and it rewrites only entries of its own list.

**Security.** The change adds no parser, reads no packet field and reaches no network. `NullLock` holds no state and `__slots__` is empty, so one shared instance serves every caller.

**Backend.** `Processor.__getattr__` still guards on `"fingerprinters" in self.__dict__`, and `self.thread_safe` is set before that key exists. Neither `lock` nor `thread_safe` collides with an attribute a fingerprinter already held.

**Comment accuracy.** One comment in `ja4plus/processor.py` named "the call above" for a call that sits below it. Fixed in d7a0380.

**Whitespace.** `git diff -w` over `ja4l.py`, `ja4h.py` and `ja4x.py` shows the `thread_safe` parameter, the `with self._lock:` lines and comment rewraps, and no other change. The 809-value byte-identical replay over 38 captures is the second reading of the same claim.

Crank-Git:

**Measured independently rather than on the verdict's word.** The thread-safety suite reports
`22 passed` clean. Replacing `self._lock = threading.RLock() if thread_safe else NULL_LOCK`
with `self._lock = NULL_LOCK`, so every fingerprinter lock becomes a no-op, **fails 5 runs
out of 5.** The locks are load-bearing.

The register is untouched at 137 keys, Changelog round 84 is present, and both suites match
the base.

## The disclosure that matters, and what it obliges

**Eight threads over five concatenated captures differ from one thread on 4 of 5 runs, and
the worker measured the cause: the eviction order of the bounded tables, not a race.**
Per single capture the match is exact on 3 of 3 runs.

**That is a true statement about this package after #39, and it is not a defect of #40.**
Bounding a table makes its output order-dependent once traffic exceeds the bound. The
acceptance criterion is met in the arrangement the issue itself names — each thread owning
whole connections through `get_shard_key`, which is what that method exists for.

**It is a contract statement, and #43 owns the contract.** A caller who spreads packets of
one connection across threads, on a workload that exceeds a table bound, may read a
different fingerprint set than a single thread reads. **The project manager has recorded
this on #43** so it lands in the documented contract rather than being discovered later.

## One practice partially met, disclosed rather than hidden

**The mechanical `with self._lock:` insertions landed before the concurrency cases.** The
worker said so plainly instead of claiming tests-first. It then proved every one of the
eleven locks by removal, 20 runs each, which is a stronger measurement than a failing-first
case. **Accepted on that basis.** A practice missed and stated is a different thing from a
practice missed and reported as met.

## What the first sweep exposed, and this is the reason the sweep exists

**Six locks reported 0 failures out of 20 when removed, and so did a base-class lock.** They
measured nothing. The six gained direct-call mutual-exclusion cases and the base-class lock
was deleted. **A lock whose removal fails nothing is not protection; it is cost.**

Two further readings worth keeping: a plain `Lock` in place of the `RLock` deadlocks in 3
runs of 3, which is why the issue specifies `RLock`; and the measured cost of
`thread_safe=True` is **1.22%, or 1.63 microseconds per packet**, over 1963 packets.

**No interaction with #42's module lock.** No fingerprinter calls `ja4plus.ja4db.lookup` —
`ja4ssh.lookup_hassh` reads a local dictionary, and `ja4plus/cli.py` calls the client after
`process_packet` returns — so no thread holds one lock while waiting for the other.

Issue #279, opened 2026-08-08
State the memory ceiling for the one-million-packet case


Crank-Git:
Split from #38 by the project manager on 2026-08-08, so that Epic 3's gate issue can merge
without inventing a published number.

## The question

**What memory ceiling does this package state for the one-million-packet case?**

`docs/specs/features/03-concurrency-safety.md` leaves the ceiling open and asks #38 to
propose one. #38 measured the projection but **did not publish a ceiling**, because a
ceiling is a number this package states to its users and the user has not seen it.

## The measurement #38 produced

Measured on 2026-08-08, against the defaults #38 shipped — 10000 entries and 600 seconds.

| Component | Measured |
|---|---|
| An idle `Processor`, with scapy imported | 91.03 MiB |
| One full JA4SSH-shaped state table | 51.72 MiB |
| Nine JA4L-shaped state tables, 10.69 MiB each | 96.21 MiB |
| **Projection** | **238.96 MiB** |

**#38 proposed 512 MiB**, which is about 2.1 times the projection.

## Why the project manager did not simply take the proposal

512 MiB is defensible and the headroom is reasonable. **It is also a number this package
publishes, and no measurement forces it** — the projection forces a floor near 239 MiB and
nothing forces the distance above it. That is a product judgement.

**Nothing is blocked either way.** #38 ships the measurement and states that the ceiling is
open. #41 reports the statistics an operator reads, and it needs no ceiling to do it.

## What to decide

Confirm **512 MiB**, or name another number, or decide that this package states the
measurement and publishes no ceiling at all.

Crank-Git:
## Decision, 2026-08-08. The package states a 512 MiB ceiling for the one-million-packet case

**The user chose 512 MiB on 2026-08-08**, which is the number #38 proposed.

#38 measured a 238.96 MiB projection against the defaults it shipped, 10000 entries and 600
seconds: an idle `Processor` with scapy imported at 91.03 MiB, one full JA4SSH-shaped state
table at 51.72 MiB, and nine JA4L-shaped tables at 10.69 MiB each for 96.21 MiB. **512 MiB is
about 2.1 times the projection**, and that headroom is the reason the number is not 256.

## What the change writes, and what it must not

**Write the ceiling into `docs/specs/features/03-concurrency-safety.md`**, which leaves it
open today and asks #38 to propose one. State it as a ceiling this package holds at the
shipped defaults, and state the defaults beside it — a ceiling with no stated configuration is
[...]

PR #283, opened 2026-08-08
Report processor statistics

Crank-Git:
`Processor.stats()` returns one `ProcessorStats` for each of the ten methods. Each one
states the packet count of the method and one `TableStats` for each state table it
holds. FR-concurrency-safety-11 and FR-concurrency-safety-12 state the requirement, and
issue #41 owns it.

## What changed

- `ja4plus/utils/state_table.py` gains `StateTable`, the base class every state table
  holds, and `TableStats`, the six counts one table reports.
- `ja4plus/utils/tcp_stream.py` makes `TCPStreamReassembler` inherit `StateTable`.
- `ja4plus/fingerprinters/base.py` gains `BaseFingerprinter.state_tables`.
- `ja4plus/processor.py` gains `ProcessorStats` and `Processor.stats`.
- `tests/test_processor_stats.py` holds 38 cases.
- `docs/api_reference.md`, `docs/specs/features/03-concurrency-safety.md` and
  `docs/specs/spec.md` record the result. Changelog round 85.

## The set of state tables is fifteen, and not thirteen

A walk of `Processor()` reports the 13 `BoundedStateTable` instances that Changelog
round 82 records, and it reports 2 `TCPStreamReassembler` instances beside them, on
`JA4HFingerprinter.reassembler` and `JA4XFingerprinter.reassembler`. Each reassembler
holds per-connection data across packets, so the `## Terms` table of
`docs/specs/spec.md` names each one a state table. Round 82 recorded the reassembler as
an `OrderedDict` that keeps its own bounds, and that reading holds: this pull request
leaves the bounds where they are and adds the counts.

```
flat walk 14
one level walk 15
BoundedStateTable 13
TCPStreamReassembler 2
```

The flat walk misses `SynAckTracker.times`, so `state_tables` descends one level and
reports that table under the name `syn_ack_times.times`. It finds a table by its type
and by no list, because #39 found that a list of tables written in an issue named six
where the code held thirteen.

## How `stats()` reads shared state, and what the guard buys

`stats()` holds the lock of one fingerprinter across the read of that fingerprinter,
and it holds two locks never. `BoundedStateTable.__setitem__` stores the entry and
counts the insert after that store, so a reader between the two steps sees
`inserts == entries + evictions + removals - 1`.

`tests/test_processor_stats.py` forces that race rather than sampling it. A
`PausingTable` holds the writer inside the window with a `threading.Event`, and a
second thread then reads the statistics. Removing the guard means
`Processor(thread_safe=False)`, which gives every fingerprinter the shared `NULL_LOCK`.

```
with the guard: reader returned early 0/20, snapshot broken 0/20
guard removed: reader returned early 20/20, snapshot broken 20/20
```

Two committed cases hold the same experiment:
`test_the_reader_waits_for_the_writer_and_reports_a_consistent_snapshot` and
`test_a_reader_that_acquires_nothing_reports_a_broken_snapshot`.

## How a returned connection differs from a first sighting

`StateTable` holds an `OrderedDict` of the keys it evicted and has not seen since, least
recent first. A key that arrives and sits in that memory is a return. A key that arrives
and does not sit there is a first sighting. A key the caller removed leaves no memory,
so a connection that returns after `cleanup_connection`, `del`, `pop` or `clear` counts
as a first sighting. The caller asked for that removal, and no eviction lost its
packets.

Nothing that survives across packets grows without a limit, so the memory holds the
entry bound of its own table. The fifteen tables hold 46400 remembered keys between
them, and `tracemalloc` reports 187 bytes for one key, so the memory costs 8.3 MiB when
every table is full and every entry has been replaced. Only 4 of the 15 tables carry the
10000 default.

The counters cost 1.8% of one single-threaded replay: 1963 packets over four captures
report a median of 272.8 ms with the counters and 267.9 ms with each counter replaced by
a no-op. One `stats()` call costs 16.8 microseconds.

The order of the three steps of an insert was decided by a failing case. A first form
read the memory inside the counter call, after the eviction pass, and
`test_a_stream_that_returns_after_an_eviction_counts_one` reported 0 where it expects 1
on a table of one entry: the eviction that the arrival drove had already pushed the
arriving key out of a one-key memory. A table now calls `take_evicted_key` before it
evicts, then stores the entry, then counts the insert.

## No fingerprint moved

A replay of all 38 committed captures dumps 809 values before the change and 809 after,
and the two dumps are byte-identical. The dump holds no wall-clock field, and two runs
on the base commit `0d3480e` produced identical bytes, which is the reading that proves
it.

`tests/foxio_deviations.json` holds 137 keys before and after, and this pull request
does not touch it. The conformance suite reports 137 `xfailed` before and after.

## How it was tested

```
pytest tests/ -m "not spec_validation"   1532 passed, 8 xfailed   (base: 1494 passed, 8 xfailed)
pytest tests/ -m spec_validation         1477 passed, 143 skipped, 137 xfailed   (unchanged)
ruff format --check ja4plus/ tests/      127 files already formatted
pytest --cov=ja4plus                     86%, 522 missed of 3676   (base: 85%)
```

## The acceptance criteria

- **`Processor().stats()` returns one entry per method.** Met.
  `test_the_report_holds_one_entry_for_each_method` compares the keys of the report
  against `Processor.fingerprinters` and reads a length of 10.
- **Each entry reports the state-table entry count, the eviction count and the packet
  count.** Met. `test_each_entry_reports_the_entry_count_the_evictions_and_the_packets`
  reads `.entries`, `.evictions` and `.packets` on every entry.
  `test_the_report_names_every_state_table_of_every_method` proves the report names all
  fifteen tables, by a walk that reads none of the code it measures.

## The practices, plainly

The tests followed the reporting code rather than preceding it, so TDD is only partly
met. The eviction, return and guard cases came after the surface they measure. Two of
them then failed against that surface and moved it, and the guard measurement above is
the evidence that they measure something.

Coverage rises from 85% to 86%. Commits follow the conventional form. The documentation
lands with the change.

This project defines no E2E harness, so no E2E specification ships.

## Scope

Part of #14. The suites and the contract document belong to #43. Epic 4 makes
`ProcessorStats` a typed dataclass; a plain object is what this issue ships. The memory
ceiling belongs to #279.


Crank-Git:
## Self-review

Two Sonnet children reviewed the diff by lens, and I ran the vacuity checks myself. Both
lenses returned clean, and each returned one item I acted on.

### Correctness

The reviewer traced every mutation site of `BoundedStateTable._entries` and
`TCPStreamReassembler.streams`, and it found no counter path that is missed. It
confirmed that `evict_aged` and the eviction loop of `__setitem__` mutate the inner
`OrderedDict` directly, so neither reaches `__delitem__` and neither double-counts.

**Finding, accepted.** `BoundedStateTable` inherits `pop`, `popitem`, `setdefault` and
`update` from `MutableMapping`, and no case measured them.
`SynAckTracker.drop` calls `pop`, at `ja4plus/fingerprinters/ja4ts.py:74`, so the path
is live. `test_the_dictionary_operations_the_table_inherits_count_their_removals` now
covers all four, and it reads `pop` with an absent key too.

### Resource safety and concurrency

The reviewer confirmed that every read of `Processor.stats` sits inside `with fp.lock`,
that `Processor._packet_counts` is written and read under the lock of the fingerprinter
that owns its key, that `stats` and `reset` hold one lock at a time, and that no new
code holds a reference to a packet object.

**Finding, accepted.** The memory of the evicted keys was unmeasured. I measured it:
the fifteen tables hold 46400 remembered keys between them, and `tracemalloc` reports
187 bytes for one key, so the memory costs 8.3 MiB when every table is full and every
entry has been replaced. Only 4 of the 15 tables carry the 10000 default:
`ja4s._quic_dcids`, `ja4l.connections`, `ja4l.grouping_keys` and `ja4ssh.connections`.
`docs/api_reference.md` and `docs/specs/features/03-concurrency-safety.md` now state the
figure.

**Finding, declined.** The reviewer read the `move_to_end` call of `count_eviction` as
redundant, because a fresh key lands at the end already. It covers a key the memory
still holds, which happens when a subclass evicts a key that no `take_evicted_key` call
removed. Without it the memory drops the newest key first. A comment now states the
reason.

### Can these cases fail?

`.claude/rules/conformance.md` asks the question, so I measured it rather than asserting
it.

- `StateTable.take_evicted_key` replaced by a function that returns False: 5 cases fail.
- `StateTable.count_eviction` replaced by a no-op: 11 cases fail.

### Cost

The counters cost 1.8% of one single-threaded replay. A replay of 1963 packets over four
captures reports a median of 272.8 ms with the counters and 267.9 ms with each counter
replaced by a no-op, which is 2.50 microseconds for each packet. One `stats()` call
costs 16.8 microseconds.

### Gates after the review

```
pytest tests/ -m "not spec_validation"   1532 passed, 8 xfailed
pytest tests/ -m spec_validation         1477 passed, 143 skipped, 137 xfailed
ruff format --check ja4plus/ tests/      127 files already formatted
```

A replay of all 38 committed captures dumps 809 values, byte-identical to the base
commit `0d3480e`.

Crank-Git:

**Measured independently rather than on the verdict's word.** Clean: `38 passed`. Inserting
an early `return` at the top of `count_eviction` fails **11 cases** — exactly the count the
verdict reports. The register is untouched at 137 keys, Changelog round 85 is present, and
the conformance suite is unchanged.

## The set of state tables is fifteen, and the method of counting is the finding

#179 recorded **six**. #39 measured **thirteen**. This issue measured **fifteen** — the
thirteen plus `JA4HFingerprinter.reassembler` and `JA4XFingerprinter.reassembler`.

**More useful than the number: a flat `vars()` walk reports fourteen and silently misses
`SynAckTracker.times`**, so `state_tables()` descends one level and names it. **A sweep that
looks one level deep is itself a comparison that cannot fail** — it would have reported a
complete answer while missing a table. That is the third time this count has moved, and each
move came from a better method rather than from new code.

## The forced race is real and the proof is committed

`BoundedStateTable.__setitem__` stores the entry and then counts the insert, so a reader
caught between the two sees `inserts == entries + evictions + removals - 1`. The case holds
a writer inside that window on an `Event` while a second thread calls `stats()`.

**With the guard: 0 of 20 broken. With the guard removed: 20 of 20 broken.** Both are
committed cases, not a one-off measurement. `stats()` holds one fingerprinter lock at a time
and never two, so it cannot deadlock against #40's locks or #42's.

## The returned-connection reading, which the issue asked for and is easy to fake

A key that arrives while it sits in the table's evicted-key memory is a **return**; a key
that does not is a **first sighting**; and a key the caller removed itself leaves no memory,
so it is a first sighting. **Two cases prove the distinction in both directions**, and
stubbing `take_evicted_key` to return `False` fails 5.

**The memory of that memory is measured, not assumed:** 46400 remembered keys across the
fifteen tables at 187 bytes each, 8.3 MiB worst case. The counters cost 1.8% over 1963
packets.

## One practice partly met, disclosed

**The reporting surface was written before the cases that measure it.** The worker said so in
the pull request body and in Changelog round 85, and **two of those cases then failed against
the surface and moved it** — which is the evidence that they were not written to fit. The
removal proofs stand in for the ordering. Accepted on the same basis as #40.

## A declined review finding, correctly

The review proposed removing `move_to_end` from `count_eviction` as redundant. **The worker
declined it and left a comment saying why.** A declined finding with a stated reason is worth
more than a silently applied one.

PR #284, opened 2026-08-08
Add the thread-safety and memory-bound suites and document the contract

Crank-Git:
Issue: #43. Part of #14.

Add `tests/test_memory_bounds.py` and `TestTheConcurrencyContract`, and state the
concurrency contract in `README.md` and `docs/api_reference.md`. #43 is the last issue
of Epic 3.

**No file under `ja4plus/` changes.** `git diff -- ja4plus/` is empty, so no fingerprint
can move.

## The contract

The contract is narrower than "thread safe", and it has three clauses.

1. Give each thread whole connections. `Processor.get_shard_key` returns the key that
   arrangement routes on, and eight threads then read the value set one thread reads.
2. A caller that splits the packets of one connection across threads gets undefined
   results.
3. Feed one processor the packets of one timeline.

`thread_safe=False` is a promise the caller makes, not a mode the library checks. The
documentation says so. **The documentation states no memory ceiling**, because #279 owns
that number.

## The premise of round 84 is disproven

Round 84 records that eight threads over five concatenated captures differ from one
thread on 4 of 5 runs, and it names the entry count bound as the cause. A re-measurement
over the same 2047 packets reads **5 of 5**, and it names the **age bound** instead.

| Reading | Runs that differ |
|---|---|
| Both bounds as shipped | 5 of 5 |
| The entry count bound raised, the age bound as shipped | 4 of 5 |
| The age bound raised, the entry count bound as shipped | 0 of 5 |
| Both bounds raised | 0 of 5 |

The third reading still evicts 168 entries of `JA4XFingerprinter.scan_offsets` on its
entry count while the two value lists stay equal, so the entry count bound causes none of
the difference.

**The mechanism is the clock of the concatenation, and it is no race.** The five captures
were recorded across 183 days, so the concatenation jumps the packet timestamp by months,
and `BoundedStateTable.evict_aged` measures an age against the timestamp of the most
recent packet. A concatenation moved onto one timeline reads **0 of 5** and produces 310
values where the jumping concatenation produces 309. **One thread loses that value too**,
which is the reading that removes the race, so the committed case runs on one thread.

**Every reading above was taken twice**, once against `0d3480e` and again against
`0e377b0` after #41 sub-merged. The conclusion holds on both.

Clause 1 reproduces round 84: eight sharded threads match one thread on **3 of 3** runs
for each of `latest.pcapng`, `ssh2.pcapng`, `browsers-x509.pcapng`, `http1.pcapng` and
`tls-sni.pcapng`.

## Clause 1 needs no lock

Every fingerprinter lock removed fails **0 of 20** runs of `TestTheConcurrencyContract`,
and the lock of `Processor.process_packet` removed fails 1 of 20. A thread that
owns whole connections touches keys no other thread touches. This is why
`thread_safe=False` is safe for a caller who shards, and the class docstring records it
rather than implying the class measures a lock. The locks guard the caller who shares one
processor without that arrangement, and round 84 records 20 of 20 for those cases.

## Every case is proven by its removal

| Removal | Runs that fail | Cases that fail |
|---|---|---|
| The entry count eviction of `BoundedStateTable.__setitem__` | 3 of 3 | 8 of 20 memory cases |
| The age pass of `BoundedStateTable.evict_aged` | 3 of 3 | 1 of 20 memory cases |
| The `max_streams` eviction of `TCPStreamReassembler.add_segment` | 3 of 3 | 4 of 20 memory cases |
| The returned connection count of `StateTable.count_insert` | 3 of 3 | 1 of 20 memory cases |
| The nested descent of `BaseFingerprinter.state_tables` | 3 of 3 | 1 of 20 memory cases |
| The direction sort of `Processor.get_shard_key` | 20 of 20 | 4 of 7 contract cases |
| The age pass of `BoundedStateTable.evict_aged` | 20 of 20 | 1 of 7 contract cases |

**The first sweep found one case shape that could not fail.** The memory file held no
case that measured the age bound, and removing the age pass failed 0 of 3 runs.
`TestTheAgeBoundFreesAnIdleConnection` repairs it, and it holds a control that replays
one connection in steps below the age.

**A walk replaces a hardcoded table list.**
`test_the_processor_holds_no_mapping_without_a_bound` walks a live `Processor()` and
fails on any mapping that is neither a `BoundedStateTable` nor one of the two stream
tables, so a table a later change adds cannot pass unread.

**That case earned its keep during this issue.** An earlier form named the two stream
tables in an allowlist, and the rebase onto #41 failed it on `_evicted_keys`, a mapping
that is in fact bounded. The rule now reads the type of the holder rather than a name.

## Acceptance criteria

- [x] `pytest tests/test_thread_safety.py` passes. 29 passed in 8.94 s.
- [x] `pytest tests/test_memory_bounds.py` passes. 20 passed in 2.90 s.
- [x] The README states the concurrency contract in one paragraph. `README.md` holds
      "The concurrency contract" and "The memory bound".

## How this was tested

```
pytest tests/ -m "not spec_validation"   1559 passed, 8 xfailed   (base 0e377b0: 1532 passed, 8 xfailed)
pytest tests/ -m spec_validation         1477 passed, 143 skipped, 137 xfailed   (unchanged)
ruff format --check ja4plus/ tests/      127 files already formatted
coverage --cov=ja4plus                   86%, 520 missed of 3676   (base: 522 of 3676)
tests/foxio_deviations.json              137 keys before and after
```

The two files add 27 cases, which takes the unit suite from 24.8 s to 30.1 s. Round 85
records 1531 for the base and a repeated reading here is 1532.
The soak case keeps the arrangement of #40: it defaults to 2.0 s and reads
`JA4PLUS_THREAD_SOAK_SECONDS`.

## #41 sub-merged, and the contract states what it built

This branch is rebased onto `0e377b0`. `docs/api_reference.md` now holds a `stats()` row,
the six fields of `ProcessorStats`, and the definition of a **returned connection**.

**The table count was re-measured and it is fifteen**, which matches round 85.
`Processor.stats()` reports 15 tables across 10 methods, and the walk of
`tests/test_memory_bounds.py` reads 15 by an independent route.
`test_the_walk_of_this_file_agrees_with_the_report_of_the_processor` compares the two, so
a change that hides a table from one of them fails. The count moved three times, from six
in #179 to thirteen in #39 to fifteen in #41, and each move followed a better walk rather
than new code.

Changelog round 86. No external interface was called, so this pull request cites none.


Crank-Git:
Sub-merge gate, project manager, 2026-08-08. **Accepted. This closes Epic 3.**

**Measured independently.** Clean: `49 passed` across both suites. Inserting an early
`return 0` in `BoundedStateTable.evict_aged` fails **1 case** in the memory suite — the
age-bound case the worker added **only after its own removal sweep proved the file could not
fail without it.** `ja4plus/` is untouched, the register is untouched at 137 keys, and the
README states no memory ceiling.

## The plan's central premise was wrong, and the worker measured the right answer

**Changelog round 84 and the project manager's brief both state that the concatenated
difference comes from "the eviction order of the bounded tables".** Four readings over the
same 2047 packets separate the two bounds:

| Reading | Runs that differ |
|---|---|
| Both bounds as shipped | 5 of 5 |
| Entry count bound raised | 4 of 5 |
| **Age bound raised** | **0 of 5** |
| Both raised | 0 of 5 |

**The age bound alone causes it. The entry count bound causes none of it** — the third
reading still evicts 168 `scan_offsets` entries on its entry count while the value lists stay
equal.

**The mechanism is the measurement's own artifact:** the five captures span **183 days**, so
concatenating them jumps the clock and `evict_aged` ages out state the next capture needs.
Placing the captures on one timeline reads 0 of 5.

**And one thread loses a value too — 309 against 310.** That is what removes "race" as an
explanation entirely, so the committed case runs single-threaded.

**This is the ninth premise a worker disproved this session and the second the project
manager wrote.** The Changelog is append-only, so round 84 stands and **round 86 carries the
correction**, which is the form rounds 42 and 43 already set.

## Sharding needs no lock, and that is the real argument for `thread_safe=False`

**Removing every fingerprinter lock fails 0 of 20 runs of the contract class.** A thread that
owns whole connections touches keys no other thread touches. **The worker recorded that in
the class docstring rather than letting the cases imply they measure a lock** — a case that
passes with and without the guard is exactly the shape this project has recorded sixteen
times, and naming it is better than deleting it.

## Two gates caught real things during this issue

- **Its first memory suite could not fail on the age bound**, 0 of 3 on removal. Repaired
  with a case and a control.
- **`test_the_processor_holds_no_mapping_without_a_bound` failed on the rebase**, because #41
  added `_evicted_keys`. **The case worked as designed.** The rule now reads the holder's type
  rather than a name allowlist, which is the stronger form.

## One discrepancy to record

The worker measures the base unit suite at **1532 passed** where Changelog round 85 records
**1531**. Round 85 is merged and the Changelog is append-only, so **round 86 carries the
corrected figure.** Nothing rests on it.

Issue #285, opened 2026-08-08
Bound the JA4TS prefixes table that the Epic 3 merge orphans


Crank-Git:
Found by the project manager at the Epic 3 gate on 2026-08-08, resolving the merge of `dev`
into `epic/14-concurrency-and-resource-safety`.

## The defect the merge would introduce

**#246 added a sixteenth state table that Epic 3 never saw.** The two branches were live at
the same time and neither could read the other.

`ja4plus/fingerprinters/ja4ts.py` on `dev` holds `SynAckTracker.prefixes`, a plain dictionary
keyed by connection that stores part a through part d of the first SYN-ACK, which the RST
value of #246 reads. **On `dev` it is bounded transitively and correctly**: `_expire` and
`_evict` bound `times`, and both call `drop(key)`, which removes the key from `times` **and**
from `prefixes`.

**Epic 3 removes that mechanism.** #39 replaced `self.times = {}` with a `BoundedStateTable`,
and #38's table **evicts internally** — it calls neither `drop` nor any caller hook. So after
the merge:

- `times` stays bounded.
- **`prefixes` grows without limit**, because nothing removes a key that `times` evicted.

**That is the exact rule Epic 3 exists to enforce**, stated in `CLAUDE.md`: a structure that
survives across packets holds a maximum entry count and a maximum age.

## Why no gate caught it

Both branches were green. **#39 and #41 swept for state tables and found thirteen and then
fifteen — neither sweep could see `prefixes`, because it did not exist on their base.**
`tests/foxio_deviations.json` does not move and no fingerprint changes, so no conformance
case fails either.

**Two live integration branches cannot see each other.** Session 9 recorded the same shape
and the batch gate caught it then; the gate caught it again here.

## What to build

- [ ] Merge `origin/dev` into `epic/14-concurrency-and-resource-safety` and resolve every
      conflict in `ja4plus/fingerprinters/ja4ts.py`.
- [ ] **`prefixes` holds a maximum entry count and a maximum age**, in lockstep with `times`,
      so a key evicted from one is absent from the other.
- [ ] **Prove the leak exists before you fix it.** Build the case that fills `times` past its
      bound and reads `len(prefixes)`; show it fails on the merged tree before your repair
      and passes after.
- [ ] `Processor.stats()` reports the table, and the count of state tables rises from 15.
      **Re-measure that count; do not assume it becomes 16.**
- [ ] The `reset` conflict keeps Epic 3's lock and clears both tables.
[...]

PR #286, opened 2026-08-08
Bound the JA4TS prefix table that the Epic 3 merge orphans

Crank-Git:
Merges `origin/dev` into `epic/14-concurrency-and-resource-safety` and repairs the state
table that the merge orphans. Issue #285.

## The merge

Four conflicts. Three in `ja4plus/fingerprinters/ja4ts.py` and one in
`docs/specs/spec.md`, which matches what the project manager measured.

1. `SynAckTracker.__init__` keeps the `BoundedStateTable` of #39 and the `prefixes` map
   of #246.
2. `SynAckTracker._expire` and `SynAckTracker._evict` go, because `BoundedStateTable`
   runs both passes. Every caller was re-read. `record` already reached
   `times.on_packet(now)` through the automatic merge. `reset_value` called
   `_expire(now)` and now calls `times.on_packet(now)`, so a RST packet still drives the
   age pass.
3. `JA4TSFingerprinter.reset` keeps the lock of #39 and the docstring of #246, and
   `SynAckTracker.clear` empties both tables.
4. The Changelog appends in numeric order: rounds 76 to 80 from batch 21, then 81 to 86
   from Epic 3, then round 87 for this issue. No row is dropped and none is renumbered.

Git flagged no conflict at two call sites of `SynAckTracker.record` in
`tests/test_bounded_state_table_adoption.py`, and both read
`TypeError: SynAckTracker.record() missing 1 required positional argument: 'prefix'`.
The merge commit states the third argument at each.

## The defect, measured before the repair

Commit `fe83bbb` holds the merge with no repair. Four cases fail on it, and two of the
four report the leak.

```
tests/test_memory_bounds.py::TestTheProcessorHoldsNoUnboundedStateTable::test_the_processor_holds_no_mapping_without_a_bound
E   AssertionError: these mappings hold no bound: ['ja4ts.syn_ack_times.prefixes']

tests/test_ja4ts_reset.py::TestTheResetBounds::test_the_stored_parts_hold_one_entry_for_each_tracked_connection
E   AssertionError: 1100 not less than or equal to 1000
```

`tests/test_ja4ts_prefix_bound.py` is the case this issue asks for. Against `fe83bbb` it
reads 5 failed, 3 passed:

```
E   AssertionError: assert 60 <= 25        # the entry count bound, prefixes holds 60
E   AssertionError: assert 2 == 1          # the age pass leaves the prefix behind
E   assert False = isinstance({}, BoundedStateTable)
E   AssertionError: assert ['syn_ack_times.times'] == ['syn_ack_times.prefixes', 'syn_ack_times.times']
```

Against this branch it reads 10 passed.

Reproduce the failing state:

```
git checkout fe83bbb -- ja4plus/fingerprinters/ja4ts.py ja4plus/utils/state_table.py
.venv/bin/pytest tests/test_ja4ts_prefix_bound.py -q
```

## How the two tables stay in lockstep, and what that costs

`BoundedStateTable` gains an eviction hook. That is a change to the structure #38 built,
and every case of #38 still passes.

- `on_eviction` is a new constructor argument. The table calls it with the key of every
  entry it evicts, on either bound. A caller removal through `pop`, `del` or `clear`
  calls nothing, because that caller already knows.
- `evict_key(key)` removes one entry, counts it as an eviction, and calls the hook. Both
  eviction sites run through it, so the count bound and the age pass report alike.
- `SynAckTracker.prefixes` is a `BoundedStateTable`, and the hook of `times` removes the
  key from it. `record` stores into `times` before `prefixes`, so the hook runs before
  the new prefix arrives and `prefixes` never passes the count of `times`.

**The cost.** `BoundedStateTable` grows one argument and one method, and a hook runs
inside an eviction, so a hook that writes to the same table produces undefined results.
The class docstring states that. `prefixes` runs no age pass of its own and never
reaches its own count bound, so both of its bounds are a second limit that the hook
keeps the table away from. `prefixes` also holds its own memory of evicted keys, at
1000 keys, which raises the worst-case memory of one processor from 8.3 MiB to 8.5 MiB.

**Three other shapes were read and declined, with the cost of each.**

1. **A second table with its own bounds and no hook.** Cost: it loses lockstep. A read
   moves the least recently used order of `times` and not of `prefixes`, so at the count
   bound the two evict different keys. Measured:
   `test_a_read_of_one_table_moves_no_key_out_of_step_with_the_other` fails, and
   `test_the_rst_value_survives_an_eviction_that_the_entry_count_drives` reads
   `KeyError: ('10.0.0.2', 443, '10.0.0.1', 50000)`.
2. **One table whose value holds the pair.** Cost: lockstep is free and correct, and
   `Processor.stats()` then reports fifteen tables rather than sixteen. The prefix data
   would carry no separate entry count and no separate eviction count, and the
   acceptance criterion of #285 asks for both.
3. **A sweep that drops from `prefixes` every key `times` does not hold.** Cost: the
   sweep reads every entry on every packet, so it costs the entry count on every packet.

## The state table count

Re-measured against the code, not assumed. `Processor.stats()` reports **16** tables
across 10 methods: 14 `BoundedStateTable` instances and the 2 `TCPStreamReassembler`
instances. The independent walk of `tests/test_memory_bounds.py` reads 16 as well, and
`test_the_walk_of_this_file_agrees_with_the_report_of_the_processor` compares the two.
`report["ja4ts"].tables` reads `['syn_ack_times.prefixes', 'syn_ack_times.times']`.

## Proven by removal

With `self.on_eviction(key)` removed from `BoundedStateTable.evict_key`, and
`__pycache__` cleared between the runs, the suite reads 5 failed, 1636 passed:

```
FAILED tests/test_ja4ts_prefix_bound.py::test_the_age_pass_removes_the_prefix_of_an_aged_connection
FAILED tests/test_ja4ts_prefix_bound.py::test_a_read_of_one_table_moves_no_key_out_of_step_with_the_other
FAILED tests/test_ja4ts_prefix_bound.py::test_the_rst_value_survives_an_eviction_that_the_entry_count_drives
FAILED tests/test_state_table.py::TestTheEvictionHook::test_the_table_calls_the_hook_on_the_entry_count_bound
FAILED tests/test_state_table.py::TestTheEvictionHook::test_the_table_calls_the_hook_on_the_age_pass
```

## The invariant holds

- `tests/foxio_deviations.json` holds **137 keys** and the file does not move.
  `git diff` against the merge base reports no change to it. The conformance suite reads
  **137 xfailed**.
- A replay of the **38** committed captures produces **797** values on `origin/dev` and
  **797** on this branch, and every value matches. The record holds the capture name,
  the method, the fingerprint and the endpoint pair, and it holds no timestamp, so it
  holds no wall-clock field and no capture clock field.
- The RST value of #246 is still produced. `tests/test_ja4ts_reset.py` passes whole, and
  `test_the_rst_value_survives_an_eviction_that_the_entry_count_drives` reads
  `64240_2_1460_0_0-R0`.

## The five gates

```
.venv/bin/pytest tests/ -m "not spec_validation"   1641 passed, 8 xfailed, 93 subtests passed
.venv/bin/ruff format --check ja4plus/ tests/      133 files already formatted
```

Coverage with `--cov=ja4plus`: **86%** (3733 statements, 519 missed). The epic base
`8beea06` reads 86% (3676, 520) and `origin/dev` reads 85% (3519, 521). Coverage does
not fall.

## Documentation

- `docs/specs/spec.md`: Changelog round 87; two new terms, `eviction hook` and
  `lockstep`; #285 added to the Epic 3 row of the issue map.
- `docs/specs/features/03-concurrency-safety.md`: the table row for
  `SynAckTracker.prefixes`, the lockstep mechanism, the count 15 to 16, the memory
  figure, and the removal proof.
- `docs/api_reference.md`: the count 15 to 16 in three places, the memory figure, and
  the `on_eviction` and `evict_key` rows of the `state_table` table.

## Practices

- **TDD**: met. `tests/test_ja4ts_prefix_bound.py` was written and run red against the
  merge commit `fe83bbb` before the repair.
- **Coverage must not fall**: met. 86% against 86% on the epic base.
- **Conventional commits**: met.
- **Documentation lands with the change**: met.

No external interface was called, so this change cites none.


Crank-Git:
## Self-review

I ran this review myself and spawned no review child. Every finding below reads the diff
of `54a76a4` against `8beea06`.

### Correctness

1. **`evict_key` runs inside the count-bound loop of `__setitem__`, and the hook could
   re-add the key.** That would loop without end. Not a defect for the one caller:
   `SynAckTracker._drop_prefix` writes another table. The class docstring now states
   that a hook which writes to its own table produces undefined results. **Kept.**
2. **`evict_aged` calls `evict_key` while it walks the entries.** The walk reads
   `list(self._entries.items())`, a snapshot, so the mutation reaches no live iterator.
   **No change.**
3. **`next(iter(self._entries))` replaces `popitem(last=False)`.** An empty mapping would
   raise `StopIteration`. The loop guard is `len(self._entries) >= self.max_connections`
   and the constructor refuses a count below one, so the mapping holds an entry whenever
   the line runs. **No change.**
4. **`take_evicted_key` still runs before the eviction loop.** `StateTable` states that
   order, because an eviction can drop the memory of the key that arrives. The repair
   moved neither call. **No change.**
5. **`SynAckTracker.reset_value` reads `self.prefixes[key]` with no fallback.** A broken
   lockstep would raise `KeyError`, and `generate_ja4ts` catches `ValueError`,
   `TypeError`, `IndexError` and `AttributeError`, so the error would reach the caller.
   A fallback to `None` would hide the state instead of reporting it, and no path
   produces it. **Recorded in the code rather than changed**, commit `7e1d3a5`.
6. **`prefixes` runs no age pass of its own.** Nothing in `ja4plus/` calls `evict_aged`
   across every table, so its `_now` stays None and its wall clock never reads. If a
   later caller ran that pass, `prefixes` would empty and finding 5 would fire. The
   class docstring of `SynAckTracker` states the reading. **No change.**
7. **`SynAckTracker.drop` fires no hook.** It calls `pop` on both tables itself, so the
   two stay in step and neither eviction count moves. That is the correct reading:
   the caller asked for the removal. **No change.**

### Security

8. **A flood of crafted connections cannot grow `prefixes`.** Measured:
   `test_the_prefix_table_stays_inside_the_entry_count_bound` reads 60 connections
   against a bound of 25 and holds 25.
9. **`evict_key` returns False for a key the table does not hold.** A hook that arrives
   twice for one key evicts once and raises no error.

### Interface

10. **`on_eviction` defaults to None**, so every existing construction of
    `BoundedStateTable` reads the same behaviour. The 13 other tables pass no hook.
11. **The eviction count of `prefixes` rises through the hook**, not the removal count.
    `evict_key` exists for that reason: FR-concurrency-safety-12 asks for the count of
    entries the table lost without the caller asking, and a `pop` inside the hook would
    have reported zero evictions forever on a table that loses entries constantly.

### Cases

12. **Two of my first cases could not fail.** `test_the_two_tables_hold_the_same_keys`
    and `test_the_prefix_table_stays_inside_the_entry_count_bound` both pass with the
    hook removed, because `prefixes` holds a count bound of its own and the flood
    inserts in one order with no read between. I kept both as boundary cases and added
    `test_a_read_of_one_table_moves_no_key_out_of_step_with_the_other` and
    `test_the_rst_value_survives_an_eviction_that_the_entry_count_drives`, which
    reorder one table with a retransmission. Both fail with the hook removed, and the
    second reads `KeyError: ('10.0.0.2', 443, '10.0.0.1', 50000)`.

Crank-Git:
Sub-merge gate, project manager, 2026-08-08. **Accepted. Epic 3 is now ready for its gate.**

**Measured independently.** On the merged tree: `1477 passed, 143 skipped, 137 xfailed` and
`1641 passed, 8 xfailed`. `origin/dev` is an ancestor of this branch, the register is
byte-identical to `dev` at 137 keys, and the Changelog reads **76 to 87 in numeric order**
with no row dropped or renumbered.

## The defect was real and the proof reproduces

Removing `on_eviction=self._drop_prefix` from `SynAckTracker.__init__` fails **3 cases** of
`tests/test_ja4ts_prefix_bound.py`. The worker's own red run against the merge commit
`fe83bbb` read `AssertionError: assert 60 <= 25`.

## A precise reading, because the worker's claim is slightly stronger than what I measure

The verdict says #43's memory suite caught the leak, and **that is true of the original
form** — on `fe83bbb`, `prefixes` was a plain dictionary and
`test_the_processor_holds_no_mapping_without_a_bound` named it:
`these mappings hold no bound: ['ja4ts.syn_ack_times.prefixes']`.

**It does not catch the loss of lockstep.** With the hook removed but `prefixes` still a
`BoundedStateTable`, that suite reports **20 passed**. Measured just now.

**That is correct behaviour, not a gap.** They are two different properties: "every mapping
holds a bound" and "two tables evict the same keys". **The second needed the dedicated cases
this issue wrote**, and the boundary between them is worth having in the record so a later
reader does not assume the memory suite covers both.

## What this issue found that the brief did not name

**Two semantic conflicts git did not flag.** Two `SynAckTracker.record` call sites in
`tests/test_bounded_state_table_adoption.py` read
`TypeError: record() missing 1 required positional argument: 'prefix'`. **Git reports a
conflict where two branches edit the same lines; it reports nothing where one branch changes
a signature the other branch calls.** The project manager's conflict count named three hunks
and one file, and it was incomplete.

## The count of state tables is sixteen, and it was measured rather than assumed

The brief said "do not assume it becomes 16". The worker walked a live `Processor()` with a
script reading the report and the attributes independently: **16 reported, 14 bounded plus 2
reassemblers.** `STATE_TABLE_COUNT` and `BOUNDED_TABLE_COUNT` are pinned, and a case compares
the walk against the report so a change hiding a table from either one fails.

**Six in #179, thirteen in #39, fifteen in #41, sixteen here.** Four counts, and every rise
came from a better method or from work landing on a branch the previous count could not see.

## The RST value survives

`tests/test_ja4ts_prefix_bound.py::test_the_rst_value_survives_an_eviction_that_the_entry_count_drives`
reads `64240_2_1460_0_0-R0`, and the 38-capture replay produces **797 values on `dev` and 797
here, byte-identical**.

Issue #44, opened 2026-08-06
Add FingerprintResult with deprecated item access


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/04-typed-api.md`](docs/specs/features/04-typed-api.md):

> FR-typed-api-1 — The package exports a frozen dataclass named `FingerprintResult`.

> FR-typed-api-2 — `FingerprintResult` carries the fields the spec's data model lists.

> FR-typed-api-5 — A `FingerprintResult` supports item access by field name, so that code written against the dictionary keeps working for one major version.

> FR-typed-api-6 — Item access on a `FingerprintResult` emits a `DeprecationWarning`.

## What to build

New file `ja4plus/types.py`.

```python
@dataclass(frozen=True)
class FingerprintResult:
    type: str
    fingerprint: str
    raw: str | None = None
    raw_original_order: str | None = None
    src_ip: str = ""
    src_port: int = 0
    dst_ip: str = ""
    dst_port: int = 0
    timestamp: datetime | None = None
```

The field names are the snake-case form of the port's `FingerprintResult` struct, under parity rule 2. The field that names the method is `type`, not `method`, because the port calls it `Type` and the current dictionary already uses that key. `timestamp` is the field the port carries and this project does not.

The result is frozen: it describes something that already happened. Item access covers reading only; a result does not support item assignment.

Verified against https://github.com/Crank-Git/ja4plus-go/blob/master/types.go (retrieved 2026-08-06).

## Acceptance criteria

- [ ] `result.type` and `result["type"]` return the same value.
- [ ] `result["type"]` emits a `DeprecationWarning`.
- [ ] `result["method"]` raises `KeyError`.
- [ ] `result.fingerprint = "x"` raises `FrozenInstanceError`.
- [ ] A test asserts that `FingerprintResult` field names match the list in `docs/specs/spec.md`.

## Links

- Feature: [`docs/specs/features/04-typed-api.md`](docs/specs/features/04-typed-api.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #15

Crank-Git:
## Plan

**Approach.** Add `ja4plus/types.py` holding the frozen dataclass `FingerprintResult` with the
[...]

Issue #15, opened 2026-08-06
Epic 4: Typed public interface


Crank-Git:
## Goal

Give version 1.0.0 an interface that does not change again before version 2.0.0.

## Exit criteria

`mypy --strict ja4plus/` reports nothing. `py.typed` ships in the wheel. `Processor.process_packet` returns typed results whose field names match the port. The processor exposes the parse failures it collected. `ja4plus/collector.py` is removed.

## Spec

- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Feature: [`docs/specs/features/04-typed-api.md`](docs/specs/features/04-typed-api.md)

Depends on: #14

## Sub-issues

- [x] #44
- [x] #45
- [x] #46
- [x] #47
- [x] #48

Crank-Git:
Unblocked. **Epic 3 (#14) closed on 2026-08-08**, merged into `dev` as `ed7d83f` with nine of nine continuous-integration jobs green.

**#48 is closed as absorbed by #191**, decided by the user on 2026-08-08, so this epic's remaining sub-issues are **#44, #45, #46 and #47**. `docs/specs/spec.md` still lists #48 in the epic child table; that table records the specification's original decomposition and Changelog round 71 records the absorption, so it is history rather than a schedule.

**Read #41 before scheduling #44.** It shipped `ProcessorStats` as a plain object and recorded that **Epic 4 makes it a typed dataclass** — that work belongs here, and #44 or #45 should name it.

PR #292, opened 2026-08-08
Return typed results from the processor and expose parse failures

Crank-Git:
## What

`Processor.process_packet` returns a list of `FingerprintResult` in place of a list of
dictionaries. The new `Processor.process_packet_with_errors` returns the results and the
errors the fingerprinters raised.

```python
def process_packet(self, packet: Packet) -> list[FingerprintResult]: ...
def process_packet_with_errors(
    self, packet: Packet
) -> tuple[list[FingerprintResult], list[Exception]]: ...
```

The method order matches the port under parity rule 2. `process_packet` delegates to
`process_packet_with_errors` and drops the errors, so one code path builds the results.

## Why

`ja4plus/processor.py:81` swallowed every fingerprinter exception and logged it at
DEBUG, so a caller could not tell a packet that produces no fingerprint from a packet
that failed a parse. `docs/specs/features/04-typed-api.md` states FR-typed-api-3 and
FR-typed-api-4.

Verified against: https://github.com/Crank-Git/ja4plus-go/blob/master/processor.go
(retrieved 2026-08-06, through the reading the issue body records).

## How tested

TDD. `tests/test_processor_typed_results.py` came first and reported
`13 failed, 1 passed` against the base branch. The one that passed states that
`process_packet` raises no error when a fingerprinter raises, which the base branch
already satisfied. The other thirteen state the new behaviour. The three cases of the
second commit came first too, and they reported `3 failed`.

All five gates pass, and `pytest` also passes under `-W error::DeprecationWarning`.

| Gate | Result |
|---|---|
| `pytest tests/ -m "not spec_validation"` | 1673 passed, 8 xfailed, 93 subtests passed |
| `pytest tests/ -m spec_validation` | 1477 passed, 143 skipped, 137 xfailed |
| `ruff format --check ja4plus/ tests/` | 136 files already formatted |
| `mypy ja4plus/` | Success: no issues found in 28 source files |

**No fingerprint value moved.** The conformance suite reported
`1477 passed, 143 skipped, 137 xfailed` on the base branch before the change, and it
reports the same three numbers after it.

**Coverage rose.** Measured with `pytest tests/ --cov=ja4plus` on the base commit
`af07a1a` and on this branch.

| | `ja4plus/processor.py` | Whole package |
|---|---|---|
| Base | 92% (113 statements, 9 missed) | 86.19% (3758 statements, 519 missed) |
| This branch | 95% (133 statements, 6 missed) | 86.34% (3778 statements, 516 missed) |

The six statements this branch still misses are all pre-existing: the `AttributeError`
guard of `__getattr__`, and the two error handlers of `close_open_windows` and
`cleanup_connection`.

## A defect this branch found and fixed

The second commit fixes a defect the first commit introduced.

`CLAUDE.md` states: "No code holds a reference to a packet object after `process_packet`
returns." An exception reaches the caller with its `__traceback__`, the traceback holds
the frame of every call it passed, and those frames hold the packet as a local. A
monitor that keeps the errors of every packet would therefore hold every packet it read
— the exact failure mode the rule exists to stop.

`process_packet_with_errors` now clears the traceback of every error it returns, and of
every error in the `__cause__` and `__context__` chain, because a fingerprinter that
raises inside an `except` block chains a second error whose traceback holds its own
frames. **The type, the message and the chain stay**, so `repr(error)` reads the same and
FR-typed-api-4 still holds. `docs/api_reference.md` carries the warning, and a caller who
needs the stack logs the error inside the loop that reads it.

`tests/test_processor_typed_results.py` measures it with a weak reference to the packet.
The case fails without the fix.

## Decisions a reviewer should check

1. **`close_open_windows` still returns dictionaries.** A window entry carries a
   `connection` key, and no `FingerprintResult` field holds a connection key. The issue
   names `process_packet` alone. `docs/api_reference.md` and `CHANGELOG.md` state this.
2. **`timestamp` holds `None`.** The processor reads no packet timestamp today, and
   `packet.time` of a synthesized scapy packet holds the time the object was built
   rather than a capture time. Setting it from that value would report a timestamp the
   packet does not carry. `docs/api_reference.md` states this.
3. **`ProcessorStats` stays in `ja4plus/processor.py`.** The feature file lists it under
   `ja4plus/types.py`, but the issue does not, so this leaves it for a later issue of
   Epic 4.
4. **The returned errors carry no traceback**, for the reason above. A reviewer who
   wants the stack should say so, because that trades the packet rule against it.

## A premise of the handoff brief that measurement contradicts

The brief stated that `ja4plus/cli.py`, the test suite and the conformance harness all
read the result of `Processor.process_packet`. Two of the three do not.

- `ja4plus/cli.py` calls `Processor` nowhere. Its three call sites,
  `ja4plus/cli.py:129`, `ja4plus/cli.py:235` and `ja4plus/cli.py:296`, call
  `fp.process_packet` and `fp.close_open_windows` on one fingerprinter. A fingerprinter
  returns a string, and this change does not touch it. `cli.py` needed no edit.
- The conformance harness `tests/conformance_index.py` imports no `Processor` either.
  Its one call site, `tests/conformance_index.py:234`, calls
  `fingerprinter.close_open_windows()`.
- Every file under `examples/` drives one fingerprinter too, so none needed an edit.

The measurement used `pytest -W error::DeprecationWarning` to find every caller rather
than a grep. It named four test files, and all four now read the attribute:
`tests/test_processor.py`, `tests/test_ja4ts_part_e.py`, `tests/test_ja4ts_reset.py`
and `tests/test_thread_safety.py`.

One caller broke rather than warned. `tests/test_processor.py` asserted
`"fingerprint" in r`. A `FingerprintResult` defines `__getitem__` and no `__contains__`,
so `in` falls back to the sequence protocol, calls `__getitem__(0)` and raises
`KeyError`. The case now reads the attribute and asserts the endpoint values, which the
dictionary form never checked.

`README.md` carried a second break. Its example read `r.get("raw")`, and a
`FingerprintResult` holds no `get` method. The example now reads `r.raw`.

Part of #15. Depends on #44.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,11 @@ project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   results alone, so a caller could not tell a packet that produces no fingerprint from a
   packet that failed a parse. The new method returns both lists, and one method that
   raises poisons no other method. The Go port returns the pair from one call, and parity
-  rule 2 keeps it. `docs/specs/features/04-typed-api.md` states FR-typed-api-4.
+  rule 2 keeps it. **Every returned exception carries no traceback.** A traceback holds
+  the frame of every call it passed, and those frames hold the packet, so a monitor that
+  keeps the errors of every packet would hold every packet it read. The type, the
+  message and the error chain stay. `docs/specs/features/04-typed-api.md` states
+  FR-typed-api-4, and `CLAUDE.md` states the packet rule.
 
 - **`FingerprintResult` is the typed result of the public interface** (#44). The new
   module `ja4plus/types.py` holds a frozen dataclass with nine fields, and `ja4plus`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,28 @@ project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Changed
+
+- **`Processor.process_packet` returns a list of `FingerprintResult`** (#45). Round TBD.
+  The method returned a list of dictionaries through version 0.6.0. A caller who reads
+  `result["fingerprint"]` keeps working for one major version, and item access emits a
+  `DeprecationWarning` that names the attribute form. Read `result.fingerprint` instead.
+  The results follow the fixed method order `ja4`, `ja4s`, `ja4h`, `ja4t`, `ja4ts`,
+  `ja4l`, `ja4x`, `ja4ssh`, `ja4d`, `ja4d6`, and that order is part of the interface.
+  `Processor.close_open_windows` still returns a list of dictionaries, because a window
+  carries a connection key and no `FingerprintResult` field holds one. The processor
+  reads no packet timestamp, so `timestamp` holds `None`. No fingerprint value moved:
+  the conformance suite reports the same 1477 passed, 143 skipped and 137 xfailed.
+  `docs/specs/features/04-typed-api.md` states FR-typed-api-3.
+
 ### Added
+
+- **`Processor.process_packet_with_errors` returns the results and the errors** (#45).
+  Round TBD. `process_packet` logs a fingerprinter error at DEBUG and returns the
+  results alone, so a caller could not tell a packet that produces no fingerprint from a
+  packet that failed a parse. The new method returns both lists, and one method that
+  raises poisons no other method. The Go port returns the pair from one call, and parity
+  rule 2 keeps it. `docs/specs/features/04-typed-api.md` states FR-typed-api-4.
 
 - **`FingerprintResult` is the typed result of the public interface** (#44). The new
   module `ja4plus/types.py` holds a frozen dataclass with nine fields, and `ja4plus`
@@ -16,8 +37,7 @@ project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   code written against the dictionary of version 0.6.0 keeps working for one major
   version. **Item access emits a `DeprecationWarning` that names the attribute form.**
   The item access covers reading only, and a result supports no item assignment.
-  `Processor.process_packet` still returns a list of dictionaries, and #45 changes it.
-  No fingerprint value moved. `docs/specs/features/04-typed-api.md` states
+  #45 makes `Processor.process_packet` return these results. No fingerprint value moved. `docs/specs/features/04-typed-api.md` states
   FR-typed-api-1, FR-typed-api-2, FR-typed-api-5 and FR-typed-api-6.
 
 - **JA4SSH emits the window a connection holds open when the capture ends**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,10 +28,10 @@ project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   packet that failed a parse. The new method returns both lists, and one method that
   raises poisons no other method. The Go port returns the pair from one call, and parity
   rule 2 keeps it. **Every returned exception carries no traceback.** A traceback holds
-  the frame of every call it passed, and those frames hold the packet, so a monitor that
-  keeps the errors of every packet would hold every packet it read. The type, the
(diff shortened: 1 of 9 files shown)


Crank-Git:
## Self-review, lens 1: correctness

I ran a correctness review over commit `e65a3fa`. It returned five findings. Four are
non-issues, and one is real. This comment records all five, so a later reader sees what
the review checked rather than only what it found.

**1. A remaining caller that treats a result as a dictionary — non-issue.**
The review swept `ja4plus/` and `tests/`. Every remaining `result["..."]` reads either
`Processor.close_open_windows`, which still returns dictionaries, or `tests/test_types.py`,
which measures the deprecated `__getitem__` on purpose. `ja4plus/cli.py` calls `Processor`
nowhere.

**2. Thread safety — non-issue.**
`process_packet_with_errors` holds `fp.lock` across the packet count, the call and the
read of `last_raw`, which is the invariant the earlier code held. `process_packet` takes
no lock of its own, so the delegation introduces no second acquisition and loses no
packet count.

**3. A test that would pass on a broken implementation — none found.**
The review named no vacuous case. It checked the cases that monkeypatch a fingerprinter
and reported that they measure the ordering and the error handling of the processor,
which is what they claim to measure.

**4. A returned error holds the packet alive — REAL. Fixed in `710006e`.**
`errors.append(e)` handed the caller an exception whose `__traceback__` holds the frame
of `process_packet_with_errors`, and that frame holds `packet` as a local. The review
proved it: after `del pkt; gc.collect()` the packet was still alive.

`CLAUDE.md` states "No code holds a reference to a packet object after `process_packet`
returns", and a monitor that keeps the errors of every packet would hold every packet it
read. Commit `710006e` clears the traceback of the returned error and of every error in
its `__cause__` and `__context__` chain. The type, the message and the chain stay, so
FR-typed-api-4 still holds. Three cases now measure the rule with a weak reference, and
all three fail without the fix.

**5. A moved fingerprint value or a changed field — none.**
The conformance suite reports the same 1477 passed, 143 skipped and 137 xfailed.

Crank-Git:
## Self-review, lens 2: the traceback fix

I ran a second review over `710006e`, the commit that clears the traceback. It found no
blocker and no major. Five findings follow, and `293114d` addresses the two that asked
for a change.

**1. `_drop_traceback` terminates on a cyclic error chain — non-issue, proved.**
The review built `a.__context__ = b; b.__context__ = a` and called the helper. It
returned at once and cleared both tracebacks. The identity list stops the walk.

**2. The mitigation covers the traceback and nothing else — addressed in `293114d`.**
An error that carries the packet in `args`, or on an attribute of its own, still holds
it. The review swept `ja4plus/` and found no fingerprinter that builds such an error:
`ja4plus/fingerprinters/base.py:113` holds the only `raise`, and it raises
`NotImplementedError` with a string. The docstring of `_drop_traceback` now states the
limit, so a later fingerprinter author reads it.

**3. The helper mutates no shared exception — non-issue.**
The review looked for a cached or module-level exception instance and found none. Every
raise builds a new object, so the in-place clear reaches no other caller. Two threads
therefore clear no common object.

**4. The three new cases fail without the fix — non-issue, proved twice.**
The review ran them against a no-op `_drop_traceback` and all three failed. It then ran
them against a partial fix that clears the top-level traceback alone, and
`test_a_returned_error_clears_the_traceback_of_the_error_it_followed` still failed. The
chain walk is therefore measured and not assumed.

**5. Three sentences passed the 25-word cap — addressed in `293114d`.**
`.claude/rules/ste.md` rule 1 states the cap. The three sentences sat in `CHANGELOG.md`,
`docs/api_reference.md` and `tests/test_processor_typed_results.py`. Each one is now two
sentences. The review found no factual error in the documentation changes.

### The five gates after `293114d`

```
pytest tests/ -m "not spec_validation"   1673 passed, 8 xfailed, 93 subtests passed
pytest tests/ -m spec_validation         1477 passed, 143 skipped, 137 xfailed
ruff format --check ja4plus/ tests/      136 files already formatted
mypy ja4plus/                            Success: no issues found in 28 source files
```

The unit suite also passes under `-W error::DeprecationWarning`, so no caller inside
this repository reads a result by item access.

Issue #267, opened 2026-08-08
Decide whether this project publishes a JA4X raw form


Crank-Git:
Split from #228 by the project manager on 2026-08-08. **#228 asked two questions and the
user answered one.** Decision 1, the empty extension list, is decided and #228 builds it.
Decision 2 is this issue, and it is undecided.

## The question

**Does this project publish a JA4X raw form?**

R11 of `docs/specs/foxio/JA4X.md` holds the reading.

| Source at commit `27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8` | Raw form |
|---|---|
| `rust/ja4x/src/lib.rs` | Writes `ja4x_r`, the three unhashed lists joined with `_` |
| `wireshark/source/packet-ja4.c:1726` | Registers the field `ja4.ja4x_r` |
| FoxIO Python | Writes no raw form for JA4X |
| `ja4plus` | Writes none |

**Two of the three FoxIO implementations publish it and this project does not.**

## Why it needs the user

It adds a value the package does not publish today, so it is a schema addition rather than
a defect repair. `ja4plus` already publishes a raw form for other methods, so the question
is whether JA4X joins them, and not whether the concept belongs to the package.

**Read the question against the Go port before deciding.** `CLAUDE.md` states that where
FoxIO specifies nothing the port decides interface. Here FoxIO does specify, and two of its
three implementations agree, so the specification decides intent and the port does not
outrank it. The port is still worth reading for the field name.

## Why it is not urgent

**No vector value moves.** Adding a value adds comparisons; it changes none. The issue
gates nothing, and #228 closes without it.

## What blocks it

The user's ruling. Nothing else.

Crank-Git:
## Decision, 2026-08-08. This project publishes a JA4X raw form

**`ja4plus` writes `JA4X_r`.** It holds the three unhashed lists joined with `_`, which is
the form `rust/ja4x/src/lib.rs` writes and the form the dissector registers at
`packet-ja4.c:1726` as `ja4.ja4x_r`. Two of the three FoxIO implementations publish it, and
FoxIO specifies the value, so parity rule 1 decides it and the port does not outrank it.

The name `JA4X_r` follows the `JA4S_r` form this package already publishes. `ja4plus` writes
`JA4_o`, `JA4_ro`, `JA4S_r`, `JA4S_o`, `JA4S_ro` and `JA4H_ro` today, so the raw form is an
established shape here and not a new concept.

[...]

Issue #50, opened 2026-08-06
Version the output schema and document it


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/05-structured-output.md`](docs/specs/features/05-structured-output.md):

> FR-structured-output-3 — Every JSON object carries a `schema_version` field.

> FR-structured-output-7 — The table format is for a person reading a terminal, and carries no stability promise.

> FR-structured-output-8 — The JSON and CSV formats carry a stability promise, stated in the documentation.

> FR-structured-output-13 — The documentation records the schema and its version.

## What to build

New file `docs/output-schema.md`.

The schema version starts at 1. It rises when a field is removed or its meaning changes; adding a field does not raise it. A JSON object omits no field, and a field with no value is `null`. A new CSV column is appended, never inserted.

The table format is for a person reading a terminal and carries no stability promise. JSON Lines and CSV carry one, and the documentation must say which is which.

## Acceptance criteria

- [ ] Every JSON object holds `"schema_version": 1`.
- [ ] Every JSON object holds all 11 documented fields.
- [ ] `docs/output-schema.md` documents every field.
- [ ] A test compares the CSV header against the documented column list.

## Links

- Feature: [`docs/specs/features/05-structured-output.md`](docs/specs/features/05-structured-output.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #16
Depends on #49

PR #293, opened 2026-08-08
Publish the JA4X_r raw form

Crank-Git:
## What

`ja4plus` publishes `JA4X_r`. The value holds the three unhashed lists of the JA4X
fingerprint, joined with `_`. Each list holds the hex object identifiers in wire order,
joined with `,`.

Refs #267.

## Why

The user decided the form on 2026-08-08, and the decision is reversible. Two of the three
FoxIO implementations publish the value, and the FoxIO Python implementation publishes
none, so parity rule 1 decides the form and the Go port does not outrank it.

- `rust/ja4x/src/lib.rs` writes `let ja4x_r = with_raw.then(|| parts.join("_"));`.
- `wireshark/source/packet-ja4.c:1726` registers the field `ja4.ja4x_r`.

R11 of `docs/specs/foxio/JA4X.md` holds the transcription, and D1 of that page closes.

## How it is tested

**The new value is compared, not merely emitted.** No expected-output file under
`tests/foxio_vectors/` holds a `JA4X_r` key, and no field of a Rust snapshot holds one,
because the FoxIO Rust implementation writes `ja4x_r` only when the caller asks for it.
The same function hashes the three parts for `ja4x` and joins the same three parts for
`ja4x_r`, so the raw form is the preimage of the snapshot value.
`TestTheJa4xRawFormTheRustSnapshotImplies` in `tests/test_foxio_rust_parity.py` hashes
each part of each produced raw form and reads the snapshot value.

| Measurement | Count |
|---|---|
| Raw forms the five local snapshots imply | 43 |
| Values `ja4plus` reproduces exactly, in snapshot order | 43 |
| Values that differ | 0 |
| Streams whose value count differs | 0 |

**The gate is proven by its reversal.** Three mutations of `generate_ja4x_raw` each make
the conformance suite fail, measured on 2026-08-08 and then reverted.

| Mutation | Cases the conformance suite fails |
|---|---|
| Join the three parts with `\|` rather than `_` | 46 |
| Join the issuer list with `;` rather than `,` | 44 |
| Sort the extension list | 43 |

Two checks inside the suite measure the same two directions, so a reader who never runs
the mutation still holds the measurement:
`test_a_sorted_list_reaches_no_rust_snapshot_value` and
`test_another_list_separator_reaches_no_rust_snapshot_value`.

**No JA4X value moves.**

| Suite | Before | After |
|---|---|---|
| `pytest tests/ -m spec_validation` | 1477 passed, 143 skipped, 137 xfailed | 1529 passed, 143 skipped, 137 xfailed |
| `pytest tests/ -m "not spec_validation"` | 1641 passed, 8 xfailed | 1652 passed, 8 xfailed |

The register holds 137 keys before and after, and it gains no entry.

## The five local gates

| Gate | Result |
|---|---|
| `.venv/bin/pytest tests/ -m "not spec_validation"` | 1652 passed, 8 xfailed, 93 subtests passed |
| `.venv/bin/pytest tests/ -m spec_validation` | 1529 passed, 143 skipped, 137 xfailed |
| `.venv/bin/ruff check ja4plus/ tests/` | All checks passed! |
| `.venv/bin/ruff format --check ja4plus/ tests/` | 134 files already formatted |
| `.venv/bin/mypy ja4plus/` | Success: no issues found in 27 source files |

Coverage rises. `ja4plus/fingerprinters/ja4x.py` reads 92% before and 93% after, and the
package total holds 86%.

## Two readings a reviewer must check

1. **An empty list reaches the raw form as an empty part.** R8 gives the zero sentinel
   `000000000000` to the hashed form alone, because `hash12` of the FoxIO Rust
   implementation runs on the hashed form alone.
2. **A JA4X result holds one raw value under two keys.** R10 sorts no list, so `raw` and
   `raw_original_order` are equal. JA4S holds the same shape for the same reason, and
   FoxIO publishes no `JA4X_ro` key.

## One call site moved

`read_certificate` is now the reader of one certificate, and it returns the fingerprint
and the raw form from one parse. `fingerprint_certificate` returns its first value, so
`ja4plus/__init__.py` and `ja4plus/cli.py` call the same name they always called.
`tests/test_parity.py` records the certificate bytes through `read_certificate`, because
the scan calls that method now.

## Out of scope

The output schema version is #50, and this pull request does not write it. No other
fingerprinter changes.

Verified against: https://github.com/FoxIO-LLC/ja4/blob/main/rust/ja4x/src/lib.rs (retrieved 2026-08-08, commit `27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8`)
Verified against: https://github.com/FoxIO-LLC/ja4/blob/main/wireshark/source/packet-ja4.c (retrieved 2026-08-08, commit `27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8`)

The Changelog round reads `TBD`. The project manager assigns it.


Crank-Git:

A Sonnet child read the diff against `rust/ja4x/src/lib.rs` at the pinned commit, the
callers of the refactored method, the lock discipline of `ja4plus/processor.py`, and every
new assertion. It returned no blocking finding and no should-fix finding. The four
readings it confirmed:

1. **`generate_ja4x_raw` reproduces `ja4x_r`.** `ja4x.py` builds
   `[",".join(issuer_rdns), ",".join(subject_rdns), ",".join(extensions)]` and joins with
   `_`, which is `let parts = [issuer_rdns, subject_rdns, extensions];` and
   `parts.join("_")` at `rust/ja4x/src/lib.rs:47` to `:49`. An empty list writes an empty
   part in both, because `hash12` at `lib.rs:162` to `:169` runs on the hashed form alone.
2. **The refactor moves no behaviour.** `fingerprint_certificate` returns
   `self.read_certificate(cert_data)[0]`, and `read_certificate` holds the same guard and
   the same return value as the old method. Every caller reads that first value:
   `ja4plus/__init__.py:56`, `ja4plus/cli.py:358`, and six test modules. 169 cases over
   those modules pass.
3. **`last_raw` pairs with the fingerprint the processor reports.** `_record` is the one
   writer, it writes the attribute and appends the entry with no branch between them, and
   `Processor.process_packet` reads the attribute inside the same `with fp.lock:` block
   that made the call. The lock is the reentrant lock of `base.py:50` to `:56`, so the
   second acquisition inside `ja4x.py:157` deadlocks nothing.
4. **No new assertion is vacuous.** Every comparison reads `index_produced`, which runs
   the fingerprinter over a real capture. `test_a_sorted_list_reaches_no_rust_snapshot_value`
   and `test_another_list_separator_reaches_no_rust_snapshot_value` each carry a count
   assertion, so a loop that matches no case fails rather than passes.

## One finding, and it is out of scope

`ja4x.py:446`, `ja4x.py:476` and `x509_utils.py:258` write
`except (ValueError, TypeError, Exception) as e:`, which catches every exception.
`CLAUDE.md` states that a fingerprinter does not catch bare `Exception`. **The diff moves
one of the three lines and it writes none of them**, so this pull request changes nothing
there. #294 records the finding and the repair it needs.

Crank-Git:
## Self-review, writing standard lens

A Sonnet child read every added line of prose against `.claude/rules/ste.md` and the
`## Terms` table of `docs/specs/spec.md`. It reported two should-fix findings and two
nits. `3d2334e` fixes both should-fix findings, and this comment rejects both nits with
the reason.

### Fixed

1. **`ja4x.py`, the reset comment.** The sentence broke mid-clause and "such a value"
   named nothing. It now reads: `The reset drops these two attributes too. Each one holds
   a value of the capture the reset discards, and a later reader would read it as its own.`
2. **`docs/specs/foxio/JA4X.md`, R11.** One sentence held 28 words and three claims, and
   "FoxIO specifies it" read against the heading `the image states none`. The page now
   quotes the user's ruling verbatim and states that "FoxIO specifies the value" names the
   two implementations and not the image.

### Rejected, with the reason

3. **The docstring of `fingerprint_certificate` opens with an imperative verb.** The text
   is byte-identical to the text the method carried before this pull request, and the diff
   moved the method rather than writing it. `CLAUDE.md` asks for a surgical change, so a
   rewrite of unchanged prose belongs to another issue.
4. **`the FoxIO Rust snapshot value` is a four-word noun cluster.** The phrase is the
   established name in `tests/test_foxio_rust_parity.py`, which carries it in its class
   names and in its existing docstrings. One concept keeps one word under the standard, so
   a new synonym here would cost more than the cluster.

### The one factual check the child could not run

The child reported that it could not run `pytest -m spec_validation`, because it read the
system Python and not the virtual environment of this worktree. **The counts are measured
in this worktree**, with `.venv/bin/pytest`: 1477 passed, 143 skipped and 137 xfailed on
the base, and 1529 passed, 143 skipped and 137 xfailed on this branch. The child did
reproduce the case counts it could reach: 43 values over 19 streams of 5 captures.

Issue #47, opened 2026-08-06
Ship the py.typed marker and define the public interface


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/04-typed-api.md`](docs/specs/features/04-typed-api.md):

> FR-typed-api-9 — The package ships a `py.typed` marker file.

> FR-typed-api-10 — The wheel contains the `py.typed` marker.

> FR-typed-api-12 — `ja4plus/__init__.py` lists every public name in `__all__`.

> FR-typed-api-13 — A name absent from `__all__` is not part of the interface the project promises.

## What to build

New file `ja4plus/py.typed`, shipped as package data through `pyproject.toml`. The marker follows PEP 561.

`ja4plus/__init__.py` lists every public name in `__all__`. A name absent from `__all__` is not part of the interface the project promises, which is what makes the 1.0.0 stability promise enforceable.

Verified against https://peps.python.org/pep-0561/ (retrieved 2026-08-06).

## Acceptance criteria

- [ ] `unzip -l dist/*.whl` lists `ja4plus/py.typed`.
- [ ] Every name in `ja4plus.__all__` is importable from `ja4plus`.
- [ ] A caller running `mypy --strict` against their own code resolves the annotations.

## Links

- Feature: [`docs/specs/features/04-typed-api.md`](docs/specs/features/04-typed-api.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #15
Depends on #46

Issue #46, opened 2026-08-06
Annotate the package and make mypy --strict pass


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/04-typed-api.md`](docs/specs/features/04-typed-api.md):

> FR-typed-api-7 — Every public function and method in `ja4plus/` carries type annotations.

> FR-typed-api-8 — `mypy --strict ja4plus/` reports no error.

## What to build

`--strict` applies to `ja4plus/` only. The test suite is not type checked.

The Python floor is 3.9, so `X | None` in an annotation needs `from __future__ import annotations` or the `Optional` form. Pick one and use it throughout.

`scapy` ships no type information, so its imports need an ignore entry in the `mypy` configuration rather than a cast at each call site.

Epic 0 already runs `mypy` without `--strict`. This issue turns it on.

## Acceptance criteria

- [ ] `mypy --strict ja4plus/` reports no error.
- [ ] Every public function and method in `ja4plus/` carries type annotations.

## Links

- Feature: [`docs/specs/features/04-typed-api.md`](docs/specs/features/04-typed-api.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #15
Depends on #45

Crank-Git:
## Carried forward from #24

#24 made `mypy ja4plus/` report no error without `--strict`. It reached zero with
two overrides in `pyproject.toml`, and this issue must decide what to do with them.

1. **`follow_imports = "skip"` for `scapy.*`.** 31 of the 32 errors were
   `Module "scapy.all" has no attribute "IP"`, which describes scapy's dynamic layer
   loading, not this project. The override makes every scapy name `Any`, so a
   `--strict` run will report nothing about any packet field this project reads.
   Decide between keeping the override and writing local stubs for the scapy names
   the package uses.
2. **`ignore_missing_imports` for `requests.*`.**
3. **`python_version` is unset.** `mypy` 2.3.0 rejects the value `3.9` with
   `Python 3.9 is not supported (must be 3.10 or higher)`, so the Python 3.9 test
   job is the only 3.9 gate. `mypy` checks nothing about 3.9 compatibility today.

## Also carried forward

#24 disabled two ruff rules under the more-than-50-findings rule in
[...]

Issue #45, opened 2026-08-06
Return typed results from the processor and expose parse failures


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/04-typed-api.md`](docs/specs/features/04-typed-api.md):

> FR-typed-api-3 — `Processor.process_packet` returns a list of `FingerprintResult`.

> FR-typed-api-4 — `Processor.process_packet` exposes the parse failures it collected.

## What to build

`ja4plus/processor.py:81` swallows every fingerprinter exception and logs it at `DEBUG`. The port returns them: `ProcessPacket(packet) ([]FingerprintResult, []error)`.

Python has no multiple-return form that reads well as a default, so `process_packet` keeps returning results and `process_packet_with_errors` returns both:

```python
def process_packet(self, packet: Packet) -> list[FingerprintResult]: ...
def process_packet_with_errors(self, packet: Packet) -> tuple[list[FingerprintResult], list[Exception]]: ...
```

The method order is part of the interface. A packet that carries no address layer gives empty address strings and zero ports.

Verified against https://github.com/Crank-Git/ja4plus-go/blob/master/processor.go (retrieved 2026-08-06).

## Acceptance criteria

- [ ] `Processor().process_packet(pkt)` returns a list whose elements are `FingerprintResult`.
- [ ] `process_packet_with_errors` returns the exception raised by a fingerprinter given a packet crafted to raise.
- [ ] Results come back in the fixed method order the processor already uses.

## Links

- Feature: [`docs/specs/features/04-typed-api.md`](docs/specs/features/04-typed-api.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #15
Depends on #44

PR #295, opened 2026-08-08
Annotate the package and make mypy --strict pass

Crank-Git:
## What

`mypy --strict ja4plus/` reports no error. Every function and method in `ja4plus/` carries type annotations.

This issue's red-to-green measurement is the error count itself, because an annotation carries no test of its own.

| Commit | `mypy --strict ja4plus/` |
|---|---|
| `c103900`, the base | **720 errors in 25 of 28 source files** |
| `f80ccde`, the shared foundation | 567 errors |
| `b2ee07c`, this branch | **Success: no issues found in 28 source files** |

The 720 broke down as 437 `no-untyped-call`, 274 `no-untyped-def`, 8 `var-annotated` and 1 `type-arg`. The first group is a consequence of the second, because a call reads as untyped until the function it names carries annotations.

## Why

FR-typed-api-7 and FR-typed-api-8, quoted in #46.

## How

**One annotation form across the package.** `ja4plus/types.py` chose `from __future__ import annotations` with the `X | None` form in #44. Every module now follows it. No module writes `Optional`, and the Python 3.9 floor holds.

**No annotation changes a value.** Where mypy reported `Returning Any from function declared to return "X"`, the fix annotates a local and returns it:

```python
# scapy ships no type information, so each field reads as `Any`. The local
# annotation states the type the scapy field holds, and it changes no value.
ttl: int = packet[IP].ttl
return ttl
```

A converter such as `int(...)` around a scapy field would change a value, so no call site holds one. The diff adds no `cast` and no `assert`.

**Three type aliases name shapes the code already built.**

- `ja4plus.fingerprinters.base.Lock` names the guard a fingerprinter holds. typeshed publishes no name for the object `threading.RLock` returns, and every call site uses `with` and no other operation.
- `ja4plus.ja4db.LookupResult` names one match the mapping file holds.
- `ja4plus.cli.ResultRow` names one row the command writes.

**Four `# type: ignore` comments, each with its error code and its reason.** Three sit on `BoundedStateTable.keys`, `.values` and `.items`, which return a list where `MutableMapping` returns a live view. The fourth sits on `JA4SSHFingerprinter.interpret_fingerprint`, whose docstring already states that None reaches the error dictionary through the `except` block.

**`scapy` and `cryptography` needed no new configuration.** The `follow_imports = "skip"` override that Epic 0 wrote for `scapy.*` already covers `--strict`. `cryptography` ships its own type information, so no call site holds a cast. This answers the "check whether `cryptography` needs the same" item of the plan: it does not.

**`process_packet` returns `str | None` on all ten methods and on the base class.** Six already stated it; the other five stated `Any`. The typed interface reads better for the narrow type, and mypy stays clean with it.

**The gate command changed, so the three files that state it changed with it.**

- `.github/workflows/test.yml` runs `mypy --strict ja4plus/`.
- `CLAUDE.md` states the same command.
- `[tool.mypy]` of `pyproject.toml` holds `strict = true` with `files = ["ja4plus"]`, so an editor reports what the gate reports and the test suite stays out.

## How tested

Five gates, all from a virtual environment inside the worktree.

| Gate | Base commit | This branch |
|---|---|---|
| `pytest tests/ -m "not spec_validation"` | 1673 passed, 8 xfailed, 93 subtests passed | **identical** |
| `pytest tests/ -m spec_validation` | 1477 passed, 143 skipped, 137 xfailed | **identical** |
| `ruff check ja4plus/ tests/` | All checks passed! | All checks passed! |
| `ruff format --check ja4plus/ tests/` | 136 files already formatted | 136 files already formatted |
| `mypy --strict ja4plus/` | Found 720 errors in 25 files | **Success: no issues found in 28 source files** |

**Line coverage rises from 86% to 87%**, measured with `pytest tests/ --cov=ja4plus`, the directory form that `.claude/rules/conformance.md` requires. Statements rise from 3791 to 3842, and misses hold at 516 and 517.

**The Python 3.9 floor is measured, not assumed.** A 3.9.6 environment built with `uv` imports every module, builds a `BoundedStateTable`, and resolves all four runtime-evaluated aliases:

```
contextlib.AbstractContextManager[typing.Any] dict[str, str] tuple[typing.Any, ...] tuple[str, int, str, int]
```

The whole suite runs there and reports `1 failed, 3149 passed, 143 skipped, 145 xfailed`. **That one failure is on the base commit too**, so this change does not cause it: `tests/test_ja4ssh_interpret.py::test_a_part_that_holds_more_digits_than_python_converts_returns_an_error` needs `sys.set_int_max_str_digits`, which 3.9.6 predates. CI runs a current 3.9.x.

## Two premises of the plan, re-measured

**`ProcessorStats` stays in `ja4plus/processor.py`.** `docs/specs/features/04-typed-api.md` lists it under `ja4plus/types.py`, and #45 left it because its own issue did not name it. Neither #46 nor #47 names the move either, read from `gh issue view 46` and `gh issue view 47`. This branch annotates it where it sits and leaves the move to the issue that names it.

**`docs/specs/features/00-foundation.md` is unchanged, on purpose.** It states "`mypy` runs without `--strict` in this epic. Epic 4 turns `--strict` on", and it lists `mypy --strict` under "Out of scope". Both sentences describe Epic 0 and both stay true. FR-foundation-9 asks that `mypy` run on every pull request, which the strict command satisfies.

## One observation, filed nowhere and fixed nowhere

`x509_utils.get_name_attribute` returns what `cryptography` types as `str | bytes`. The narrow type reports two call sites that format the value with an f-string, where a bytes value would render as `b'...'`. Each caller reads `ORGANIZATION_NAME`, `COMMON_NAME` or `COUNTRY_NAME`, and each of those carries a string, so no call reaches the case. #46 changes no behaviour, so the return type stays wide and a comment records the reading.

## Acceptance criteria

- [x] `mypy --strict ja4plus/` reports no error. Evidence: `Success: no issues found in 28 source files`.
- [x] Every public function and method in `ja4plus/` carries type annotations. Evidence: `--strict` turns on `disallow_untyped_defs`, so the same clean run proves it; the base reported 274 `no-untyped-def`.

Part of #15


Crank-Git:
## Self-review

I spawned two review children, one on the behaviour lens and one on the annotation-and-writing lens. **Neither returned.** I ran both reviews myself, and this comment records what I found. `14135b7` fixes every finding.

### Method

I extracted all 1059 changed lines under `ja4plus/` and classified each one. A line is inert when it is a parameter annotation, a return annotation, a variable annotation, an import that only an annotation reads, a `from __future__ import annotations` line, a comment, or a `ruff format` reflow. Everything else I read in context.

### Finding 1 — BLOCKER. Two `assert` statements in the SSH segment reader

`ja4plus/utils/ssh_utils.py:156` and `ja4plus/utils/ssh_utils.py:201` gained `assert self._next_seq is not None` to narrow an `int | None`.

An `assert` is an executable statement, and it raises. `CLAUDE.md` states the rule: **a parser that cannot read a packet returns nothing, and it does not raise.** The invariant holds today, so no packet reaches the raise, but the change added a raise path to a per-packet parser that had none.

Fixed: each one becomes a guard that returns the documented empty result, `[]` for `_read_segment` and `None` for `_take_ready`. The package now contains no `assert`.

### Finding 2 — MINOR. An unreachable guard with no explanation

`ja4plus/fingerprinters/ja4ssh.py` gained `if ip is None: return None` after `get_ip_layer(packet)`.

I traced it. The guard at the top of `_process_packet` already requires `packet.haslayer(TCP) and (packet.haslayer(IP) or packet.haslayer(IPv6))`, and scapy's `Packet.__contains__` calls `haslayer`, so `get_ip_layer` reads the same two layers and returns a layer. The test reaches no packet, so behaviour is unchanged. It stays, because removing it would need a `# type: ignore`, which is worse. It now carries a comment that says why it exists.

### Finding 3 — MINOR. Two annotation forms for one concept

`cli.py` read `Sequence` from `typing` while `state_table.py` and `tls_utils.py` read `Callable`, `Iterator`, `MutableMapping` and `Iterable` from `collections.abc`. Fixed: `cli.py` reads `collections.abc` too.

### Finding 4 — MINOR. Five `process_packet` methods returned `Any`

`base.py`, `ja4.py`, `ja4s.py`, `ja4ts.py` and `ja4x.py` annotated the central method `-> Any` where the other five already stated `-> str | None`. `Any` on the central method of a typed-interface epic buys nothing. All ten and the base class now state `str | None`, and mypy stays clean.

### Finding 5 — MINOR, not fixed on purpose. A too-wide return in `x509_utils`

`get_name_attribute` returns what `cryptography` types `str | bytes`. I tried the narrow type; it reported two call sites that format the value with an f-string, where a bytes value renders as `b'...'`. Each caller reads `ORGANIZATION_NAME`, `COMMON_NAME` or `COUNTRY_NAME`, and each carries a string, so no call reaches the case. #46 changes no behaviour, so the type stays `Any` and a comment records the reading. This is an observation, not a defect I can demonstrate.

### Checked and cleared

- **`ja4l.py`: `return _quic_client_initial(...)` became a bare call plus `return None`.** I read `_quic_client_initial`. Every path returns None, including the early `return None`. The two forms are identical.
- **`ja4h.py`: the guard `if fingerprint is None:` became `if fingerprint is None or http_info is None:`.** The line above is `fingerprint = _generate_ja4h_from_info(http_info) if http_info else None`, so a falsy `http_info` already forces `fingerprint` to None. Python short-circuits `or`, so the second operand is read only when `fingerprint` is not None, and `http_info` is then truthy. The two forms are identical.
- **`state_table.py`: `MutableMapping` became `MutableMapping[Any, Any]` as a base class.** PEP 585 subscription returns a `types.GenericAlias`, and the class it builds keeps the same MRO and the same `isinstance` results. Measured on Python 3.9.6: the table constructs, stores and reads.
- **No converter anywhere.** The diff adds no `int(...)`, `str(...)` or `cast(...)` around a value. Every `Returning Any` fix annotates a local and returns it.
- **No new packet reference survives `process_packet`.** No hunk adds an attribute that stores a packet.
- **Four `# type: ignore`, each with a code and a reason.** Three sit on `BoundedStateTable.keys`, `.values` and `.items`, which return a list where `MutableMapping` returns a live view; the docstrings already stated that. The fourth sits on `interpret_fingerprint`, whose docstring already stated that None reaches the error dictionary through the `except` block. None hides a defect.
- **One annotation form across the package.** Every module uses `from __future__ import annotations` with `X | None`. No module writes `Optional`. Three files carry no future import: two `__init__.py` files that hold no annotation, and `loopback.py`, whose one annotation is `-> None`.
- **Writing standard.** I read every comment the diff adds. Each states why rather than what, uses the active voice and the present tense, keeps the articles, and stays inside the sentence limits.

### Gates after the fixes

| Gate | Result |
|---|---|
| `pytest tests/ -m "not spec_validation"` | 1673 passed, 8 xfailed, 93 subtests passed |
| `pytest tests/ -m spec_validation` | 1477 passed, 143 skipped, 137 xfailed |
| `ruff format --check ja4plus/ tests/` | 136 files already formatted |
| `mypy --strict ja4plus/` | Success: no issues found in 28 source files |
| Python 3.9.6, whole suite | 3116 passed, 143 skipped, 145 xfailed |

Every number matches the base commit.

Issue #271, opened 2026-08-08
Read the four JA4D6 subfield fields at any nesting depth


Crank-Git:
Found while #231 ruled on D8 of `docs/specs/foxio/JA4D.md`. This issue is untriaged.

#231 added DHCPv6 option 9, Relay Message, to the containers that part b of JA4D6
recurses into. **It scoped that ruling to part b, because the four subfield rules R16,
R19 and R20 stay uncertain and #231 must resolve none of them.** Two divergences
therefore stay open, and both need a relay message to observe.

## The dissector reads four fields at any nesting depth

`wireshark/source/packet-ja4.c:967-969` calls `proto_all_finfos(tree)` and walks every
field of the whole dissection tree. Each test below matches on the field name alone. Four JA4D6 inputs read that way.

| Field | Subfield | `packet-ja4.c` |
|---|---|---|
| `dhcpv6.duid.bytes` | Part a, subfield 2 | `:1547-1559` |
| `dhcpv6.iata` | Part a, subfield 3 | `:1560-1562` |
| `dhcpv6.client_domain` | Part a, subfield 4 | `:1563-1565` |
| `dhcpv6.requested_option_code` | Part c | `:1574-1578` |

`ja4plus/fingerprinters/ja4d6.py:194-221` reads all four at the top level alone. A
RELAY-FORW message that carries a client SOLICIT inside option 9 therefore gives the
dissector the inner Client DUID length and the inner Option Request List, and gives
`ja4plus` `0000` and `00`.

**Measured on 2026-08-08 against the relay case of `tests/test_ja4d_decisions.py`.**
`ja4plus` produces `rlayf0000nn_18-9-1-6_00`, where part b now matches and the other
three parts do not.

## The dissector concatenates the message type of a relay message

`wireshark/source/packet-ja4.c:1537-1546` appends a five-character name for every
`dhcpv6.msgtype` field. A relay message holds two: the outer type and the inner type. The
dissector therefore writes `rlayfsolct` into subfield 1, which holds ten characters.

**That breaks the schema the image states.** R2 of `docs/specs/foxio/JA4D.md` gives part a
eleven characters and subfield 1 five of them. `ja4plus` writes `rlayf` alone.

## Why this needs the user

**R16, R19 and R20 rest on the Wireshark dissector alone**, and
`docs/specs/foxio/README.md` marks all three uncertain. A nested reading would settle
what each rule measures, which is the question the second corroboration would answer.
[...]

Issue #231, opened 2026-08-08
Decide the eleven JA4D and JA4D6 readings the images leave open


Crank-Git:
Found while #204 transcribed `technical_details/JA4D.png` and
`technical_details/JA4D6.png`. This issue is untriaged.

`docs/specs/foxio/JA4D.md` holds the transcription, the eleven disagreements and the five
uncertain rules. **No disagreement below changes a value on any vector this repository
holds.** All ten FoxIO reference values match today, measured on 2026-08-08.

## Why this needs the user

**JA4D6 reaches one FoxIO implementation.** The FoxIO Python implementation emits neither
method, the FoxIO Rust implementation emits neither, and `zeek/README.md:15` states
`JA4D6 &rarr; ja4d.log (awaiting Zeek DHCPv6 suppport)`. The Wireshark dissector is the
only source for five JA4D6 rules, so `docs/specs/foxio/README.md` marks them uncertain and
keeps the vector fallback. Each decision below moves a fingerprint this project publishes.

## The eleven disagreements

Each item names the packet that separates the two readings. No such packet is in
`tests/foxio_vectors/`.

| Item | `file:line` | This project | The reference |
|---|---|---|---|
| D1 | `ja4plus/fingerprinters/ja4d.py:168` | Reads UDP port 67 or 68 | The dissector and the Zeek script apply no port test |
| D2 | `ja4plus/fingerprinters/ja4d.py:141-142` | Emits nothing without option 53 | `zeek/ja4d/main.zeek:43-45` emits `00000`; `wireshark/source/packet-ja4.c:1498` agrees with this project |
| D3 | `ja4plus/fingerprinters/ja4d.py:136-137` | Reads the presence of option 81 | `wireshark/source/packet-ja4.c:1521` reads `dhcp.fqdn.name`, the name inside the option |
| D4 | `ja4plus/fingerprinters/ja4d.py:132-133` | Keeps the last option 57 | `wireshark/source/packet-ja4.c:1508-1512` concatenates every option 57 |
| D5 | `ja4plus/fingerprinters/ja4d.py:138-139` | Keeps the last option 55 | `wireshark/source/packet-ja4.c:1530-1534` concatenates every option 55 |
| D6 | `ja4plus/fingerprinters/ja4d.py:42-44` | Cites FoxIO pull requests 267 and 270 | R9 of the transcription now holds the FoxIO statement |
[...]

PR #296, opened 2026-08-08
Read the four JA4D6 subfield fields at any nesting depth

Crank-Git:
Builds #271. The user ruled on all three decisions on 2026-08-08, and this branch lands
them.

## What changes

**D1 and D2 — part a and part c read an inner relay message.** Part a reads the Client
DUID, the IA_TA option and the Client FQDN at any nesting depth, and part c reads the
Option Request List the same way. `wireshark/source/packet-ja4.c:967-969` calls
`proto_all_finfos(tree)` and walks every field of the whole dissection tree, matching on
the field name alone. #231 already made part b recurse into DHCPv6 option 9, so all three
parts now read the same way.

One walk now reads all six inputs. `_read_subfield_option` holds the four subfield reads
that a second top-level loop held before, and `_walk_options` calls it at every depth. D9
and D10 keep their rulings, because the walk reads the outer option before the inner one.

**D3 — subfield 1 holds the outer message type alone.**
`wireshark/source/packet-ja4.c:1537-1546` appends a five-character name for every
`dhcpv6.msgtype` field, and a relay message holds two, so the dissector writes
`rlayfsolct` and a part a of sixteen characters. R2 of `docs/specs/foxio/JA4D.md` gives
part a eleven characters and gives subfield 1 five of them. This project declines the
concatenation as a provable reference defect, under `.claude/rules/conformance.md` and
rule 1 of `CLAUDE.md`. `ja4plus` writes `rlayf`, and no code changed for this ruling.

## Why the tests are the evidence

**No FoxIO implementation other than the Wireshark dissector writes a JA4D6 value.**
`zeek/README.md:15` reads `JA4D6 -> ja4d.log (awaiting Zeek DHCPv6 suppport)`, FoxIO ships
no Python and no Rust for the method, and no deleted text file covers it.

**No vector separates any of the three readings.** `tests/foxio_vectors/dhcpv6.pcap`
carries no relay message. All six JA4D6 reference values match before this change and
after it, so the conformance suite measures none of it.

`tests/test_ja4d_decisions.py` gains 7 cases. Each ruling was reversed locally, the suite
was run, and the reversal was reverted.

|---|---|
| D1 — the three part a subfield fields read the top level alone | 3 |
| D2 — part c reads the top level alone | 2 |
| D3 — subfield 1 concatenates every message type | 5 |

No reversal fails one case of the conformance suite, which is the reading that shows the
vector set cannot measure this. 5 of the 7 cases failed against the base before the
implementation landed. The other 2 guard a reading the merged walk could otherwise lose:
the outer Client DUID length wins over the inner one, and part a holds eleven characters.

## What the documents record

- `docs/specs/foxio/JA4D.md` gains the section `JA4D6 — the rulings of #271`, and R16,
  R18, R19 and R20 each record the nesting depth this decision settles.
- **R16, R19 and R20 keep the `uncertain` mark and keep the vector fallback.** The brief
  asked to clear the mark where the decision settles the rule. Measured against
  `docs/specs/foxio/README.md:172-174`, the mark counts FoxIO-authored corroborations, and
  a decision adds none. `docs/specs/foxio/README.md` now states that reading and states
  plainly that no second implementation corroborates JA4D6.
- The divergence register of `docs/specs/spec.md` gains one row for D3. D1 and D2 carry no
  row, because each follows the dissector.
- The Changelog gains one row, and the round number reads `TBD`.

## How this was tested

```
.venv/bin/pytest tests/ -m "not spec_validation"   1659 passed, 8 xfailed, 93 subtests passed
.venv/bin/pytest tests/ -m spec_validation         1529 passed, 143 skipped, 137 xfailed
.venv/bin/ruff format --check ja4plus/ tests/      134 files already formatted
.venv/bin/mypy ja4plus/                            no issues found in 27 source files
```

The conformance counts equal the base counts exactly, so no reference value moved. The
unit suite rises from 1652 to 1659, which is the 7 cases this branch adds. Coverage of the
project holds at 86%, the count of missed statements falls from 519 of 3755 to 516 of
3742, and coverage of `ja4plus/fingerprinters/ja4d6.py` rises from 94% to 96%.
`tests/foxio_deviations.json` holds 137 keys before and after, and no key names JA4D6.

Verified against: https://github.com/FoxIO-LLC/ja4/blob/main/wireshark/source/packet-ja4.c (commit `27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8`, read from a local checkout, retrieved 2026-08-08)


Crank-Git:
## Self-review — correctness

A review by the correctness lens found no defect. Four checks ran, and each is recorded
here so a later reader can repeat it.

**1. The new part c loop reads the same bytes as the loop it replaces.** The old form read
a sliced copy, `for i in range(0, len(opt_data) - 1, 2)`, and the new form reads the buffer
directly, `for i in range(pos, pos + opt_len - 1, 2)`. `_walk_options` already proves
`pos + opt_len <= end <= len(data)` before the call, so the highest index the loop reads is
`data[pos + opt_len - 1]`. A comparison of the two forms over 20000 random offsets and
lengths reports 0 differences.

**2. No unintended value moves.** A differential run compares the base module against the
new one over 20000 random DHCPv6 payloads. **8133 payloads carry no container, and 0 of
them differ.** The remaining 11867 may nest, and none of those differ either, because
random bytes rarely form a readable option inside a container. The seven committed cases
carry the nested packets that do differ.

**3. The order holds.** The walk visits a container and then its children, which is the
order the dissection tree reports. The outer Client DUID therefore still decides subfield 2
under D9, and part c still joins every Option Request List in wire order under D10. Two
committed cases hold that reading.

**4. Every packet is hostile, and none raises.** A fuzz of 30000 random UDP payloads on
port 546 raises nothing. The reviewer added an empty payload, a payload shorter than the
header, an option 6 of length 0 and of length 1, a truncated option header, a length field
that passes the end of the buffer, an IA_NA shorter than its own header, a relay option of
length 0, and relay chains of 50 and 2000 levels. Each returns a value or `None`.
`_DHCPV6_MAX_NESTING_DEPTH` bounds the chain, and this change does not touch it.

The reviewer found no dead code and no leftover from the refactor.

Crank-Git:
## Self-review — the writing standard and the citations

A review by the writing lens read every line of prose this branch adds, plus every new
comment, docstring and test name.

**Finding, accepted and fixed: sentences above the limit.** `.claude/rules/ste.md` sets 25
words for a description sentence. The review named offending sentences in
`docs/specs/foxio/JA4D.md`, `docs/specs/foxio/README.md`, the divergence register row and
the Changelog row. Commit `1dd94da` splits each one, and one citation list became a
vertical list. A check over every added sentence of the diff now reports none above 25
words. The review notes that the surrounding Changelog rows carry the same density; this
branch repairs the text it adds and leaves the rows it does not own.

**Verified, no change needed: every `ja4d6.py:NNN` citation.** The review read each of the
twenty citations this branch adds or changes against the current
`ja4plus/fingerprinters/ja4d6.py`, and every one points at the code the row describes.

**Verified, no change needed: the facts and the tables.** The review traced
`_walk_options` and `_read_subfield_option` by hand against all seven new cases and found
no claim that contradicts the code. It found no malformed table row. It reports the suite
counts as unconfirmed, because it ran no suite; the counts in the pull-request body come
from the five gates run inside this worktree, and they repeat.

**Rejected: nothing.** Every finding the review raised is real, and each is fixed.

Issue #272, opened 2026-08-08
Rule on the six JA4L entries that #34 left open


Crank-Git:
Opened by the project manager on 2026-08-08, so that #255 can refresh the owner snapshot
without deciding anything.

## Why this issue exists

**#34 is closed and six undecided register entries still name it.** An entry whose owner is
closed names nothing a worker can act on, and
`TestTheRegisterOwners.test_no_entry_names_an_epic_or_a_closed_issue` rejects exactly that.
**The check passes today only because `tests/foxio_deviation_owners.json` records #34 as
open, and the snapshot is stale.** #255 refreshes it, and these six need a live owner first.

**This issue decides nothing.** It records that the question is open and names where it
lives. Marking an entry `decided` to make a check pass is what #193 refused and what #251
refused again.

## The six entries

`CVE-2018-6794.pcap/JA4L-C`, `CVE-2018-6794.pcap/JA4L-S`, `https-connect.pcap/JA4L-C`,
`https-connect.pcap/JA4L-S`, `tls-handshake.pcapng/JA4L-S` and `ssh2.pcapng/JA4L-S`.

Changelog round 56 measured all six and settled none.

## The reading, and why it splits five from one

**#200 measured on 2026-08-08 that five of the six are artifacts of the reference.**
`python/ja4.py:339` runs `delete_keys(['JA4L-S','JA4L-C'], final)` when the generating run
names another method, so **three reference files publish no JA4L key at all.** This project
was not emitting more than the reference; **the reference published nothing to compare.**

**The sixth, `ssh2.pcapng/JA4L-S`, is a real extra value.** It is not an artifact and it
needs its own reading.

## The question for the user

1. **Are the five artifacts declined on #200's reading**, with a `Divergence register` row
   recording that the comparison is unreachable because the reference deletes the key?
2. **What is the reading for `ssh2.pcapng/JA4L-S`**, where this project publishes a value
   the reference does not?

## What blocks it

The user's ruling. Nothing else. **This issue gates nothing** — #255 re-homes the entries
here and refreshes the snapshot, and the suite goes green either way.

Crank-Git:
## Decisions, 2026-08-08. The user delegated both questions to the project manager

**The user delegated this issue on 2026-08-08**, after ruling on #215, #267 and #271 in the
same session. Both readings below follow rules this project has already written down, and
neither invents scope.

### Question 1 — the five artifacts are declined, and the register records why

[...]

PR #300, opened 2026-08-08
Write the JA4T and JA4TS form the user decided

Crank-Git:
Builds #215. The user ruled on D1, D2 and D4 on 2026-08-08, and D3 and D5 ride with
them as plain defects. #226 keeps D6 and D7, and #246 keeps the RST value.

## What changed

`ja4plus/utils/tcp_options.py` is new. It holds part a through part d of both methods,
so `ja4ts.py` no longer repeats the body of `ja4t.py`. The FoxIO references build the two
methods from one function too: `ja4t()` in `wireshark/source/packet-ja4.c` serves both
fields, and `zeek/ja4t/main.zeek:195-236` repeats one block.

| Reading | The rule |
|---|---|
| D1 | An empty option list, an absent maximum segment size and a zero window scale each take the two-digit form. |
| D2 | The reader takes the raw TCP option bytes, so each End of Option List byte adds one entry. |
| D3 | Part b holds every option kind. |
| D4 | One connection produces one JA4T value, from its first SYN. |
| D5 | A repeated option keeps the first value. |

## D1 was measured before it was built, and the count is not one

The decision comment states that the count of captures the two forms separate "is not
known until the change is measured". **It is 12 values across 6 captures, and 7 of the 12
move only because the window scale is zero.**

| Capture | Values | Shape |
|---|---|---|
| `CVE-2018-6794.pcap` | 3 JA4TS | The SYN-ACK carries no option. |
| `gre-erspan-vxlan.pcap` | 1 JA4T, 1 JA4TS | The packet carries no option. |
| `gre-sample.pcap` | 1 JA4T, 1 JA4TS | The option list is not empty and the window scale is zero. |
| `sshv1.pcap` | 1 JA4T, 1 JA4TS | The option list is not empty and the window scale is zero. |
| `v6.pcap` | 1 JA4T, 1 JA4TS | The option list is not empty and the window scale is zero. |
| `ssh2.pcapng` | 1 JA4TS | The option list holds the segment size alone. |

All 12 diverge from the FoxIO Rust implementation. **Only 1 of the 12 reaches a FoxIO
reference comparison**, because this repository holds a Rust snapshot for 7 captures and
no reference file holds a JA4TS value. That one is
`gre-erspan-vxlan.pcap/0:65174/JA4T.1`, which stays in `tests/foxio_deviations.json` as a
decided divergence and keeps `xfail(strict=True)`. The register therefore gains no entry
from D1.

## Two premises of the brief were wrong, and the evidence is here

**1. The brief scoped D1 to `ja4t.py`, and D1 reaches `ja4ts.py` too.** 13 of the 28
moved values are JA4TS values. The image titles itself `JA4T/S`, and both FoxIO
implementations build the two methods from one function, so one form covers both. A
change to `ja4t.py` alone would leave two forms in one package.

**2. D1 is not only about the empty option list.** The heading of the ruling names the
empty option list, and `wireshark/source/packet-ja4.c:664` and `zeek/ja4t/main.zeek:204`
also write part c with `%02d` and part d as `00` whenever the window scale is zero. The
issue body's own heading reads "D1 — the empty option list, **and a zero part**", and
R11 of `docs/specs/foxio/JA4T.md` already stated the whole rule. This pull request builds
the whole rule.

## A reading the brief did not name, and it needs a reader

`docs/specs/foxio/deleted-text-specifications.md` quotes the deleted
`technical_details/JA4T.md`:

> If any field does not exist, then the output is 00. For example, a packet with a Window
> of 1024 and no TCP options, and therefore no Window scale would be:
> JA4T = 1024_00_00_00

**The prose reads `00` as the mark of an absent field. Both implementations read the
value instead**, because each stores the window scale as an integer that defaults to
zero and neither records whether the packet carried the option. A packet that carries a
Window Scale option of value 0 therefore writes `00` in both implementations and `0`
under the prose alone. **7 of the 12 moved values are of that shape**, and
`gre-sample.pcap`, `sshv1.pcap` and `v6.pcap` each carry such a packet.

`ja4plus` follows the two executable references the user cited, so that one tool's output
compares against another tool's output. R11 of `docs/specs/foxio/JA4T.md` records the
reading and the alternative, and the `Divergence register` states that the decision is
reversible.

## The whole vector set was replayed, before and after

The 38 captures produce **809 values before and 784 after, and 28 differ**. Every one of
the 28 is a JA4T value or a JA4TS value, and no value of another method moved. The 25
values the count loses are the duplicate JA4T values of `ssh2.pcapng` that D4 stops.

- `ssh2.pcapng` JA4T falls from 44 values to 19, which is the count the Rust snapshot
  holds. The distinct values do not move.
- `chrome-cloudflare-quic-with-secrets.pcapng` moves from `65535_2-1-3-1-1-8-4-0_1440_6`
  to `65535_2-1-3-1-1-8-4-0-0_1440_6`, which is the Rust value.
- 16 values move to the reference form under D2, and 12 move away from it under D1.

`tests/foxio_deviations.json` falls from **137 keys to 135**, against 135 `xfailed`
cases. The D2 entry and the D4 entry resolve to a pass and leave the register. The
undecided entries fall from nine to six, and the six are the JA4L entries #272 owns.

## Each ruling is proven by its reversal

The tests came first: 15 of the 31 cases in `tests/test_ja4t_form.py` failed against
`ja4t.py` and `ja4ts.py` as they stood, and commit `c0bdc31` holds them. The self-review
added a 32nd, and reversing the read order fails it.

|---|---|
| D1, the one-digit form | 36 |
| D2, one entry for several pad bytes | 11 |
| D3, the six named option kinds | 2 |
| D4, one value for each SYN | 60 |
| D5, the last repeated option | 2 |

Each reversal ran the whole suite and the worktree was clean after each one.

## The state table D4 adds

`JA4TFingerprinter.connections` is a `BoundedStateTable` with a maximum entry count of
10000 and a maximum age of 600 seconds, which `connections` in
`ja4plus/fingerprinters/ja4ssh.py` already held. It follows the form Epic 14 shipped and
adds no second form. The eviction reads the packet timestamp. `cleanup_connection` drops
both orderings of the key, and `reset` clears the table. The processor now walks 17 state
tables and 15 of them are a `BoundedStateTable`.

## The Zeek comparison

`tests/compare_zeek_baselines.py` needs a `FoxIO-LLC/ja4` checkout that this worktree
does not hold, so **the comparison was not re-run and this pull request states no
measured number for it**. `docs/specs/foxio/zeek.md` records what the row-level evidence
already settles: no row of the ten JA4T rows or the ten JA4TS rows changes its verdict,
because every SYN and every SYN-ACK of the three `conn.log` captures carries a window
scale above zero.

## How this was tested

```
.venv/bin/pytest tests/ -m "not spec_validation"   1697 passed, 8 xfailed, 93 subtests
.venv/bin/ruff format --check ja4plus/ tests/      136 files already formatted
.venv/bin/mypy ja4plus/                            no issues in 28 source files
```

The base reported 1659 passed and 8 xfailed on the unit suite, 1529 passed and 137
xfailed on the conformance suite, and 137 register keys.

**Coverage holds at 86%**, measured with `--cov=ja4plus`, the directory form the rule
requires. The missed statements move from 516 of 3742 to 519 of 3790.
`ja4plus/utils/tcp_options.py` reports 98%.

## Sources

Verified against: https://raw.githubusercontent.com/FoxIO-LLC/ja4/27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8/wireshark/source/packet-ja4.c (retrieved 2026-08-08, pinned commit)
Verified against: https://raw.githubusercontent.com/FoxIO-LLC/ja4/27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8/zeek/ja4t/main.zeek (retrieved 2026-08-08, pinned commit)

The Changelog round is `TBD`, and the project manager assigns it.


Crank-Git:
## Self-review, the hostile-input lens

The lens read `ja4plus/utils/tcp_options.py` and `ja4plus/fingerprinters/ja4t.py`
against the four rules of `CLAUDE.md` and the state rules of
`.claude/rules/conformance.md`. It returned clean on all four questions.

- **The loop terminates on every input.** Each branch advances the offset by one byte
  or by a length of two or more, so the offset rises on every pass. A 10000-byte
  adversarial buffer completes in under one millisecond.
- **No index reaches past the buffer.** A truncated length byte and a length that
  reaches past the option field each stop the reader. `option_bytes` rejects a data
  offset below five and a data offset past the packet, and a `TCP(dataofs=15)` packet
  truncated to 25 bytes returns an empty byte string.
- **The table stores a boolean.** `connections[key] = True` holds no packet reference,
  and the key is the address pair and the port pair. The bounds are 10000 entries and
  600 seconds.
- **No `assert`, no bare `except Exception`, no unbounded structure.**
- **`cleanup_connection` drops both orderings**, and `pop` with a default raises nothing
  on a key the table does not hold.

**The lens found one thing the code did not state, and the review fixed it before this
comment.** `generate_ja4t` marked the connection before it read the packet, so a packet
the reader could not read consumed the connection and the SYN that followed produced
nothing either. `8dc7e92` moves the read ahead of the gate and adds
`test_a_syn_the_reader_cannot_read_leaves_the_connection_open`. Reversing the order
fails that one case.

Crank-Git:
## Self-review, the correctness lens

The lens read the diff at `5c5c9c0`, which is before the ordering fix. It hand-walked
the byte parser against real TCP option encodings and confirmed byte 12 of the TCP
header against RFC 9293 section 3.1.

**It found the same ordering hazard the hostile-input lens found, and it named the
consequence more precisely.** `_first_syn_of_connection` wrote `connections[key] = True`
before `tcp_prefix` ran, so a raise inside the reader would mark the connection
permanently and every later SYN of it would produce nothing. `8dc7e92` fixed it and
`test_a_syn_the_reader_cannot_read_leaves_the_connection_open` holds it.

**The lens also states that the hazard was not reachable today**, because the reader
raises on none of the inputs it tried. That reading matches the hostile-input lens. The
fix stands, because a later change to the reader would make it reachable and silent.

The lens found nothing else.

- The boundary logic of `read_options` and `option_bytes` holds on every input it built:
  a corrupted length byte of `0xff`, a data offset past the buffer, and multiple End of
  Option List pad bytes. Every slice and index is in bounds behind the length guard.
- Byte 12 reads `0x70` on a header of 20 bytes and 8 option bytes, which is a data offset
  of 7 and 28 bytes. The reading is correct.
- The exception clause still names what the new path can raise, and the new path adds
  nothing outside it.
- JA4T and JA4TS call one function for part a through part d, so the two cannot drift.

**Both lenses returned.** No lens is reported here that did not run.

Issue #302, opened 2026-08-08
Resolve the three TBD Changelog rounds that reached dev


Crank-Git:
Found by the worker that merged `dev` into `batch/291-foxio-rulings` on 2026-08-08. **It
reported the finding and deliberately did not act on it, because the rows belong to another
worker and its own brief scoped it to three different rows.** That was the right call and this
issue records the work.

## The measurement

**`CHANGELOG.md` carries three unresolved `Round TBD` markers**, at lines 11, 25 and 43. They
arrived on `dev` with Epic 4 and they were written by #45 and #47.

**The rounds they name are already numbered.** `docs/specs/spec.md` records Epic 4 as rounds
89 and 90, assigned by the project manager when the epic merged. So the numbers exist and only
`CHANGELOG.md` still says `TBD`.

Reproduce it with one command from the root of the repository.

```bash
grep -n "Round TBD" CHANGELOG.md
```

## Why this matters, and why it is small

**A `TBD` that ships is a promise the document does not keep.** `CHANGELOG.md` is the file a
user reads to learn what changed in a release, and version 1.0.0 is close. A reader who follows
`Round TBD` reaches nothing.

It is small because no number has to be invented. **Read the round from
`docs/specs/spec.md` for each of the three rows and write it.** Do not renumber anything and do
not rewrite the prose of the rows — it belongs to #45 and #47.

## Why the rows exist at all

Every worker in this project writes `TBD` rather than choosing a Changelog round, because two
members of one batch have claimed the same round number before. The project manager assigns
each number at sub-merge. **These three rows were missed because the assignment covered
`docs/specs/spec.md` and not `CHANGELOG.md`.**

**Prevent the next one.** Add a check that fails when a `TBD` round reaches `dev`, or state in
`CLAUDE.md` that the assignment covers both files. A repair with no guard is a repair this
project makes again.

## Acceptance criteria

- [ ] `grep -n "Round TBD" CHANGELOG.md` returns nothing.
- [ ] Each of the three rows names the round `docs/specs/spec.md` already records for that work.
- [ ] No prose of the three rows changes, and no round number in `docs/specs/spec.md` moves.
- [ ] A check or a written rule prevents a `TBD` round reaching `dev` again, and the check
      fails on a tree where a `TBD` is reintroduced.

## Links

- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)

[...]

Issue #49, opened 2026-08-06
Write output records with separate address and port fields


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/05-structured-output.md`](docs/specs/features/05-structured-output.md):

> FR-structured-output-1 — Every output format carries the source address, source port, destination address and destination port as separate fields.

> FR-structured-output-2 — The JSON format writes one object per line.

> FR-structured-output-4 — The CSV format writes the same columns whatever flags the user passed.

> FR-structured-output-5 — A column with no value is empty rather than absent.

## What to build

New file `ja4plus/output.py`, holding one writer per format.

Today `ja4plus/cli.py:123` emits `{"source": ..., "type": ..., "fingerprint": ...}` where `source` packs both addresses and both ports into one string, so a downstream tool must parse it apart again. Lines 128 to 134 add the `identified_as` field only when `--lookup` was passed, so the CSV header changes with the command line.

The fixed CSV column order:

```
schema_version,timestamp,type,fingerprint,raw,raw_original_order,src_ip,src_port,dst_ip,dst_port,identified_as
```

The mockup at `docs/specs/mockups/01-cli-output.html` shows all three formats. It is guidance; the column order and the JSON field set are the contract.

## Acceptance criteria

- [ ] `ja4plus analyze <capture> --format json` writes one JSON object per line.
- [ ] The CSV header is identical with and without `--lookup`.
- [ ] The CSV header matches the documented column order exactly.
- [ ] `ja4plus analyze <capture> --format json | jq -e '.src_port'` succeeds.
- [ ] A column with no value is empty rather than absent.

## Links

- Feature: [`docs/specs/features/05-structured-output.md`](docs/specs/features/05-structured-output.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #16

Crank-Git:
## Plan

**Approach.** Add `ja4plus/output.py` holding one writer per format. The writers read
`FingerprintResult`, which Epic 4 shipped, and write the fixed column set the issue body
names. `ja4plus/cli.py:123` stops packing both addresses and both ports into one `source`
string, and the CSV header stops changing with the command line.

**Files.** `ja4plus/output.py` (new), `ja4plus/cli.py`, and the tests.

[...]

Issue #16, opened 2026-08-06
Epic 5: Structured output


Crank-Git:
## Goal

Give the capture analyst an output format another tool can read.

## Exit criteria

The command-line program writes JSON Lines, CSV and a table. The schema is documented and versioned.

## Spec

- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Feature: [`docs/specs/features/05-structured-output.md`](docs/specs/features/05-structured-output.md)

Depends on: #15

## Sub-issues

- [x] #49
- [x] #50
- [x] #51
- [x] #52

PR #304, opened 2026-08-08
Record the output schema and hold its version rule (#50)

Crank-Git:
Part of #16. Builds #50. Targets `epic/16-structured-output`. Every head commit carries `[skip ci]`.

## What this adds

New page `docs/output-schema.md`. It records the schema version, the eleven fields, the raw form each of the ten methods writes, the rule that raises the version, and which format carries a stability promise.

#49 wrote `SCHEMA_VERSION = 1` into `ja4plus/output.py` and asserted nothing about it. This pull request owns the rule, the page and the checks that hold both. **No file under `ja4plus/` changes.**

## The rule, stated so a check enforces it

The page states four sentences, and `TheSchemaDocument::test_the_document_states_the_rule_that_raises_the_version` reads each one back:

- The schema version rises when a field goes away or its meaning changes.
- A new field raises no version.
- A new CSV column appends to the end of the row.
- A JSON object omits no field, and a field with no value is `null`.

`SCHEMA_HISTORY` in `tests/test_output_schema.py` freezes the column list of each released version. `TheSchemaVersionRule::test_a_released_column_keeps_its_position_until_the_version_rises` requires the frozen list to stay a prefix of `CSV_COLUMNS` while `SCHEMA_VERSION` holds. An append passes. An insertion, a rename or a removal fails until the version rises and the history records the new list.

`TheSchemaDocument` parses the Markdown field table of the page and compares it against the written CSV header and the written JSON object. The page and the code therefore cannot drift apart.

## The comparison is proven by its reversal

Ten mutations, measured on 2026-08-08 in this worktree with the corrected harness. The baseline is `41 passed`.

| Mutation | Result |
|---|---|
| Swap `src_ip` and `src_port` in `CSV_COLUMNS` | 7 failed |
| Append `new_field` to `CSV_COLUMNS` | 4 failed, and the version rule stays green, which is the specified behaviour |
| Drop `raw_original_order` from the JSON object | 5 failed |
| Raise `SCHEMA_VERSION` to 2 with no history entry | 6 failed |
| Rename `src_ip` to `source_ip` in the page alone | 3 failed |
| Change the `csv` stability promise to `No` | 1 failed |
| Delete the sentence `A new field raises no version.` | 1 failed |
| Swap the JSON values of `raw` and `raw_original_order` | 1 failed |
| Swap the CSV values of `raw` and `raw_original_order` | 2 failed |
| Write `src_port` into `dst_port` | 2 failed |

The append case is the one that matters most: it fails the page checks, because a new field must be documented, and it passes the version checks, because a new field raises no version.

The last three cases exist because of the self-review. A swap of two values changes the meaning of both fields and moves no name, so the name checks alone passed it. See the review comment on this pull request.

**A warning for the next round that proves a gate by reversal here.** My first harness reported that a swap of two CSV columns left the suite green. The tests were right and the harness was wrong: the swap preserves the file size, the copy preserved the mtime, and Python reused a stale `.pyc`, so the mutated module never loaded. Every length-preserving mutation silently did nothing. The harness now clears `__pycache__` and sets `PYTHONDONTWRITEBYTECODE=1`.

## A correction to the base

**The base of this branch was red before the first commit.** `tests/test_output_schema.py` failed two cases at `50ec718`, and passed all 23 at `eeb96bf`. The merge of `dev` carried `f63265e Write the JA4T and JA4TS form the user decided (#300)` onto fixtures that #49 had pinned to the earlier JA4T form. D1 of that ruling gives a zero window scale the two-digit form, so `8760_2-1-1-4_1460_0` became `8760_2-1-1-4_1460_00`. This pull request corrects the stale literals. It decides nothing; the user already decided the form in #300.

## The schema this page describes is the schema the writers emit

The page describes what `ja4plus` writes today, read from the writers rather than from the earlier text. `ja4x` now writes `JA4X_r` into both raw fields, which #267 added and which the `dev` sync carried here. `ja4h` writes `JA4H_ro` into `raw_original_order` and `null` into `raw`, because it publishes no `JA4H_r` value. The other six methods write `null` in both.

## How this was tested

All five gates, run in a virtual environment in the worktree.

```
.venv/bin/pytest tests/ -m "not spec_validation"   1801 passed, 8 xfailed, 93 subtests passed
.venv/bin/ruff format --check ja4plus/ tests/      142 files already formatted
.venv/bin/mypy --strict ja4plus/                   Success: no issues found in 30 source files
```

The conformance suite is identical to the base, and the register invariant holds: `tests/foxio_deviations.json` holds 135 keys against 135 `xfailed` cases, before and after. The unit suite moves from `1781 passed, 2 failed` to `1801 passed`: 18 new cases, plus the two the base correction repairs.

Coverage holds at 88%. It cannot fall, because `git diff --stat origin/epic/16-structured-output -- ja4plus/` is empty.

The self-review found that `ja4plus/output.py` uses a noun the `## Terms` table forbids. That code is #49's and already merged, so #306 records it rather than this pull request changing it.

`CHANGELOG.md` carries `Round TBD`, per #302. The project manager assigns the round.

Verified against: `docs/specs/features/05-structured-output.md` (FR-structured-output-3, -7, -8, -13) and `.claude/rules/ste.md`, both read in this worktree on 2026-08-08.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,11 +43,17 @@ project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   same. A new column appends to the end of the row and raises no version.
   `TheSchemaDocument` parses the page and compares its field table against the written
   CSV header and the written JSON object, so the page and the code cannot drift apart.
-  **Seven mutations each make the file fail**, measured on 2026-08-08: a swap of two CSV
-  columns fails 6 cases, an appended column fails 4, a dropped JSON field fails 4, a
-  raise of `SCHEMA_VERSION` with no history entry fails 4, a renamed column in the page
-  alone fails 3, a changed stability promise fails 1, and a deleted rule sentence fails
-  1. **The `json` and `csv` formats carry the stability promise, and the `table` format
+  Two further cases read the value of each field, because a swap of two values changes
+  the meaning of both fields and moves no name. **Ten mutations each make the file
+  fail**, measured on 2026-08-08 against a baseline of 41 passed: a swap of two CSV
+  columns fails 7 cases, an appended column fails 4, a dropped JSON field fails 5, a
+  raise of `SCHEMA_VERSION` with no history entry fails 6, a renamed column in the page
+  alone fails 3, a changed stability promise fails 1, a deleted rule sentence fails 1, a
+  swap of two JSON values fails 1, a swap of two CSV values fails 2, and a write of
+  `src_port` into `dst_port` fails 2. The appended column is the case that proves the
+  rule: it fails the page checks, because a new field must be documented, and it leaves
+  the version checks green, because a new field raises no version.
+  **The `json` and `csv` formats carry the stability promise, and the `table` format
   carries none**, which the page states and a check reads back. The page records that
   `ja4`, `ja4s`, `ja4h` and `ja4x` write a raw form and the other six methods write
   `null`; `ja4x` now writes `JA4X_r`, which #267 added.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,23 @@ project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Added
 
+- **The output schema carries a version, and `docs/output-schema.md` records it** (#50).
+  Round TBD. The new page states the schema version, the eleven fields, the raw form
+  each of the ten methods writes, and the rule that raises the version. **The rule is a
+  check and not prose alone.** `SCHEMA_HISTORY` in `tests/test_output_schema.py` freezes
+  the column list of each released version, and `TheSchemaVersionRule` fails when a
+  released column moves, changes name or goes away while `SCHEMA_VERSION` stays the
+  same. A new column appends to the end of the row and raises no version.
+  `TheSchemaDocument` parses the page and compares its field table against the written
+  CSV header and the written JSON object, so the page and the code cannot drift apart.
+  **Seven mutations each make the file fail**, measured on 2026-08-08: a swap of two CSV
+  columns fails 6 cases, an appended column fails 4, a dropped JSON field fails 4, a
+  raise of `SCHEMA_VERSION` with no history entry fails 4, a renamed column in the page
+  alone fails 3, a changed stability promise fails 1, and a deleted rule sentence fails
+  1. **The `json` and `csv` formats carry the stability promise, and the `table` format
+  carries none**, which the page states and a check reads back. The page records that
+  `ja4`, `ja4s`, `ja4h` and `ja4x` write a raw form and the other six methods write
+  `null`; `ja4x` now writes `JA4X_r`, which #267 added.
 - **The package ships the `py.typed` marker and declares `__all__`** (#47). Round TBD.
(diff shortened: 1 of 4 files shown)


Crank-Git:
## Self-review, two lenses

Two Sonnet children reviewed this diff on 2026-08-08: one on the writing standard and the accuracy of the page, one on the correctness of the checks. Both returned. Findings and what I did with each.

### Accepted, and fixed

**1. A swap of two values defeated every check.** The correctness lens edited `JsonLinesWriter.write` to write `result.raw_original_order` under the key `"raw"` and the reverse, left `SCHEMA_VERSION` at 1, and the whole file stayed green at `39 passed`. That is a meaning change with no version rise — precisely the case `docs/output-schema.md` says the rule catches. Root cause: every check read field **names** and no check read field **values**.

Fixed by two new cases, `TheSchemaVersionRule::test_every_json_field_carries_the_value_of_the_field_it_names` and `::test_every_csv_column_carries_the_value_of_the_column_it_names`. Each writes a `FingerprintResult` whose eleven values are distinct and non-null, and reads each value back under its own name. The swap now fails, and so does a write of `src_port` into `dst_port`. This was the most valuable finding in the round.

**2. A dead escape hatch.** `assertTrue(prefix_holds or SCHEMA_VERSION > max(SCHEMA_HISTORY))` could never take the second branch, because two other cases jointly force `SCHEMA_VERSION == max(SCHEMA_HISTORY)`. It read as a relief valve and was not one. Replaced with a plain `assertEqual` against `SCHEMA_HISTORY[SCHEMA_VERSION]`, which states the same rule with no unreachable branch.

**3. The heading match was a substring match.** `"## Fields" in text` would also match a later heading `## Fields To Add` and could select the wrong table. Now anchored to a whole line, and the count must be exactly 1.

**4. The page used `record` as a noun.** `docs/specs/spec.md:81` lists `record` in the `Do not use` column of the term `result`. The page now writes `output line`. **I did not change the `## Terms` table**, because the shared controlled vocabulary is not mine to edit from one issue.

### Accepted, and filed rather than fixed

`ja4plus/output.py` uses the same forbidden noun, in its module docstring and in three method docstrings. That code is #49's and it is already merged onto this branch, so it is outside this issue. Filed as a separate untriaged issue.

### Rejected, with the reason

**Two copies of the column list** (`DOCUMENTED_COLUMNS` and `documented_columns()`). The reviewer suggested collapsing them. I am keeping both. They are deliberately independent: one is a literal transcription of the specification, and one is parsed from the shipped page. A single source would compare the code against itself. When they disagree, the failure is the signal.

### A finding of my own: the mutation harness could not fail

My first mutation run reported that a swap of two CSV columns left the suite green. The tests were correct and the harness was wrong. A swap of `"src_ip"` and `"src_port"` preserves the file size, and the copy preserved the mtime, so Python reused a stale `.pyc` and the mutated module never loaded. The same held for `SCHEMA_VERSION = 1` becoming `= 2`. Every length-preserving mutation silently did nothing, which reads exactly like a check that cannot fail.

The harness now removes `__pycache__` and sets `PYTHONDONTWRITEBYTECODE=1` before each run. **The ten mutation results in the pull-request body come from the corrected harness.** I am recording this because any future round that proves a gate by reversal in this repository can hit it.

Issue #305, opened 2026-08-08
Repair the stale JA4T values that the dev merge left in test_output_schema.py


Crank-Git:
## What happens

Two cases of `tests/test_output_schema.py` fail on `epic/16-structured-output`. They fail
on the base commit `50ec718`, before any work of #51.

```
FAILED tests/test_output_schema.py::TheCsvHeader::test_lookup_changes_the_identified_as_value_and_no_other_column
FAILED tests/test_output_schema.py::TheCsvRows::test_the_csv_row_holds_the_address_and_the_port_in_separate_columns
```

The second reads:

```
AssertionError: '8760_2-1-1-4_1460_00' != '8760_2-1-1-4_1460_0'
```

## Why

The failure is a semantic merge conflict, and no file conflicted.

`tests/test_output_schema.py` arrived with #49 in commit `eeb96bf`, and it exists on
`epic/16-structured-output` alone. It expects the JA4T value `8760_2-1-1-4_1460_0` for the
first packet of `tests/data/http.cap`.

Commit `50ec718` merged `dev` into `epic/16-structured-output`. That merge carried #300,
which wrote the JA4T and JA4TS form the user decided on 2026-08-08: a window scale of zero
writes `00` rather than `0`. Round 93 of `docs/specs/spec.md` records the ruling.

The program now writes `8760_2-1-1-4_1460_00`, and every other test of the repository
expects the two-digit form. `tests/test_ja4t_form.py:79` reads `65535_2_1460_00` and
`tests/test_ja4ts.py:61` reads `14600_2-1-3-4-1-1_1460_00`.

The first failure follows the second: its stub lookup client is keyed on the stale value,
so the lookup matches nothing.

## What to do

Read whether the two-digit form is right for these two cases, then move the three stale
values of `tests/test_output_schema.py` onto it. The three sit at lines 123, 137 and 166.

A third value at line 172 feeds a writer directly and asserts no fingerprint, so it moves
or stays without changing the result.

## Scope

#51 neither caused this nor repairs it. #51 measured that its own change moves no value:
the program writes the same 863 records over the 43 committed captures, byte for byte.

Crank-Git:
Closed as redundant on 2026-08-08 by the project manager.

**#50 landed the repair this issue asked for.** Commit `afd0e56` moved every stale JA4T
literal of `tests/test_output_schema.py` onto the two-digit form that #300 decided and
Changelog round 93 records.

[...]

Issue #52, opened 2026-08-06
Separate results from diagnostics and add the output option


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/05-structured-output.md`](docs/specs/features/05-structured-output.md):

> FR-structured-output-6 — The `--format` option accepts `table`, `json` and `csv`.

> FR-structured-output-9 — The program writes results to standard output and diagnostics to standard error.

> FR-structured-output-10 — The `--output` option writes results to a file.

## What to build

The program writes results to standard output and every diagnostic to standard error, so a pipe stays clean. `cmd_analyze` currently prints its table header and its progress messages to standard output.

When `--output` names an existing file, refuse to overwrite unless `--force` is passed.

A pipe that closes early must exit quietly. Handle `BrokenPipeError` rather than letting it reach the top level as a traceback.

The `--format` option accepts `table`, `json` and `csv`; all three already exist and keep their names.

## Acceptance criteria

- [ ] `ja4plus analyze <empty capture> --format csv` writes the header and no data row.
- [ ] `ja4plus analyze <capture> --output existing.json` exits non-zero.
- [ ] `ja4plus analyze <capture> --output existing.json --force` overwrites.
- [ ] Piping the output into `head -1` produces no traceback.

## Links

- Feature: [`docs/specs/features/05-structured-output.md`](docs/specs/features/05-structured-output.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #16
Depends on #49

Issue #51, opened 2026-08-06
Build the command-line program on Processor and surface fingerprinter errors


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/05-structured-output.md`](docs/specs/features/05-structured-output.md):

> FR-structured-output-11 — The program builds one `Processor` rather than a dictionary of fingerprinters.

> FR-structured-output-12 — The program does not swallow a fingerprinter error silently.

## What to build

`ja4plus/cli.py` builds its own dictionary of fingerprinters at `_build_fingerprinters` and duplicates the per-packet loop in `cmd_analyze` (lines 192 to 200) and `cmd_live` (lines 248 to 256). Both swallow every exception with `except Exception: pass`.

Moving to `Processor` removes the duplication, gives the command-line program the connection eviction Epic 3 added, and makes `process_packet_with_errors` from Epic 4 the source of the diagnostics.

The `--types` option must keep working: select which methods to report from the processor's results rather than by constructing a subset.

## Acceptance criteria

- [ ] The command-line program builds one `Processor` rather than a dictionary of fingerprinters.
- [ ] A fingerprinter error writes one line to standard error that names the method, and the run continues.

## Links

- Feature: [`docs/specs/features/05-structured-output.md`](docs/specs/features/05-structured-output.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #16
Depends on #49

Issue #306, opened 2026-08-08
Settle the word for the serialized output line


Crank-Git:
## What is wrong

`ja4plus/output.py` uses `record` as a noun. The `## Terms` table of `docs/specs/spec.md` forbids it.

`docs/specs/spec.md:81` reads:

```
| result | noun | One typed object that carries a fingerprint and its connection. | record, row, output |
```

The fourth column is the `Do not use` column, and it names `record`.

The module uses the forbidden noun in five places:

- `ja4plus/output.py:1` — "The writers that turn a `FingerprintResult` into an output record."
- `ja4plus/output.py:121` — "Write one record."
- `ja4plus/output.py:129` — "The base class writes no record."
- `ja4plus/output.py:151` — the local name `record` in `JsonLinesWriter.write`.
- `ja4plus/output.py:212` — "The writer that aligns the records for a person reading a terminal."

`.claude/rules/ste.md` states that the controlled vocabulary covers code comments and docstrings.

## Why this is a question and not a defect to fix in place

The word names a real concept that `result` does not name. A `FingerprintResult` holds nine fields, and the line the command-line program writes holds eleven: it adds `schema_version` and `identified_as`. The two are different things, so one word cannot serve both.

That leaves two shapes, and the choice belongs to the person who owns the vocabulary:

1. **Add `record` to the `## Terms` table** as its own term, meaning the serialized output line of the command-line program, and remove `record` from the `Do not use` column of `result`. The word is then legal for one meaning and one part of speech.
2. **Keep the table as it stands** and reword `ja4plus/output.py` to another word.

#50 took shape 2 for `docs/output-schema.md` alone, which now writes `output line`. #50 changed no file under `ja4plus/`, so the module still carries the forbidden noun, and the page and the module now use two words for one concept.

## What to do

Pick shape 1 or shape 2, then make `ja4plus/output.py` and `docs/output-schema.md` agree. `tests/test_output_schema.py` also writes `record` as a local name, and it follows whichever shape wins.

## Acceptance criteria

- [ ] `docs/specs/spec.md` and `ja4plus/output.py` agree on one word for the serialized output line.
- [ ] `docs/output-schema.md` uses that same word.
- [ ] No file under `ja4plus/` uses a noun that the `Do not use` column of the `## Terms` table names.

## Links

- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
[...]

PR #307, opened 2026-08-08
Build the command-line program on Processor and surface fingerprinter errors

Crank-Git:
Part of #16. Builds #51.

## What this changes

`ja4plus/cli.py` built its own dictionary of fingerprinters in `_build_fingerprinters`,
and `cmd_analyze` and `cmd_live` each ran a copy of the per-packet loop. Both copies
caught every error with `except Exception` and continued without a word, which
`CLAUDE.md` forbids.

One `Processor` replaces both. The program now gets the connection eviction of Epic 3
and the errors of Epic 4. FR-structured-output-11 and FR-structured-output-12 close.

A method that fails to read a packet writes one line to standard error:

```
Warning: ja4h could not read a packet: the parser read a length field it cannot trust
```

The run continues. The program keeps no exception: it writes each line and holds
nothing, so nothing grows across a capture. #45 already cleared the traceback of every
returned exception, so no error the program reads holds a packet.

`--types` now selects the results the program reports rather than the methods it
builds. The processor runs every method whatever the option names, so the program
evicts the connections of a method the user filtered out and it reports that method's
errors. The reported order is still the order the user wrote, and a name the user writes twice
keeps its first position. The self-review found that second case and fixed it in
`e01ae81`; no committed capture could have caught it. The first PR comment holds the
whole review.

## One premise of the plan was wrong

The plan asks `process_packet_with_errors` to be the source of the diagnostic. It
cannot name the method.

`process_packet_with_errors` returns `list[Exception]`. An exception names no method,
and the list holds an entry only for the methods that raised, so no caller can recover
the name from the order. The acceptance criterion asks for a line that names the
method.

`Processor.process_packet_with_method_errors` therefore returns the same results and
one pair of the method name and the exception. `process_packet_with_errors` keeps its
signature and drops the name, so every case of `tests/test_processor_typed_results.py`
holds, including the three that compare exception identity.

This is why the change touches `ja4plus/processor.py`, which the "Data touched" list of
`docs/specs/features/05-structured-output.md` does not name. A pair of values survives
#52 moving the diagnostics; a log line would need parsing.

## No fingerprint moved, measured twice

A replay of the 43 committed captures under `tests/foxio_vectors/` and `tests/data/`:

```
base   863 records
after  863 records
diff -r  0 files differ, byte for byte across all 43 files
```

A second matrix of 6 captures against the 3 formats and 6 `--types` orderings, two of
them the ten method names written forward and backward:

```
base   2453 lines
after  2453 lines
diff -r  0 files differ
```

The second matrix is what proves the `--types` order holds. The processor runs
`ja4t` and `ja4ts` before `ja4l`, and `VALID_TYPES` holds `ja4l` first, so a plain
adoption of the processor order would have moved the output.

## The diagnostic is proven by its reversal

`tests/test_cli_processor.py` holds 13 cases. Against the base commit, 11 of the 13
fail. With the swallow restored on the new code, by replacing the body of
`_report_errors`, 5 of the 13 fail:

```
FAILED TheProgramReportsAFingerprinterError::test_the_program_names_the_method_that_raised
FAILED TheProgramReportsAFingerprinterError::test_the_program_reports_an_error_of_a_method_the_types_option_excludes
FAILED TheProgramReportsAFingerprinterError::test_the_program_writes_one_line_for_every_packet_that_raises
FAILED TheProgramReportsAFingerprinterError::test_the_program_writes_the_diagnostic_to_standard_error
FAILED TheProgramReportsAFingerprinterError::test_the_program_writes_the_message_the_error_carries
```

The tests came first. The whole file was red before any line of `ja4plus/cli.py`
changed.

**The mutation was proven live before it was read.** #50 found that a length-preserving
edit with a preserved mtime lets CPython load a stale `.pyc`, so a mutation reports a
green suite while the mutated module never runs. This measurement cleared every
`__pycache__` outside `.venv/`, set `PYTHONDONTWRITEBYTECODE=1`, and read
`inspect.getsource` of the loaded `_report_errors`:

```
MUTATION present in loaded module: True
print-to-stderr present in loaded module: False
```

The revert was proven the same way, and it reads `False` then `True`.

## One removal

`ja4plus.cli._packet_endpoints` is gone, because the processor reads the endpoints and
the copy became unreachable. The three cases of `tests/test_output_schema.py` that
measured it now read `ja4plus.processor._packet_endpoints`, which returns the two
addresses first and the two ports second.

The processor's reader was probed against five malformed packets, including a bare
`Raw` payload, an IP header that declares TCP and carries none, and a truncated TCP
header. It raised on none of them, so the removal loses no guard.

## Gates

Measured on the merged tree, after `afd0e56` (#50) sub-merged and repaired the base.

```
pytest tests/ -m "not spec_validation"   1814 passed, 8 xfailed, 93 subtests, 0 failed
ruff format --check ja4plus/ tests/      143 files already formatted
mypy --strict ja4plus/                   Success: no issues found in 30 source files
coverage                                 88% (486 of 3897 missed)
```

The register invariant holds: `tests/foxio_deviations.json` holds 135 keys against 135
`xfailed`.

The unit suite rises from 1801 passed on the merged base to 1814 passed, which is the 13
cases this branch adds. **The two failures reported earlier are gone.** #50 moved the
stale JA4T literals of `tests/test_output_schema.py` onto the two-digit form that #300
shipped, which was the exact repair #305 asked for, so #305 is redundant.

One case flaked once during the base measurement:
`tests/test_thread_safety.py::TestTheConcurrencyContract::test_eight_sharded_threads_read_one_timeline_the_way_one_thread_reads_it`
failed on one run of the base source and passed on the next run of the same source and
in isolation. No file this branch changes reaches it, and it passes on every run of this
branch.

## Boundary with #52

This change decides nothing about where diagnostics go beyond what the feature spec
already states. FR-structured-output-9 asks for results on standard output and
diagnostics on standard error, and the edge-case table asks for one line that names the
method. #52 owns the `--output` option and the separation of results from diagnostics.

## `TBD` markers

Both files carry one, per the assignment:

- `CHANGELOG.md` — two entries, each reading `Round TBD`.
- `docs/specs/spec.md` — one changelog row whose round number is `TBD`.


Crank-Git:
## Self-review

I ran the correctness lens over the diff myself. Two child reviewers were spawned and
neither returned, so nothing below rests on a review I did not run.

### One finding, fixed in e01ae81

**`--types` with a repeated name reported the wrong order.**

Version 0.6.0 built a dict in `_build_fingerprinters`, so a repeated key kept the
position of its first write. `--types ja4t,ja4,ja4t` reported the JA4T value first.

The first form of `_reporting_order` was a dict comprehension over `enumerate`, where a
later write wins. It reported the JA4 value first.

```
['ja4t', 'ja4', 'ja4t']
  old        ['ja4t', 'ja4']
  last-wins  ['ja4', 'ja4t']   match=False
  first-wins ['ja4t', 'ja4']   match=True
```

**No committed capture could have caught this**, because no replay writes a repeated
name. The 43-capture diff and the 54-run matrix were both green while the defect was
present. `test_a_method_the_user_wrote_twice_keeps_its_first_position` now covers it,
and it fails against the first form.

### The other questions, each answered

1. **Does `process_packet_with_errors` still return the same exception objects?** Yes.
   It returns `[error for _, error in errors]` over the same objects. The three cases of
   `tests/test_processor_typed_results.py` that compare identity pass.
2. **Can one packet produce two results of the same method?** No. The processor appends
   at most one result for each method for each packet, so the sort in `_select` has no
   tie to break.
3. **Does the program accumulate anything across a capture?** No. `_report_errors`
   writes each line and returns, and `_fingerprint_one_packet` returns one list for one
   packet. Nothing survives the packet.
4. **Is `entry.get("connection") or ""` equivalent to `entry.get("connection", "")`?**
   Yes for every input the processor produces. `Processor.close_open_windows` always
   writes the key and writes `None` where the fingerprinter omitted it, so the old form
   would have returned `None` and the new form returns `""`.
5. **Does `cmd_live` build one processor?** Yes. The closure and the trailing-window
   call read the same object.
6. **Could running all ten methods change a reported value when `--types` names a
   subset?** No. The ten fingerprinters hold no shared state. Measured: the matrix
   includes `--types ja4ssh` alone, and its output is byte-identical.
7. **Does `dataclasses.replace` preserve `raw`, `raw_original_order` and the
   endpoints?** Yes. The JSON output carries all four fields and is byte-identical
   across all 43 captures.

### Hostile input

`ja4plus.cli._packet_endpoints` caught `AttributeError`, `IndexError`, `TypeError` and
`ValueError`, and `ja4plus.processor._packet_endpoints` catches nothing. The program now
reads the processor's copy, so I probed the processor's copy against five malformed
packets: a bare `Raw` payload, an IP header that declares TCP and carries none, an IP
header with a three-byte payload, a truncated TCP header, and `proto=6` with four bytes.

It raised on none of them and returned the empty address and the port zero for each, so
the removal loses no guard. I filed nothing, because I could not demonstrate a raise.

Crank-Git:
## Re-measured against the merged base

`afd0e56` (#50) sub-merged and repaired the two base failures. This branch now carries
that merge, in `c706437`.

One conflict, in `CHANGELOG.md`: #50 and #51 each add an entry under `### Added`. Both
entries stay. `tests/test_output_schema.py` auto-merged cleanly — #50's two-digit JA4T
literals and my three `ja4plus.processor._packet_endpoints` retargets both survive.

### All five gates, on the merged tree

```
pytest tests/ -m "not spec_validation"   1814 passed, 8 xfailed, 93 subtests, 0 failed
ruff format --check ja4plus/ tests/      143 files already formatted
mypy --strict ja4plus/                   Success: no issues found in 30 source files
coverage                                 88% (486 of 3897 missed)
```

**The two failures are gone.** The register invariant holds: `tests/foxio_deviations.json`
holds 135 keys against 135 `xfailed`.

The unit suite rises from 1801 passed on the merged base to 1814 passed. The difference
is 13, which is exactly the count of cases in `tests/test_cli_processor.py`.

### The replay still moves no value

The 43 committed captures produce 863 records on the merged tree, and `diff -r` against
the baseline taken before any edit reports no differing file.

### The mutation was proven live before it was read

Per the warning carried from #50: a length-preserving edit with a preserved mtime lets
CPython load a stale `.pyc`, so a mutation reports a green suite while the mutated
module never runs.

Every `__pycache__` outside `.venv/` was removed, `PYTHONDONTWRITEBYTECODE=1` was set,
and `inspect.getsource` of the loaded `_report_errors` was read in the same interpreter
configuration:

```
mutation applied:  MUTATION present in loaded module: True
                   print-to-stderr present in loaded module: False
                   -> 5 failed, 8 passed

mutation reverted: MUTATION present in loaded module: False
                   print-to-stderr present in loaded module: True
                   -> 13 passed
```

### One flake observed, not caused here

`tests/test_thread_safety.py::TestTheConcurrencyContract::test_eight_sharded_threads_read_one_timeline_the_way_one_thread_reads_it`
failed on one run of the **base** source and passed on the next run of the same source
and in isolation. No file this branch changes reaches it, and it passes on every run of
this branch. Recorded in the spec round, not filed.

### #305 is redundant

I measured it: #50 moved the three stale literals of `tests/test_output_schema.py` onto
the two-digit form, which is exactly the repair #305 asked for. It can close unbuilt. I
have not touched it, and I have not touched #306.

PR #311, opened 2026-08-08
Separate results from diagnostics and add the output option (#52)

Crank-Git:
Part of #16. Builds #52, the last member of Epic 5.

## What this changes

The command-line program writes results to standard output and every diagnostic to
standard error. `--output FILE` writes the results to a file instead. The program refuses
to overwrite a file that exists unless `--force` is passed. A reader that closes the pipe
early ends the run without a traceback. The `--format` option keeps `table`, `json` and
`csv`.

## Why the table header stays on standard output

The brief names the table header as a line that moves to standard error. That premise was
re-measured and declined. The `Screens & states` table of
`docs/specs/features/05-structured-output.md` gives the table format the states
`Header and rows; no results; error on standard error`, and the edge-case table gives the
`csv` format `the header only` for a capture with no fingerprint. The header is part of
the output of each format. A header on standard error would leave the file that
`--output` names without its `csv` header, which no parser could read.
`tests/test_cli.py::TestAnalyzePcap::test_analyze_table_has_header` states the same
reading and it is unchanged.

`cmd_analyze` held no progress message on the base commit `4c798d0`. The one progress
message that did sit on standard output is the line of `db update`,
`Downloading latest fingerprint database from FoxIO...`, and it now goes to standard
error.

## Why every option now runs after the subcommand name

Two acceptance criteria write the option after the subcommand:
`ja4plus analyze <capture> --format csv` and
`ja4plus analyze <capture> --output existing.json --force`. `argparse` reads an option of
the main parser before the subcommand alone, so `_add_output_options` registers
`--format`, `--types`, `--lookup`, `--output` and `--force` on the main parser with their
defaults and on the `analyze`, `live` and `cert` subparsers with `argparse.SUPPRESS`.

`argparse.SUPPRESS` is the whole mechanism. A subparser that carries a real default writes
that default over the value the main parser already parsed, and
`ja4plus --format json analyze capture.pcap` would then report the table format.
`test_the_analyze_command_keeps_the_format_option_written_before_the_name` holds that
reading.

## The self-review found a symbolic-link defect

The first form of the guard tested `os.path.exists(path)` and then opened with the mode
`w`. `os.path.exists` follows a link and reports False for a link whose target is absent,
so a link planted at the path the operator names passed the guard and the open created the
target. No race was needed. `cmd_live` runs as root, which makes the reading a privilege
boundary.

The guard now opens with the mode `x` when the user passes no `--force`. That mode fails
on an existing path and on a symbolic link, and it closes the check-to-open race at the
same time. `test_the_command_writes_through_no_symbolic_link` and
`test_the_command_refuses_a_file_that_arrives_after_the_test` hold both readings.

Both review children returned. The correctness lens reported no defect and confirmed the
`argparse.SUPPRESS` reading by running the program in every option position. The security
lens reported the symbolic-link defect and six sentences over the STE word cap, and every
one is repaired.

## The refusal is proven by its removal

With the guard, the eight cases of `TheProgramRefusesToOverwriteAFileThatExists` report
8 passed. With `if os.path.exists(path) and not force:` replaced by `if False:` and the
mode `x` replaced by `w`, they report 5 failed and 3 passed:

```
FAILED test_the_command_exits_with_the_status_1
FAILED test_the_command_leaves_the_file_byte_for_byte_unchanged
FAILED test_the_command_names_the_file_on_standard_error
FAILED test_the_command_refuses_a_file_that_arrives_after_the_test
FAILED test_the_command_writes_through_no_symbolic_link
```

The two `--force` cases pass by design. `test_the_command_writes_nothing_to_standard_output`
measures the redirection rather than the guard.

The closed-pipe handler is proven the same way. With the handler, the five cases of
`TheProgramExitsQuietlyWhenTheWriteFails` and
`TheProgramExitsQuietlyWhenTheReaderClosesThePipe` report 5 passed; with
`except BrokenPipeError` changed to `except KeyboardInterrupt`, they report 5 failed.

**The mutation was proven live before either result was trusted.** Every `__pycache__`
outside `.venv/` was removed and `PYTHONDONTWRITEBYTECODE=1` was set, and
`inspect.getsource` on the loaded function reported `GUARD_PRESENT`, then `GUARD_ABSENT`,
then `GUARD_PRESENT` again after the revert. A first run of that check reported
`GUARD_ABSENT` three times, because the check still read the text of an earlier form of
the guard. The check was repaired and the run repeated.

## No fingerprint moved

A replay of the 40 committed captures under `tests/foxio_vectors/` and `tests/data/`
across the three formats writes **2649 standard-output lines before the change and 2649
after**. `diff -r` reports **no differing file** across the 120 runs, and the comparison
covers the standard-output text, the standard-error text and the exit status of each run
in separate files.

A matrix of 6 captures by 3 formats by 6 `--types` orderings gives **108 runs and 3684
standard-output lines before and after**, and `diff -r` reports no differing file.

No line moved between the two channels for any committed capture, because every one of
them produces at least one result.

## How this was tested

The tests came first. 18 of the first 23 cases failed against the base commit `4c798d0`.
The 5 that passed are the regression guard on the option written before the subcommand
name, two cases that the base satisfies because it rejects `--output` before it opens a
file, and two pipe cases that the base satisfies because its `except Exception` catch
turns the traceback into a message. The self-review added 2 more, which brings
`tests/test_cli_output_option.py` to 25.

The pipe case needs a capture larger than the pipe buffer. `tests/data/http.cap` writes
its whole output in one write and never reaches the closed pipe, so
`tests/foxio_vectors/tls-handshake.pcapng` carries the case at 644 lines and 87.6
kilobytes.

All five gates, run in the worktree:

```
.venv/bin/pytest tests/ -m "not spec_validation"   1839 passed, 8 xfailed, 93 subtests passed
.venv/bin/ruff format --check ja4plus/ tests/      144 files already formatted
.venv/bin/mypy --strict ja4plus/                   Success: no issues found in 30 source files
```

The unit suite rises from 1814 passed to 1839 passed. The conformance suite is unchanged.
`tests/foxio_deviations.json` holds 135 keys before and after, against 135 `xfailed`, so
the register invariant holds. Coverage holds at 88%, and `ja4plus/cli.py` rises from 59%
to 65%.

Verified against: https://docs.python.org/3/library/signal.html#note-on-sigpipe (CPython
3, Note on SIGPIPE, retrieved 2026-08-08). The handler follows that recipe, including its
exit status of 1.

## Two notes for the reviewer

`ja4plus --output out.json db info` now parses and ignores `--output`, where it was an
`argparse` error before. The base already accepted `--format`, `--types` and `--lookup`
before `db` and ignored all three, so the shape is not new. #57 owns the `db` subcommand.

A capture the command cannot read leaves a partial output file, which is what a shell
redirection of standard output also produces. `docs/output-schema.md` states it and asks
the reader to read the exit status.

## Files that hold a `TBD`

- `docs/specs/spec.md` — the new `Changelog` row, round `TBD`.
- `CHANGELOG.md` — the new `### Changed` entry, `Round TBD.`


Crank-Git:
## Self-review record

Two Sonnet children ran the correctness lens and the security lens over the diff. Both
returned. Their findings and the ruling on each:

**Accepted and fixed**

1. *Security, high.* `os.path.exists` followed a symbolic link and reported False for a
   link whose target is absent, so a planted link passed the overwrite guard and
   `open(path, "w")` created the target. No race was needed, and `cmd_live` runs as root.
   Fixed: the open mode is `x` when the user passes no `--force`.
   `test_the_command_writes_through_no_symbolic_link` holds the reading.
2. *Security, medium.* The check-to-open race let a second writer lose its file. The same
   mode `x` closes it. `test_the_command_refuses_a_file_that_arrives_after_the_test`
   holds the reading.
3. *Writing standard.* Six sentences ran over the STE word cap, in `ja4plus/cli.py`,
   `README.md`, `docs/output-schema.md` and `tests/test_cli_output_option.py`. Each one
   is split.

**Accepted as a note, not a change**

4. A capture the command cannot read leaves a partial output file. A shell redirection of
   standard output produces the same result, so the behaviour holds and
   `docs/output-schema.md` states it and asks the reader to read the exit status.
5. `ja4plus --output out.json db info` parses and ignores `--output`. The base already
   accepted `--format`, `--types` and `--lookup` before `db` and ignored all three, so
   the shape is not new. #57 owns the `db` subcommand.

**Reported clean**

The correctness lens ran the program in every option position and found no defaulting
defect, no swallowed `SystemExit`, no leaked file handle and no path where `count` is
unbound. Neither lens found a new bare `Exception` catch, a new `assert` inside the
package, an unbounded structure, or a packet reference held after `process_packet`
returns.

Issue #309, opened 2026-08-08
Read the module x509 import in extract_certificate_info


Crank-Git:
Found on 2026-08-08 while #294 narrowed the handler at `ja4plus/utils/x509_utils.py:267`.
The defect sits on a different line, so #294 reports it and repairs none of it.

## What the code holds

`extract_certificate_info` imports `x509` and `default_backend` inside the function, at
`ja4plus/utils/x509_utils.py:262-263`. The import sits inside the branch that runs only
when `extract_certificate_from_bytes` returns nothing. Python reads the two names as
locals of the whole function, because the function binds them.

`ja4plus/utils/x509_utils.py:277` then calls `x509.load_der_x509_certificate(cert_data,
default_backend())`. That line runs only when `extract_certificate_from_bytes` returns a
certificate, which is the branch the import never reaches. The call therefore raises
`UnboundLocalError`.

`ja4plus/utils/x509_utils.py:279` catches it with `except Exception`, so the function
returns `None` for every packet that carries a certificate it found.

## The measurement

A packet that holds one TLS certificate message reaches the defect. The reader finds the
certificate and then returns nothing.

```
extract_from_bytes -> b'0\x82\x02\x910\x82\x01y'
  Found certificate handshake at position 0
  Handshake type: 11, length: 667
  Certificate list length: 664
  First certificate length: 661
Direct certificate parsing failed: cannot access local variable 'x509' where it is not associated with a value
extract_certificate_info -> None
```

The module already imports `x509` and `default_backend` at `ja4plus/utils/x509_utils.py:14-15`,
so the two local imports remove the working names.

## Why no fingerprint moves

No caller in `ja4plus/` calls `extract_certificate_info`. `JA4XFingerprinter` reads a
certificate with `read_certificate`, which imports nothing inside a branch. The defect
therefore reaches no fingerprint today, and it reaches any caller of the public utility.

## What the repair needs

Remove the two imports at `ja4plus/utils/x509_utils.py:262-263` and read the module-level
names. Write the case first: a packet that holds one TLS certificate message produces the
certificate details, and the reader writes no `UnboundLocalError` record.

Read the two wide handlers of the same function as part of the repair.
`ja4plus/utils/x509_utils.py:279` and `:284` each write `except Exception`, which is the
form `CLAUDE.md` forbids and #294 removed from three other lines.

Crank-Git:
[...]

Issue #308, opened 2026-08-08
Batch: the hygiene findings workers reported out of scope


Crank-Git:
A loose batch formed by the project manager on 2026-08-08. **Every member was found by a
worker building something else, and every one was reported rather than repaired out of scope.**
That is the behaviour this project wants, and this batch is where the reports become work.

## Members, in order

- [x] #294 — name the exceptions the X.509 reader expects. Three sites write
  `except (ValueError, TypeError, Exception)`, which `CLAUDE.md` forbids. Found by #267.
- [ ] #306 — settle the word for the serialized output line. `record` sits in the `Do not use`
  column of the term `result` in `docs/specs/spec.md`. Found by #50.
- [x] #309 — read the module `x509` import in `extract_certificate_info`. The defect #294's wide catch was hiding. Found by #294.
- [x] #314 — remove the X.509 helpers no caller uses. Split from #309 by the user's decision.
- [x] #316 — name the errors the two X.509 byte reader handlers expect. Split from #314.
- [ ] #319 — read the fourteen bare `Exception` handlers the package still holds. Reported by #316.
- [ ] #302 — resolve the three `TBD` Changelog rounds that reached `dev` in `CHANGELOG.md`,
  **and add the guard that stops a fourth.** Found by the #291 merge worker.

## Why they are sequenced and not parallel

**All three write `docs/specs/spec.md`, and two write `CHANGELOG.md`.** Members that share a
file are sequenced so the project manager resolves one conflict rather than three. #302 runs
last on purpose: it assigns round numbers, and the rounds its siblings write must exist first.

## What no member does

**No member changes a fingerprint, and no member changes the published output schema.** #50
froze that schema at version 1. #306 renames prose and identifiers; **it does not rename a JSON
field or a CSV column**, because that is a schema change that raises the version.

**No member picks its own Changelog round.** The project manager assigns each at sub-merge.

Crank-Git:
## Ship-partial decision by the project manager, 2026-08-08

**This batch ships its four completed members now, and its three remaining members move to a
follow-up batch #320.** PR #321 holds the four.

**Why.** The batch was formed with three members and grew to seven, because **every member of
the X.509 chain found the next one**: #267 found #294, #294 found #309, #309's question produced
#314, and #314 found #316. That is the batch working as intended, but `batchSize` is 4 and a
batch that grows while it runs never closes.

**The four that ship are one coherent change** — the X.509 reader, from the deceptive
[...]

Issue #294, opened 2026-08-08
Name the exceptions the X.509 reader expects


Crank-Git:
Found on 2026-08-08 during the self-review of #267. The finding is out of the scope of
that issue, so this issue records it and #267 changes none of the three lines.

## What the code holds

Three handlers write `except (ValueError, TypeError, Exception) as e:`. `Exception` is a
superclass of the other two names, so each handler catches every exception.

| Line | Method |
|---|---|
| `ja4plus/fingerprinters/ja4x.py:446` | `JA4XFingerprinter.get_cert_details` |
| `ja4plus/fingerprinters/ja4x.py:476` | `JA4XFingerprinter.read_certificate` |
| `ja4plus/utils/x509_utils.py:258` | The X.509 reader |

## Why it matters

`CLAUDE.md` states the rule: "A fingerprinter catches the parse errors it expects. It
does not catch bare `Exception`." `.claude/rules/conformance.md` states the same rule.

A handler that catches every exception hides a defect of this project beside a defect of
the packet. `MemoryError` and `KeyboardInterrupt` reach no reader either, because
`KeyboardInterrupt` is no `Exception` but `MemoryError` is.

## What the repair needs

Read which exceptions `cryptography` raises for a certificate this project cannot read,
and name those alone. `x509.load_der_x509_certificate` raises `ValueError` for a
certificate it cannot parse. `cert.extensions` raises `DuplicateExtension` and
`UnsupportedGeneralNameType`, which both inherit `Exception` and not `ValueError`. D2 of
`docs/specs/foxio/JA4X.md` records the duplicate-attribute case.

## The gate

No JA4X value moves. The conformance suite reports 1529 passed, 143 skipped and 137
xfailed on `batch/291-foxio-rulings` with #267, and it must report the same counts after
the repair.

Crank-Git:
Triaged 2026-08-08 by the project manager as `status:ready`.

**The finding is real and it names a rule this project writes down.** `CLAUDE.md` states that
a fingerprinter catches the parse errors it expects and does not catch bare `Exception`.
`except (ValueError, TypeError, Exception)` catches everything, and naming `ValueError` and
`TypeError` beside `Exception` reads as a narrow catch while behaving as a wide one. That is
worse than a plain bare catch, because a reader checking the rule sees two named errors and
moves on.

**#267 found it and correctly did not repair it.** #267 owned the JA4X raw form; it moved one
of the three lines and wrote none of them. A repair belongs in its own change.

[...]

PR #313, opened 2026-08-08
Read the module x509 import in extract_certificate_info

Crank-Git:
Part of #308. Repairs #309.

## The defect, measured

On the base commit `74de3d7`, `ja4plus/utils/x509_utils.py:262` and `:263` held the two
imports inside the branch that a payload of more than 100 bytes enters. A branch-local
import makes the name a local of the whole function, so the parse at
`ja4plus/utils/x509_utils.py:285` raised `UnboundLocalError`.

**The failure is on the path that succeeds, and not on a failure path.** The reader
raised for every packet whose payload carried a certificate it found, and the
`except Exception` below the parse returned nothing and wrote no log record.

The verbose form of the unrepaired function writes the message for a TLS Certificate
record that carries one self-signed certificate:

```
Direct certificate parsing failed: cannot access local variable 'x509' where it is not associated with a value
```

The line numbers of the issue body and of the round above it in `docs/specs/spec.md` are
stale, because the #294 merge moved them. This round measured the site.

## The repair

The module imports `x509` and `default_backend` at line 14 and line 15, so the
branch-local import bought nothing. The two lines are deleted, and a comment above the
module import states the reason a reader cannot see.

## How it is tested

`tests/test_x509_certificate_info.py` holds three cases, and the middle case failed
against the base:

- `test_the_reader_finds_the_certificate_of_a_tls_certificate_record`
- `test_the_packet_reader_returns_the_details_of_a_certificate_it_found` — **failed on
  the base**
- `test_the_packet_reader_returns_nothing_for_a_packet_that_carries_no_payload`

The case was written first and watched fail. **The repair is proven by its reversal, and
the reversal was proven live.** Putting the two lines back fails the case, three times of
three. Following #51 and #294, `inspect.getsource` of the loaded function reports the
branch-local import present after each reversal and absent after each restore, with
`PYTHONDONTWRITEBYTECODE=1` and a cleared `__pycache__`.

```
=== round 1 ===
before-reversal loaded source holds the branch-local import: False
1 passed
after-reversal  loaded source holds the branch-local import: True
1 failed
after-restore  loaded source holds the branch-local import: False
```

## Gates

| Gate | Result |
|---|---|
| `pytest tests/ -m "not spec_validation"` | 1781 passed, 8 xfailed, 93 subtests passed |
| `pytest tests/ -m spec_validation` | 1531 passed, 143 skipped, 135 xfailed |
| `ruff format --check ja4plus/ tests/` | 142 files already formatted |
| `mypy --strict ja4plus/` | no issues found in 29 source files |

**No fingerprint moves.** The conformance suite reports the same numbers before the
change and after it, and `tests/foxio_deviations.json` holds 135 keys against 135
xfailed cases. **Coverage rises**: the project holds 88% before and 89% after, and
`ja4plus/utils/x509_utils.py` rises from 34% to 54%.

## The open question, which the user decides

**No caller inside `ja4plus/` calls `extract_certificate_info`**, and no case reached its
certificate path before this round. That is why the defect survived a green suite, and it
is the sixteenth recorded instance of a comparison that is never made.

**The two readers of the interface disagree.**

- `__all__` of `ja4plus/__init__.py` names 25 public names for version 1.0.0, and
  `extract_certificate_info` is not one of them.
- `docs/api_reference.md:540` documents it under `ja4plus.utils.x509_utils`.

This pull request repairs the import and removes nothing, because the repair is safe
under either answer and the removal is not. A name that leaves the package before version
1.0.0 is the user's decision.

## What this pull request leaves

Two wide handlers remain in `extract_certificate_info`. Each one writes a plain
`except Exception`, so #294 narrowed neither. Their fate follows the removal decision, so
this round files no issue for them.

`extract_certificate_from_bytes` holds the same branch-local import shape. It raises
nothing, because the block that imports the names also uses them, and no case covers that
ASN.1 path. This round leaves it rather than change code it cannot measure.

Verified against: https://cryptography.io/en/latest/x509/reference/ (`cryptography`
50.0.0, retrieved 2026-08-08)


Crank-Git:
## Self-review

Three lenses ran over the diff, and this worker ran each one itself.

### Correctness

**The removal changes no name binding.** The branch below the deleted lines reads
`x509.DuplicateExtension` in its handler, and the module-level `from cryptography import
x509` binds the same module object. The 21 cases of
`tests/test_ja4x_named_exceptions.py` and `tests/test_x509_certificate_info.py` pass
together.

**The repair opens no new failure to a caller.** The parse at the repaired path still
sits under a handler, so `extract_certificate_info` returns nothing rather than raise.
Rule 2 of `CLAUDE.md` holds: a parser that cannot read a packet returns nothing.

**One of the three new cases measures the repair, and two do not.** Only
`test_the_packet_reader_returns_the_details_of_a_certificate_it_found` fails against the
base. The other two state the contract and pass on both sides, and this comment records
that rather than let the count read as three measurements.
`.claude/rules/conformance.md` asks what would fail if the code were wrong, and the
reversal answers it: the case fails three times of three.

### Security

The change deletes two import lines and adds no input path. The new case builds its own
key and reaches no network. No packet reference survives the call, no `assert` reaches
the package, and no wide handler is added.

### Interface

**The two readers of the public interface disagree, and that is the real finding.**
`__all__` of `ja4plus/__init__.py` names 25 public names for version 1.0.0 and omits
`extract_certificate_info`, while `docs/api_reference.md:540` documents it. No caller
inside `ja4plus/` calls it. The pull-request body carries the question, and the user
decides it.

### Left, with the reason

Two `except Exception` handlers remain inside `extract_certificate_info`. Each writes the
plain form, so #294 narrowed neither, and their fate follows the removal decision.

`extract_certificate_from_bytes` holds the same branch-local import shape and raises
nothing, because the block that imports the names also uses them. No case covers that
ASN.1 path, so this round measures no change to it and makes none.

Issue #314, opened 2026-08-08
Remove the X.509 helpers no caller uses before version 1.0.0


Crank-Git:
Split from #309 by the project manager on 2026-08-08. **#309 repaired the import and returned
the removal question, which is what its brief required. The user answered it on 2026-08-08.**

## The decision

**`extract_certificate_info` leaves the package before version 1.0.0, together with the sibling
helpers that no caller uses and `__all__` does not name.**

## The evidence the decision rests on

| Reader | What it says |
|---|---|
| `ja4plus/__init__.py:104-136` | `__all__` names 25 entries and **omits** the function |
| `ja4plus/` | **No caller.** `#309` searched the package and found none |
| The suite | **No case executed its certificate path** until #309 wrote one |
| `docs/api_reference.md:540` | Documents it, under `ja4plus.utils.x509_utils` |

**The last two lines together are the reason this matters.** A documented function that nothing
calls and nothing tests carried a live `UnboundLocalError` on its success path for the whole
history of this project, and a plain `except Exception` hid it. **It is the fifteenth instance
of the shape this project records: a comparison that is never made reads as a comparison that
passes.** Version 1.0.0 promises 25 names; this is not one of them.

## What to remove, and what to check first

**Measure before you delete. Every removal below is conditional on there being no caller.**

1. `extract_certificate_info` in `ja4plus/utils/x509_utils.py`, and the test #309 added for it.
2. **The sibling helpers of that module that no caller uses and `__all__` does not name.** #309
   reports three. **Confirm the list yourself** — search `ja4plus/`, `tests/` and `examples/`
   for each name before removing it. **A helper with a caller stays**, whatever `__all__` says,
   and you report that rather than breaking the caller.
3. The `docs/api_reference.md` entries for every name removed.

**Two things ride on this and disappear with it, which is half the value of the decision:**

- **Two plain `except Exception` handlers remain inside `extract_certificate_info`.** #294
  narrowed neither, because it targeted the deceptive `(ValueError, TypeError, Exception)`
  form. Removing the function removes them.
- **`extract_certificate_from_bytes` holds the same branch-local import shape** #309 repaired.
  It raises nothing today only because the block that imports the names also uses them, and no
  case covers that path. **If that function survives the caller check, repair the import shape
  rather than leaving a second copy of a defect this project has already paid for.**

## What must not move

**No fingerprint moves and no published name leaves.** `__all__` holds 25 names before and
[...]

Issue #297, opened 2026-08-08
Enable the ruff F401 rule


Crank-Git:
## Context

`pyproject.toml` turns the `ruff` rule F401 off, and its comment names #47 as the issue
that turns the rule back on. #47 declares `__all__` in `ja4plus/__init__.py`, which
clears the re-export findings. The rule still reports 58 findings, and none of the 58
sits in `ja4plus/__init__.py`.

Measured on `issue/47-py-typed` at `ruff 0.14.5`:

```
ruff check --select F401 ja4plus/ tests/
Found 79 errors.        # before #47 declares __all__
Found 58 errors.        # after
[*] 58 fixable with the `--fix` option.
```

The 58 sit in 28 files. `ja4plus/utils/x509_utils.py` holds 5,
`ja4plus/fingerprinters/ja4.py` and `ja4plus/fingerprinters/ja4h.py` hold 3 each, and
the rest sit in the test suite. #47 removes none of them, because an import a
fingerprinter no longer reads is a separate change from the public interface.

## What to build

Remove the 58 unused imports. Then remove the `F401` entry from the `ignore` list of
`[tool.ruff.lint]` in `pyproject.toml`, and update the comment that names #47.

Run the conformance suite, because the change touches files under
`ja4plus/fingerprinters/` and `ja4plus/utils/`.

## Acceptance criteria

- [ ] `ruff check ja4plus/ tests/` passes with `F401` absent from the `ignore` list.
- [ ] `pytest tests/ -m "not spec_validation"` reports the same counts as `dev`.
- [ ] `pytest tests/ -m spec_validation` reports the same counts as `dev`.

## Links

- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)

Crank-Git:
Sub-merged into `batch/362-hygiene-three` as `23bbed8`. Labelled `status:batched`. The issue closes when the batch pull request merges into `dev`.

**F401 is enforced and 54 unused imports are gone.** `pyproject.toml` now ignores `I001` alone.

## Every recorded number in this issue was stale

| Measurement | This issue recorded | Measured on the base |
|---|---|---|
| F401 findings | 58 | **54** |
| Files holding one | 28 | **27** |
| `ja4plus/utils/x509_utils.py` | 5 | **3** |
| `I001` findings, in the `pyproject.toml` comment | 82 | **76** |

The `ruff` version differed too: the issue measured with 0.14.5 and the base measures with **0.16.2**.

[...]

PR #315, opened 2026-08-08
Remove the X.509 helpers no caller uses before version 1.0.0

Crank-Git:
Part of #308. Refs #314.

The user decided on 2026-08-08 that `extract_certificate_info` leaves the package before
version 1.0.0, together with the sibling helpers no caller uses and `__all__` does not
name.

## The measurement came before the delete

One grep of the whole repository, excluding `.git/`, `.venv/` and `ja4plus.egg-info/`,
reports every reference to each of the seven names the module held.

| Name | References the grep found | Verdict |
|---|---|---|
| `extract_certificate_info` | Its definition, 2 cases of `tests/test_x509_certificate_info.py`, 5 cases of `tests/test_ja4x_named_exceptions.py`, `docs/api_reference.md:540` | **Removed.** No caller inside `ja4plus/` |
| `get_certificate_issuer` | Its definition alone | **Removed.** No caller anywhere |
| `get_certificate_subject` | Its definition alone | **Removed.** No caller anywhere |
| `get_name_attribute` | Its definition and 3 calls, all 3 inside the two functions above | **Removed.** No caller outside the group |
| `oid_to_hex` | `ja4plus/fingerprinters/ja4x.py:20`, plus 3 test files | **Stays.** A fingerprinter reads it |
| `get_cert_details` | 2 calls inside `extract_certificate_info`, 1 monkeypatch in a test of that function | **Stays.** It held a caller at the measurement, and `docs/api_reference.md:538` documents it |
| `extract_certificate_from_bytes` | 1 call inside `extract_certificate_info`, 1 case of the reader test | **Stays.** The issue body names it a survivor |

**#309 reported three siblings, and this branch re-measured the list rather than trust the
count.** The count is three. `get_certificate_issuer` and `get_certificate_subject` hold
no caller today. `get_name_attribute` holds three callers, and all three sit inside those
two functions, so the three form a closed group that nothing outside reaches.

**A helper with a caller stays.** `get_cert_details` now holds no caller at all, because
its only two callers sat inside the removed function. A decision on it is not covered by
the user's decision, so this branch leaves it and reports it.

## The two things that disappear with the function

- **Two plain `except Exception` handlers leave with `extract_certificate_info`.** #294
  narrowed neither, because it targeted the deceptive `(ValueError, TypeError, Exception)`
  form.
- **Two plain `except Exception` handlers survive in the module**, at
  `ja4plus/utils/x509_utils.py:148` and `ja4plus/utils/x509_utils.py:154`, inside
  `extract_certificate_from_bytes`. This branch leaves both. The outer one holds the whole
  body, and a narrow list there changes what the reader raises for hostile input, which
  needs its own cases and its own issue.

## The second copy of the import shape is repaired

`extract_certificate_from_bytes` imported `x509` and `default_backend` inside the ASN.1
branch, and the module imports both names already, so the two lines bought nothing.

**The shape is latent there, and no case turns it red.** The block that imports the two
names also reads them, so the read follows the import for every input. The repair
therefore changes no behaviour, and no red-to-green case exists to write.

**The ASN.1 path carried no case at all**, which is the defect this module has now paid
for twice, so this branch covers it.
`test_the_asn1_reader_finds_a_certificate_that_no_record_carries` builds one certificate
that the record reader rejects, and it requires the ASN.1 reader to return that
certificate. The fixture builds certificates until the record reader rejects one, because
a certificate body carries the byte pair the record reader scans for by chance. Six runs
of the file report `4 passed`.

## How this was tested

All five gates, on the final tree:

```
pytest tests/ -m "not spec_validation"   1776 passed, 8 xfailed, 93 subtests passed
ruff format --check ja4plus/ tests/      142 files already formatted
mypy --strict ja4plus/                   Success: no issues found in 29 source files
```

Proof that the removal is safe rather than merely green:

- **`import ja4plus` plus a `getattr` of each of the 25 `__all__` names** reports no
  missing name. `__all__` names 25 unique entries before and after.
- **`mypy --strict ja4plus/`** is clean, so no reference to a removed name survives.
- **A grep of each removed name after the removal** returns only the Changelog rows and
  the docstrings that record the removal. No code holds a reference.
- **The register invariant holds.** `tests/foxio_deviations.json` holds 135 keys against
  135 `xfailed` cases, before and after. The conformance suite is byte-identical.
- **Coverage rises**, as the removal of untested code should. The project holds 88%
  before and 90% after, and the missed statements fall from 463 of 3904 to 402 of 3842.
  `ja4plus/utils/x509_utils.py` rises from 44% to 66%.
- The unit suite falls from 1781 passed to 1776 passed. Eight cases leave with the
  function they measured, and three new cases cover the ASN.1 reader.

## TDD

**TDD applies oddly to a deletion, and this branch says so rather than invent a test.**
The red-to-green here is the caller search and the import check, both above. #309's case
`test_the_packet_reader_returns_the_details_of_a_certificate_it_found` measured
`extract_certificate_info`, so it leaves with the function it measured, together with the
five cases of `tests/test_ja4x_named_exceptions.py` that reached the same function.
`tests/test_x509_certificate_info.py` is now `tests/test_x509_certificate_reader.py`,
because the function it named is gone and it measures `extract_certificate_from_bytes`.

## Notes

- `binascii`, `hashlib` and `ExtensionOID` stay unread in the module. They were unread
  before this branch, and #297 owns the F401 findings. This branch removed `NameOID` and
  `scapy.all.Packet`, because its own change orphaned both.
- `docs/api_reference.md` was checked in both directions. It documented one of the four
  removed names, at line 540, and that row is gone. It documents `oid_to_hex`,
  `get_cert_details` and `extract_certificate_from_bytes`, and the package still holds
  all three.
- `TBD` markers: `docs/specs/spec.md` holds a new round row, and `CHANGELOG.md` holds a
  new `### Removed` entry. Both write `TBD` for the round number.
- No fingerprint moves. No entry of `__all__` moves.

Verified against: https://cryptography.io/en/latest/x509/reference/ (`cryptography`
50.0.0, retrieved 2026-08-08).


Crank-Git:
## Self-review

**I ran every lens myself rather than delegate it, and I say so here.** The lenses are
correctness, the test change, the import audit, the documentation, and the `CLAUDE.md`
rules. Three findings follow. Two are reports the issue asks for, and one is a note.

### Finding 1 — `get_cert_details` now holds no caller at all

`get_cert_details` held two callers at the time of the measurement, and both sat inside
`extract_certificate_info`. The removal takes both away, so the function now holds no
caller anywhere.

**I left it, and I did not decide it.** The issue names the removal set as
`extract_certificate_info` plus the three siblings, and it states that a helper with a
caller stays. `docs/api_reference.md:538` documents the function. A removal of it is not
covered by the user's decision of 2026-08-08, so it belongs to the user rather than to
this branch.

### Finding 2 — two plain `except Exception` handlers survive in the module

They sit at `ja4plus/utils/x509_utils.py:148` and `ja4plus/utils/x509_utils.py:154`,
inside `extract_certificate_from_bytes`. Acceptance criterion 3 asks this branch to state
whether any survives, and these two do.

I left both. The inner one holds the ASN.1 parse, and a narrow list there is small work.
The outer one holds the whole body, and a narrow list there changes what the reader
raises for hostile input, against `CLAUDE.md` rule 2. That needs its own cases and its
own issue, and #316 records it.

### Finding 3 — the shared certificate builder now uses an elliptic-curve key

`tests/test_x509_certificate_reader.py` builds its certificates with
`ec.SECP256R1()` where the file built them with RSA 2048 before. The retry fixture builds
up to 10 certificates, and 10 RSA keys cost far more than the file should. The record
reader measures byte framing, which no key type changes, and the case still passes.

### The import audit

An AST audit of `ja4plus/utils/x509_utils.py` reports that `NameOID` and
`scapy.all.Packet` are gone, because this branch orphaned both. `Any`, `default_backend`,
`logging` and `x509` are all still read. `binascii`, `hashlib` and `ExtensionOID` stay
unread, because they were unread before this branch and #297 owns the F401 findings.

### What I checked and found clean

- `import ja4plus` plus a `getattr` of each of the 25 `__all__` names reports no missing
  name.
- `mypy --strict ja4plus/` is clean, so no reference to a removed name survives.
- A grep of each removed name returns only Changelog rows and docstrings.
- `tests/foxio_deviations.json` holds 135 keys against 135 `xfailed` cases.
- No `assert` reaches the package, no new bare `Exception` reaches the package, and no
  state table changed.

Issue #56, opened 2026-08-06
Add the capture filter and a real privilege check


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/06-live-capture.md`](docs/specs/features/06-live-capture.md):

> FR-live-capture-11 — The `--bpf` option applies a capture filter.

> FR-live-capture-12 — The command reports a clear error when it lacks the privilege to open the interface.

> FR-live-capture-13 — The command works on Linux and on macOS.

## What to build

`ja4plus/cli.py:217` reads `os.geteuid() != 0`. Two problems: `os.geteuid` does not exist on Windows, so the command raises `AttributeError` instead of reporting that Windows is unsupported; and a Linux host may grant `CAP_NET_RAW` without granting user identity zero, so a permitted operator is refused.

Check privilege by attempting the capture and reading the failure. On Linux the capability is `CAP_NET_RAW`; on macOS the operator needs read access to the `/dev/bpf*` devices.

The `--bpf` option applies a capture filter through `sniff(filter=...)`.

Also remove `examples/monitoring_daemon.py`; the supported command replaces it, and Epic 8 documents that.

## Acceptance criteria

- [ ] Running without capture privilege exits with status 1 and names the required privilege.
- [ ] Naming a nonexistent interface exits with status 1 and lists the available interfaces.
- [ ] An invalid capture filter exits with status 1 and reports the filter error.
- [ ] Running on Windows reports that Windows is not supported and exits with status 1.

## Links

- Feature: [`docs/specs/features/06-live-capture.md`](docs/specs/features/06-live-capture.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #17
Depends on #53

Issue #55, opened 2026-08-06
Report monitor statistics


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/06-live-capture.md`](docs/specs/features/06-live-capture.md):

> FR-live-capture-8 — The command reports statistics on exit.

> FR-live-capture-9 — The `--stats-interval` option makes the command report statistics on a schedule.

> FR-live-capture-10 — The command writes statistics to standard error.

## What to build

The statistics line:

```
[ja4plus] packets=1284302 fingerprints=48211 connections=8134 evicted=112094 dropped=0 uptime=3600s
```

The statistics thread is the only thread the command starts, and it starts only when `--stats-interval` is passed.

Dropped-packet counts come from the capture layer when it reports them, and are `null` when it does not. Measure whether `scapy` reports them on both Linux and macOS, and state the finding in the pull request.

The counts come from `Processor.stats()`, added in Epic 3.

## Acceptance criteria

- [ ] The final statistics line reports the packet count, the fingerprint count, the connection count and the eviction count.
- [ ] `--stats-interval 1` writes a statistics line every second.
- [ ] Statistics go to standard error.

## Links

- Feature: [`docs/specs/features/06-live-capture.md`](docs/specs/features/06-live-capture.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #17
Depends on #53

PR #317, opened 2026-08-08
Add the watch command with a connection table and eviction

Crank-Git:
## What

`ja4plus watch <interface>` reads packets from an interface until the operator stops it.
`ja4plus live` stays as an alias of it. The command owns a connection table, and that
table holds a maximum entry count and a maximum age. `--max-connections` and
`--connection-timeout` set the two bounds.

New file `ja4plus/watch.py` holds the monitor loop and the connection table. New file
`tests/test_watch.py` holds 19 cases.

Part of #17. Refs #53.

## Why

`cmd_live` built its processor and called `Processor.cleanup_connection` never, so a
monitor on a busy interface held the state of every connection it ever read and grew
until the host stopped it.

`CLAUDE.md` states the rule: a state table has a maximum entry count and a maximum age,
and nothing that survives across packets grows without a limit. #279 sets the stated
ceiling of the package at 512 MiB for the one-million-packet case, against a measured
238.96 MiB. A monitor whose table held no bound would break that number.

## How

The connection table is a `BoundedStateTable`, the form Epic 14 shipped. This change
adds no second form of a state table.

- The count bound removes the least recently used connection as soon as the table
  reaches `--max-connections`.
- The age bound removes a connection that sends no packet for `--connection-timeout`
  seconds of **capture time**. A capture replays faster than real time, so a wall clock
  would evict state the capture still needs.
- The eviction hook of the table calls `Processor.cleanup_connection`. One eviction
  therefore drops the entry of the connection table **and** the per-connection state of
  all ten methods together. An eviction that dropped the entry alone would be the leak
  wearing a bound as a disguise.

Eviction runs on packet arrival. The command starts no thread.

`Monitor.handle_packet` records the connection first and reports the packet second. An
eviction pass that ran after the report would drop the state of the packet that just
arrived.

The connection key reads the outer address layer and the innermost port layer, through
`ja4plus/utils/packet_utils.py`, because the ten methods read the same two layers. A key
read from other layers would name a connection whose state `cleanup_connection` never
drops, and a tunnelled connection is where the two readings differ. The two directions
of one connection sort to one key.

One parser serves `watch` and `live`, so the two accept the same options and behave the
same way. `ja4plus --help` prints `watch (live)`.

## How tested

No test opens a capture interface. A background worker can open none, and this change
asks for none. Every case drives the loop from an injected packet source, and the two
command-level cases replace `read_interface` with a call that replays a list. **The
capture layer itself is therefore unmeasured here.**

The tests came first. The whole file failed to collect against the base commit
`250a2ac`, because `ja4plus/watch.py` did not exist.

### Gates

```
.venv/bin/pytest tests/ -m "not spec_validation"   1858 passed, 8 xfailed  (base: 1839 passed, 8 xfailed)
.venv/bin/pytest tests/ -m spec_validation         1531 passed, 143 skipped, 135 xfailed  (unchanged)
```

No fingerprint moved. `tests/foxio_deviations.json` holds 135 keys against 135 `xfailed`,
before and after.

Coverage holds at 88% overall. `ja4plus/watch.py` reports 94%; the three uncovered lines
are the body of `read_interface`, which calls `scapy.all.sniff`. `ja4plus/cli.py` rises
from 65% to 73%.

### The bound, measured

- 100000 packets across 50000 connections leave the table holding 10000 connections,
  which is the default `--max-connections`, and the eviction count rises. That is the
  acceptance criterion at its stated size, and it runs in 0.92 s.
- 1000000 packets across 50000 connections report 10000 tracked connections, 990000
  evictions and **7.20 MiB** held by the monitor, measured with `tracemalloc`. The
  package states 512 MiB for that case.

### Three reversals, each proven live first

and `inspect.getsource` on the loaded member reported the text **present**, then
**absent** under the mutation, then **present** again after the revert. #50 found that a
length-preserving edit lets CPython load a stale `.pyc`, so a green result under a
mutation that never loaded proves nothing.

| Mutation | Member | Cases that fail |
|---|---|---|
| body replaced by `return` | `Monitor._drop_connection` | 3 |
| `max_connections` replaced by `1000000000` | `Monitor.__init__` | 4 |
| `on_packet` announcement replaced by `pass` | `Monitor.handle_packet` | 2 |

Unchanged source: 19 passed. After the revert: 19 passed.

## Two readings this change records

**The `--connection-timeout` default of 300 no longer matches the library.**
`DEFAULT_MAX_CONNECTION_AGE` in `ja4plus/utils/state_table.py` is 600, because #179
raised it above the 320.714503 second gap of `ssh-r.pcap`. The issue body and
`docs/specs/features/06-live-capture.md` both state 300 and both call it the library
value. The command ships the stated 300, and Changelog round 97 records that the stated
reason is stale. The effect is the stricter direction: the monitor sheds an idle
connection before the state tables of the ten methods do.

**Three options of this feature belong to siblings and are absent.** #54 owns the
termination signals and the flush, #55 owns `--stats-interval` and the statistics line,
and #56 owns `--bpf` and the privilege check. The privilege check therefore still reads
`os.geteuid`, which the feature file asks #56 to replace with a failed capture attempt.
`examples/monitoring_daemon.py` stays until the command carries all three.

## Documentation

- `docs/specs/spec.md`: the `## Terms` table gains `monitor` and `connection table`, and
  the Changelog gains round **TBD** (written as row 97).
- `CHANGELOG.md`: an `Unreleased` entry that carries round **TBD**.
- `docs/usage.md`: a section that documents the command, its two options and the two
  bounds.
- `docs/api_reference.md`: the CLI list, the option tables and a `ja4plus.watch` section.
- `README.md`: the CLI examples.

`monitor` is a noun in the `## Terms` table, so the help text of the subcommand reads
`Read packets from a network interface`.

Verified against: https://scapy.readthedocs.io/en/latest/api/scapy.sendrecv.html (scapy
2.6, retrieved 2026-08-08), for `sniff(prn=..., iface=..., store=0)`.


Crank-Git:
## Self-review

No review child ran. I ran the four lenses over the diff myself, because a review that
does not return is a review that did not happen.

### Correctness

**The connection key must name the same connection the ten methods key on.** This is the
one reading that decides whether the eviction reaches anything. `ja4t.py`, `ja4l.py` and
the other stateful methods build their keys from
`ja4plus.utils.packet_utils.packet_endpoints`, which reads the **outer** address layer
and the **innermost** port layer. `connection_key` calls the same function, so a
tunnelled connection keys the same way on both sides of `cleanup_connection`. A key built
from `Processor.get_shard_key`, which reads `packet[TCP]` and therefore the outer port
layer, would have named a different connection for `gre-sample.pcap`,
`gre-erspan-vxlan.pcap` and `tcpdump-geneve.pcap`.

**The order inside `handle_packet` is load-bearing.** The monitor announces the packet,
records the connection, and reports the packet last. A report that ran before the record
would let the age pass evict the connection of the packet that just arrived, because that
connection still carries its old last-seen time until the write.
`test_the_monitor_reports_one_packet_once` and
`test_a_connection_that_keeps_sending_stays_in_the_table` hold the reading.

**The eviction hook writes to no table this class owns.** `BoundedStateTable` states that
a hook which writes to its own table produces undefined results. `_drop_connection` calls
`Processor.cleanup_connection`, which touches the tables of the ten methods and not the
connection table.

### Security

The privilege check is unchanged and still refuses a user identity other than zero. #56
owns the replacement the feature file asks for. No path in this change opens a file, and
the `--output` guard of #52 runs before the interface opens.

`--max-connections` accepts any whole number of one or more, so an operator who states a
very large number gets a table that grows to it. That is the same contract
`BoundedStateTable` gives a library caller, and it is the operator's own statement.

### Two findings, both fixed in `5a42822`

**1. The README stated a rule that this change made false.** It read "Every option runs
before the subcommand name and after it." That was true of the five output options, which
`_add_output_options` registers on the main parser and on each subparser. It is false for
`--max-connections` and `--connection-timeout`, which belong to `watch` alone:

```
$ ja4plus --max-connections 5 watch eth0
ja4plus: error: argument COMMAND: invalid choice: '5' (choose from analyze, watch, live, cert, db)
```

Registering the two on the main parser would make `ja4plus --max-connections 5 analyze
capture.pcap` parse and do nothing, which is worse than the refusal. The README now names
the five and states where the two run.

**2. `Monitor` acquires no lock, and #55 adds a thread.** `tracked_connections` builds a
list from the table. A statistics thread that calls it while the capture thread writes the
table reads a list the writer changed under it. No second thread exists today, so this is
no defect today. The docstring now carries the warning, so #55 meets the constraint before
it writes the thread rather than after.

### Frontend

No user-facing surface other than the command line. `ja4plus --help` prints
`watch (live)`, so the alias is documented where an operator looks for it.

Issue #320, opened 2026-08-08
Stop the monitor on an interface that carries no traffic


Crank-Git:
## What #54 found

`ja4plus watch` stops on `SIGINT` and on `SIGTERM`. The handler sets a flag, and `scapy`
reads that flag through the `stop_filter` argument of `sniff`.

`scapy` applies `stop_filter` to a packet, and to nothing else. Its capture loop blocks
in `select` with no timeout while the interface carries no traffic. The loop below holds
that behaviour, in `scapy/sendrecv.py`, in `AsyncSniffer._run`:

```python
while sniff_sockets and self.continue_sniff:
    if timeout is not None:
        remain = stoptime - time.monotonic()
        ...
    sockets = select_func(list(sniff_sockets.keys()), remain)
```

`remain` is None when the caller passes no `timeout`, so `select` waits without an end.
A monitor on a quiet interface therefore reads the flag after the next packet arrives,
and not when the signal arrives. An operator who runs `systemctl stop` on such a monitor
waits for the service timeout, and the host then sends `SIGKILL`.

Version 0.6.0 did not have this behaviour: it installed no handler, so the default
`SIGINT` handler raised `KeyboardInterrupt` inside `select` and ended the capture at
once. #54 replaced that handler, because a `KeyboardInterrupt` that arrives during a
write truncates the line.

## What to decide

`docs/specs/features/06-live-capture.md` names `sniff(stop_filter=...)` as the mechanism,
so #54 shipped it and changed nothing else. Three options exist, and each one carries a
cost the feature file does not cover.

1. Open the capture socket in `read_interface` and call `sniff` with `opened_socket` and
   a short `timeout` in a loop. The socket stays open across the calls, so no packet is
   lost. The unmeasured part of `ja4plus/watch.py` grows, because no test opens an
   interface.
2. Run the capture through `AsyncSniffer` and call `stop` from the main thread. The
   behaviour rules of the feature file state that the statistics thread is the only
   thread the command starts, so this option changes a stated rule.
3. Accept the behaviour, and document it. `docs/usage.md` states it today.

## Acceptance criteria

- [ ] A monitor on an interface that carries no traffic exits within one second of
      `SIGTERM`.
- [ ] The monitor loses no packet that arrives while it stops.
- [ ] The command starts no thread other than the one `--stats-interval` starts.

## Links

- Feature: [`docs/specs/features/06-live-capture.md`](docs/specs/features/06-live-capture.md)
[...]

Issue #54, opened 2026-08-06
Handle termination signals and flush the output on exit


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/06-live-capture.md`](docs/specs/features/06-live-capture.md):

> FR-live-capture-5 — The command exits cleanly when it receives `SIGINT`.

> FR-live-capture-6 — The command exits cleanly when it receives `SIGTERM`.

> FR-live-capture-7 — The command flushes its output before it exits.

## What to build

A signal handler sets a flag. It does not call `sys.exit`, because a signal may arrive while the output buffer is half written. The loop finishes the line it is writing, flushes, then exits.

`scapy`'s `sniff` reads the flag through its `stop_filter` argument.

The deprecated `ja4plus/collector.py` called `sys.exit` from inside its handler at line 40. Epic 4 removed that module; do not repeat the pattern.

Verified against https://scapy.readthedocs.io/en/latest/api/scapy.sendrecv.html (scapy 2.6, retrieved 2026-08-06).

## Acceptance criteria

- [ ] Sending `SIGTERM` to the monitor produces exit status zero.
- [ ] Sending `SIGINT` to the monitor produces exit status zero.
- [ ] The output file written before `SIGTERM` holds every fingerprint the monitor reported.

## Links

- Feature: [`docs/specs/features/06-live-capture.md`](docs/specs/features/06-live-capture.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #17
Depends on #53

Issue #53, opened 2026-08-06
Add the watch command with a connection table and eviction


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/06-live-capture.md`](docs/specs/features/06-live-capture.md):

> FR-live-capture-1 — The command `ja4plus watch <interface>` reads packets from an interface until it is stopped.

> FR-live-capture-2 — The command evicts a connection that has sent no packet for longer than the configured age.

> FR-live-capture-3 — The `--max-connections` option sets the maximum number of tracked connections.

> FR-live-capture-4 — The `--connection-timeout` option sets the maximum connection age in seconds.

> FR-live-capture-14 — The command `ja4plus live` remains as an alias of `ja4plus watch`.

## What to build

New file `ja4plus/watch.py`, holding the monitor loop and the connection table.

`ja4plus/cli.py:215` `cmd_live` never calls `cleanup_connection`, so a monitor on a busy interface grows until the host stops it. The command owns the connection table, records the last-seen time per connection, and calls `Processor.cleanup_connection` on eviction.

Eviction runs on packet arrival. The command starts no timer thread for eviction.

Defaults match the library: 10000 connections, 300 seconds.

Test the loop by feeding packets directly. Do not open a capture interface in a test.

## Acceptance criteria

- [ ] `ja4plus watch --help` documents every option this feature lists.
- [ ] Feeding 100000 packets across 50000 connections holds the tracked-connection count at or below `--max-connections`.
- [ ] A connection that stops sending is absent from the connection table after `--connection-timeout` seconds of capture time.
- [ ] `ja4plus live eth0` behaves the same as `ja4plus watch eth0`.

## Links

- Feature: [`docs/specs/features/06-live-capture.md`](docs/specs/features/06-live-capture.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #17

PR #323, opened 2026-08-08
Stop the monitor on a termination signal and flush the output

Crank-Git:
## What

`ja4plus watch` stops on `SIGINT` and on `SIGTERM`, and each ends the run with the status
zero. The command flushes its output before it exits.

`ja4plus/watch.py` gains `StopRequest` and `stop_on_signal`. `ja4plus/cli.py` passes
`StopRequest.stop_after` to `read_interface`, which passes it to `sniff` as `stop_filter`,
and it flushes the result stream before it returns. `tests/test_watch.py` gains 17 cases.

Part of #17. Refs #54.

## Why

A signal handler that calls `sys.exit` ends the run at the point the signal arrived, and
that point holds half a line whenever the signal arrives during a write. Version 0.6.0
called `sys.exit` inside the handler of `ja4plus/collector.py`, and Epic 4 removed that
module.

The command flushed the output only when it wrote a result, so a monitor that produced no
fingerprint left its header in the buffer.

## How

The handler sets a flag and returns. `scapy` reads that flag through the `stop_filter`
argument of `sniff`, and it applies the filter to a packet **after** it reports the
packet. The loop therefore finishes the line it writes. The command flushes after the loop
returns.

`stop_on_signal` restores the handler it replaced, whether the body returns or raises, so
a library caller that runs a monitor twice leaves no handler behind.

(retrieved 2026-08-08). The `sniff` entry of that page states: "stop_filter: Python
function applied to each packet to determine if we have to stop the capture after this
packet." The order of the `prn` call and the filter comes from the pinned source in this
worktree, `scapy/sendrecv.py`, in `AsyncSniffer._run`:

```python
if prn:
    result = prn(p)
    ...
# check
if (stop_filter and stop_filter(p)) or \
        (0 < count <= self.count):
    self.continue_sniff = False
```

**The pinned `scapy` is 2.7.0, and the issue body cites 2.6.** The page the body names is
the `latest` page, its `sniff` entry states the same behaviour, and
`.claude/rules/external-apis.md` pins "2.4 or later". The citation holds, and this note
records the difference.

## How tested

**No test opens a capture interface.** A background worker can open none. Every case
drives the command from an injected packet source, and that source applies `stop_filter`
in the order the `scapy` source above holds. **The capture layer itself is therefore
unmeasured**, which is the same three lines round 97 records for #53.

**A flush that the interpreter runs at shutdown is no flush this project implemented.**
Python flushes every stream at shutdown, and `close` flushes a file, so a case that ends
the program and then reads a complete file passes against a command that flushes never.
The cases therefore write to a recorder that records the order of the calls, and they
assert `flush` and then `close`. The one case that fails without this change writes the
`csv` header and no fingerprint, because no per-result flush runs on that path.

**The cases also prove that the handler does not exit.** Three cases send a real `SIGTERM`
or `SIGINT` to the test process while the loop holds a packet, and they read the output
the loop must still write. A handler that exited would pass a case that asked only whether
the capture stopped. Each of those cases installs a handler of its own first, so a build
that installs no handler fails one assertion rather than ending the test run.

**The tests came first.** The whole file failed to collect against the base commit
`379c8a0`: `ImportError: cannot import name 'StopRequest' from 'ja4plus.watch'`.

### Gates

```
.venv/bin/pytest tests/ -m "not spec_validation"   1875 passed, 8 xfailed  (base: 1858 passed, 8 xfailed)
.venv/bin/pytest tests/ -m spec_validation         1531 passed, 143 skipped, 135 xfailed  (unchanged)
```

No fingerprint moved. `tests/foxio_deviations.json` holds 135 keys against 135 `xfailed`.
`ja4plus.__all__` names 25 entries. Coverage holds at 88% overall, and `ja4plus/watch.py`
rises from 94% to 96%; its 3 uncovered lines are the body of `read_interface`, which calls
`scapy.all.sniff`.

### Four reversals, each proven live first

Every `__pycache__` outside `.venv/` was cleared, `PYTHONDONTWRITEBYTECODE=1` was set, and
`inspect.getsource` on the loaded member reported the marker **present**, then **absent**
under the mutation, then **present** again after the revert. #50 found that a
length-preserving edit lets CPython load a stale `.pyc`.

| Mutation | Member | Cases that fail |
|---|---|---|
| The final `stream.flush()` removed | `ja4plus.cli.cmd_watch` | 1 |
| `stop.stop_after` removed from the `read_interface` call | `ja4plus.cli.cmd_watch` | 1 |
| `self._requested = True` replaced by `pass` | `ja4plus.watch.StopRequest` | 5 |
| `self._requested = True` replaced by `raise SystemExit(0)` | `ja4plus.watch.StopRequest` | 8 |

The first mutation ran once with a marker that named text outside the member, and the
probe reported `ABSENT` before and after. That reading was discarded and the mutation was
run again with a marker inside the member.

## A finding, filed as #320

**The monitor reads the flag on packet arrival alone.** `scapy` blocks in `select` with no
timeout, so an interface that carries no traffic holds the monitor until the next packet
arrives. `docs/specs/features/06-live-capture.md` names `sniff(stop_filter=...)` as the
mechanism, so this issue ships that mechanism and changes nothing else. #320 records the
gap, states the three options and the cost of each, and stays untriaged.

## Changelog

`docs/specs/spec.md` holds round **98** as `TBD`, below the round 97 that #53 wrote as
`TBD` on this branch. `CHANGELOG.md` holds one `TBD` marker for #54.


Crank-Git:
## Self-review

I ran each lens myself and read the whole diff. Review children have stalled repeatedly
this session, so no child ran and none is reported.

### Correctness

1. **Fixed: `StopRequest.request` carried two default arguments that nothing used.**
   `signal.signal` always calls a handler with two arguments, and every case passes both.
   The defaults were flexibility that no caller asked for, so the signature now reads
   `request(self, signal_number: int, frame: Optional[FrameType])`. Commit `eef43ee`.

2. **Accepted: the command prints `\nCapture stopped.` on the flag path as well.**
   `SIGINT` no longer raises `KeyboardInterrupt`, so the message would be lost without
   this line, and an operator who reads the earlier `Starting live capture` line expects
   the matching one. The leading newline exists because a terminal echoes `^C`, and a
   `SIGTERM` that arrives from `kill` therefore writes one blank line to standard error
   first. The exit summary belongs to #55, so this change adds no other text.

3. **Accepted: `stop_on_signal` runs inside the `except Exception` guard of `cmd_watch`.**
   `signal.signal` raises `ValueError` outside the main thread, and the guard reports it
   as `Error during capture: signal only works in main thread of the main interpreter`
   and exits 1. That message names the cause, and the guard predates this change.

4. **Checked: the flush that this change adds can raise.** A broken pipe on standard
   output raises `BrokenPipeError` from the new `stream.flush()`, outside the `try` of
   `cmd_watch`. `main` catches it and calls `_exit_on_broken_pipe`, which is the path
   #52 built. The `finally` of `_result_stream` raised the same error at the same point
   before this change.

### Security

5. `StopRequest` holds one boolean. It reads no packet field, and `stop_after` keeps no
   reference to the packet it receives, so no code holds a packet after `process_packet`
   returns. The change opens no file, reads no environment variable, and reaches no
   network.

6. `stop_on_signal` replaces the `SIGINT` handler of the whole process for the length of
   the capture. It restores the handler it replaced in a `finally`, so a library caller
   that embeds `cmd_watch` keeps its own handler afterwards. One case measures the
   restore after a normal return, and one measures it after an exception.

### The project rules

7. No new state table. `CLAUDE.md` bounds a structure that survives across packets, and
   `StopRequest` holds one boolean.
8. No `assert` in `ja4plus/`, and no new bare `Exception` catch.
9. No fingerprinter and no parser changed, so no fingerprint can move. The conformance
   suite reports the same 1531 passed, 143 skipped and 135 xfailed.

Issue #326, opened 2026-08-08
Report the dropped-packet count of the capture layer


Crank-Git:
## What was measured

`ja4plus watch` writes `dropped=null` in every statistics line. #55 built the field and
measured the capture layer, and `scapy` 2.7.0 reports no drop count to a caller of
`sniff`.

1. On macOS the capture socket reads a drop count, and `sniff` keeps the socket to
   itself. `_L2bpfSocket.get_stats` at `scapy/arch/bpf/supersocket.py:297` reads the
   `BIOCGSTATS` ioctl and returns the received count and the drop count.
   `AsyncSniffer._run` at `scapy/sendrecv.py:1205` holds the socket it opens in the
   local name `sniff_sockets`, so a caller reaches that socket through the
   `opened_socket` argument alone. `ja4plus/watch.py` passes no `opened_socket`.
2. On Linux `scapy` reads no drop count. It defines `PACKET_STATISTICS = 6` at
   `scapy/arch/linux/__init__.py:95` and it calls `getsockopt` with that option nowhere.

Reading 1 ran on macOS 25.6.0 against `scapy` 2.7.0. Reading 2 is a reading of the
`scapy` source, and no Linux host ran it.

## What to build

`MonitorStats` already accepts a `dropped_source` callable, and `ja4plus/cli.py` passes
none. The command must open the capture socket itself, pass it to `sniff` as
`opened_socket`, and pass the drop count of that socket as `dropped_source`.

#56 opens the capture socket for the privilege check, so that work decides where the
socket lives. Build this issue after #56.

The Linux path needs a second measurement on a Linux host before it reports a number. A
`getsockopt(SOL_PACKET, PACKET_STATISTICS)` call reads the count from an `AF_PACKET`
socket, and that call resets the kernel counter, so the monitor must add each reading to
a running total rather than report the last one.

## Acceptance criteria

- [ ] A monitor on macOS reports a whole number in the `dropped` field.
- [ ] A monitor reports `dropped=null` where the capture layer reports no count.
- [ ] The pull request states which platform ran each measurement.

## Links

- Feature: [`docs/specs/features/06-live-capture.md`](docs/specs/features/06-live-capture.md)
- Found by #55.

## Amendment, 2026-08-09. The issue is split and the Linux half is measurable

**The user ruled on 2026-08-09: the Linux measurement runs on the granted Linux host,
bounded to the loopback interface.** #423 holds the macOS half.

### What the project manager measured on `bigboy` before writing this

[...]

Issue #17, opened 2026-08-06
Epic 6: Live capture


Crank-Git:
## Goal

Make a long-running monitor a supported mode rather than an example script.

## Exit criteria

`ja4plus watch` reads an interface, evicts connections by age, handles a termination signal, and reports its own statistics.

## Spec

- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Feature: [`docs/specs/features/06-live-capture.md`](docs/specs/features/06-live-capture.md)

Depends on: #16

## Sub-issues

- [x] #53
- [x] #54
- [x] #55
- [x] #56
- [x] #320

PR #329, opened 2026-08-08
Report monitor statistics on exit and on a schedule

Crank-Git:
Part of #17. Builds #55, on top of #53 and #54.

## What this adds

The monitor writes one statistics line when it exits, and `--stats-interval SECONDS`
adds a line for each interval that passes. Every line goes to standard error, which #52
established as the diagnostic channel.

```
[ja4plus] packets=1284302 fingerprints=48211 connections=8134 evicted=112094 dropped=0 uptime=3600s
```

`ja4plus/watch.py` gains `MonitorStats`, `StatisticsSnapshot`, `format_statistics`,
`write_statistics`, `StatisticsReporter` and `report_statistics`. `ja4plus/cli.py` gains
the `--stats-interval` option and writes the exit summary.

## How the thread reads counters another thread writes

`MonitorStats` holds the four counts and one `threading.Lock`. Every write and every
read acquires it, so a reader reads the counts of one instant.

The connection table is the harder half. It is a dictionary that the capture thread
rewrites on each packet, and a second reader of it reads entries the writer is moving.
`Monitor.handle_packet` therefore calls `record_packet(len(table), table.evictions)`
itself, after it writes the table and before it reports the packet. The statistics
thread reads `MonitorStats` and never the table.
`test_the_snapshot_reads_no_connection_table` replaces the table with an object that
raises on every read, and the snapshot holds.

The thread stops cleanly. It waits on a `threading.Event` rather than on a sleep, so
`stop` ends it inside a millisecond, and `report_statistics` wraps the `read_interface`
call, so a termination signal stops the monitor and the thread together. A `stop` that
sets no event fails 7 cases.

**A case that reads an invariant beside a writer measures nothing here, and this was
measured before it was trusted.** One reader against one writer, over 20000 writes, read
97635 snapshots and found 0 broken invariants with the lock and 0 without it, at the
default switch interval and at one microsecond. CPython leaves one thread inside three
attribute stores that hold no call and no backward jump, so the reader never samples the
gap. That case was discarded.
`test_a_write_waits_while_a_read_holds_the_guarded_region` replaces it: a clock that
blocks holds the reader inside the guarded region, and the writer reaches no count until
the reader leaves. It is the case that fails when either lock is removed.

## The dropped-packet count is `null`, and that is a measurement

`scapy` 2.7.0 reports no drop count to a caller of `sniff`.

1. **macOS.** The capture socket reads a drop count and `sniff` keeps the socket to
   itself. `_L2bpfSocket.get_stats` at `scapy/arch/bpf/supersocket.py:297` reads the
   `BIOCGSTATS` ioctl and returns the received count and the drop count.
   `AsyncSniffer._run` at `scapy/sendrecv.py:1205` holds the socket it opens in the
   local name `sniff_sockets`, so a caller reaches that socket through the
   `opened_socket` argument alone.
2. **Linux.** `scapy` reads no drop count at all. It defines `PACKET_STATISTICS = 6` at
   `scapy/arch/linux/__init__.py:95` and it calls `getsockopt` with that option nowhere.

**Reading 1 ran on macOS 25.6.0. Reading 2 is a reading of the `scapy` source, and no
Linux host ran it.** `MonitorStats` accepts a `dropped_source`, `ja4plus/cli.py` passes
none, and the field reads `null`. #326 owns the socket work that reports a count, and it
follows #56, which opens the capture socket for the privilege check.

Verified against: https://scapy.readthedocs.io/en/latest/api/scapy.sendrecv.html (scapy
2.7.0, retrieved 2026-08-08), and the installed `scapy` 2.7.0 source at the three
`file:line` references above.

## The counts are real

The counts of each case differ from each other, so a field wired to another counter
fails the case that names it.

- Seven packets across five connections, with a table of three, give `packets=7`,
  `fingerprints=4`, `connections=3`, `evicted=2`.
- One TCP handshake through the command gives `packets=2`, `fingerprints=3`,
  `connections=1`, and the fingerprint count is compared against the count of result
  lines the command wrote.

## Twelve mutations, each proven live first

and `inspect.getsource` on the loaded member reported the marker absent, then present,
then absent again after the revert. The two deletion mutations were probed in the other
direction, present then absent.

|---|---|
| The packet record swaps the two table counts | 5 |
| The packet counter counts nothing | 6 |
| The snapshot acquires no lock | 1 |
| The packet record acquires no lock | 1 |
| The reporter writes a line before the first interval passes | 1 |
| The exit summary reaches standard output | 5 |
| The fingerprint counter counts one for each packet | 1 |
| `report_statistics` starts the thread without the option | 1 |
| The stop sets no event | 7 |
| The uptime reads the packet count | 1 |
| The trailing window reaches no count | 1 |
| The interval guard admits a value that is no finite number | 2 |

## The self-review found two defects

1. **The exit summary counted the trailing window never.** `cmd_watch` writes the JA4SSH
   window that a capture leaves open, which #214 decided, and it counted that window
   never. `ssh.pcapng` produces two fingerprints through the monitor and the summary
   reported one. `test_the_final_line_counts_the_trailing_window` pins it.
2. **The interval guard admitted a value that is no finite number.**
   `--stats-interval nan` passed the guard, and `Event.wait` returns at once for `nan`,
   so the thread writes a statistics line without an end. `--stats-interval inf` passed
   the guard, and `Event.wait` raises `OverflowError` for `inf`, so the thread ends and
   the monitor keeps running. The guard now reads `math.isfinite`.

## What stays unmeasured

**No case opens a capture interface.** Every case drives the monitor from an injected
packet source, so the three lines of `read_interface` that call `scapy`'s `sniff` stay
uncovered, as rounds 97 and 98 record. The drop count of a real capture socket is
therefore unmeasured too, and #326 owns it.

`--stats-interval 1` is measured at a shorter interval. A case that waited one second
for each line would cost the suite more than it measures, so the cases use 0.05 seconds
and read the gaps between the lines.

## Gates

```
.venv/bin/pytest tests/ -m "not spec_validation"   1916 passed, 8 xfailed, 96 subtests passed
.venv/bin/ruff format --check ja4plus/ tests/      147 files already formatted
```

**No fingerprint moved.** The conformance suite reports 1531 passed, 143 skipped and 135
xfailed before and after, and `tests/foxio_deviations.json` holds 135 keys against those
135 `xfailed` cases. `ja4plus.__all__` holds 25 entries before and after.

**Coverage rises.** The base measures 88% over 4051 statements, and this branch measures
89% over 4141. `ja4plus/watch.py` rises from 96% to 98%. Both numbers come from
`.venv/bin/pytest tests/ -m "not spec_validation" --cov=ja4plus`, run against the base
commit and against this branch.

The known base flake
`tests/test_thread_safety.py::TestTheConcurrencyContract::test_eight_sharded_threads_read_one_timeline_the_way_one_thread_reads_it`
did not appear in any run.

## Files that hold a `TBD`

- `docs/specs/spec.md` — Changelog round `99 | TBD`. Rounds 97 and 98 belong to #53 and
  #54 on this branch, and this row takes 99. The project manager renumbers the epic's
  rows at the batch gate.
- `CHANGELOG.md` — the `Round TBD.` marker of the new `### Added` entry.

## Cost

`Monitor.handle_packet` acquires one lock for each packet. The per-packet path already
acquires ten inside `Processor`, one for each method.


Crank-Git:
Self-review, run by me over the diff, one lens at a time.

**Correctness.** Two defects, both fixed in this branch.

1. `cmd_watch` wrote the trailing JA4SSH window that #214 emits, and counted it never. `ssh.pcapng` produces two fingerprints through the monitor and the exit summary reported one. `test_the_final_line_counts_the_trailing_window` pins it, and a mutation that drops the count fails that case.
2. `--stats-interval` accepted `nan` and `inf`. `Event.wait` returns at once for `nan`, so the thread writes a statistics line without an end. It raises `OverflowError` for `inf`, so the thread ends and the monitor keeps running. The guard now reads `math.isfinite`, and a mutation that drops the guard fails 2 cases.

**Concurrency.** The first invariant case measured nothing, and the measurement found that before the case was trusted: with the lock and without it, one reader read 97635 snapshots over 20000 writes and found 0 broken invariants. CPython leaves one thread inside three attribute stores that hold no call and no backward jump. That case was discarded and replaced by `test_a_write_waits_while_a_read_holds_the_guarded_region`, which holds the reader inside the guarded region by hand. Removing either lock fails it.

**Security.** The diff parses no packet field and reaches no network. `--stats-interval` is the one new input, and it is validated.

**Cost.** `Monitor.handle_packet` now acquires one lock for each packet. The per-packet path already acquires ten inside `Processor`.

**Scope.** #56 keeps the capture filter and the privilege check. #326 keeps the drop count, because it needs the capture socket that #56 opens. #320 stays unrepaired.

Issue #319, opened 2026-08-08
Read the fourteen bare Exception handlers the package still holds


Crank-Git:
Reported by #316 on 2026-08-08 while it narrowed the last two handlers of
`ja4plus/utils/x509_utils.py`. **#316 named the sites and deliberately filed nothing, because
they are pre-existing and outside its issue.** That was right, and this issue records the work.

## The measurement, taken on `dev` by the project manager

`grep -rn "except Exception" ja4plus/` finds **18 sites**. **Four of them disappear when the
hygiene batch #308 merges** — two removed with `extract_certificate_info` by #314, and two
narrowed by #316. **Fourteen remain.**

| File | Lines | Count |
|---|---|---|
| `ja4plus/cli.py` | 368, 419, 474, 510, 558 | 5 |
| `ja4plus/processor.py` | 223, 259, 323 | 3 |
| `ja4plus/utils/quic_utils.py` | 381, 471, 622 | 3 |
| `ja4plus/ja4db.py` | 162, 189 | 2 |
| `ja4plus/__init__.py` | 83 | 1 |

**Line numbers move. Find the code.**

## This issue is a reading, not a sweep

**Do not narrow all fourteen.** `CLAUDE.md` states the rule for one role: *"A fingerprinter
catches the parse errors it expects. It does not catch bare `Exception`."* The fourteen sites do
not all hold that role, and treating them alike would be wrong in both directions — it would
force narrowing where the wide catch is the design, and it would let a real defect pass as
"already reviewed".

**Read each site and put it in one of three groups.** The verdict states the group and the
reason for every one of the fourteen.

1. **A parser reading hostile input.** The rule binds it. `ja4plus/utils/quic_utils.py` is the
   clear case: every packet is hostile input and a parser that cannot read a packet returns
   nothing. **Narrow these, by #294's and #316's method.**
2. **Deliberate collection of an error the caller receives.** `ja4plus/processor.py` is the
   likely case: #45 made the processor return the parse failures it collected, and
   `process_packet_with_method_errors` names the method behind each. **A catch that exists to
   hand an error onward is not the defect the rule describes** — but it must be documented as
   such at the site, so the next reader does not narrow it.
3. **Top-level reporting.** `ja4plus/cli.py` reports a failure to the user and exits; the
   package boundary in `ja4plus/__init__.py` and the lookups in `ja4plus/ja4db.py` may be the
   same shape. **A command-line program that turns any unexpected error into a traceback is
[...]

PR #331, opened 2026-08-08
Add the capture filter and a real privilege check

Crank-Git:
Part of #17. Builds #56.

## What this changes

`--bpf FILTER` passes a Berkeley Packet Filter expression to `sniff(filter=...)`, so the
capture layer drops every packet the filter rejects and the monitor never reads it.

The command reads no user identity. It attempts the capture and reads the failure.
Version 0.6.0 read `os.geteuid() != 0` at `ja4plus/cli.py:452`, and that check held two
defects. `os.geteuid` is absent on Windows, so the command raised `AttributeError` there.
A Linux host also grants `CAP_NET_RAW` without granting the user identity zero, so the
check refused a permitted operator.

`examples/monitoring_daemon.py` is removed. `ja4plus watch` is the supported monitor, and
`docs/usage.md` documents it.

## Why it is shaped this way

`describe_capture_failure` reads one exception object and four strings. It calls no
capture function and it knows nothing about the way the socket opens.

#320 changes `read_interface` to open the capture socket itself and to call `sniff` with
`opened_socket` and a short `timeout`. The same exceptions still travel out of the same
call, so that change leaves the classification where it is. The `Raises:` section of
`read_interface` states the contract between the two.

The bare `except Exception` clause of `cmd_watch` stays. The self-review removed it and
then restored it: #319 owns the fourteen bare catches of this module, and a removal here
would turn every unclassified failure into a stack trace. The new clause sits above it
and reads `CAPTURE_FAILURES` alone. `BrokenPipeError` inherits `OSError`, so the clause
that re-raises it precedes both.

## What the capture layer reports

Verified against `scapy` 2.7.0, the version this virtual environment installed. The
interface table of `.claude/rules/external-apis.md` names `2.4 or later`, and #54 found
the same 2.7.0.

| Failure | Class | Source |
|---|---|---|
| The host refuses the `/dev/bpf` device. | `Scapy_Exception` | `scapy/arch/bpf/core.py:59` |
| The host holds no such interface. | `ValueError` | `scapy/interfaces.py:434` |
| `libpcap` refuses the filter. | `Scapy_Exception` | `scapy/arch/common.py:129` |
| A socket wraps the filter failure. | `Scapy_Exception` | `scapy/arch/bpf/supersocket.py:218`, `scapy/arch/linux/__init__.py:232` |

The Linux socket calls raise `OSError` and `scapy` wraps neither, so `EPERM` and `EACCES`
name a refused privilege and `ENODEV` names an absent interface.

`filter=None` is safe on both platforms: `scapy/arch/bpf/supersocket.py:213` and
`scapy/arch/linux/__init__.py:228` each attach a filter only where one is not None.

(scapy 2.6 and later, retrieved 2026-08-08). The four line numbers come from the
installed `scapy` 2.7.0 source, read on 2026-08-08.

## How this is tested

No test opens a capture interface.

Three cases produce a real `scapy` failure from a call that opens none, and they feed
that exception object to the classifier.

- `get_dev_bpf()` reported `Permission denied: could not open /dev/bpf0. Make sure to be
  running Scapy as root ! (sudo)`.
- `resolve_iface('nosuchif0')` reported `Interface 'nosuchif0' not found !`.
- `compile_filter('tcp port', linktype=1)` reported `Failed to compile filter expression
  tcp port (-1)`.

One case replaces `scapy.all.sniff` and reads its arguments, so `read_interface` runs for
real and opens no interface. The command-level cases inject the failure into the capture
call.

**The host is macOS 25.6.0 and it granted no capture privilege.** The `/dev/bpf` case
skips on any other platform and under the user identity zero.

**The Linux behaviour is a reading of the `scapy` source and no Linux host ran it.** The
`CAP_NET_RAW` refusal, the `EPERM` number and the `ENODEV` number are claims and not
measurements, and the two cases that cover them construct the `OSError` by hand.

`get_if_list()` reported 20 interfaces on this host without the capture privilege, which
is the reading that makes the interface list answerable.

## Three reversals

and `inspect.getsource` on the loaded member reported PRESENT under each mutation and
ABSENT after each revert.

| Reversal | Failing cases |
|---|---|
| The privilege reading replaced by `if False` | 5 |
| `filter=capture_filter` replaced by `filter=None` | 1 |
| The Windows reading replaced by `if False` | 5 |

## The removal

`git grep` over every tracked file reported two references to
`examples/monitoring_daemon.py` outside the file itself.

- The Purpose section of `docs/specs/features/06-live-capture.md`. This branch rewrote it.
- Changelog round 97 of `docs/specs/spec.md`, which records that the file stays until the
  command carries all three options. It stays as written.

No file under `tests/`, no file under `.github/` and no line of `README.md` named it.

## Gates

Every command ran in the worktree against a virtual environment built there.

```
.venv/bin/pytest tests/ -m "not spec_validation"   1950 passed, 8 xfailed, 96 subtests passed
.venv/bin/ruff format --check ja4plus/ tests/      148 files already formatted
```

No fingerprint moved. The conformance suite reports the same three counts before and
after, and `tests/foxio_deviations.json` holds 135 keys against those 135 xfailed.

Coverage holds at 89%, and the total misses fall from 476 of 4141 statements to 466 of
4178. `ja4plus/watch.py` rises from 98% to 99%, and `ja4plus/cli.py` rises from 75% to
78%. The removed example carried no coverage, because the measurement reads `ja4plus/`
alone.

## Changelog markers

`docs/specs/spec.md` holds Changelog row 100 with `TBD` in its date column.
`CHANGELOG.md` holds one `### Added` entry that reads `Round TBD`. The project manager
assigns both at the batch gate.


Crank-Git:
## Self-review

I ran the three lenses over the diff myself and recorded each finding below. I spawned
no review child.

### Correctness — three findings, all fixed

1. **The removal of the bare `except Exception` was out of scope, and it was a
   regression.** The first form of `cmd_watch` replaced that clause with the
   `CAPTURE_FAILURES` clause. Every failure outside the three classes then reached the
   operator as a stack trace. #319 owns the fourteen bare catches of this module.
   The clause is restored, below the new one, and `f4633ec` holds both.
   One case now covers it: `TheCommandReportsAFailureOfNoCaptureClass`.

2. **A comment named the wrong line of the `scapy` source.** `_PRIVILEGE_TEXT` cited
   `scapy/arch/bpf/core.py:60` where the docstring of `describe_capture_failure` cited
   `:59`. The `raise Scapy_Exception(` statement is at line 59. Both read `:59` now.

3. **The filter failure has two wrappers and the comment named one.** Reading
   `scapy/arch/bpf/supersocket.py:218` found that the macOS socket also writes
   `Cannot set filter:`, not the Linux socket alone. The comment, the feature file and
   the pull-request body all name both now.

### Documentation — two findings, both fixed

4. **An unmeasured performance claim.** `docs/usage.md` read "A filter costs less than a
   monitor that reads every packet and discards most of them". No measurement supports
   it. `4c58526` states the mechanism instead.

5. **A `setcap` recipe that does not work.** A first form of the start-up-error section
   carried `sudo setcap cap_net_raw,cap_net_admin=eip "$(which ja4plus)"`. `ja4plus` is a
   Python console script, and the kernel ignores a capability set on a file that begins
   with `#!`. I removed the recipe rather than publish an instruction I did not run on a
   Linux host.

### Security — no finding

The change opens no network connection and writes no file. The `--bpf` expression reaches
the `libpcap` compiler through `sniff(filter=...)` and no shell, so no expression reaches
a command line. `available_interfaces()` reads the interface list of the host and reports
it to the operator who ran the command, which discloses nothing that operator cannot
already read.

The classifier repeats the text of the failure the capture layer reported. That text
holds an interface name and a `/dev/bpf` device name, and it holds no credential.

### One reading I could not settle here

`unsupported_platform_message` refuses `cygwin` beside `win32`. Cygwin holds no
`/dev/bpf` device and no `AF_PACKET` socket, so the capture fails there, and
FR-live-capture-13 names Linux and macOS alone. No Cygwin host ran this. If the project
would rather let Cygwin attempt the capture and read the failure, that is a one-line
change to `_WINDOWS_PLATFORMS`.

PR #335, opened 2026-08-08
Stop the monitor on an interface that carries no traffic

Crank-Git:
Part of #17. Builds #320.

## What this changes

A monitor on an interface that carries no traffic exits within one second of `SIGINT` or
of `SIGTERM`. It exited when the next packet arrived, and an operator who ran
`systemctl stop` on such a monitor waited for the service timeout. The host then sent
`SIGKILL`, and a killed process skips the flush #54 built.

`read_interface` now opens the capture socket itself and calls `sniff` with
`opened_socket` and a timeout of 0.25 seconds, in a loop. It reads the stop request after
each call, and `scapy` still reads it after each packet through `stop_filter`.

The user chose this option on 2026-08-08 and declined the two others.

## Why the socket stays open

`AsyncSniffer._run` closes the sockets it opened itself and no other socket:

```python
if opened_socket is None:
    for s in sniff_sockets:
        s.close()
elif close_pipe:
    close_pipe.close()
```

The socket therefore stays open across the `sniff` calls, and the host buffer holds every
packet that arrives between two of them. A loop that reopened the socket would drop each
one. That property is the reason the option works, not a detail of it.

The gap it repairs sits in the same loop. `scapy` reads `remain` as None while the caller
states no timeout, so `select` waits without an end:

```python
while sniff_sockets and self.continue_sniff:
    if timeout is not None:
        remain = stoptime - time.monotonic()
        if remain <= 0:
            break
    sockets = select_func(list(sniff_sockets.keys()), remain)
```

## What #56 keeps

`--bpf` still applies. `libpcap` now compiles the expression when the socket opens, and
`read_interface` passes no `filter` to `sniff`, because `AsyncSniffer._run` reads that
argument only where it opens the socket itself. An expression stated there would reach no
socket.

`open_capture_socket` calls `resolve_iface(name).l2listen()(type=ETH_P_ALL, iface=name,
filter=...)`, which is the call `AsyncSniffer._run` makes. The same three failure classes
travel out of `read_interface`, so `describe_capture_failure`, `available_interfaces` and
the filter reading of #56 all still hold.

## Three findings of the self-review

**A closed socket.** `scapy` closes a socket whose read raised and drops it from its
list. A loop that called `sniff` again would wait on a closed socket for as long as the
monitor runs. The loop now reads `capture_socket.closed`, which `SuperSocket` holds as a
class attribute, and returns.
`test_the_capture_returns_when_the_capture_layer_drops_the_socket` pins it.

**A poll interval of zero.** `AsyncSniffer._run` breaks before it waits while the
remaining time is at or below zero, so `sniff` would return at once and the loop would
call it without a wait. `read_interface` now refuses a poll interval that is no positive
finite number, which is the guard #55 shipped for `--stats-interval` after `nan` produced
the same defect. Three cases pin it. `ja4plus watch` exposes no option for the value.

**The file descriptors of the control pipe.** `AsyncSniffer._run` builds one `ObjectPipe`
for each `sniff` call against a socket that reports a blocking select, and that pipe
holds two file descriptors. A monitor calls `sniff` four times every second and runs for
weeks, so `test_three_hundred_calls_hold_the_open_descriptor_count` reads `/dev/fd` after
each of 300 calls and reports the same count as before the loop.

## Verified against

Verified against `scapy` 2.7.0, the version this virtual environment installed, read on
2026-08-08:

| Reading | Source |
|---|---|
| `sniff` runs `AsyncSniffer._run` in the calling thread and starts no thread. | `scapy/sendrecv.py:1434` |
| `_run` opens a socket only where the caller supplies none. | `scapy/sendrecv.py:1268` |
| `_run` waits in `select` with `remain` set to None while the caller states no timeout. | `scapy/sendrecv.py:1331` |
| `_run` closes a socket whose read raised, and it drops that socket. | `scapy/sendrecv.py:1370` |
| `_run` closes the sockets it opened itself and no other socket. | `scapy/sendrecv.py:1391` |
| `_run` builds one control pipe for each call against a blocking socket. | `scapy/sendrecv.py:1300` |
| The macOS socket attaches the filter as it opens, and it wraps the failure as `Cannot set filter:`. | `scapy/arch/bpf/supersocket.py:215` |
| `SuperSocket` holds `closed` as a class attribute, and `close` sets it. | `scapy/supersocket.py:86` |

(scapy 2.6 and later, retrieved 2026-08-08).

## How this is tested

No case opens a capture interface. Every case of `tests/test_watch_stop.py` supplies the
capture socket, so the real `scapy` capture loop runs against a socket the test file
controls.

**The one-second bound is measured, not reasoned about.** Two cases send a real `SIGTERM`
and a real `SIGINT` to the test process from inside the wait of the capture socket, and
each one reads `time.monotonic` at the send and again after `read_interface` returns. Two
more cases measure the same bound through the whole command-line program.

**The interface carries no traffic at all in those cases.** Each one asserts that the
capture loop called `recv` never. A case that needed a packet to pass would measure the
mechanism #54 already shipped.

**The loop boundary is measured with a socket that a reopen would replace.**
`open_socket` builds a new socket for each call. The packet reaches the queue of the first
socket at the instant its timeout expires, and the case asserts that the report holds one
packet and that the run opened one socket.

## Six reversals, each proven live first

|---|---|
| The stop-request check of the loop reads `if False and ...`. | 11 |
| The loop reopens the socket for each `sniff` call. | 3 |
| `timeout=poll_interval` reads `timeout=None`. | 11 |
| The closed-socket guard reads `if False and ...`. | 1 |
| `stop_requested=stop.requested` leaves the `read_interface` call of `ja4plus/cli.py`. | 4 |
| The poll-interval guard reads `if False and ...`. | 3 |

Under the reopen mutation the boundary case reports
`AssertionError: 0 != 1 : the loop boundary lost the packet`, which is the failure the
second acceptance criterion names.

and `inspect.getsource` on the loaded member reported PRESENT under each mutation and
ABSENT after each revert. Four of the six mutations run the monitor without an end, so
those runs were bounded by `pytest-timeout`. The virtual environment holds it and the
project declares it nowhere.

The tests came first, and the whole file failed to collect against the base commit
`8b2bad7`, because `ja4plus/watch.py` held no `DEFAULT_POLL_INTERVAL` and no
`open_capture_socket`.

## What stays uncovered

The three lines of `read_interface` that rounds 97, 98 and 99 record as uncovered are now
zero. `ja4plus/watch.py` reports one uncovered line, line 131 of `connection_key`, which
this issue did not touch.

**The real socket open is covered and it is not measured.** The three cases that read
`open_capture_socket` replace `scapy.interfaces.resolve_iface`, so no case builds a real
`L2bpfListenSocket` and no case reads a real interface. The host is macOS 25.6.0 and it
granted no capture privilege.

## A reading the criteria do not cover

A packet the host buffered at the instant the signal arrives reaches no reader. The
monitor closes the socket and exits, and the mechanism #54 shipped drops the same buffer.
The second acceptance criterion reads the loop boundary, which the issue body states, and
no stop point of either design reads that buffer.

## Gates

```
.venv/bin/pytest tests/ -m "not spec_validation"   1974 passed, 8 xfailed, 96 subtests
.venv/bin/ruff format --check ja4plus/ tests/      149 files already formatted
```

No fingerprint moved. `tests/foxio_deviations.json` holds 135 keys against those 135
xfailed. `__all__` of `ja4plus` names 25 entries before and after. The unit suite rises
from 1950 passed to 1974 passed, which is the 24 cases this issue adds. Coverage holds at
89%, the total misses hold at 466, and no file falls.

## Changelog

`TBD` markers sit in `docs/specs/spec.md`, at Changelog round 101, and in `CHANGELOG.md`,
in the entry for #320. The project manager assigns both at the batch gate.


Crank-Git:
## Self-review, by lens

I ran every lens myself and none of it came from a child agent.

### Correctness

**A defect: `scapy` drops a socket whose read raised.** `AsyncSniffer._run` closes that
socket at `scapy/sendrecv.py:1370` and removes it from its list, and `sniff` then returns
because `sniff_sockets` is empty. A loop that called `sniff` again would wait on a closed
socket for as long as the monitor runs. Fixed: the loop reads `capture_socket.closed`.
`test_the_capture_returns_when_the_capture_layer_drops_the_socket` fails without it.

**A defect: a poll interval of zero turns the loop into a spin.** `AsyncSniffer._run`
breaks before it waits while the remaining time is at or below zero, so `sniff` returns
at once. `read_interface` now refuses a value that is no positive finite number, which is
the guard #55 shipped for `--stats-interval`. Three cases fail without it.
`ja4plus watch` exposes no option for the value, so no operator reaches it today.

**Checked and clean.** `stop_after` holds no packet after it returns, so the rule of
`CLAUDE.md` still holds. The change adds no state table and no bare `except Exception`.
It adds no `assert` in the package. `open_capture_socket` opens one socket and
`read_interface` closes it in a `finally`, so a failure of the report path leaks no
socket.

### Resource

**The control pipe of `scapy` costs two file descriptors for each `sniff` call.**
`AsyncSniffer._run` builds one `ObjectPipe` at `scapy/sendrecv.py:1300` for a socket that
reports a blocking select, which every real capture socket does. A monitor calls `sniff`
four times every second and runs for weeks, so this is worth a measurement rather than an
argument. Over 300 calls against a socket that reports a blocking select, `/dev/fd` holds
5 entries before the loop, 5 after each call and 5 after the loop.
`test_three_hundred_calls_hold_the_open_descriptor_count` holds the measurement.

### The acceptance criteria, measured at the shipped poll interval

`DEFAULT_POLL_INTERVAL` is 0.25 seconds. Against a socket that reports no packet at all,
with the signal sent from inside its wait:

```
SIGTERM: 0.2552 s to the return; recv calls=0
SIGINT: 0.2513 s to the return; recv calls=0
```

The `recv` count of zero is the part that matters: the interface carried no traffic, so
the mechanism #54 shipped could not have produced the return.

### Security

`--bpf` reaches `libpcap` through the socket rather than through `sniff`, and the
expression still comes from the operator's own command line. The trust boundary does not
move. The capture socket still needs `CAP_NET_RAW` on Linux or read access to
`/dev/bpf*` on macOS, and #56 still reports a refusal.

### What I did not measure

No case opens a capture interface, so no case builds a real `L2bpfListenSocket`. The
three cases that read `open_capture_socket` replace `scapy.interfaces.resolve_iface`. The
host is macOS 25.6.0 and it granted no capture privilege. The Linux socket path is a
reading of the `scapy` source, and no Linux host ran it.

Issue #334, opened 2026-08-08
Decide whether a declined Python value forfeits precedence to a Rust snapshot


Crank-Git:
Found by the run of #332 and not repaired there, because it needs a decision the user has
not made.

## The gap

#332 wrote one exception into `.claude/rules/external-apis.md`: **a FoxIO Python value
that `tests/foxio_deviations.json` declines under a decided entry forfeits its precedence,
and a Zeek baseline may hold the reference value for that method on that connection.** The
decision the user made names the Zeek baseline alone, so #332 wrote it for the Zeek
baseline alone.

**`docs/specs/foxio/JA4SSH.md:473-478` already applies the same reading to the FoxIO Rust
snapshot, and the rule text names no exception there.** The page reads:

> **`tests/foxio_vectors/ssh2.pcapng.json` holds a different second value, and the conflict
> resolves.** `.claude/rules/external-apis.md` states that `python/test/testdata/` decides
> where it and a Rust snapshot both carry a value for one method on one stream. Here the
> Python value is `c36s36_c0s0_c2s0`, which #97 declines, so the precedence rule points at a
> value this project already declined. The decline outranks it, and the Rust snapshot
> decides.

`.claude/rules/external-apis.md:74-78` states the Rust precedence and names no exception:

> Where both carry a value for one method on one stream, `python/test/testdata/` decides.

`docs/specs/features/01-spec-conformance.md:202-206` states the same rule and names no
exception either.

## The question the user decides

**Does a declined FoxIO Python value forfeit its precedence against the FoxIO Rust
snapshot too, on the same condition #332 wrote for the Zeek baseline?**

- **Yes** — the rule gains the same exception for the Rust snapshot, and
  `docs/specs/foxio/JA4SSH.md:473-478` follows from the rule.
- **No** — `docs/specs/foxio/JA4SSH.md:473-478` states a reading the rule denies, and the
  page is repaired.

**Do not answer it from which value looks right.** #332 states the condition as a recorded
decline that a reader checks in the register, and the same care applies here.

## What must not move

**No fingerprint moves. No file under `ja4plus/` changes.** No baseline and no snapshot is
adopted as a vector; this issue changes only which source may hold a reference value. The
register invariant holds: the key count of `tests/foxio_deviations.json` equals the
`xfailed` count, 135 against 135.

## Links

- Found by: #332
- The rule: `.claude/rules/external-apis.md`
- The decline that raises it: #97
[...]

Issue #332, opened 2026-08-08
Write the precedence exception for a declined FoxIO Python value


Crank-Git:
Split from #327 by the project manager on 2026-08-08. **#327 repaired the page, found that a
vector-adoption decision no longer followed from its evidence, and returned the question rather
than answering it. The user answered it on 2026-08-08.**

## The decision

**A FoxIO Python value that the `Divergence register` declines under a recorded decision
forfeits its precedence.** Where the register declines it, a Zeek baseline may hold the
reference value for that method on that connection.

**Python keeps its precedence everywhere else.** This adds one exception; it does not weaken the
rule.

## The reasoning the user gave, stated so the next reader inherits it

**Precedence exists to break a tie between sources this project trusts. A value this project has
already ruled wrong is not a tie.** `.claude/rules/external-apis.md:63-64` reads *"Where both
directories carry a value for a method, `python/test/testdata/` decides"* and names no
exception — **because the case had not arisen when the rule was written.**

## The case that raised it

| Source | Second JA4SSH value of `ssh2.pcapng` |
|---|---|
| `ja4plus` | `c36s52_c42s76_c51s2` |
| Zeek baseline | `c36s52_c42s76_c51s2` |
| FoxIO Python | `c36s36_c0s0_c2s0` |

`tests/foxio_deviations.json` declines the Python value under **#97**, at the key
`ssh2.pcapng/14:57377/JA4SSH.2`. **Two implementations agree and the third is one this project
already ruled against**, so the rule as written would have let a declined value block a vector
forever.

## What to build

1. **Write the exception into `.claude/rules/external-apis.md`**, beside the precedence sentence
   it qualifies. State the condition exactly: **the register declines the Python value under a
   recorded decision that names an issue.** A value that is merely *undecided* is not declined,
   and this exception does not reach it.
2. **Set the rating in `docs/specs/foxio/zeek.md`.** #327 left it reading `Undecided` and wrote
   the question into the page; replace both with the decision and the reason, and cite this
   issue.
3. **Report every other row the exception reaches.** The register holds 135 keys. **Search for
   any other method and connection where the register declines the Python value and a Zeek
   baseline holds one**, and list what you find. **Do not adopt any baseline as a vector** —
   adoption is its own decision, and this issue changes only which source may hold a reference.

## The trap to avoid

**Do not restate the exception as "the Zeek baseline decides when Python is wrong".** That is
wider than the decision and it invites a later reader to apply it from their own judgment. **The
[...]

Issue #322, opened 2026-08-08
Batch: the hygiene findings that follow #308


Crank-Git:
A follow-up batch formed by the project manager on 2026-08-08, when #308 shipped its four
completed members. **No member here is new work; all three were already ready and none is
dropped.** #308 records the ship-partial decision.

## Members, in order

- [x] #306 — settle the word for the serialized output line. `record` sits in the `Do not use`
  column of the term `result` in `docs/specs/spec.md`. Found by #50.
- [x] #324 — repair the Zeek baseline comparison script that #49 broke, and make a gate run it. Found by #306.
- [x] #327 — refresh the stale JA4SSH reading of the Zeek page. Found by #324, whose repair produced the new measurement.
- [x] #332 — write the precedence exception the user decided. Split from #327.
- [x] #334 — widen the precedence exception to every FoxIO source. Decided by the user; found by #332.
- [x] #341 — record the kind of decline in the register. Proposed by #334, decided by the user.
- [ ] #346 — measure the capability-declined rows so the field alone carries the bar. Split from #341.
- [ ] #279 — state the memory ceiling the user decided. Never scheduled until now.
- [ ] #359 — decide whether the lookup cache pays for eviction statistics nothing reads. Found by #60.
- [ ] #319 — read the fourteen bare `Exception` handlers the package still holds, and narrow the
  ones the rule binds. Reported by #316. **This is a reading, not a sweep.**
- [ ] #302 — resolve the three `TBD` Changelog rounds that reached `dev` in `CHANGELOG.md`, **and
  add the guard that stops a fourth.**

## The order, and why #302 is last

**#302 assigns round numbers, so the rounds it must not disturb have to stop moving first.** Its
siblings each write a round, and the project manager assigns those at sub-merge. #302 running
last means it repairs a settled table rather than a moving one.

**#319 may not fit in one issue.** Its own body says to read all fourteen sites, report the
grouping, narrow only the group the `CLAUDE.md` rule binds, and **stop and report if that group
holds more than about four sites** rather than letting one issue grow without a bound. If it
stops, the project manager splits it and this batch ships what is done.

## What no member does

**No member changes a fingerprint, no member changes the published output schema, and no member
touches `__all__`**, which names 25 entries and is the version 1.0.0 interface promise.

**No member picks its own Changelog round.** The project manager assigns each at sub-merge, in
both `docs/specs/spec.md` and `CHANGELOG.md` — assigning in one file alone is the defect #302
exists to repair.

[...]

PR #340, opened 2026-08-08
Make the precedence exception source-neutral

Crank-Git:
## What

The precedence exception of #332 becomes source-neutral. Where
`tests/foxio_deviations.json` declines the FoxIO Python value under a decided entry,
**any other FoxIO implementation may hold the reference value** for that method on that
connection. Two bars stand above the rule, as the user settled on 2026-08-08.

- **Bar 1 — a capability decline bars the row.** The exception reaches a row only where
  the decline records a disagreement about the value. #129 is the other kind: `ja4plus`
  reads no encrypted request by decision, so naming a decrypted value as the reference
  would create divergences no implementation change could close.
- **Bar 2 — a disagreement between the remaining sources bars the row.** Where the
  remaining sources hold different values, no source holds the reference and the row
  stays declined exactly as today.

## The reach: 6 rows of the 135

The widened rule **before** the bars reaches **58** of the 135 register keys, which
confirms the earlier measurement. Both bars leave **6**.

| Row | Decline | The source that may hold the reference |
|---|---|---|
| `ssh-r.pcap/2:46396/JA4SSH.1` | #96 | Rust and Wireshark, which agree |
| `ssh-scp-1050.pcap/0:49237/JA4SSH.3` | #96 | Rust and Wireshark, which agree |
| `ssh-scp-1050.pcap/0:49237/JA4SSH.4` | #96 | Rust and Wireshark, which agree |
| `ssh2.pcapng/14:57377/JA4SSH.2` | #97 | Rust and Zeek, which agree |
| `ssh2.pcapng/33:51810/JA4L-S.1` | #225 | Wireshark alone |
| `tls3.pcapng/25:61884/JA4L-S.1` | #225 | Wireshark alone |

`ssh2.pcapng/14:57377/JA4SSH.2` survives both bars, as the ruling expected. It is
confirmed, not assumed: Rust and Zeek both hold `c36s52_c42s76_c51s2`, and the Wireshark
testdata holds one JA4SSH value on that connection, which is occurrence 1.

## The search that proves the list complete

The search read **all 135 register keys** against 38 Rust snapshots, 37 Wireshark
testdata files and 7 Zeek baselines at the pinned commit
`27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8`. The funnel:

| Step | Keys |
|---|---|
| Register keys | 135 |
| The occurrence form declines no value | 76 remain |
| Every one of the 76 carries `"decided": true` | 76 remain |
| No other FoxIO source holds the value | 58 remain |
| Bar 1, the capability decline of #129 | 23 remain |
| Fact 1, the Python file holds no value | 22 remain |
| The standing Zeek JA4L bar, then bar 2 | **6 remain** |

**The Wireshark testdata records `frame.number` and no connection**, so a capture-level
reading cannot attribute a value to a connection. The search resolved every frame against
its capture and read the 5-tuple, which gives connection granularity for all three
sources. That matters: the earlier capture-level reading could not tell that the single
`ssh2.pcapng` JA4SSH value belongs to occurrence 1.

## Three findings

**1. The register records no field that separates the two kinds of decline.** Bar 1 is
therefore checkable only by naming the issue today, which
`tests/test_precedence_exception.py` does in `CAPABILITY_DECLINES`. The rule states a
proposal: give each entry a `"capability": true` field that `tests/foxio_deviations.py`
reads beside `decided` and `tests/test_foxio_deviations.py` gates the same way. **The
user decides whether to build it.** Nothing here infers the kind of a decline from prose.

**2. One value-form key declines no Python value.**
`gre-erspan-vxlan.pcap/0:65174/JA4T.1` carries the value form, but the FoxIO Python
expected-output file holds no JA4T value for that connection, and #215 declines the
**FoxIO Rust** value `8192__0_0`. Condition 1 of #332 read the key form as a proxy for
"the Python file holds a value to decline", and on that row the proxy is false. Left
unfixed, the exception would have named `8192__0_0` as the reference, contradicting the
decision #215 records. The rule now reads the fact from the vectors, and a case proves it.

**3. The bar on a Zeek JA4L value reaches a source it does not name.** The bar rests on
three rules, the second of which is "it appends a third part that the Python reference
does not publish". **The Wireshark dissector appends a third part to all 44 of its
`ja4.ja4l` values and all 44 of its `ja4.ja4ls` values** — a delta on a TCP connection
(`16_128_14777`) and the marker `quic` on a QUIC connection (`90_128_quic`). So the
reasoning reaches the dissector.

- **Effect on the reach: a bar that reached the dissector would remove the two #225 rows
  and leave 4.** Both are rows where the dissector stands alone.
- **Against widening it:** #225 records that this project adopted the `quic` marker from
  the dissector on purpose, and its cause cites `wireshark/test/testdata/` as publishing
  the marker on 18 of 88 values. Widening the bar would contradict that decision.
- **This change moves nothing there.** The user decides.

## What did not move

No fingerprint moves. No file under `ja4plus/` changes — `git diff HEAD~2 -- ja4plus/` is
empty. No vector is adopted and no register entry changes: `git diff HEAD~2 --
tests/foxio_deviations.json` is empty. The register invariant holds at **135 keys against
135 xfailed**, measured. `ja4plus.__all__` names exactly **25** entries, measured.

## Testing

`tests/test_zeek_precedence_exception.py` becomes `tests/test_precedence_exception.py`,
because it no longer reads the Zeek baseline alone. It grows from 4 cases to 19.

Written test-first: the reach computation and both bars were green on the first run,
and the 6 prose cases were red until the rule and the pages were written.

**The check can fail.** Remove `ssh2.pcapng/14:57377/JA4SSH.2` from `SOURCE_VALUES`:

```
=== UNDER THE MUTATION ===
inspect.getsource characters: 17606
marker occurrences in the loaded source: 0
FAILED tests/test_precedence_exception.py::TestTheReachOfTheException::test_the_exception_reaches_six_register_keys
FAILED tests/test_precedence_exception.py::TestTheThirdPartFinding::test_a_bar_that_reached_the_dissector_would_remove_two_rows
2 failed, 17 passed

=== AFTER THE REVERT ===
inspect.getsource characters: 17730
marker occurrences in the loaded source: 1
19 passed
```

The marker is the Zeek value of that row, which appears nowhere else in the file. An
earlier marker matched the undecided-entry fixture too, so it did not discriminate; this
one does, in both directions.

## Gates

```
.venv/bin/pytest tests/ -m "not spec_validation"   1862 passed, 8 xfailed, 93 subtests passed
.venv/bin/mypy --strict ja4plus/                   no issues found in 30 source files
```

Line coverage is 88 percent, unchanged, because no file under `ja4plus/` changed.

## Changelog

`TBD` markers are in `docs/specs/spec.md` (one round row) and `CHANGELOG.md` (one
`### Changed` entry).

## Verified against

```
Verified against: https://github.com/FoxIO-LLC/ja4/tree/main/python/test/testdata (commit 27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8, retrieved 2026-08-08)
Verified against: https://github.com/FoxIO-LLC/ja4/tree/main/rust/ja4/src/snapshots (commit 27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8, retrieved 2026-08-08)
Verified against: https://github.com/FoxIO-LLC/ja4/tree/main/wireshark/test/testdata (commit 27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8, retrieved 2026-08-08)
Verified against: https://github.com/FoxIO-LLC/ja4/tree/main/zeek/tests/Traces (commit 27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8, retrieved 2026-08-08)
```

Refs #334. Part of #322.


Crank-Git:
Self-review, run by this worker across three lenses because a review child was not spawned. Four defects found, all in this branch's own work, all fixed before the final push.

1. **Dead constant.** `CAPABILITY_KEYS_A_SOURCE_COVERS = 35` was defined and read by no case — prose in the shape of code, which is the exact pattern this issue exists to remove. Replaced with `CAPABILITY_VALUE_KEYS = 37` and `test_the_capability_bar_holds_out_thirty_seven_value_form_keys`, which computes the count from the register and fails when the register moves.
2. **An overclaiming test name.** `test_the_dissector_writes_a_third_part_on_every_ja4l_value_measured` reads only the 23 rows of `SOURCE_VALUES`, not all 88 dissector values. Renamed to `..._every_ja4l_value_the_table_holds`.
3. **A count that came from prose and not from the search.** The capability bar cited 30 rows, which the issue body carries. The measurement says #129 holds 37 decided value-form keys and another source covers 35. Fixed in a2fe6c1.
4. **A rule stated as a negative.** "changes what the bar covers on nobody's own judgment" breaks STE rule 15. Fixed in a2fe6c1.

A fifth item is recorded and not fixed: the #332 entries in `CHANGELOG.md` and the #332 round row in `docs/specs/spec.md` still name `tests/test_zeek_precedence_exception.py`, which this branch renames. They are the record of what #332 shipped, and rewriting them would misrepresent it. The #334 entry in each file states the rename, and it sits above the #332 entry in both.

Issue #341, opened 2026-08-08
Record the kind of decline in the register


Crank-Git:
Proposed by #334 and decided by the user on 2026-08-08. **#334 shipped the bar and reported that
its condition was not machine-checkable; this issue makes it so.**

## The measurement

**Bar 1 of the precedence exception reads a hardcoded issue number.**
`tests/test_precedence_exception.py` holds `CAPABILITY_DECLINES = {129}`, because
`tests/foxio_deviations.json` records no field distinguishing **a decline about the value** from
**a decline about a capability this project chose not to build.**

**The bar itself is right, and the user decided it.** #129 declines because `ja4plus` reads no
encrypted request, and naming the Wireshark decrypted values as reference would create 35
divergences no implementation change could close. **The defect is that the condition is a
constant in a test rather than a fact in the register.**

## Why this is worth building rather than noting

**A new capability decline would fall into the exception silently**, unless someone remembered to
edit a test that names it nowhere near where the decline is recorded. **That is the failure mode
this project keeps meeting**: a rule stated where nothing checks it. Four instances of prose
ahead of its evidence were recorded on 2026-08-08 alone.

## What to build

**Give each entry of `tests/foxio_deviations.json` an explicit `capability` field**, read by
`tests/foxio_deviations.py` beside `decided` and gated by `tests/test_foxio_deviations.py` the
same way `decided` already is.

**Bar 1 then reads one field of one entry.** Remove `CAPABILITY_DECLINES` from
`tests/test_precedence_exception.py`.

**Decide the default and state it.** An entry with no `capability` field is not a capability
decline — **but say so in the schema rather than relying on a falsy read**, and consider whether
the gate should require the field on every decided entry so a new one cannot omit it by accident.
That is the difference between a field and a checked field.

**Set the field on every existing entry.** 135 keys. **The only capability decline today is
#129** — verify that rather than assuming it, and report what you find. **An entry whose kind you
cannot determine from its recorded reason is a finding, not a coin toss**: report it and return
`needs-feedback`.

## Prove the field is load-bearing

**Set `capability` to `false` on a #129 entry and confirm the reach computation changes**, then
revert. The reach is **6 rows** today; a #129 entry admitted would raise it. Report both numbers.

**With the mutation proven live** — clear every `__pycache__` outside `.venv/`, set
`PYTHONDONTWRITEBYTECODE=1`, and confirm with `inspect.getsource`, discriminating in both
[...]

PR #344, opened 2026-08-08
Record the kind of decline in the register (#341)

Crank-Git:
Bar 1 of the source-neutral precedence exception read a hardcoded issue number.
`tests/test_precedence_exception.py` held `CAPABILITY_DECLINES = {129}`, because
`tests/foxio_deviations.json` recorded no field that separates a value decline from a
capability decline. A new capability decline would have reached the exception unless
somebody remembered to edit a test that named it nowhere near the decline.

Refs #341. Part of #322.

## What changed

**`tests/foxio_deviations.json` carries a `capability` field on all 135 entries.**
`true` records a capability this project chose not to build. `false` records a
disagreement about the value.

**`tests/foxio_deviations.py` reads the field beside `decided`.** `Deviation.capability`
defaults to `False`, and the module docstring states that default as a rule rather than
as a falsy read: an entry with no field is a value decline. The reader rejects a field
that is not `true` or `false`, the way it already rejects `decided`.

**`unkinded_declines` in `tests/test_foxio_deviations.py` requires the field on every
decided entry.** It reads the register file itself and not the `Deviation` map, because
the reader turns an absent field into the default and no caller could then tell an
omission from a decision. An undecided entry declines nothing, so it may omit the field.

**`CAPABILITY_DECLINES` is gone.** `exception_reach` reads `deviation.capability`, and
fact 3 of the rule now reads one field of one entry.

## How the kind of each entry was decided

The kind came from the recorded cause of each of the 135 entries, and not from the issue
number. A cause that states `reads no encrypted request` or `reads no encrypted
certificate` records a capability decline, because `ja4plus` reads no key material by a
decision Changelog round 26 settled.

**43 entries carry `"capability": true`, and all 43 name #129.** The field and the
constant `{129}` agree exactly.

**No entry was undeterminable.** The other ten issues each record a value decline:

| Issue | Entries | What the cause records |
|---|---|---|
| #96, #97, #105 | 9 | A FoxIO reference defect this project declines to reproduce. |
| #138, #151 | 39 | A stream the FoxIO Python file omits, which the Rust snapshot holds. |
| #162 | 16 | An ALPN byte the two FoxIO implementations disagree on. |
| #214, #215, #225 | 22 | A form this project chose, against a reference that writes another. |
| #272 | 6 | An open count difference, which #272 decides. |

## Two premises this issue held, and the measurement of each

**The reach is 6 before and after, and the six rows are unchanged.** Both numbers are
measured, not carried over.

**The issue states that a `#129` entry admitted to the exception would raise the reach.
That premise is false.** `SOURCE_VALUES` holds no `#129` key at all, because #334
measured the 35 source values of #129 and left them out of the table under the bar. With
every one of the 43 `#129` entries flipped to `"capability": false`, the reach stays 6:
fact 4 finds no source value and bars each row before fact 3 matters. The capability bar
on `#129` is therefore enforced twice today, once by the field and once by an absent
measurement.

**The bar is proven live in the direction that discriminates.** Setting
`"capability": true` on the `#96` row `ssh-r.pcap/2:46396/JA4SSH.1` in the committed
register drops the reach from 6 to 5 and fails 7 cases, including
`test_the_exception_reaches_six_register_keys`. The revert reproduces the register byte
for byte, `sha256` prefix `6273fc7b` before and after.
`test_the_bar_reads_the_field_and_not_the_issue_number` holds the same mutation as a
permanent case over a temporary register.

## How the mutation was proven live

Every `__pycache__` outside `.venv/` was removed, `PYTHONDONTWRITEBYTECODE=1` was set,
and `inspect.getsource` confirmed the module under execution. The marker discriminates in
both directions:

| Marker | Base file | Head file |
|---|---|---|
| `if deviation.capability:` | 0 | 1, and unique in the module |
| `CAPABILITY_DECLINES` | 5 | 0 |

## What did not move

No fingerprint moves. No file under `ja4plus/` changes. No vector is adopted.
`ja4plus.__all__` names the same 25 entries. `tests/foxio_deviations.json` holds 135 keys
before and after, against 135 `xfailed` cases.

## How this was tested

```
.venv/bin/pytest tests/ -m "not spec_validation"   1876 passed, 8 xfailed, 93 subtests passed
.venv/bin/mypy --strict ja4plus/                   no issues found in 30 source files
```

Line coverage is 88 percent, unchanged, and no file under `ja4plus/` changed.

**The cases were written first and they bite.** Against commit `65b62e0` they reported
18 failed, 148 passed. The implementation commit `4ed4b26` turns them green at 166
passed.

## Documentation

- `tests/foxio_deviations.py` gains a `## The kind of decline` section that states the
  schema, the default and the gate.
- `.claude/rules/external-apis.md` replaces the proposal of #334 with the built rule, and
  fact 3 now names the field.
- `docs/specs/spec.md` records a Changelog round with the number `TBD`, and the two
  `## Terms` rows name the field.
- `CHANGELOG.md` records the change under `### Changed` with the round `TBD`.

Verified against: https://github.com/FoxIO-LLC/ja4/tree/main/python/test/testdata (pinned commit `27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8`, retrieved 2026-08-08)


Crank-Git:
## Self-review

I ran each lens myself and did not spawn review children, because review children have
stalled repeatedly in this session and a review that does not return is worse than no
review. Lenses: correctness of the field and the gate, the writing standard, the scope
bars this issue sets, and the register invariants.

**Two findings against my own diff, both fixed in `a7e1930`.**

**Finding 1, and the more serious one.**
`test_every_capability_decline_states_the_capability_it_records` demanded that every
entry carrying `"capability": true` state the boundary in the words of #129, `reads no
encrypted request` or `reads no encrypted certificate`. **That reintroduces the exact
coupling this issue removes.** A capability decline of another kind, such as the JA4TScan
scope decline #197 owns, would have failed a case that names it nowhere near the decline.
The case is removed. `TestTheEncryptedHttpDeviations` already gates the wording of #129
against 36 keys, so nothing is lost.

**Finding 2.** The gate function was `unkinded_declines`. `unkinded` is an invented past
participle, and `.claude/rules/ste.md` allows no such word. It is `unrecorded_kinds`,
which reads in real words and matches `unmarked_decisions` and `unusable_owners`.

**One finding accepted rather than fixed.** `CAPABILITY_ISSUE` and `CAPABILITY_ENTRIES`
in `tests/test_foxio_deviations.py` name #129 and 43. They look like the constant this
issue deletes, and they are not: they record the state of the register and carry no bar,
the way `test_the_register_holds_six_undecided_entries` already does. A new capability
decline fails them loudly, which is the outcome this issue wants, where
`CAPABILITY_DECLINES` failed silently. A warning comment now states that difference and
states the repair.

## One premise of the issue body is false

The issue states that admitting a #129 entry to the exception would raise the reach.
**It does not.** `SOURCE_VALUES` holds no #129 key at all, because #334 measured the 35
source values of #129 and left them out of the table under the bar. With every one of the
43 #129 entries flipped to `"capability": false` in the committed register, the reach
stays 6: fact 4 finds no source value and bars each row before fact 3 is ever read.

The bar is proven in the direction that does discriminate, on the committed register:
`"capability": true` on the #96 row `ssh-r.pcap/2:46396/JA4SSH.1` drops the reach from 6
to 5 and fails 7 cases.

## One finding outside this issue's scope, filed rather than fixed

The `## The marker rule` section of `tests/foxio_deviations.py` reads `38 of the 128
decided entries`. The register holds **129** decided entries. The 38 is correct and the
denominator is stale. #345 records it. #341 changed neither line.

Issue #343, opened 2026-08-08
Decide whether a random handshake body produces a fingerprint


Crank-Git:
Found by #338 while it widened the hostile-input set. **#338 does not repair it, because
the repair decides which byte strings are a ClientHello, and that is a fingerprint
decision.**

## The measurement

A payload that carries a valid TLS record header, a valid handshake header, and a random
body produces a well-formed JA4 value on every draw.

```python
load = b"\x16\x03\x01\x01\x00\x01\x00\x00\xfc" + os.urandom(252)
JA4Fingerprinter().process_packet(IP() / TCP(sport=12345, dport=443) / Raw(load=load))
```

Over 5000 draws:

| Result | Count |
|---|---|
| A well-formed JA4 value | 5000 |
| Nothing | 0 |
| A raise | 0 |

Sample values: `t00i330000_3b02dd174198_000000000000`,
`t00i000000_000000000000_000000000000`, `t00i490000_9d1cd381e981_000000000000`.

The version token is `00` on 4999 of the 5000 draws. 784 of the 5000 carry no cipher
suite and no extension, so `t00i000000_000000000000_000000000000` names a client that
offered no cipher suite. No real client sends that.

**Every length field in this payload is self-consistent**, so `FR-correctness-audit-8`
to `FR-correctness-audit-10` do not reject it. The record length and the handshake
length both agree with the bytes present. The body alone is random.

## Why #338 stopped here

Two readings of the hostile-input rule conflict, and the project holds a committed
assertion for the second one.

1. Input that is not a ClientHello produces nothing. Under this reading the values above
   are fabricated fingerprints and the parser must reject them.
2. The parser reads the fields it can and emits a value from them. Only a length that
   reads past the packet stops it.

`tests/fuzz/test_length_fields.py::test_an_extension_that_declares_more_bytes_than_the_packet_produces_no_exception`
asserts reading 2. It inflates one extension length and asserts `produced is not None`.
Reading 1 would invert that assertion.

**FoxIO decides behaviour, and no vector and no FoxIO material settles this case.**
`CLAUDE.md` rule 1 applies, so this needs a reference measurement or a user decision.

## What a repair would cost, measured

A narrow guard is available and it moves no committed vector. Over 168 ClientHellos in
`tests/foxio_vectors/`:

| Property | Count of the 168 |
|---|---|
| Version token `00` | 0 |
| No cipher suite | 0 |
| No extension | 0 |

[...]

Issue #342, opened 2026-08-08
Batch: the findings of the deliberate sweep


Crank-Git:
A batch formed by the project manager on 2026-08-08 from the **deliberate sweep the user asked
for**, after four instances of one shape surfaced by accident while Epics 4 to 6 were built.

**The shape: a comparison that is never made reads as a comparison that passes.** This project
has recorded sixteen instances. The sweep looked for the rest on purpose.

## Members

- [x] #338 — assert that unparseable input produces no fingerprint. **`priority:high`**
- [x] #348 — make the four QUIC rejection cases able to fail. Found by #337.
- [x] #343 — record that a structurally valid ClientHello produces a fingerprint. Decided by the user; measured by #338.
- [x] #337 — assert the QUIC v2 Initial contract the test only names
- [x] #339 — give the `cleanup_connection` cases a reader

## Why #338 leads

**It is the only one whose failure would produce a wrong answer rather than hide a gap.**
`tests/test_edge_cases.py` accepts *"Either None or a result"* for 256 random bytes, so a
regression that fabricated a JA4 value from unparseable input would pass. **This package exists
so that one tool's output can be compared against another tool's, and a fabricated fingerprint is
evidence of a client that does not exist.**

#337 and #339 hide gaps rather than produce wrong answers, so they follow.

## Every member carries the heavy proof

The user tiered verification on 2026-08-08: the heavy proof binds work that can move a
fingerprint, change a published field, or touch a FoxIO reading. **All three of these are
conformance-affecting.** #339 in particular is what Epic 6's eviction and the user's 512 MiB
ceiling rest on.

## What the sweep already knew, which is the finding above the findings

**`docs/mutation_sweep.md` had already flagged both high findings** — at lines 1250, 606 and 609.
#206 holds the 976 candidates that report produced, and 248 of them are explained by the report's
own limits. **These were not, and nobody read them.** The tool found the defects and the report
went unread, which is the same shape one level up.

Issue #339, opened 2026-08-08
Give the cleanup_connection cases a reader


Crank-Git:
Found by the deliberate sweep the user asked for on 2026-08-08.

## The measurement

**Twelve cases across five modules call `cleanup_connection` or a stateful method and assert
nothing about the state afterwards.** Each asserts only that the call did not raise.

`tests/test_cleanup_connection.py:17,40,99,124,144,152`, `tests/test_ja4d.py:276`,
`tests/test_edge_cases.py:399,410`, `tests/test_ja4x_reassembly.py:10,21`,
`tests/test_ipv6_support.py:62,76`. **Line numbers move; find the cases.**

```python
fp.cleanup_connection("1.2.3.4", 100, "5.6.7.8", 443, "tcp")
# No-op on empty state
```

**A `cleanup_connection` that removes the wrong key, leaks the entry, or mutates a sibling
connection passes every one of them.**

## Why this matters more now than when it was written

**Epic 6 built the watch command's eviction on `cleanup_connection`.** #53 evicts a connection
from a bounded table and calls it to drop the per-method state; the user's stated 512 MiB ceiling
(#279) rests on that call doing what its name says.

**#53 proved the call happens.** It did not prove the call is correct, and these twelve cases are
where that would be proven.

`docs/mutation_sweep.md:523-532` already lists the `test_cleanup_connection.py` cases as
candidates needing a reader. **The other cases were never sampled by that tool**, so they are
unconfirmed by it and read the same way in source.

## What to build

**Give each case a reader.** After the call, assert what the state holds: the key is gone, the
sibling connection is untouched, the entry count moved by exactly one. **Read
`ja4plus/utils/state_table.py` and the fingerprinters' own tables for what is observable.**

**Cover the three failure modes the current cases cannot distinguish**: removing the wrong key,
leaving the entry, and disturbing a neighbouring connection. **A case that only checks the
target key is gone does not catch the third.**

**If a fingerprinter's state is not observable from outside, say so and propose how** rather than
asserting something weaker.

## Prove the cases can fail

**Three mutations, each proven live and discriminating in both directions**: make
`cleanup_connection` a no-op, make it remove a neighbouring key, and make it remove nothing while
reporting success. **Report which cases each one kills.** A mutation that kills no case names a
gap that remains.

**This is conformance-affecting work — the memory ceiling rests on it — so the heavy proof
applies.**

## What must not move

[...]

Issue #338, opened 2026-08-08
Assert that unparseable input produces no fingerprint


Crank-Git:
Found by the deliberate sweep the user asked for on 2026-08-08. **This is the finding with the
worst consequence of the five.**

## The measurement

`tests/test_edge_cases.py::test_random_bytes` and `::test_very_large_packet` feed unparseable
input to `JA4Fingerprinter` and **assert nothing about the result**:

```python
data = os.urandom(256)
packet = IP() / TCP(sport=12345, dport=443) / Raw(load=data)
# Should not raise, just return None
self.ja4.process_packet(packet)
# Either None or a result, but should not crash
```

**The comment states the defect out loud.** *"Either None or a result"* accepts a fingerprint
produced from 256 random bytes as a passing outcome.

## Why this is worse than a weak test

**`CLAUDE.md` rule 2 states it: a parser that cannot read a packet returns nothing. It does not
raise.** These cases hold the second half and abandon the first.

**This package exists so that one tool's output can be compared against another tool's output.**
A JA4 value produced from random bytes is a **fabricated fingerprint** — a downstream tool, a
threat-intelligence pipeline or an operator reading `ja4plus watch` would treat it as evidence
of a real client. **A regression that fabricated values would pass this suite silently**, and
`docs/mutation_sweep.md:606,609` already lists both cases as candidates.

## What to build

**Assert the contract: unparseable input produces nothing.**

**Measure the current behaviour first and report it before changing any test.** If `ja4plus`
already returns nothing for both inputs, this issue tightens a test and moves no behaviour. **If
it produces a value for either, that is a live defect and the repair is in `ja4plus/`, not in
`tests/`** — say so plainly and report what it produced.

**Then widen the input set.** Two inputs do not measure a contract about hostile input. Add at
least: empty payload, a truncated TLS record, a TLS record claiming a length beyond the packet,
long runs of `0x00` and of `0xff`, and a valid handshake with one field corrupted. **#316 fuzzed
its reader over 13680 calls and reported `raised: {}`** — follow that example in spirit; the
volume is yours to choose, but two is not enough.

**Search for the same shape elsewhere.** `grep -rn "should not crash\|Either None or" tests/`
and report every case whose only assertion is the absence of an exception. **Repair the ones in
this file; report the rest** so the project manager can scope them.

## Prove the cases can fail

**Make the fingerprinter return a value for unparseable input, confirm the cases go red,
revert** — with the mutation proven live, discriminating in both directions.

[...]

PR #346, opened 2026-08-09
Assert that unparseable input produces no fingerprint

Crank-Git:
Closes nothing on its own. #338 is a member of the sweep batch #342.

## What the measurement found

**The premise was measured before any test changed, and the answer is that no behaviour
moves.** Both inputs the issue names produce nothing today.

| Input | 20000 draws | Result |
|---|---|---|
| `os.urandom(256)` | 20000 | Nothing, on every draw. No raise. |
| `b"\x16\x03\x03" + b"\x00" * 5000` | 1 | Nothing. No raise. |

The six shapes the issue names also produce nothing: an empty payload, a truncated TLS
record, a record that declares more bytes than the packet holds, a long run of `0x00`, a
long run of `0xff`, and a ClientHello with one corrupted byte.

**No file under `ja4plus/` changes.** `git diff` reports the package byte-for-byte equal
to the base commit.

## What changes

`tests/test_edge_cases.py` — `test_random_bytes` and `test_very_large_packet` now read
the result. `test_random_bytes` reads 64 draws, because the payload is random and one
draw measures one byte string.

`tests/fuzz/test_synthetic_hostile_input.py` — 17 cases over the six named shapes, plus
random bytes behind a record header, plus a byte sweep of a real ClientHello. Every case
counts the calls of `parse_tls_handshake`, because `tests/fuzz/README.md` states that a
case which reaches no parser proves nothing.

The corrupted-byte case does not demand silence. A corrupted cipher value stays readable
and moves the fingerprint, so the case demands that the reader produces nothing or a
value the JA4 form accepts.

## The cases are proven live, in both directions

`process_packet` was changed to return `t13d1516h2_000000000000_000000000000` where it
returns nothing.

| Run | Result |
|---|---|
| The repaired cases, mutation live | 22 failed, 1 passed |
| The two cases as written on the base commit, same mutation live | 2 passed |

**The second row is the point.** The old cases were blind to a fingerprinter that
fabricates. The repaired cases catch it.

Every `__pycache__` outside `.venv/` was removed and `PYTHONDONTWRITEBYTECODE=1` was
set. `inspect.getsource` on the loaded `JA4Fingerprinter.process_packet` reported
`MUTATION_LIVE`, then `MUTATION_ABSENT` after the revert. The marker
`FABRICATED_MUTATION_338` matched nothing in the repository before the mutation, so it
discriminates.

## One finding is filed and not repaired: #343

A payload of a valid record header, a valid handshake header and a random body produces
a well-formed JA4 value on **5000 draws of 5000**. Samples:
`t00i000000_000000000000_000000000000`, `t00i330000_3b02dd174198_000000000000`.

Every length field in that payload is self-consistent, so `FR-correctness-audit-8` to
`FR-correctness-audit-10` do not reject it. Repairing it decides which byte strings are a
ClientHello, and
`tests/fuzz/test_length_fields.py::test_an_extension_that_declares_more_bytes_than_the_packet_produces_no_exception`
already asserts the opposite reading. **That is a fingerprint decision and no vector
settles it**, so #343 owns it.

The cost of a repair is measured in #343: over 168 ClientHellos in
`tests/foxio_vectors/`, none carries the version token `00`, none carries no cipher
suite, and none carries no extension.

## Cases in `tests/` whose only assertion is the absence of an exception

17 before this change, 14 after. The three repaired are `test_ja4t_no_ip`,
`test_synack_before_syn` and `test_duplicate_syn`, each now reading its measured value.

Of the 14 that stay, 8 belong to #339 (`cleanup_connection`). The verdict lists the
other 6.

## How it is tested

```
pytest tests/ -m "not spec_validation"   2019 passed, 8 xfailed   (base: 2002 passed, 8 xfailed)
pytest tests/ -m spec_validation         1531 passed, 143 skipped, 135 xfailed (equal to base)
ruff format --check ja4plus/ tests/      152 files already formatted
mypy --strict ja4plus/                   no issues in 31 source files
```

The register invariant holds: `tests/foxio_deviations.json` holds 135 keys against 135
`xfailed`. `ja4plus.__all__` names 25 entries. Coverage is 92% before and after.

## TBD markers

`docs/specs/spec.md` holds one `TBD`, in Changelog round 106. `CHANGELOG.md` is
unchanged.


Crank-Git:
## Self-review

I ran the correctness lens and the test-quality lens over the diff myself. Two findings
were real and `4663ffe` repairs both. One finding was a false positive of my own scan and
I decline it.

### Finding 1 — a case that fails on one run in many (repaired)

`test_random_bytes_produce_no_fingerprint`,
`test_random_bytes_behind_a_record_header_produce_no_fingerprint` and
`tests/test_edge_cases.py::TestMalformedTLS::test_random_bytes` drew their payloads from
`os.urandom`.

The parser walks every record of the segment, so a random payload can carry `0x16` at any
offset and enter the handshake read. **25000 measured draws produced nothing**, so the
probability is low. It is not zero, and a suite that fails once in many runs reports a
defect nobody can reproduce.

Every random payload now comes from `random.Random(338)`. The count of draws is unchanged
and the coverage of byte strings is unchanged.

### Finding 2 — the fabrication detector accepted a trailing line feed (repaired)

`JA4_FORM` ended with `$`. In Python, `$` matches at the end of the string **and** before
a final line feed. A fabricated value that carried a trailing line feed would have passed
`test_one_corrupted_byte_produces_nothing_or_a_well_formed_fingerprint`.

The pattern now opens with `\A` and ends with `\Z`.

### Finding 3 — declined

My survey script reported
`tests/test_thread_safety.py:408::test_a_processor_that_promises_one_thread_acquires_no_lock`
as a case with no assertion. **That reading is wrong and the case is well built.** The
assertion sits inside `FailOnAcquire`, which the case installs in place of
`threading.RLock`, and `test_a_thread_safe_processor_acquires_the_lock` is an explicit
control that proves the case measures the lock. I report it here so the count is not read
as six defects.

### The mutation proof was run again

The seeded cases were re-measured against the same mutation, because the payload source
changed. `inspect.getsource` on the loaded `process_packet` reported `MUTATION_LIVE`, the
cases reported **22 failed, 1 passed**, and the revert reported `MUTATION_ABSENT`.

### Gates after the repair

```
pytest tests/ -m "not spec_validation"   2019 passed, 8 xfailed
ruff format --check ja4plus/ tests/      152 files already formatted
mypy --strict ja4plus/                   no issues in 31 source files
```

`git diff HEAD -- ja4plus/` is empty, so the package stays byte-for-byte equal to the
base commit.

Issue #348, opened 2026-08-09
Make the QUIC rejection cases able to fail


Crank-Git:
Found while #337 repaired `test_v2_initial_type_not_rejected`. The scope of #337 is the QUIC
version 2 Initial contract, so this issue holds the rest of the class.

## The measurement

`parse_quic_initial` and `parse_quic_server_initial` reject a packet at four guards: the
length guard, the short-header guard, the version negotiation guard and the packet type
guard. Every rejection case builds a packet of dummy bytes, so the packet fails decryption
whichever guard runs. **The case therefore passes with its guard removed, and it measures
nothing.**

Three mutations prove it. Each ran against `origin/batch/342-sweep-findings`, with every
`__pycache__` outside `.venv/` removed and `PYTHONDONTWRITEBYTECODE=1` set.

| Mutation in `ja4plus/utils/quic_utils.py` | Case | Result |
|---|---|---|
| `if not (first_byte & 0x80)` reads `0x81`, so a short header reaches the parser | `test_short_header` | 2 passed |
| `if packet_type != 0x00` reads `if packet_type != packet_type`, so every version 1 type reaches the parser | `test_non_initial_type` | 1 passed |
| `if version == 0` reads `if version == 999999`, so a version negotiation packet reaches the parser | `test_version_negotiation` | 1 passed |

`TestParseQuicServerInitial` holds the same four shapes at
`tests/test_quic_utils.py:172` to `:201`, and this issue owns them too.

## What to build

Build each rejection case from a packet that decrypts. `tests/quic_builder.py` gained
`client_initial_crypto` and `client_hello` on #337, and `server_initial` was already there.
A case that carries a protected ClientHello returns that ClientHello when the guard goes,
so the case measures the guard alone.

`tests/test_quic_utils.py::TestParseQuicInitial::test_v2_non_initial_rejected` shows the
form after #337 repairs it.

## What must not move

No fingerprint moves. This issue changes no file under `ja4plus/`.

## Acceptance criteria

- [ ] Every rejection case of `TestParseQuicInitial` fails when its guard goes, proven with
      a live mutation.
- [ ] Every rejection case of `TestParseQuicServerInitial` fails when its guard goes, proven
      with a live mutation.
- [ ] The conformance suite reports 1531 passed, 143 skipped and 135 xfailed.

Crank-Git:
Triaged 2026-08-08 by the project manager as `status:ready`, and added to the sweep batch #342.

[...]

Issue #337, opened 2026-08-08
Assert the QUIC v2 Initial contract the test only names


Crank-Git:
Found by the deliberate sweep the user asked for on 2026-08-08, after four instances of this
shape surfaced by accident during Epics 4 to 6.

## The measurement

`tests/test_quic_utils.py::test_v2_initial_type_not_rejected` **calls `parse_quic_initial` and
discards the return value.** The case holds no assertion of any kind.

```python
parse_quic_initial(bytes(pkt))
```

Its own docstring states the contract it is meant to hold: *"QUIC v2 Initial uses packet type
0x01, not 0x00. Must not be rejected."*

**The regression this case exists to catch is the one it cannot catch.** A change that rejected
every QUIC v2 Initial packet at the packet-type check would leave this case green.

**The project already knew.** `docs/mutation_sweep.md:1250` names this case as an unaddressed
candidate. #206 holds the 976 candidates the sweep produced, and 248 of those are explained by
the report's own limits — **this one is not, and nobody read it.** That is the same shape as the
finding itself.

## What to build

**Capture the return value and assert what the contract states.** The case must distinguish
*rejected at the type check* from *reached the decryption stage and failed there*, because those
are different outcomes and only the first is the defect.

**Read `ja4plus/utils/quic_utils.py` to find how that distinction is observable** — the function
returns nothing in both cases today, so the case may need to observe the branch rather than the
return. **If the distinction is not observable from outside, say so and propose how to make it
observable rather than asserting something weaker.**

**The comment in the case claims a method it does not use:** *"We verify by checking a v2
non-Initial IS rejected."* **Check whether that verification exists anywhere.** If it does, cite
it; if it does not, the comment is a fourth instance of prose asserting a check that is not there.

## Prove the case can fail

**Reject every v2 Initial at the type check, confirm the case goes red, revert.** With the
mutation proven live — clear every `__pycache__` outside `.venv/`, set
`PYTHONDONTWRITEBYTECODE=1`, and confirm with `inspect.getsource` on the loaded function,
discriminating in both directions.

**This is conformance-affecting work, so the heavy proof applies** under the tiering the user set
on 2026-08-08.

## What must not move

No fingerprint moves. The register invariant holds: the key count of
`tests/foxio_deviations.json` equals the `xfailed` count. `__all__` names exactly 25 entries.

## Acceptance criteria

- [ ] The case asserts the outcome its docstring names, and holds at least one assertion.
[...]

PR #349, opened 2026-08-09
Assert the QUIC version 2 Initial contract (#337)

Crank-Git:
Builds #337. Sweep batch #342.

## The defect

`tests/test_quic_utils.py::test_v2_initial_type_not_rejected` called `parse_quic_initial`
and discarded the return. The case held no assertion, so a change that rejected every QUIC
version 2 Initial packet at the type check left it green.

## The hard part, and the answer

**The distinction is observable, and the return alone does not carry it.**
`parse_quic_initial` returns nothing when it rejects a packet at the type check, and it
returns nothing when it reaches the decryption and fails there. A packet of dummy bytes
therefore returns nothing whichever branch runs.

Two observables tell the branches apart.

1. `parse_quic_initial` logs `QUIC parsing failed: <error>` at the DEBUG level when the
   decryption raises, and it logs nothing when the type check rejects the packet.
2. **A packet that decrypts returns the ClientHello it carries.**

**This pull request uses the second observable, because the first states what the reader
logged and not what the reader produced.** `tests/quic_builder.py` gains `client_hello` and
`client_initial_crypto`, and `_protect` now holds the payload protection and the header
protection that `server_initial` held. A version 2 Initial packet that carries a protected
ClientHello now produces that ClientHello, with the server name `v2.example.com`.

**The contract already holds, so no file under `ja4plus/` moves.** `git diff --stat HEAD --
ja4plus/` is empty on every commit of this branch.

## The comment that claimed a check

The case read *"We verify by checking a v2 non-Initial IS rejected."*
**That case exists.** It is `tests/test_quic_utils.py:132`
`TestParseQuicInitial::test_v2_non_initial_rejected` on the base commit. **So the comment
is no fifth instance of prose that names an absent check.**

**The case it named could not fail either.** Its packet is 30 zero bytes, so the reader
returns nothing whether it rejects the packet at the type check or accepts it and fails the
decryption. The repaired case carries the same protected ClientHello as the Initial case
under the type code 3, which RFC 9369 Section 3.2 gives a Handshake packet. A reader that
drops the type check returns that ClientHello, so the case now measures the type check and
nothing else.

## Proof that the cases can fail

Every run cleared each `__pycache__` outside `.venv/` and set `PYTHONDONTWRITEBYTECODE=1`.
`inspect.getsource` on the loaded `parse_quic_initial` reported the marker
`MUTATION-337-LIVE`, and `grep -rn "MUTATION-337-LIVE" .` matched nothing in the repository
beforehand.

| Mutation in `parse_quic_initial` | Cases of this branch | Cases of the base commit |
|---|---|---|
| `if packet_type != 0x01` reads `0x99`, so the reader rejects every version 2 Initial packet | `test_v2_initial_yields_the_client_hello` **fails**, the other two pass | both pass |
| `if packet_type != 0x01` reads `if packet_type != packet_type`, so the reader accepts every version 2 type | `test_v2_non_initial_rejected` **fails**, the other two pass | both pass |

The revert is proven too: `git diff HEAD -- ja4plus/` is empty and the marker matches
nothing.

## What did not move

| Measurement | Base | This branch |
|---|---|---|
| `pytest tests/ -m spec_validation` | 1531 passed, 143 skipped, 135 xfailed | 1531 passed, 143 skipped, 135 xfailed |
| `tests/foxio_deviations.json` keys | 135 | 135 |
| `len(ja4plus.__all__)` | 25 | 25 |
| `pytest tests/ -m "not spec_validation"` | 2019 passed, 8 xfailed | 2020 passed, 8 xfailed |
| Coverage of `ja4plus` | 92%, 336 lines missed | 92%, 336 lines missed |
| Coverage of `ja4plus/utils/quic_utils.py` | 89%, 36 lines missed | 89%, 36 lines missed |

The unit suite rises by one case: this branch removes one case and adds two.

## Gates

```
.venv/bin/pytest tests/ -m "not spec_validation"   2020 passed, 8 xfailed, 96 subtests passed
```

## Out of scope, and filed

**#348.** The length guard, the short-header guard and the version negotiation guard of
`parse_quic_initial` and `parse_quic_server_initial` hold the same shape as the case this
pull request repairs. Three mutations prove it: with the short-header guard changed
`test_short_header` passes, with every version 1 type accepted `test_non_initial_type`
passes, and with the version negotiation guard changed `test_version_negotiation` passes.

## Documentation

`docs/specs/spec.md` gains Changelog round 107, dated 2026-08-08, opening with `TBD`.
Round 106 belongs to #338 and this branch does not renumber it. **`CHANGELOG.md` holds no
entry**, because the change moves no user-facing behaviour, and #338 wrote none for the
same reason.

Verified against: https://www.rfc-editor.org/rfc/rfc9369.html (RFC 9369, Section 3.1
"Version Field" and Section 3.2 "Long Header Packet Types", retrieved 2026-08-08)


Crank-Git:
## Self-review

I ran the three lenses myself over the whole diff, rather than through a child agent.

### Correctness

1. **`_protect` moves 15 lines out of `server_initial`, and seven files call
   `server_initial`.** `tests/build_quic_ja4l_captures.py`,
   `tests/test_ja4l_quic_repeated_connection.py`, `tests/test_ja4l_quic_server_hello.py`,
   `tests/test_ja4l_udp_direction.py`, `tests/test_ja4s_quic_direction.py`,
   `tests/test_ja4s_quic_split_server_hello.py` and `tests/test_quic_server_initial.py`
   read it. The moved lines are byte-identical and the `trailer` stays at the call site,
   so the packet a caller gets does not change. The unit suite passes 2020 cases, which
   covers all of them.

2. **`client_initial_crypto` raises `KeyError` for a version the table does not hold.**
   `INITIAL_TYPE_CODES` holds version 1 and version 2, which is every version this
   project reads. I add no handler, because the project writes no error handling for a
   case that cannot occur.

3. **`client_initial_crypto` derives the version 2 secrets for a version the table does
   not hold**, through `1 if version == QUIC_VERSION_1 else 2`. `server_initial` already
   held that form, so the two builders agree. #348 covers neither, and I open no issue for
   a builder that only this repository calls.

4. **`0xC0 | (type_code << 4)` writes the type into bits 4 and 5 and leaves the packet
   number length at one byte**, which is what `_protect` encodes. RFC 9000 Section 17.2
   gives the layout.

### Security

The diff changes no file under `ja4plus/`, reads no network, and writes no file. The
AES-ECB call in `_protect` is the header protection mask of RFC 9001 Section 5.4.2, it is
pre-existing code that this pull request moves, and it encrypts one 16-byte sample.

### Duplication

`tests/test_comprehensive.py::build_client_hello` builds a TLS **record** that carries a
ClientHello. `client_hello` here returns the handshake **message**, because
`parse_quic_initial` wraps the message in a record of its own. I keep the two apart rather
than import from a module that pytest collects as a test module.

PR #350, opened 2026-08-09
Give each QUIC Initial header guard a case that can fail

Crank-Git:
Part of #342. Refs #348. Found by #337.

## What this changes

Every rejection case of `parse_quic_initial` and `parse_quic_server_initial` built a
packet of dummy bytes, so the reader returned nothing whichever guard ran. Each guard
case now carries a protected handshake message the reader decrypts, so the guard the
case names is the only rejection.

No file under `ja4plus/` changes.

## The guard count is eleven, not four

The issue names four guards. The readers hold eleven rejection conditions.

| Reader | Line | Condition |
|---|---|---|
| `decrypt_quic_server_initial_crypto` | `quic_utils.py:440` | `if len(udp_payload) < 5 or not client_dcid:` — two conditions |
| `decrypt_quic_server_initial_crypto` | `quic_utils.py:444` | `if not (first_byte & 0x80):` |
| `decrypt_quic_server_initial_crypto` | `quic_utils.py:448` | `if version == 0:` |
| `decrypt_quic_server_initial_crypto` | `quic_utils.py:454` | `if packet_type != 0x01:` — version 2 arm |
| `decrypt_quic_server_initial_crypto` | `quic_utils.py:458` | `if packet_type != 0x00:` — version 1 arm |
| `parse_quic_initial` | `quic_utils.py:570` | `if len(udp_payload) < 20:` |
| `parse_quic_initial` | `quic_utils.py:574` | `if not (first_byte & 0x80):` |
| `parse_quic_initial` | `quic_utils.py:578` | `if version == 0:` |
| `parse_quic_initial` | `quic_utils.py:586` | `if packet_type != 0x01:` — version 2 arm |
| `parse_quic_initial` | `quic_utils.py:589` | `if packet_type != 0x00:` — version 1 arm |

`parse_quic_server_initial` holds no header guard of its own. It calls
`decrypt_quic_server_initial_crypto`, and that function holds all six.

## How the cases work

The AEAD tag covers the header, so a case that changes a byte of a built packet destroys
the tag. The reader then returns nothing for that reason instead of for the guard the
case names. Each case therefore builds the rejected header. `tests/quic_builder.py`
gains three parameters:

- `long_header` clears bit 7, so the reader reads a long-header packet as a short header.
- `key_version` protects a packet under the version 1 salt while the header carries
  version 0.
- `type_code` on `server_initial`, which `client_initial_crypto` already held.

## Proof by live mutation

Each of the eleven conditions came out of the reader in turn. Marker `MUTATION-348-LIVE`,
verified with `inspect.getsource` on the loaded module in both directions, with every
`__pycache__` outside `.venv/` removed and `PYTHONDONTWRITEBYTECODE=1` set. `grep` found
the marker nowhere in the repository beforehand and nowhere afterward.

**On the base commit `9a48020`, ten of the eleven conditions killed no case.** The whole
QUIC suite of 53 cases stayed green under every one. Only the version 2 type guard of
`parse_quic_initial` was measured, and #337 built that case.

**On this branch every condition kills the case that names it.**

| Guard removed | Case that fails |
|---|---|
| server length, `:440` | `TestParseQuicServerInitial::test_a_truncated_long_header_returns_none_rather_than_raising`, `test_the_server_reader_returns_none_for_a_truncated_long_header` |
| server connection ID, `:440` | `TestParseQuicServerInitial::test_empty_client_dcid_returns_none`, `test_the_server_reader_rejects_an_empty_connection_id` |
| server short header, `:444` | `TestParseQuicServerInitial::test_short_header_returns_none`, `test_the_server_reader_rejects_a_short_header` |
| server version, `:448` | `TestParseQuicServerInitial::test_version_negotiation_returns_none`, `test_the_server_reader_rejects_a_version_negotiation_packet` |
| server type, version 2, `:454` | `TestParseQuicServerInitial::test_a_version_2_wrong_packet_type_returns_none`, `test_the_server_reader_rejects_a_version_2_handshake_packet` |
| server type, version 1, `:458` | `TestParseQuicServerInitial::test_wrong_packet_type_returns_none`, `test_the_server_reader_rejects_a_handshake_packet` |
| client length, `:570` | `TestParseQuicInitial::test_a_truncated_long_header_returns_none_rather_than_raising` |
| client short header, `:574` | `TestParseQuicInitial::test_short_header` |
| client version, `:578` | `TestParseQuicInitial::test_version_negotiation` |
| client type, version 2, `:586` | `TestParseQuicInitial::test_v2_non_initial_rejected` |
| client type, version 1, `:589` | `TestParseQuicInitial::test_non_initial_type` |

## One acceptance criterion stays open

Criterion 1 reads:

> Each of the four guards has a case whose packet would produce a result if that guard
> were removed.

**The two length guards admit no such case, and arithmetic decides it.** A client Initial
packet that carries a ClientHello holds at least 33 bytes: 10 header bytes, a 16-byte
AEAD tag, a 3-byte CRYPTO frame header and a 4-byte handshake header. The guard rejects
at 20 bytes, so no packet under the threshold can produce a result. The server guard
rejects at 5 bytes, and the same floor applies.

The length guards do hold a measurable behaviour. Both readers unpack the version field
outside their `try` block, so the length guard is the only thing that stops a
`struct.error` reaching the caller. `parse_quic_initial(b"\xc0\x00\x00\x00")` returns
None today, and it raises `struct.error` with the guard removed. That is the case this
branch ships, and the mutation table above proves it red.

`CLAUDE.md` states the contract it measures: "A parser that cannot read a packet returns
nothing. It does not raise."

**The question for the project manager:** does the exception case satisfy criterion 1 for
the length guards, or does the criterion need a different form? The other nine conditions
meet the criterion as written.

The old `test_too_short` cases stay, and this branch keeps them. They pass with the
length guard removed, because the short-header guard rejects `b"\x00" * 10` as well.

## Gates

```
.venv/bin/pytest tests/ -m "not spec_validation"   2028 passed, 8 xfailed, 96 subtests passed
```

The unit suite rises from 2020 passed to 2028 passed. The conformance suite is unchanged
against the base, and `tests/foxio_deviations.json` holds 135 keys against 135 xfailed
cases. `__all__` names 25 entries. No fingerprint moves.

Coverage rises. `ja4plus/utils/quic_utils.py` holds 89% before and 93% after. The project
holds 92% before and after, with 336 missed statements before and 324 after.

## TDD

TDD is inverted here, as the brief expects. The guards already work, so no case could
fail before it was written. The cases are proven by live mutation instead, as #337 and
#338 both did.

## Documentation

`docs/specs/spec.md` gains Changelog round 108. `CHANGELOG.md` gains nothing, because it
records user-visible behaviour and no behaviour changes. Rounds 106 and 107 kept the same
boundary.

## Verified against

The new version 2 cases rest on the version 2 type codes, so this branch confirmed them
rather than recalling them.

```
Verified against: https://www.rfc-editor.org/rfc/rfc9369.html (RFC 9369, retrieved 2026-08-08)
```

RFC 9369 Section 3.1 gives the version number `0x6b3343cf`, which matches
`QUIC_VERSION_2`. Section 3.2 gives the long header type codes: Initial `0b01`, 0-RTT
`0b10`, Handshake `0b11`, Retry `0b00`. `INITIAL_TYPE_CODES` maps version 2 to 1, and
each new version 2 Handshake case writes the type code 3. Both match.

RFC 9000 Section 17.2 and RFC 9001 Sections 5.2, 5.3 and 5.4 already carry the layout,
the key derivation and the two protection steps that `tests/quic_builder.py` names. This
branch changed neither, and the reader round-trips against the builder for every case in
the table above.


Crank-Git:
## Self-review

I read the diff myself under three lenses rather than delegating, because a review that
does not return is worse than no review. Four checks and two observations follow.

### Correctness

**`server_initial` now writes the version 2 Initial type code, where it wrote 0 before.**
That is a behaviour change for any caller that passes `version=QUIC_VERSION_2`. No caller
did. `rtk git grep "server_initial("` across `tests/` returns 19 call sites and every one
of them takes the version 1 default, except the case this branch adds. The change is
therefore safe, and `test_the_server_reader_reads_a_version_2_initial_packet` is the case
that proves the builder now produces a version 2 Initial packet the reader accepts.

**`key_version` preserves the previous derivation.** The old line read
`1 if version == 1 else 2` and the new default reads `1 if version == QUIC_VERSION_1
else 2`. `QUIC_VERSION_1` is 1, so no existing packet changes.

**`server_initial(version=0)` raises `KeyError` unless the caller states `type_code`.**
`INITIAL_TYPE_CODES` holds no entry for version 0. `client_initial_crypto` already
behaved this way, so the two builders agree, and both version 0 cases state the type
code. I left the behaviour rather than adding a fallback, because a silent default would
hide the very thing the version 0 case measures.

**The new `first_byte` helper collides with nothing.** The only prior use of that name was
a local variable in `test_wrong_packet_type_returns_none`, and this branch removes it.

### The short-header case does not claim to be a real short header

`long_header=False` clears bit 7 of a long-header packet. It does not build a QUIC short
header, and the case does not say it does. That is the right shape: the guard reads bit 7
and nothing else, so the case gives it a packet that differs from an accepted packet in
bit 7 and in nothing else.

One consequence worth stating. RFC 9001 Section 5.4.1 masks the low four bits for a long
header and the low five for a short header. `remove_header_protection` always masks four,
and `_protect` always masks four, so the case round-trips. A reader that ever handles a
true short header would need the five-bit mask, and that is not this issue.

### Two observations, both already owned

**`decrypt_quic_server_initial_crypto:471` and `parse_quic_initial:622` each catch bare
`Exception`.** `CLAUDE.md` states that a fingerprinter catches the parse errors it
expects. #319 owns the remaining fourteen wide handlers, so this branch files nothing new
and changes nothing.

**Both readers unpack the version field outside their `try` block.** The length guard is
the only thing that keeps a `struct.error` from the caller, which is why the two new
truncated-header cases exist. The guard is correct today. The cases now hold the contract
that `CLAUDE.md` states: "A parser that cannot read a packet returns nothing. It does not
raise." Before this branch, no case held it for either reader.

### No defect found under `ja4plus/`

I read each of the eleven conditions against the reads that follow it. At the server
threshold of 5 bytes, `_find_pn_offset` reads byte 5 and raises, and the `try` at
`quic_utils.py:467` catches it. At the client threshold of 20 bytes, the read of byte 5
at `quic_utils.py:592` is in range. Every guard rejects before the read it protects, so
no guard is wrong and this branch stops at the boundary the issue draws.

Issue #59, opened 2026-08-06
Add bulk lookup and record the source of every result


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/07-db-enrichment.md`](docs/specs/features/07-db-enrichment.md):

> FR-db-enrichment-7 — `JA4DBClient.lookup_many` accepts a sequence of fingerprints and returns a result per fingerprint.

> FR-db-enrichment-8 — A lookup result records its source.

## What to build

```python
@dataclass(frozen=True)
class LookupResult:
    application: str
    type: str
    notes: str
    source: str          # "bundled" | "cache" | "remote"

def lookup_many(self, fingerprints: Sequence[str]) -> dict[str, LookupResult | None]: ...
```

`LookupResult` carries the port's three fields plus `source`. An analyst needs to know where a name came from to judge how much to trust it.

Verified against https://github.com/Crank-Git/ja4plus-go/blob/master/lookup.go (retrieved 2026-08-06).

## Acceptance criteria

- [ ] `lookup_many` returns one entry per input fingerprint, including misses.
- [ ] `lookup_many` given 100000 fingerprints with remote lookup off performs no network request.
- [ ] Every `LookupResult` carries a `source` value of `bundled`, `cache` or `remote`.

## Links

- Feature: [`docs/specs/features/07-db-enrichment.md`](docs/specs/features/07-db-enrichment.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #18
Depends on #57

Issue #58, opened 2026-08-06
Add the command-line opt-in and the disclosure notice


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/07-db-enrichment.md`](docs/specs/features/07-db-enrichment.md):

> FR-db-enrichment-3 — The command-line option `--lookup` performs a local lookup only.

> FR-db-enrichment-4 — The command-line option `--lookup-remote` permits the remote lookup.

> FR-db-enrichment-5 — The environment variable `JA4PLUS_DB_LOOKUP` set to `1` permits the remote lookup.

> FR-db-enrichment-6 — The command reports on standard error, once, that a remote lookup is enabled.

## What to build

`--lookup` becomes local-only. `--lookup-remote` opts into the service. The environment variable `JA4PLUS_DB_LOOKUP` set to `1` does the same, for an operator who cannot change the command line.

The notice appears once per run, not once per lookup, and names the service it will contact.

`_init_lookup` at `ja4plus/cli.py:143` is where the client is built today.

## Acceptance criteria

- [ ] `ja4plus analyze <capture> --lookup` performs no network request.
- [ ] `ja4plus analyze <capture> --lookup-remote` writes the notice once to standard error.
- [ ] `JA4PLUS_DB_LOOKUP=1 ja4plus analyze <capture> --lookup` permits the remote lookup.
- [ ] `requests` missing while remote lookup is requested exits with status 1 and names the extra to install.

## Links

- Feature: [`docs/specs/features/07-db-enrichment.md`](docs/specs/features/07-db-enrichment.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #18
Depends on #57

Issue #18, opened 2026-08-06
Epic 7: Database enrichment


Crank-Git:
## Goal

Make fingerprint lookup useful offline and safe by default.

## Exit criteria

No network request happens unless the caller opts in. Bulk lookup exists. The `db update` and `db info` subcommands exist and match the port. The mapping file refresh is documented.

## Spec

- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Feature: [`docs/specs/features/07-db-enrichment.md`](docs/specs/features/07-db-enrichment.md)

Depends on: #15

## Sub-issues

- [x] #57
- [x] #354
- [x] #58
- [x] #59
- [x] #60
- [x] #61

- [x] #364 — give `LookupResult` the deprecated item access `FingerprintResult` has. Decided by the user after #59 reported the break.

PR #352, opened 2026-08-09
Make the remote lookup opt-in at the client

Crank-Git:
Part of #18. Refs #57.

## What

`JA4DBClient` performs no network request by default. `JA4DBClient(allow_remote=True)`
permits the remote lookup. The remote lookup has a timeout, and a remote lookup failure
returns nothing and raises nothing.

## Why

On the base commit `79f8c66`, `_do_lookup` called `_remote_lookup` for every fingerprint
the bundled mapping file held no entry for. `ja4plus --lookup analyze capture.pcap`
therefore sent each unknown fingerprint to `https://ja4db.com/api/read/<fingerprint>`.
A fingerprint describes traffic the operator observed, so each request told a third
party what the operator watched. No caller asked for that, and the option name reads as
a request for a local lookup. Silent-by-default is the defect. The opt-in is the repair.

## How

- `ja4plus/ja4db.py`: the constructor takes `allow_remote: bool = False` first and
  `cache_size` second, which is the order `docs/specs/features/07-db-enrichment.md`
  line 131 publishes. `_do_lookup` reads the mapping file, then answers a miss with
  `None` until the operator opts in.
- `ja4plus/ja4db.py`: `_REMOTE_TIMEOUT` holds `5.0`, and `_remote_lookup` passes it.
- `ja4plus/ja4db.py`: the constructor raises `TypeError` when `allow_remote` is no bool,
  and `_remote_lookup` escapes the fingerprint it puts in the request path. The
  self-review comment below states why.
- `ja4plus/ja4db.py`: new `_read_remote_body` accepts one shape, an object that carries
  a non-empty `application` string. It reads `type` and `notes` as strings and
  substitutes an empty string for a field of another type. It returns `None` for every
  other shape, so no unchecked value of the lookup service reaches a caller.
- `ja4plus/cli.py`: the `--lookup` help text now states that the option makes no network
  request. The option builds `JA4DBClient()`, so the command holds the new default. #58
  owns the option that permits the remote lookup and the notice that goes with it.
- `README.md`, `docs/api_reference.md`: an operator can now read that a lookup can leave
  the host, and how to stop it.
- `docs/specs/spec.md`: the `## Terms` table gains `lookup service` and `remote lookup`.

## Behaviour that changed for a caller

Version 0.6.0 read a response object without an `application` field as the application
name `Unknown`. The client now reads it as a miss. `CHANGELOG.md` records it.

## How it is tested

The tests came first. `tests/test_db_offline.py` reported 20 failed and 3 passed against
the unrepaired client, and it reports 30 passed now.

**The default is measured by the attempt, and not by the result.** The `lookup` extra is
absent from this environment, so `_remote_lookup` returns `None` on `ImportError`. A case
that blocks every outbound socket therefore passes against the unrepaired client and
proves nothing. Each such case installs a recording stand-in for `requests` in
`sys.modules` and asserts that its call list is empty.

Two reversals, with `PYTHONDONTWRITEBYTECODE=1` and `inspect.getsource` reporting
PRESENT under each mutation and ABSENT after each restore:

| Mutation | Result |
|---|---|
| `if not self._allow_remote` replaced by `if False and not self._allow_remote` | 3 cases fail, among them `test_the_default_client_looks_1000_fingerprints_up_while_every_socket_fails` |
| The version 0.6.0 body reader restored | 5 cases fail |

No case reaches `ja4db.com`. Every case that exercises the remote path installs a
stand-in in `sys.modules`.

## Gates

```
.venv/bin/pytest tests/ -m "not spec_validation"   2032 passed, 8 xfailed, 96 subtests passed
```

**No fingerprint moves.** The conformance suite reports the same 1531 passed, 143
skipped and 135 xfailed before the change and after it. `tests/foxio_deviations.json`
holds 135 keys against those 135 xfailed cases.

**Coverage rises.** The project holds 92% before and 92% after. `ja4plus/ja4db.py` rises
from 86% to 93%. The unit suite rises from 2002 passed to 2032 passed, which is the 30
cases this pull request adds.

`__all__` of `ja4plus` names 25 entries before and after. This pull request adds no
public name.

## A divergence the project manager owns

`docs/specs/features/07-db-enrichment.md` line 136 publishes a `timeout: float = 5.0`
parameter on the constructor. Line 113 of the same file states that the remote timeout
"is not configurable before version 1.0.0", and the body of #57 states the same. This
pull request ships the second reading: the timeout is the module constant
`_REMOTE_TIMEOUT`, and the constructor takes no `timeout` parameter.

## Changelog

Two files hold a marker, and both read `TBD`:

- `docs/specs/spec.md` — the round row, at the end of the changelog table.
- `CHANGELOG.md` — the `### Changed` entry, "Round TBD".

## Out of scope

`--lookup-remote`, `JA4PLUS_DB_LOOKUP` and the notice on standard error belong to #58.
`lookup_many` and the `source` field belong to #59.

Verified against: https://github.com/FoxIO-LLC/ja4/blob/main/ja4plus-mapping.csv
(retrieved 2026-08-08). The lookup service publishes no versioned API document, which
`.claude/rules/external-apis.md` records, so the client treats every unexpected shape as
a miss.


Crank-Git:
## Self-review

I ran four lenses over the diff myself: correctness and interface compatibility,
security, backend and concurrency, and documentation. No child agent ran them, and none
stalled. Two findings are real, and `32f900b` repairs both. Two findings are declined,
with the reason.

### Finding 1, real: the parameter order reintroduced the disclosure this issue repairs

`cache_size` was the first parameter of `JA4DBClient.__init__`, and
`docs/specs/features/07-db-enrichment.md` line 131 publishes `allow_remote` first. I
adopted the published order, which puts a caller that wrote `JA4DBClient(100)` for a
lookup cache of 100 entries in an unwanted place. Measured against `e5080f1`:

```
$ python -c "from ja4plus.ja4db import JA4DBClient; c = JA4DBClient(100); print(c._allow_remote, c._cache.max_connections)"
100 100000
```

The call permits every remote lookup, and it keeps the default entry count. That is the
exact defect this issue exists to repair, delivered silently to a caller that asked for
something else. A wrong value now fails loudly rather than reaching the lookup service:
the constructor raises `TypeError` when `allow_remote` is no bool. Five cases pin it,
and `docs/api_reference.md` and `CHANGELOG.md` both name the replacement,
`JA4DBClient(cache_size=100)`.

### Finding 2, real: the fingerprint reached the request path unescaped

`_remote_lookup` built `f"https://ja4db.com/api/read/{fingerprint}"`. `lookup` accepts
any string, so a caller value that carries `/` or `?` names another path on the lookup
service. `CLAUDE.md` rule 2 reads every input as hostile, and this is the one call in
the diff that leaves the host. `_remote_lookup` now calls `quote(fingerprint, safe="")`.
A JA4+ fingerprint holds no character the escape changes, so the escape moves no request
this project makes, and one case pins that.

### Finding 3, declined: the two wide `except Exception` handlers stay

`ja4plus/ja4db.py` holds two, at `_do_lookup` and at `_remote_lookup`. #319 owns the
fourteen that remain in the package. This pull request repairs neither and adds none. A
comment above the handler of `_do_lookup` now names that issue.

### Finding 4, declined: `_remote_lookup` reads no `allow_remote` of its own

`_do_lookup` holds the one gate. A second test inside `_remote_lookup` would give two
places to read for one rule. The method is private, `_do_lookup` is its one caller
inside the package, and the docstring states the contract.

PR #356, opened 2026-08-09
Add the command-line opt-in and the disclosure notice

Crank-Git:
## What

`--lookup` reads the bundled mapping file and makes no network request,
FR-db-enrichment-3.

`--lookup-remote` identifies each fingerprint, and it sends every fingerprint the mapping
file holds no entry for to `https://ja4db.com`, FR-db-enrichment-4. It asks for the
lookup as well as for the disclosure, so an operator who passes it needs no `--lookup`.

`JA4PLUS_DB_LOOKUP=1` permits the same request, FR-db-enrichment-5. It serves an operator
who runs a command line another program builds.

The command writes one disclosure notice to standard error for each run that permits the
remote lookup, FR-db-enrichment-6.

Part of #18. Depends on #57.

## Why

A fingerprint describes traffic the operator observed. A request to the lookup service
discloses that traffic to a third party. #57 closed the disclosure at the client. This
issue gives the operator the way to ask for it, and it tells the operator when the
request is on.

## The precedence, and why it is no guess

Each source permits the disclosure and neither one refuses it. FR-db-enrichment-4 states
the option and FR-db-enrichment-5 states the variable, and both read as a permission.
Three consequences follow, and the tests pin all three.

- `JA4PLUS_DB_LOOKUP=0` cancels no option. An operator who wants the local lookup passes
  `--lookup`.
- The variable permits the request on the value `1` and on no other value, which is what
  FR-db-enrichment-5 states. `true`, `yes` and `2` permit nothing.
- The variable asks for no lookup. `JA4PLUS_DB_LOOKUP=1 ja4plus analyze capture.pcap`
  looks nothing up and discloses nothing.

## How it is tested

`tests/test_cli_lookup_opt_in.py` holds 29 cases. **The tests came first**, and the file
reported 24 failed and 4 passed against the unrepaired command, at the 28 cases it held
then. The self-review added the 29th.

The privacy property is measured by the attempt and not by the result. The cases install
the recording stand-in of `tests/test_db_offline.py` in `sys.modules`, so a run that tried
a request fails the case. No case reaches `ja4db.com`.

The notice count is measured against several lookups. One lookup cannot separate one
notice per run from one notice per lookup, so the case asserts more than one recorded
request beside the one notice.

**Three reversals prove the three gates**, each run with `PYTHONDONTWRITEBYTECODE=1`.

| Reversal | Result |
|---|---|
| `JA4DBClient(allow_remote=True)` in place of the option gate | 7 failed, 22 passed |
| The notice moved to the lookup site | 7 failed, 22 passed |
| The variable read as any value that is not empty | 5 failed, 24 passed |
| The restored form | 29 passed |

## Gates

| Gate | Result |
|---|---|
| `pytest tests/ -m "not spec_validation"` | 2061 passed, 8 xfailed, 96 subtests passed. The base holds 2032 passed, 8 xfailed. |
| `pytest tests/ -m spec_validation` | 1531 passed, 143 skipped, 135 xfailed, before the change and after it. |
| `ruff format --check ja4plus/ tests/` | 153 files already formatted |

**No fingerprint moves.** `tests/foxio_deviations.json` holds 135 keys against those 135
xfailed cases.

**Coverage holds.** The project holds 92% before and 92% after, and `ja4plus/cli.py`
rises from 78% to 80%.

**This round adds no wide handler.** The one `except Exception` of `_init_lookup` is the
handler #319 owns, and a comment now names that issue.

## Documentation

An operator can learn from `README.md` that a lookup can leave the host, which option
causes it, and how to stop it. `docs/api_reference.md` holds the same for the option
table and a new section. `docs/output-schema.md` names the second option beside
`identified_as`.

## Changelog markers

`CHANGELOG.md` holds one `TBD` marker, in the new `### Added` entry for #58.
`docs/specs/spec.md` holds one `TBD` row, inserted below the `TBD` row of #57. The
project manager assigns both numbers at the batch gate.

Verified against: https://github.com/FoxIO-LLC/ja4/blob/main/ja4plus-mapping.csv
(retrieved 2026-08-08). No interface outside this repository changes in this round, and
no case reaches a network.


Crank-Git:
## Self-review

I ran three lenses over the diff myself: correctness, security and command-line
behaviour. One finding changed the code, and three findings I state and reject with the
reason.

### Finding 1, accepted — no case covered the run that passes both options

`ja4plus analyze capture.pcap --lookup --lookup-remote` is a command an operator writes,
and no case measured it. Commit 6a23b03 adds
`test_the_two_options_together_build_one_client_and_write_one_notice`. The run builds one
client and writes one disclosure notice.

### Finding 2, rejected — the notice is invisible to a pipeline that drops standard error

A run that writes `2>/dev/null` performs the disclosure and shows no notice.
FR-db-enrichment-6 states standard error, and #52 established that channel. A notice on
standard output would enter the pipe that carries the results and would corrupt the JSON
and the CSV output. The channel stays standard error, and the option itself is the
opt-in.

### Finding 3, rejected — the absent `requests` package ends the run rather than the lookup

`JA4PLUS_DB_LOOKUP=1 ja4plus analyze capture.pcap --lookup` exits with the status 1 on a
host without the `lookup` extra, where a local lookup would have succeeded. The
edge-case table of `docs/specs/features/07-db-enrichment.md` states this result, and the
acceptance criterion of #58 states it too. The behaviour is also the safe one: the
operator asked for the remote lookup, and the command reports that it cannot perform it
rather than performing a different thing in silence.

### Finding 4, rejected — `_init_lookup` imports `requests` and binds no name

The import measures the package. `JA4DBClient._remote_lookup` imports it again at the
request, so the check here is one extra import of a module that `sys.modules` already
holds. It buys the exit-1 that the acceptance criterion asks for, and a check that read
`importlib.util.find_spec` would answer a different question for a stand-in in
`sys.modules`.

### The wide handler

The `except Exception` of `_init_lookup` is the handler #319 owns. This round adds no
handler, and a comment now names that issue. `CLAUDE.md` holds the rule.

PR #357, opened 2026-08-09
Settle the lookup timeout the feature file states two ways

Crank-Git:
## What

`docs/specs/features/07-db-enrichment.md` stated two readings of the remote timeout, in one file.

| Line | What it said |
|---|---|
| 113 | The remote timeout is 5 seconds, and is not configurable before version 1.0.0. |
| 135 | The `Interfaces` block published a constructor parameter `timeout: float = 5.0`. |

A published constructor parameter is configurable by definition, so the two could not both hold. #57 found the divergence, shipped the narrower reading as the module constant `_REMOTE_TIMEOUT`, and reported the question rather than settling it.

## Why the narrower reading wins

`CLAUDE.md` parity rule 2 gives the port the interface where FoxIO specifies nothing, and a timeout is that kind of choice. The port publishes no timeout on the lookup surface.

`lookup.go` of `Crank-Git/ja4plus-go` publishes:

```go
// RemoteLookupConfig configures opt-in remote lookups against ja4db.com.
// A nil http.Client falls back to http.DefaultClient.
type RemoteLookupConfig struct {
	Endpoint string
	HTTPClient *http.Client
}

func LookupFingerprintRemote(ctx context.Context, cfg *RemoteLookupConfig, fingerprint string) (*LookupResult, error)
```

`RemoteLookupConfig` names two fields, and neither one is a timeout. The caller carries the deadline in `ctx`, which is the Go convention and not a published parameter.

The 30 second interval of `cmd/ja4plus/main.go:332` is no counter-example. It sits inside `runDBUpdate`, it covers the mapping file download and not the lookup, and a command internal publishes nothing.

## How

- The `Interfaces` block loses the `timeout` parameter. **No file under `ja4plus/` moves**, because `JA4DBClient.__init__` already names `allow_remote` and `cache_size` alone.
- The block gains four sentences that state why it publishes no timeout, so #59 edits a block that no longer contradicts itself.
- `tests/test_db_offline.py` gains `test_the_constructor_publishes_no_timeout_parameter`. It reads `inspect.signature(JA4DBClient.__init__)` and pins the parameter list. A specification statement that a test can contradict outranks one that only a reader can. #332 and #334 shipped the same shape.
- Line 113 does not move, so the comment of `ja4plus/ja4db.py:53` that cites it stays accurate.
- `docs/api_reference.md:751` already published `JA4DBClient(allow_remote=False, cache_size=100000)`, so it carried no third reading and needs no change.

## Test-driven development, honestly

This is a prose repair, and the prose carries no behaviour, so no test came first for it. The case pins the code the prose describes, and it was written after the edit rather than before it. This pull request states that rather than inventing a test for prose.

**The case is proven by its reversal.** With `timeout: float = 5.0` added to `JA4DBClient.__init__`, it fails:

```
FAILED tests/test_db_offline.py::TestTheOptInClient::test_the_constructor_publishes_no_timeout_parameter
1 failed, 30 deselected
```

With the constructor restored it passes, under `PYTHONDONTWRITEBYTECODE=1`.

## How tested

Base measured on `origin/epic/18-db-enrichment` at `3c0ce5f`, after #57 and #58 landed. All five gates run with `.venv/bin/`.

| Gate | Base | This branch |
|---|---|---|
| `pytest tests/ -m "not spec_validation"` | 2061 passed, 8 xfailed | 2062 passed, 8 xfailed |
| `pytest tests/ -m spec_validation` | 1531 passed, 143 skipped, 135 xfailed | 1531 passed, 143 skipped, 135 xfailed |
| `ruff check ja4plus/ tests/` | — | All checks passed! |
| `ruff format --check ja4plus/ tests/` | — | 153 files already formatted |
| `mypy --strict ja4plus/` | — | Success: no issues found in 31 source files |

**No fingerprint moves.** The conformance suite is identical before and after, and `tests/foxio_deviations.json` holds 135 keys against those 135 xfailed cases.

**Coverage does not fall.** The project holds 92% before and 92% after.

## Changelog

`docs/specs/spec.md` holds a `TBD` round, inserted after the `TBD` rows of #57 and #58. Neither of those rows moves.

`CHANGELOG.md` holds none. This round changes no shipped behaviour and no published interface, so it reaches no reader of that file.

Verified against: https://github.com/Crank-Git/ja4plus-go/blob/master/lookup.go (retrieved 2026-08-08)

Part of #18. Settles the question #57 reported. Unblocks #59.


Crank-Git:
## Self-review

I ran the lenses myself rather than delegating them.

**Correctness — one finding, fixed in 8d2f5ef.** The note that states why the constructor publishes no `timeout` sat between the `Interfaces` code block and the sentence `LookupResult carries the port's three fields plus source`. That sentence reads the code block, and the note separated the two. The note now follows the sentence.

**Correctness — no third reading elsewhere.** I swept `docs/` for every other statement of the timeout.

- `docs/api_reference.md:751` already published `JA4DBClient(allow_remote=False, cache_size=100000)`. It carried no timeout, so it needed no change.
- `docs/mutation_sweep.md:936` names `tests/test_ja4db.py::TestJA4DBClient::test_remote_lookup_timeout`. That is a record of a sweep and not an interface claim. Left alone.
- `docs/specs/spec.md:665` is #57's round. It records that line 136 published the parameter, which was true when it was written. A changelog row is evidence, so it is not rewritten.

**Line references — checked.** The removed parameter sat below line 113, so line 113 does not move. The comment of `ja4plus/ja4db.py:53` cites line 113 and stays accurate. #57's round cites line 131 for the constructor order, and line 131 does not move either.

**Scope — no code moves.** `git diff --stat` against the base names three files, and none is under `ja4plus/`. `JA4DBClient.__init__` already matched the settled reading, so the repair asks nothing of it.

**Test quality — the case earns its place.** A prose statement that only a reader can contradict is weaker than one a test can. The reversal proves the case contradicts the parameter, not merely the current code:

```
FAILED tests/test_db_offline.py::TestTheOptInClient::test_the_constructor_publishes_no_timeout_parameter
1 failed, 30 deselected
```

**Scope of the pin — deliberate.** The case pins the whole parameter list rather than the absence of `timeout` alone, because the `Interfaces` block publishes the whole list. #59 edits that block to publish `LookupResult` and adds no constructor parameter, so the pin does not obstruct it. A future issue that adds a parameter must change this file and the case together, which is the intended cost.

**Writing standard.** Every added sentence is 23 words or fewer, active voice, present tense. `timeout` appears as an identifier in backticks where it names the parameter.

**Decision boundary held.** The evidence pointed at removal, so this issue did not need the user. Had the port published a timeout on its lookup surface, this would have returned `needs-feedback` rather than adding a parameter to a version 1.0.0 interface.

Issue #61, opened 2026-08-06
Verify the database subcommands against the cache directory


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/07-db-enrichment.md`](docs/specs/features/07-db-enrichment.md):

> FR-db-enrichment-11 — `db info` reports the mapping file path, its entry count and its source.

> FR-db-enrichment-12 — `db update` downloads the mapping file to a cache directory and leaves the bundled file unchanged.

> FR-db-enrichment-13 — The client prefers a cached mapping file over the bundled one.

## What to build

`db update` and `db info` already exist at `ja4plus/cli.py:422` to `429` and are at parity with the port. This slice confirms their behaviour against the rules the spec sets, and fixes them where they differ.

`db update` never writes inside the installed package: a package directory may be read-only, and a wheel reinstall would discard the file. The cache directory follows the platform convention — `$XDG_CACHE_HOME/ja4plus` or `~/.cache/ja4plus` on Linux, `~/Library/Caches/ja4plus` on macOS.

The client prefers a cached mapping file over the bundled one. A cache file that is empty or corrupt falls back to the bundled file and reports the problem at `WARNING`.

The mapping file comes from https://raw.githubusercontent.com/FoxIO-LLC/ja4/main/ja4plus-mapping.csv

## Acceptance criteria

- [ ] `ja4plus db info` reports the path, the entry count and the source.
- [ ] `ja4plus db update` writes to the platform cache directory and leaves the installed package unchanged.
- [ ] After `db update`, `db info` reports `cache` as the source.
- [ ] A corrupt cache file makes `db info` report `bundled` as the source.

## Links

- Feature: [`docs/specs/features/07-db-enrichment.md`](docs/specs/features/07-db-enrichment.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #18
Depends on #57

PR #363, opened 2026-08-09
Add bulk lookup and record the source of every result

Crank-Git:
Closes the last member of Epic 7. Part of #18. Refs #59.

## What

`JA4DBClient.lookup_many` accepts a sequence of fingerprints and returns one entry per
fingerprint, FR-db-enrichment-7. `LookupResult` becomes a frozen dataclass that carries
the port's three fields plus `source`, FR-db-enrichment-8.

## Why

An analyst needs to know where a name came from to judge how much to trust it. A result
that names no source reads the same whether the mapping file inside the package answered
it or a third-party service did.

## The source names a provenance, not a location

A hit from the mapping file inside the package records `embedded`. A hit from the file
that `ja4plus db update` wrote records `cache`. A hit from the lookup service records
`remote`. `_read_mapping_file` takes the source of the file it reads, so no entry can
report a file the client did not read.

The word `bundled` stays where it describes the file inside the package. `_BUNDLED_CSV`
and `_load_bundled_db` keep their names, as #61 decided.

## `lookup_many` multiplies no network

`lookup_many` reaches the lookup service under the rule that `lookup` holds, and under no
other rule. FR-db-enrichment-1 and FR-db-enrichment-2 govern the client, and
`docs/specs/features/07-db-enrichment.md` grants the bulk call no separate permission.

- A client built with `allow_remote=False` sends nothing over 100000 fingerprints.
- A client built with `allow_remote=True` sends one request for each fingerprint the
  mapping file holds no entry for.
- The lookup cache holds a miss as well as a hit, so 1000 occurrences of one fingerprint
  cost one request. One consent stays one disclosure per distinct fingerprint.

The lookup service publishes one path, `https://ja4db.com/api/read/<fingerprint>`, and no
bulk interface. No smaller request count is reachable.

## A breaking change to `lookup`, and the specification settles it

Through version 0.6.0 `JA4DBClient.lookup` returned a dict. It now returns the frozen
`LookupResult`, so a caller reads `result.application` where it read
`result["application"]` before. The `Interfaces` block of
`docs/specs/features/07-db-enrichment.md` publishes both the dataclass and the signature
`lookup(self, fingerprint: str) -> LookupResult | None`.

`LookupResult` carries no deprecated item access. `FingerprintResult` carries one because
FR-typed-api-6 asks for it; no requirement of this feature asks for one here.

**The command-line output schema does not change.** The command writes the application
name alone, and `tests/test_output_schema.py` holds the header and the columns.

## The specification read two ways, and this repairs it

#61 published `embedded` in the behaviour rule, the `## Terms` table, `CHANGELOG.md` and
`docs/api_reference.md`, and it left two sites reading `bundled`. Both now read
`embedded`:

- `docs/specs/features/07-db-enrichment.md:164`, the `source` comment of the
  `LookupResult` block.
- `docs/specs/features/07-db-enrichment.md:216`, the acceptance criterion that names
  `LookupResult`.

## `__all__` is unchanged, and this states why

`__all__` of `ja4plus` names 25 entries before this change and after it. It names no name
of `ja4plus.ja4db`: neither `JA4DBClient` nor `lookup` is there today. A `LookupResult`
added alone would promise a type that no promised function returns, and FR-typed-api-13
makes a name there permanent until version 2.0.0. Whether version 1.0.0 promises the
lookup interface at the top level is a question for the issue that names that move, and
this change makes it no harder to answer.

## How tested

Test-driven. `tests/test_db_lookup_many.py` came first and its 15 cases failed for the
stated reasons: 8 with `AttributeError: 'JA4DBClient' object has no attribute
'lookup_many'`, 5 with `AttributeError: 'dict' object has no attribute 'source'`, 1 with
`AttributeError: type object 'JA4DBClient' has no attribute 'lookup_many'`, and 1 with
`TypeError: must be called with a dataclass type or instance`.

All three sources are read, because a case that reads `embedded` alone cannot separate a
client that reads the source from a client that writes one value.

Three mutations prove the cases measure behaviour:

| Mutation | Failures |
|---|---|
| The remote result reports `embedded`. | 5 |
| `lookup_many` reads past the lookup cache. | 1 |
| Every mapping file entry reports `embedded`. | 2 |

No test reaches `ja4db.com`. Every case takes the `no_network` fixture of #57 or the
`RecordingRequests` stand-in of the same module.

### Gates

```
pytest tests/ -m "not spec_validation"   2099 passed, 2 skipped, 8 xfailed, 96 subtests
ruff format --check ja4plus/ tests/      155 files already formatted
```

No fingerprint moves. The conformance suite reports the same counts on the base commit
`826a566` and on this branch, and `tests/foxio_deviations.json` holds 135 keys against
those 135 xfailed. The unit suite rises from 2084 to 2099, and the 15 new cases are the
whole rise. Coverage holds at 93%, and the missed statement count holds at 307 while the
statement count rises by 9.

Verified against: https://github.com/Crank-Git/ja4plus-go/blob/master/lookup.go (retrieved
2026-08-08). `LookupResult` at `lookup.go:23` holds `Application`, `Type` and `Notes`.
`dbSource` at `lookup.go:31` holds `embedded`. The port publishes no bulk lookup, so
`lookup_many` follows this project's specification alone.

## Changelog

`CHANGELOG.md` and `docs/specs/spec.md` each carry a `TBD` row for this round. No other
worker's row is renumbered.


Crank-Git:
## Self-review by lens

I ran the four lenses myself over the whole diff, rather than through child agents.

### Correctness — one defect found and repaired

`JA4DBClient.lookup` carried a docstring that read "dict with 'application', 'type',
'notes' keys, or None if unknown." The method returns the frozen `LookupResult`, so the
docstring named a shape the method no longer builds. Repaired in `09016fd`.

### Privacy — one defect found and repaired

Three cases that read the lookup service took no `no_network` fixture:
`test_a_hit_from_the_lookup_service_records_remote`,
`test_a_repeated_miss_reaches_the_service_once` and
`test_the_service_reads_one_request_for_each_distinct_miss`. The `RecordingRequests`
stand-in records a request the client makes through `requests`, and it measures nothing
that another package opens. Each case now takes the fixture, so it measures the attempt
and not the result. Repaired in `c731fe6`.

### Interface — three readings, stated rather than assumed

1. **`lookup` returns the frozen result and this breaks a version 0.6.0 caller.** The
   `Interfaces` block of `docs/specs/features/07-db-enrichment.md` publishes the
   dataclass and the signature `lookup(self, fingerprint: str) -> LookupResult | None`,
   so the specification settles it. The command-line output schema does not move, because
   the command writes the application name alone.

2. **`LookupResult` carries no deprecated item access.** `FingerprintResult` carries one
   because FR-typed-api-6 asks for it. No requirement of this feature asks for one here,
   so this change adds none rather than invent an interface.

3. **`__all__` stays at 25 entries.** It names no name of `ja4plus.ja4db` today, so it
   names neither `JA4DBClient` nor `lookup`. A `LookupResult` added alone would promise a
   type that no promised function returns, and FR-typed-api-13 makes a name there
   permanent until version 2.0.0. Not adding it is the reversible reading, and a later
   issue may publish the whole lookup interface at the top level.

### Considered and declined

`lookup_many` refuses no `str`. A `str` is a `Sequence[str]`, so
`client.lookup_many("t13d...")` reads one character per entry. #57 refuses a
non-bool `allow_remote` because that call caused a disclosure; a `str` here causes no
request and no disclosure, and the specification asks for no refusal. I record the
reading rather than add an unpublished `TypeError` to a version 1.0.0 interface. A
reviewer who wants the refusal should say so.

### Mutations

Every mutation was applied to the committed tree and reverted after the measurement.

| Mutation | Failures |
|---|---|
| `_read_remote_body` reports `embedded`. | 5 |
| `lookup_many` calls `_do_lookup` and reads past the lookup cache. | 1 |
| `_read_mapping_file` reports `embedded` for every entry. | 2 |

Issue #364, opened 2026-08-09
Give LookupResult the deprecated item access FingerprintResult has


Crank-Git:
Decided by the user on 2026-08-08, after #59 reported the breaking change and did not paper over
it. **The user ruled it a specification gap, not a deliberate difference.**

## The measurement

**#59 made `JA4DBClient.lookup` return a frozen `LookupResult` where version 0.6.0 returned a
dictionary.** A caller reading `result["application"]` now raises.

**Epic 4 met the identical situation and answered it the other way.** `FR-typed-api-5` gives
`FingerprintResult` item access by field name *"so that code written against the dictionary keeps
working for one major version"*, and `FR-typed-api-6` makes that access emit a
`DeprecationWarning`.

**No requirement of `docs/specs/features/07-db-enrichment.md` asks for the same**, and **#59 was
right not to invent one** — it reported the difference and shipped what the specification said.

## The decision

**`LookupResult` gets the same treatment as `FingerprintResult`: item access by field name,
emitting a `DeprecationWarning`.** A version 0.6.0 caller keeps working for one major version and
**is told what to change rather than meeting a `TypeError` that names nothing.**

## What to build

1. **Two requirements in `docs/specs/features/07-db-enrichment.md`**, worded to match
   `FR-typed-api-5` and `FR-typed-api-6` rather than freshly invented. **Read those two first**; a
   requirement that says the same thing differently is a divergence waiting to be found.
2. **Item access on `LookupResult`, reading only**, with the warning. **Follow `ja4plus/types.py`
   exactly** — #44 built `FingerprintResult.__getitem__`, and this project does not want two
   spellings of one behaviour.
3. **The same failure modes #44 covered.** An unknown key raises `KeyError` **without** first
   emitting the warning — #44's case runs under `warnings.simplefilter("error")` to prove that
   ordering. **Assignment still raises `FrozenInstanceError`.**

## What to check that #44 could not

**`FingerprintResult` renamed a field the dictionary published**: the method key became `type`.
**Check whether `LookupResult`'s four fields match the keys version 0.6.0's dictionary actually
published** — read the 0.6.0 code, not the specification. **A field this project renamed needs its
old key to raise `KeyError` rather than silently return the wrong value**, and if any key changed,
say so and return `needs-feedback` before deciding what the old name does.

## Verification depth

**The floor binds** (`.issue-flow.json`, `practices.verificationDepth`): five gates, TDD with the
[...]

PR #365, opened 2026-08-09
Give LookupResult the deprecated item access FingerprintResult has

Crank-Git:
Part of #18. Builds #364.

## What this changes

`LookupResult` reads by field name, FR-db-enrichment-16, and that read emits a
`DeprecationWarning`, FR-db-enrichment-17. #59 replaced the dict of version 0.6.0 with a
frozen `LookupResult`, so `result["application"]` raised
`TypeError: 'LookupResult' object is not subscriptable`. A version 0.6.0 caller now keeps
working for one major version, and it is told what to change.

## Why this spelling

`ja4plus/types.py` holds `FingerprintResult.__getitem__` from #44, under `FR-typed-api-5`
and `FR-typed-api-6`. This project holds one spelling of the behaviour, so the new method
copies it: it reads a module-level frozen set of field names, it raises `KeyError` before
it warns, it passes `stacklevel=2` so the caller's line is reported, and it returns
`getattr(self, key)`. The two new requirements are worded from those two.

## The check #44 could not make

`FingerprintResult` renamed a key the dict published, so `result["method"]` raises
`KeyError`. **`LookupResult` renamed nothing.** `git show v0.6.0:ja4plus/ja4db.py` builds
the dict at `_load_bundled_db` and at `_remote_lookup`, and each site writes the same
three keys: `application`, `type` and `notes`. Each one names a field of `LookupResult`,
so no key of version 0.6.0 returns the value of another field. The `source` field of
FR-db-enrichment-8 is an addition and no rename.

## How this is tested

The cases came first. `tests/test_db_item_access.py` reported 6 failed and 4 passed
against the unrepaired dataclass, at the 10 cases it held then, and the first failure
read `TypeError: 'LookupResult' object is not subscriptable`.

Four mutations prove the cases measure behaviour. Each one was proven live by reading the
loaded source back with `PYTHONDONTWRITEBYTECODE=1`, and the restored form passed 11 of
11 after each one.

| Mutation | Result |
|---|---|
| The warning moves before the membership test | 3 fail, among them `test_an_unknown_key_raises_a_key_error_and_emits_no_warning_first` |
| The warning is removed | 2 fail |
| `stacklevel=1` | `test_the_warning_reports_the_line_of_the_caller` fails alone |
| `getattr(self, key)` becomes `self.application` | `test_item_access_returns_the_value_the_attribute_holds` fails alone |

A case asserting only that item access works cannot separate a deprecated accessor from a
permanent one, so the warning and the ordering each carry a case.

No case reaches `ja4db.com`. The file builds one `LookupResult` and it opens no socket.

## The self-review finding

A `__getitem__` makes the old iteration protocol reach the object, so `list(result)`
calls `result[0]` and raises `KeyError` where the dict of version 0.6.0 iterated its keys.
`FingerprintResult` behaves the same way and records it nowhere. The eleventh case records
it here. No caller inside `ja4plus/` iterates a lookup result: `ja4plus/cli.py:385` is the
one site and it reads `match.application`. The method adds no `__len__`, so `if match:`
keeps its meaning.

## The five gates

```
.venv/bin/pytest tests/ -m "not spec_validation"   2110 passed, 2 skipped, 8 xfailed
.venv/bin/ruff format --check ja4plus/ tests/      156 files already formatted
```

**No fingerprint moves.** The conformance suite reports the same three counts on the base
commit `21d658e` and on this branch, and `tests/foxio_deviations.json` holds 135 keys
against those 135 xfailed cases.

**Coverage does not fall.** The project holds 93% before and 93% after, and
`ja4plus/ja4db.py` reads 95%. The unit suite rises from 2099 passed to 2110 passed, which
is the 11 cases this issue adds.

**`__all__` of `ja4plus` names 25 entries before and after.**

## Documentation

- `docs/specs/features/07-db-enrichment.md` — FR-db-enrichment-16, FR-db-enrichment-17,
  and the method in the `Interfaces` block.
- `docs/specs/spec.md` — the two requirement rows, and one round row marked `TBD`.
- `CHANGELOG.md` — one entry under `### Added`, marked `Round TBD`.
- `docs/api_reference.md` — `#### The deprecated item access` under the lookup client.

Verified against: https://github.com/Crank-Git/ja4plus-go/blob/master/lookup.go (retrieved
2026-08-08), for the three fields `LookupResult` carries.


Crank-Git:
Self-review, run by this worker over the whole diff. Three lenses, and one finding.

**Correctness.** The method copies `FingerprintResult.__getitem__` line for line, and four mutations separate each of its four properties: the membership test, the warning, the `stacklevel`, and the field the return reads. The module constant `_FIELD_NAMES` shadows no name of `ja4plus/ja4db.py`, and it is built at import time from `dataclasses.fields(LookupResult)`, so a field the dataclass gains is reachable without a second edit.

**Security.** The method reads a declared field and no other attribute, so it reaches neither `_cache` nor `_db`. It opens no socket, it parses no packet, and it changes no path that reads the mapping file or the lookup service.

**The finding, and it is real.** A `__getitem__` makes the old iteration protocol reach the object. `list(result)` therefore calls `result[0]` and raises `KeyError`, where the dict of version 0.6.0 iterated its keys. `FingerprintResult` has behaved this way since #44 and records it nowhere, so a reader meets it by accident. `test_an_integer_key_raises_a_key_error` now records it. No caller inside `ja4plus/` iterates a lookup result: `ja4plus/cli.py:385` is the one site and it reads `match.application`. The method adds no `__len__`, so `if match:` keeps its meaning.

The finding needed no code change, so the method still holds one spelling of the behaviour.

Issue #361, opened 2026-08-09
Batch: the hygiene findings that follow #322


Crank-Git:
The third hygiene batch, formed by the project manager on 2026-08-08 when #322 shipped its six
completed members. **No member here is new work and none is dropped.** #322 records the
ship-partial decision.

## Members, in order

- [x] #279 — **state the memory ceiling of 512 MiB the user decided on 2026-08-08.** Ready since
  that decision and never scheduled until now; #60 has since measured the projection at 286.02 MiB.
- [x] #347 (the list first named #346, which is a merged pull request) — measure the capability-declined rows so the `capability` field alone carries the bar.
  Split from #341, which reported that its own field cannot prove its behaviour on the one issue
  the bar was written for.
- [x] #359 — decide whether the lookup cache pays for eviction statistics nothing reads. Found by
  #60, which measured `StateTable._evicted_keys` at roughly a third to a half of the cache's
  47.06 MiB.
- [x] #258 — repair the duplicate Changelog round 9. **The user delegated it on 2026-08-08 and the project manager never scheduled it.**
- [x] #272 — rule on the six JA4L entries #34 left open. **Same: delegated, ruled, never scheduled.**
- [x] #345 — correct the decided-entry count in the marker rule. Found by #341.
- [x] #297 — enable the ruff F401 rule.
- [ ] #326 — report the dropped-packet count of the capture layer. Found by #55.
- [x] #319 — read the fourteen bare `Exception` handlers the package still holds. **A reading, not
  a sweep.**
- [x] #302 — resolve the three `TBD` Changelog rounds that reached `dev`, **and add the guard that
  stops a fourth.**

## The order

**#279 runs first because three siblings read against it.** #359 weighs a saving against the
ceiling, #60 already reported 286.02 MiB against it, and the specification does not yet state the
number — **so a worker asserting against 512 MiB today would be measuring a number from a
comment.** The #60 worker refused to do that, correctly.

**#302 runs last**, because it assigns round numbers and the rounds it must not disturb have to
stop moving first.

**#319 may not fit in one issue.** Its own body says to read all fourteen sites, report the
grouping, narrow only the group the `CLAUDE.md` rule binds, and **stop and report if that group
holds more than about four sites** rather than letting one issue grow without a bound.

## What no member does

**No member changes a fingerprint, no member changes the published output schema, and no member
touches `__all__`**, which names 25 entries and is the version 1.0.0 interface promise.

[...]

PR #367, opened 2026-08-09
State the 512 MiB memory ceiling and read it (#279)

Crank-Git:
Closes nothing on its own. Part of #361, the third hygiene batch. Refs #279.

## What this states

**This package states a memory ceiling of 512 MiB for the one-million-packet case at the
shipped defaults.** The user decided the number on 2026-08-08. `docs/specs/features/03-concurrency-safety.md`
left the ceiling open and asked #38 to propose one, and it now states the number, the
defaults it holds at, the traffic the case feeds, and the boundary of the claim.

`README.md` and `docs/api_reference.md` each carried the sentence
`This package states no memory ceiling.` Each now states the ceiling and its boundary.

## The measurement, re-measured rather than quoted

Four runs of 1000000 packets across 100000 distinct connections, on macOS 26.6.1 with
Python 3.14.3, on a ten-core Apple silicon laptop.

| Run | Peak resident memory |
|---|---|
| 1 | 383.47 MiB |
| 2 | 388.25 MiB |
| 3 | 392.05 MiB |
| 4 | 394.94 MiB |

**The highest is 394.94 MiB, which is 77 percent of the ceiling.** The four sit 11.47 MiB
apart. **The table states every run and not the best one**, because a resident memory
reading moves with the machine.

**Three earlier figures describe no branch that exists.** #38 projected 238.96 MiB, round
82 read 419.27 MiB, and #60 put the projection at 286.02 MiB. This branch reads an idle
`Processor` at about 100 MiB, 40200 entries across seventeen state tables, and 400000
values across ten value lists.

## The premise this pull request corrects

**The brief expected a raised table limit to turn the ceiling case red. It does not, and
this measured it three ways.**

| Reversal | Reading at 1000000 packets | Result |
|---|---|---|
| `ja4ssh.connections` raised from 10000 to 100000 | 383.00 MiB | Green |
| `ja4l.connections` and `ja4l.grouping_keys` raised from 10000 to 100000 | 412.06 MiB | Green |
| Every one of the fifteen entry counts raised to 100000 | No reading | The run never finished |

The third produced no reading across three attempts: the operating system killed two, and
a third reached 2500 connections in ten minutes.

**The age bound and the traffic mix hold this ceiling, and the entry count bound binds two
tables of seventeen.**

| State table | Entries | Maximum | What holds it |
|---|---|---|---|
| `ja4l.connections` | 10000 | 10000 | The entry count |
| `ja4l.grouping_keys` | 10000 | 10000 | The entry count |
| `ja4s._quic_dcids` | 6000 | 10000 | The age |
| `ja4ssh.connections` | 6000 | 10000 | The age |
| `ja4t.connections` | 6000 | 10000 | The age |

The case advances the capture timestamp 0.1 seconds for each connection, so 100000
connections span 10000 seconds and a 600-second age holds a table at about 6000 entries.
**The 5423 bytes for each JA4SSH entry that #38 measured describe a full 200-packet
window**, and the case gives each connection two SSH packets.

## What makes the comparison falsifiable

**The packet count falsifies the ceiling, and this measured the crossing rather than
projecting it: 1500000 packets across 150000 connections reads 513.06 MiB.**
`BaseFingerprinter.fingerprints` grows without a limit, at about 23 MiB for each 100000
packets past the 200000th. That replaces a projection in the specification with a reading.

**The committed control proves the entry count bound is measured.** With `CONTROL_BOUND`
raised from 100 to the shipped 10000,
`test_a_smaller_entry_count_holds_less_resident_memory` fails and reads
`a bound of 10000 added 38.26 MiB and the shipped bounds added 39.08 MiB, a ratio of 0.979, so the entry count changed nothing`.
Reverted afterwards.

**That control compares a ratio and not a MiB figure**, because a resident memory reading
moves with the platform and the gate runs on `ubuntu-latest` as well as `macos-latest`.
Five runs at the default size hold the ratio between 0.746 and 0.769, the raised bound
reads 0.955 and 0.979, and the threshold of 0.85 sits between the two. The self-review
comment records why the first form was replaced.

## How it is tested

New file `tests/memory_ceiling_run.py` feeds the packets in an interpreter of its own and
reports `resource.getrusage(RUSAGE_SELF).ru_maxrss`. **The separate interpreter is part of
the measurement**: `ru_maxrss` reports the high-water mark of the whole process, so a
reading taken inside the pytest session measures every case that ran before it.

`TestTheStatedMemoryCeiling` of `tests/test_memory_bounds.py` reads it and carries three
controls, because the default size holds 3000 connections and the ceiling comparison alone
cannot fail there.

**The case feeds 30000 packets by default, and `JA4PLUS_MEMORY_CEILING_PACKETS` sets the
count.** The full run costs 481 seconds, and a case of that length costs every later run of
the unit suite. `tests/test_thread_safety.py` holds the same arrangement for the 60-second
soak of #40, and this pull request records the runs at 1000000 above.

**Five runs at the default size** held the control ratio between 0.746 and 0.769, while
one reading held a run-to-run spread of 1.03 MiB.

## Gates

```
pytest tests/ -m "not spec_validation"   2055 passed, 8 xfailed, 96 subtests passed
ruff format --check ja4plus/ tests/      156 files already formatted
```

- **Coverage holds exactly.** 92%, 322 missed of 4135, measured on the base commit and on
  this branch.
- **The register invariant holds.** `tests/foxio_deviations.json` holds 135 keys against
  135 xfailed.
- **`__all__` names 25 entries.**
- **No file under `ja4plus/` changes and no fingerprint moves.** The unit suite rises from
  2051 passed to 2055 passed, which is the four cases this adds.

## Changelog

**Both markers read `TBD`.** `docs/specs/spec.md` carries a `TBD` changelog row, and
`CHANGELOG.md` carries a `TBD` entry under `[Unreleased]`. The project manager assigns the
round at the batch gate.

## Terms

`docs/specs/spec.md` gains `resident memory` and `memory ceiling` to the `## Terms` table.


Crank-Git:
## Self-review

**I ran the lenses myself and did not delegate them.** Four lenses over the diff:
correctness, platform robustness, scope, and the writing standard.

### One real defect found and repaired, in `241c242`

**The control margin was scale-dependent.** The first form read
`control_growth + 4.0 MiB < shipped_growth`. On macOS the two runs separate by 9.03 MiB
at the closest, so 4.0 MiB is about four times the run-to-run spread and it holds. **The
gate also runs on `ubuntu-latest` across five Python versions**, and a platform whose
resident memory readings are smaller would compress that separation toward the margin. An
absolute margin that suits one platform can sit above the whole signal on another, and the
case would then fail for the platform rather than for the bound.

The comparison now reads a ratio, which measures the bound rather than the platform.

| Reading | Ratio |
|---|---|
| Five runs at the default size, control bound 100 | 0.746 to 0.769 |
| Control bound raised to the shipped 10000 | 0.955 and 0.979 |

The threshold of `CONTROL_GROWTH_RATIO = 0.85` sits between the two, and a
`CONTROL_FLOOR_MIB = 10.0` guard stops the ratio being read off two small numbers. The
raised bound still fails the case, and it now reads
`a bound of 10000 added 38.26 MiB and the shipped bounds added 39.08 MiB, a ratio of 0.979, so the entry count changed nothing`.

### A second fragility, repaired before the first commit

`test_the_run_feeds_every_packet_the_case_states` compared the reading against
`CEILING_PACKETS` directly. The run stops on a whole connection, so a
`JA4PLUS_MEMORY_CEILING_PACKETS` value that is no multiple of ten would have failed that
control for the wrong reason. It now compares against `CEILING_RUN_PACKETS`, which rounds
down the same way the run does. `memory_ceiling_run.py` also rejects a connection count
below one rather than dividing by zero.

### Findings I read and rejected

- **The subprocess parses the last line of standard output rather than the whole of it.**
  Deliberate: scapy writes a deprecation warning today and a future release may write a
  banner to standard output. A comment states the reason.
- **The case adds about 35 seconds to the unit suite, across six continuous-integration
  jobs.** Accepted. `tests/test_thread_safety.py` made the same trade for the 60-second
  soak of #40, and the alternative is a 481-second case in every gate run.
- **The ceiling comparison cannot fail at the default size.** True, and stated in the
  class docstring and in the specification rather than hidden. The default size holds 3000
  connections against a 512 MiB ceiling. That is why three controls sit beside it, and why
  the crossing at 1500000 packets is measured and recorded.

### Scope

`git diff --stat` touches no file under `ja4plus/`. No fingerprint moves, the conformance
suite reads 1531 passed, 143 skipped and 135 xfailed before and after, the register holds
135 keys against 135 xfailed, and `__all__` names 25 entries. Coverage holds at 92 percent
with 322 missed statements of 4135, measured on the base commit and on this branch.

### The one thing a reviewer should weigh

**The stated ceiling is not falsified by the entry count bound at this traffic mix**, and
the pull request body records the three reversals that show it. A reader who wants the
entry count bound to be the load-bearing term should read the entry count table in
`features/03-concurrency-safety.md` first: the age bound holds three of the five large
tables, because the case advances the capture timestamp 0.1 seconds for each connection.

Issue #371, opened 2026-08-09
Remove the wall-clock race from the periodic statistics case


Crank-Git:
A second case of the fault #369 repaired stays in the same file, outside the class #369 names.

`tests/test_watch_statistics.py::TheStatisticsGoToStandardError::test_the_periodic_line_reaches_standard_error` runs `ja4plus watch` with `--stats-interval 0.05`, sleeps 0.3 seconds inside the capture, and asserts that at least 2 statistics lines reached standard error.

**The assertion measures how promptly the host schedules a thread.** Six wake-ups are due and the case needs 2 of them, where the case #369 repaired needed 3 of 7. The margin is wider, so the case has not failed yet, and the mechanism is the same one. `Event.wait` restarts the whole interval after each line, so the write cost and the scheduling latency accumulate, and the sweep of #369 read 4 lines of 5 due at a 0.25 second sleep on an idle host.

## Why #369 left it

#369 scoped its acceptance criteria to `TheReporterWritesOneLinePerInterval`, and it added the `wait` seam to `StatisticsReporter` for that class. This case does not build a reporter. It reaches the reporter through `cmd_watch`, so a deterministic form needs a seam that carries an injected wait from the command down to `report_statistics`. That is a wider change than the hotfix carried.

## What to change

Make the case state the schedule rather than sample it, without weakening what it proves: that `--stats-interval` on the command line reaches the reporter and that the line reaches standard error, FR-live-capture-9 and FR-live-capture-10.

Do not fix this by raising the sleep or lowering the count. #369 states the reason.

## Acceptance criteria

1. The case holds no assertion whose outcome depends on how promptly the host schedules a thread.
2. The case still proves that `--stats-interval` passed on the command line reaches the reporter.
3. The case still proves that the periodic line reaches standard error and not standard output.
4. Running the class 20 times in a row passes 20 times. Record the command and the result.
5. The five gates pass: both pytest markers, `ruff check`, `ruff format --check`, `mypy --strict ja4plus/`.
6. Coverage does not fall.

Crank-Git:
## Plan

**Approach.** The case reaches the reporter through `cmd_watch`, so it cannot use the
`wait` seam #369 added to `StatisticsReporter`. Carry an injected wait from the command
down to `report_statistics`, then have the case drive the schedule rather than sample it.

The case must keep proving both facts it exists for: that `--stats-interval` from the
command line reaches the reporter, FR-live-capture-9, and that the periodic line reaches
[...]

Issue #369, opened 2026-08-09
Remove the wall-clock race from the statistics interval cases


Crank-Git:
`dev` is red on macOS. `tests/test_watch_statistics.py::TheReporterWritesOneLinePerInterval::test_the_gap_between_two_lines_holds_the_interval` failed on the `macos-latest, 3.12` job of the run for `8ef8acc` with `AssertionError: 2 not greater than or equal to 3`. The five other jobs passed it.

## What happened

The case starts a `StatisticsReporter` at a 0.05 second interval, sleeps 0.35 seconds of wall clock, and asserts that at least three lines were written.

**The assertion measures the scheduler of the host, not the reporter.** A loaded runner delivers six wake-ups late, and the case reports a defect the package does not hold. Its sibling `test_the_reporter_writes_a_line_for_each_interval_that_passes` sleeps 0.55 seconds and asserts at least four lines, so it carries the same fault and has not failed yet.

## What to change

Make the schedule of the reporter observable without a wall-clock race. Two readings are open and the choice is yours to make with evidence:

- Drive the reporter from an injected clock and an injected wait, so the case advances time and reads an exact line count. This states the schedule rather than sampling it.
- Keep the real clock and read the gap the reporter asks for rather than the gap the host delivered.

**Do not fix this by raising the sleep or lowering the count.** A longer sleep costs the suite time and moves the failure rather than removing it, and the class docstring already states that the suite must not wait one second per line.

The case that this issue must not weaken is the one that matters: `--stats-interval 1` writes a statistics line every interval, FR-live-capture-9. A replacement that no longer proves the interval reaches the wait is a loss, not a fix.

## Acceptance criteria

1. `tests/test_watch_statistics.py::TheReporterWritesOneLinePerInterval` holds no assertion whose outcome depends on how promptly the host schedules a thread.
2. The class still proves that the interval the caller gives reaches the wait of the reporter, FR-live-capture-9.
3. The class still proves that no line is written before the first interval passes.
4. Running the class 20 times in a row passes 20 times. Record the command and the result in the verdict.
5. The five gates pass: both pytest markers, `ruff check`, `ruff format --check`, `mypy --strict ja4plus/`.
6. Coverage does not fall.

## How to verify

State the failing case first. A wall-clock race does not reproduce on demand, so reproduce it by loading the host or by shrinking the sleep until the count falls short, and record what you did. Then show the replacement passing 20 consecutive runs.

## Notes

[...]

Issue #368, opened 2026-08-09
Remove the example daemon that a merge restored and return dev to green


Crank-Git:
`dev` is red. `tests/test_watch_capture.py::TheExampleDaemonIsAbsent::test_the_repository_holds_no_monitoring_daemon_example` fails on all six test jobs, and it has failed since `8b87c20`.

## What happened

#56 removed `examples/monitoring_daemon.py` and added the case that asserts the file stays absent. That commit, `f4633ec`, reached `dev` at `79f8c66`.

`batch/322-hygiene-two` forked from `dev` before `79f8c66`, so its own continuous integration run never saw the case. Its merge, `8b87c20`, restored the file. **Both parents were green and the merged tree was not**, and no merge tool reports it, because the file and the case live apart.

The next two merges, `35bdfab` and `8ef8acc`, ran red. **`8ef8acc` merged pull request #366 over six failing test jobs.**

## What to change

Remove `examples/monitoring_daemon.py` again. `ja4plus watch` is the supported monitor and the case states why the example does not return.

Read the repository for every reference to the file before you remove it — `README.md`, `docs/`, `examples/` and `pyproject.toml`. A reference that survives the removal is a second defect. Remove or correct each one you find, and name each in the verdict.

## Acceptance criteria

1. `examples/monitoring_daemon.py` is absent from the tree.
2. `tests/test_watch_capture.py::TheExampleDaemonIsAbsent` passes.
3. No file in the repository names `monitoring_daemon`, except the case that asserts its absence.
4. The five gates pass: both pytest markers, `ruff check`, `ruff format --check`, `mypy --strict ja4plus/`.
5. Coverage does not fall.

## How to verify

State the failing case first. Run `pytest tests/test_watch_capture.py::TheExampleDaemonIsAbsent -q` and record that it fails on the base commit, then record that it passes after the removal.

## Notes

This is a hotfix. It bypasses batching. Branch from `dev`, open a normal pull request into `dev`, and let continuous integration run. Do not add `[skip ci]`.

The verification floor binds this issue. The heavy proof does not: no fingerprint moves, no published field changes, and no FoxIO reading is at stake.

Crank-Git:
## The project manager rules on acceptance criterion 3

**The worker is right and the criterion as written is wrong.** Criterion 3 reads "No file in the repository names `monitoring_daemon`, except the case that asserts its absence." That bars every mention, and three classes of mention cannot be removed without a loss.

[...]

PR #372, opened 2026-08-09
Remove the wall-clock race from the statistics interval cases

Crank-Git:
Fixes the red `dev` on macOS. Refs #369.

## What was wrong

`TheReporterWritesOneLinePerInterval` started `StatisticsReporter` at a 0.05 second interval, slept a fraction of a second, and asserted a lower bound on the count of lines that arrived. **That count measures how promptly the host schedules a thread, and the class reported it as a defect of the package.** The `macos-latest, 3.12` job of the run for `8ef8acc` read `AssertionError: 2 not greater than or equal to 3`, and the five other jobs of the same run passed.

## The failing case, reproduced before the repair

The issue asks for the race to be reproduced deliberately. The user's rules bar a worker from running a load generator, so the sleep was shrunk instead, with the assertion left unchanged at `assertGreaterEqual(len(times), 3)`:

```
  sleep  wakeups due  lines seen  assertGreaterEqual(len(times), 3)
  0.350            6           6  pass
  0.300            5           5  pass
  0.250            5           4  pass
  0.200            4           3  pass
  0.175            3           3  pass
  0.150            2           2  FAIL: 2 not greater than or equal to 3
  0.125            2           2  FAIL: 1 not greater than or equal to 3
  0.100            2           1  FAIL: 1 not greater than or equal to 3
```

**The lost wakeup at 0.25 settles it.** An idle host already delivers 4 lines where 5 are due, so the count is a function of the host and not of the reporter. `Event.wait` restarts the whole interval after each line, so the write cost and the scheduling latency accumulate.

## Which reading, and why

The issue left two readings open. **This change takes the injected wait, on the following evidence rather than on preference.**

The other reading reads the gap the reporter asks for rather than the gap the host delivered. The only way to observe that gap is to intercept the wait, so **both readings cost the same seam.** Given the seam, the injected wait is a superset of the other reading: it holds that reading's assertion, it reads an exact line count instead of a lower bound, and it removes the wall clock from all four cases rather than one.

## What changed

`StatisticsReporter` gains one parameter, `wait`. It receives the interval and returns True when the stop arrives, which matches `threading.Event.wait`. The default is `self._stop.wait`, `ja4plus watch` passes no other call, and `_run` is the one line that changed. **The shipped behaviour does not move.**

The class now drives the reporter from a scripted wait. It holds no `time.sleep`, where the earlier form slept 1.10 seconds per run.

A case binds the seam to the shipped path, because an injected wait proves nothing about FR-live-capture-9 unless the default wait is the real one. `TheReporterStopsCleanly` then measures that the stop ends the thread.

## The cases discriminate — proven by mutation

The tests came first. The six cases reported `TypeError: StatisticsReporter.__init__() got an unexpected keyword argument 'wait'` against the unrepaired class.

Five mutations then broke the reporter deliberately. Each was proven live by reading the loaded source back with `PYTHONDONTWRITEBYTECODE=1`, so no stale `.pyc` could make a mutation look applied when it was not.

|---|---|
| The reporter writes one line before it waits | 4 |
| The reporter asks the wait for a constant `0.05` | the interval case, with `AssertionError: Lists differ: [0.05, 0.05, 0.05, 0.05] != [1.0, 1.0, 1.0, 1.0]` |
| The reporter writes two lines for one interval | 3 |
| The reporter ignores the injected wait | 6 |
| `write_statistics` flushes nothing | the flush case alone |

The restored form passes 6 of 6 after each mutation. **The count assertions are exact and no longer a lower bound**, so a reporter that writes an extra line now fails, where the earlier `assertGreaterEqual` passed.

## How it was tested

- **20 consecutive runs of the class: 20 passed, 0 failed.** `.venv/bin/python -m pytest tests/test_watch_statistics.py::TheReporterWritesOneLinePerInterval -q`, run 20 times, each reporting `6 passed, 2 subtests passed` in about 0.45 s.
- `pytest tests/ -m "not spec_validation"` — 2196 passed, 2 skipped, 8 xfailed. One failure, `tests/test_watch_capture.py::TheExampleDaemonIsAbsent`, is **pre-existing on `origin/dev`** and belongs to #368, which another worker holds. It was confirmed failing on the unmodified base.
- `pytest tests/ -m spec_validation` — 1531 passed, 143 skipped, 135 xfailed. `tests/foxio_deviations.json` holds 135 keys against those 135 xfailed, so the register invariant holds. **No fingerprint moves.**
- `ruff format --check ja4plus/ tests/` — 162 files already formatted.
- `mypy --strict ja4plus/` — Success: no issues found in 31 source files.
- **Coverage holds at 93%.** `ja4plus/watch.py` reads 99% on the base and 99% here. The unit suite rises from 2194 passed to 2196, which is the two cases this adds net of the four it replaces.

## Out of scope, filed

**#371** records one case of the same fault that this change leaves. `TheStatisticsGoToStandardError::test_the_periodic_line_reaches_standard_error` sleeps 0.3 seconds and asserts at least 2 lines. It reaches the reporter through `cmd_watch` rather than building one, so a deterministic form needs a wider seam than this hotfix carries. Its margin is 2 of 6 wakeups, where the case that failed needed 3 of 7.

This change touches no file under `examples/` and no `tests/test_watch_capture.py`, which #368 owns.

## Documentation

`docs/api_reference.md` names the `wait` parameter and states what it is for. `CHANGELOG.md` and the Changelog table of `docs/specs/spec.md` each gain a row, **both with the round as the literal `TBD`** for the project manager to assign.

No external interface is called by this change, so no vendor documentation needed verification.


Crank-Git:
## Self-review, two lenses

I ran a correctness lens and a test-quality lens over the diff in parallel.

The reviewer named a concrete reporter mutation that each of the six cases catches, so **no case is tautological**. It confirmed the seam changes no shipped behaviour: `report_statistics` at `ja4plus/watch.py:405` is the one caller inside the package, and it passes no `wait`, so the shipped path makes the same `Event.wait` call as before. It confirmed the False-count of `ScriptedWait` equals the line count against `_run`, and that the test cannot hang, because the join is bounded at 5 seconds.

### Test quality — four findings accepted, one rejected

**Accepted and fixed** — three sentences of my new prose broke the 25-word cap of `.claude/rules/ste.md` rule 1: the `ScriptedWait` docstring, the class docstring, and the `CHANGELOG.md` entry. Each is now split. Two minor ones are also fixed: the docstring of `ScriptedWait` dropped the relative pronoun, and the docstring of `run_reporter` opened with two actions rather than the result.

**Accepted as a wording defect, rejected as a coverage defect** — the reviewer read the word "superset" in the spec row as a claim against the *earlier* form, and observed that no case now asserts the elapsed time between two lines. The observation is correct and the claim was never that. "Superset" was written against the other reading the issue left open, which asserts the gap the reporter asks for. **Removing the elapsed-time assertion is the point of this issue**, because acceptance criterion 1 bars exactly that assertion. The wording was ambiguous, so I sharpened the spec row to say so plainly, and the class docstring now states that the removal is deliberate and names the reason.

**Rejected** — the reviewer noted `TheReporterStopsCleanly::test_the_stop_waits_no_whole_interval` and `TheStatisticsGoToStandardError::test_the_periodic_line_reaches_standard_error` still read the clock. Both are outside the class this issue names. The first asserts that a stop returns in under 1 second against a 60 second interval, which is a liveness bound and not a promptness measurement. The second is a real instance of the same fault, and #371 records it, with the reason a fix needs a wider seam than this hotfix carries.

### The one gate that stays red

Every test job of this run fails `tests/test_watch_capture.py::TheExampleDaemonIsAbsent::test_the_repository_holds_no_monitoring_daemon_example`. **That failure is on the base**: the run for `8ef8acc` on `dev` fails the same case on all six test jobs, with `AssertionError: True is not false`. It belongs to #368, which another worker holds, and nothing in this branch touches `examples/` or `tests/test_watch_capture.py`. `conformance`, `fuzz` and `lint` are green here.

Crank-Git:
## Final continuous integration result — run `31295719336`

`lint`, `conformance` and `fuzz` are green.

All six test jobs fail, and **the only failing case on all six is the same one**:

```
   6 FAILED tests/test_watch_capture.py::TheExampleDaemonIsAbsent::test_the_repository_holds_no_monitoring_daemon_example - AssertionError: True is not false
```

That case belongs to **#368**, which another worker holds. It fails identically on the base: the run for `8ef8acc` on `dev` fails the same case on the same six jobs. Nothing in this branch touches `examples/` or `tests/test_watch_capture.py`.

**The case this pull request repairs passes on the job that reported the defect.** From the `macos-latest, 3.12` log of run `31295502684`:

```
tests/test_watch_statistics.py::TheReporterWritesOneLinePerInterval::test_the_reporter_asks_the_wait_for_the_interval_the_caller_states PASSED [ 97%]
tests/test_watch_statistics.py::TheReporterWritesOneLinePerInterval::test_the_reporter_flushes_each_line PASSED [ 97%]
tests/test_watch_statistics.py::TheReporterWritesOneLinePerInterval::test_the_reporter_waits_on_the_stop_event_when_the_caller_injects_no_wait PASSED [ 97%]
tests/test_watch_statistics.py::TheReporterWritesOneLinePerInterval::test_the_reporter_writes_no_line_before_the_first_interval_passes PASSED [ 97%]
tests/test_watch_statistics.py::TheReporterWritesOneLinePerInterval::test_the_reporter_writes_no_second_line_for_one_interval PASSED [ 97%]
tests/test_watch_statistics.py::TheReporterWritesOneLinePerInterval::test_the_reporter_writes_one_line_for_each_interval_that_passes PASSED [ 97%]
```

`AssertionError: 2 not greater than or equal to 3` appears nowhere in this run.

**This branch cannot show a green continuous integration result until #368 lands**, and that is not a defect of this change. The merge order is the project manager's decision.

Issue #347, opened 2026-08-09
Measure the capability-declined rows so the field alone carries the bar


Crank-Git:
Split from #341 by the project manager on 2026-08-08. **#341 shipped the `capability` field and
reported that one thing it should be able to prove, it cannot.** This issue closes that.

## The gap

**The capability bar is enforced twice — by the `capability` field, and by an absent
measurement.**

`tests/test_precedence_exception.py` holds `SOURCE_VALUES`, the table of Rust, Wireshark and Zeek
values for every candidate row. **#334 measured the 35 source values of #129 and deliberately
left them out**, under the same bar #341 later made checkable, and
`test_no_measured_row_records_a_capability_decline` asserts the exclusion.

**So flipping `capability` on a #129 entry cannot move the reach.** The field is proven
load-bearing on a #96 row — #341 measured 6 to 5, with the mutation live — but **its behaviour on
the one issue the bar was written for is untested. If the field were wrong for exactly that case,
nothing would catch it.**

**That is this project's signature defect** — a comparison that is never made reads as a
comparison that passes. Sixteen instances are recorded. This one is known, which is the only
reason it is a scheduled issue rather than a discovery.

## What to build

**Add the 35 #129 source values to `SOURCE_VALUES`, measured against the pinned FoxIO checkout**
at `27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8`, so the **field alone** carries the bar.

**Then `test_no_measured_row_records_a_capability_decline` must change or go**, because it asserts
the exclusion this issue removes. **Read it before you delete it** — if it holds a second meaning
beyond the exclusion, keep that meaning under a new name and say so.

**The reach must stay at 6.** The 35 rows enter the table and the field bars all 35. **If the
reach moves, the field and the exclusion disagreed, and the field is not automatically right** —
stop, report which rows moved, and return `needs-feedback`.

## The proof this issue exists to make possible

**Flip `capability` to `false` on the #129 entries and confirm the reach rises**, then revert.
**Report both numbers.** That is the measurement #341's acceptance criterion asked for and could
not have.

With the mutation proven live: clear every `__pycache__` outside `.venv/`, set
`PYTHONDONTWRITEBYTECODE=1`, confirm with `inspect.getsource`, **discriminating in both
directions**. **#341 verified its own probe by checking that `if deviation.capability:` appears
once in the head file and zero times in the base**; follow that.

**This is conformance-affecting work, so the heavy proof applies.**

[...]

Issue #129, opened 2026-08-07
Decide whether ja4plus decrypts a capture that carries TLS secrets


Crank-Git:
## What is wrong

Two FoxIO vectors carry a Decryption Secrets Block. The FoxIO reference reads the
application data of both captures, because it decrypts them with those secrets.
`ja4plus` decrypts no TLS record and no QUIC 1-RTT packet, so it produces no JA4H value
from either capture. No issue owns that gap.

## Evidence

```
http2-with-cookies.pcapng  CLIENT_RANDOM blocks: 1
http2-with-cookies.pcapng packets 1607 cleartext http 0
chrome-cloudflare-quic-with-secrets.pcapng packets 83 cleartext http 0
```

`tests/foxio_vectors/http2-with-cookies.pcapng.json` holds 15 JA4H values on stream 0.
Every one of them reads `ge20` as the method and the version code. `ja4plus` produces
none of them.

`ja4plus` reads the QUIC Initial packet of `chrome-cloudflare-quic-with-secrets.pcapng`,
because the Initial keys come from the Destination Connection ID. It reads no later
packet, because the 1-RTT keys come from the handshake.

## Why it matters

Seventeen register entries name #35 as the owner of these misses. #35 fixes a repeated
cookie name and the `HTTP/2` request line, and neither change reaches traffic that
`ja4plus` never decrypts. #128 corrects those entries and needs an issue to point them
at.

The decision also sets a security boundary. A reader of a key log file reads the plain
text of the traffic the operator captured. That is a larger capability than fingerprint
production, and it is a choice, not an oversight.

## What to build

Answer one question: does `ja4plus` read the secrets a capture carries?

- No: state the limitation in `docs/specs/spec.md`, and point the JA4H register entries
  of both captures at this issue as a documented divergence.
- Yes: read the Decryption Secrets Block of a pcapng file, decrypt the TLS records and
  the QUIC 1-RTT packets it unlocks, and produce the 15 JA4H values the reference holds.

## Acceptance criteria

- [ ] `docs/specs/spec.md` states whether `ja4plus` reads a Decryption Secrets Block.
- [ ] Every JA4H register entry on `http2-with-cookies.pcapng` and
      `chrome-cloudflare-quic-with-secrets.pcapng` names this issue and the stated cause.
- [ ] `tests/test_ja4h_spec.py::test_the_foxio_capture_produces_the_reference_ja4h_values`
      passes, or its `xfail` reason names this issue.

## Links

[...]

PR #373, opened 2026-08-09
Measure the capability-declined rows so the field alone carries the bar

Crank-Git:
Measures the 35 source values of #129 so the `capability` field alone carries the bar.
Part of #347.

## What

**The bar stood twice, and one of the two enforcers was an absent measurement.** #341
shipped the field on all 135 register entries and reported that the one thing it should
prove, it could not: #334 measured the 35 source values of #129, deliberately left them
out of `SOURCE_VALUES`, and `test_no_measured_row_records_a_capability_decline` asserted
the exclusion. Flipping the field on a #129 entry therefore moved nothing, and no case
failed.

This branch adds the 35 measured rows to `SOURCE_VALUES`, so the field is the one fact
that bars them.

## Why the reading is trustworthy

The measurement read the pinned FoxIO checkout at
`27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8`. It resolved every Wireshark frame number
against its capture with `scapy` and read the 5-tuple, which is the method #334 recorded;
the Wireshark testdata publishes `frame.number` and no connection, so a capture-level
reading would mis-key these rows.

**The reading corroborates itself.** All 16 Wireshark JA4H values and all 3 Wireshark
JA4X values equal the FoxIO Python values at the same connection, method and occurrence.
That is what a correct frame-to-connection resolution produces, and a misaligned one
would not.

**No Zeek baseline reaches these rows.** `zeek/README.md` states that the JA4X module is
off, no baseline names `http2-with-cookies.pcapng`, and the one baseline that names
`chrome-cloudflare-quic-with-secrets.pcapng` is a `conn.log` holding JA4L and JA4T columns
alone.

**Two of the 37 rows carry no source value**, and they are
`chrome-cloudflare-quic-with-secrets.pcapng/0:57098/JA4X.1` and `JA4X.2`, because neither
the Rust snapshot nor the Wireshark file holds a JA4X value for that capture. The figure
of 35 is the one #334 recorded, reached here by an independent run.

The captures are byte-identical between the pinned checkout and `tests/foxio_vectors/`:
`http2-with-cookies.pcapng` is `1f6949d5…` in both, and
`chrome-cloudflare-quic-with-secrets.pcapng` is `b7c9de12…` in both.

## The measurement #341 asked for

**The reach is 6 before the change and 6 after it.**

With `capability` false on all 43 #129 entries, written to `tests/foxio_deviations.json`
on disk, **the reach rises from 6 to 25** and returns to 6 after the restore. The register
digest reads `6273fc7ba15d3558` before, `713fe5d2c556f72c` with the mutation, and
`6273fc7ba15d3558` after.

**On the base file the same flip reads 6 and 6.** That is the discrimination in the other
direction, and it is the gap this issue closes.

The 19 rows that enter are the 16 JA4H_ro rows and the 3 JA4X rows. **The 16 JA4H rows
stay out under the disagreement bar**, because the Rust value and the Wireshark value
differ on every one of them.

## The renamed case

`test_no_measured_row_records_a_capability_decline` asserted only the exclusion this
issue removes, and its sibling `test_every_measured_row_names_a_decided_register_key`
already carried the `decided` half. It is renamed to
`test_every_measured_capability_row_names_the_issue_the_bar_exists_for`, which holds the
one meaning the exclusion did not: a measured capability row names #129, and a row under
another issue would reach a bar no measurement has read.

## How tested

`pytest tests/ -m "not spec_validation"` reports 2200 passed, 2 skipped, 8 xfailed.
`pytest tests/ -m spec_validation` reports 1531 passed, 143 skipped, 135 xfailed, against
135 register keys. `ruff check`, `ruff format --check` and `mypy --strict ja4plus/` are
clean. Coverage reads 93%.

**The case failed first.** Before the 35 values entered the table,
`test_the_field_alone_bars_the_measured_capability_rows` failed with
`AssertionError: assert {'ssh-r.pcap/...884/JA4L-S.1'} == {'chrome-clou...H_ro.13', ...}`
and 19 extra items in the right set.

**Five live mutations prove neither new case can pass vacuously.** Dropping the capability
check from `exception_reach`, dropping one JA4X row from the expected rise, dropping one
JA4X row from the measured table, reading the measured count as 34, and reading the
capability issue as 130 each fail the case they target, and each case passes clean and
after the restore. Every run cleared every `__pycache__` outside `.venv/` and set
`PYTHONDONTWRITEBYTECODE=1`, and `inspect.getsource` reports `if deviation.capability:`
once in the loaded `exception_reach`. The token
`"http2-with-cookies.pcapng/0:58847/JA4H_ro.1"` appears twice in the head file and zero
times in the base file.

## What does not move

No file under `ja4plus/` changes, no register entry changes, no vector is adopted and no
fingerprint moves. `tests/foxio_deviations.json` is byte-identical to the base.

Verified against: https://github.com/FoxIO-LLC/ja4/tree/main/wireshark/test/testdata (pinned commit 27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8, read on disk 2026-08-09)
Verified against: https://github.com/FoxIO-LLC/ja4/tree/main/rust/ja4/src/snapshots (same pinned commit)
Verified against: https://github.com/FoxIO-LLC/ja4/tree/main/zeek (same pinned commit)


Crank-Git:
## Self-review

Reviewed by three lenses: correctness of the measurement, whether the new cases can fail,
and the writing standard.

**Correctness of the measurement — no finding.** The reading is checkable without trusting
the script that produced it. All 16 Wireshark JA4H values and all 3 Wireshark JA4X values
equal the FoxIO Python values at the same connection, method and occurrence. A misaligned
frame-to-connection resolution would not produce that. The count of rows with a source
value came to 35 of 37, which is the figure #334 recorded independently, and the 2 without
are the two JA4X keys of `chrome-cloudflare-quic-with-secrets.pcapng`. The captures are
byte-identical between the pinned checkout and `tests/foxio_vectors/`.

**No value was transcribed by hand.** The 35 rows were written into the table by a script
and then read back and compared against the measurement output, which reported 0 omitted
keys and 0 wrong values. A hand transcription of 30 KB of fingerprint data is the error
mode this project exists to avoid.

**Whether the new cases can fail — no finding.** Five live mutations each kill the case
they target, and both cases pass clean and after the restore. The mutations are: dropping
the capability check from `exception_reach`, dropping one JA4X row from the expected rise,
dropping one JA4X row from the measured table, reading the measured count as 34, and
reading the capability issue as 130. Every run cleared every `__pycache__` outside
`.venv/` and set `PYTHONDONTWRITEBYTECODE=1`, and each mutated file digest differed from
the clean one.

**The renamed case is not a rename in name only.** `test_no_measured_row_records_a_capability_decline`
asserted only the exclusion this issue removes, and its sibling
`test_every_measured_row_names_a_decided_register_key` already carried the `decided` half.
The replacement asserts a statement the old table could not express: a measured capability
row names #129, and the table holds 35 of them.

**Writing standard — eight findings, all fixed in `9cf9a9f`.** Eight sentences ran past
the 25-word limit of `.claude/rules/ste.md`, in the module docstring, four comments, two
docstrings, the rule page and the Changelog entry. Each one splits. No assertion, no
measured value and no number changed.

**One point a reader should check rather than take on trust.** The 16 JA4H rows do not
enter the reach even with the field flipped, because the Rust value and the Wireshark
value differ on every one of them. The disagreement bar holds them, not the capability
field. The measurement therefore proves the field load-bearing on 19 rows of the 35, not
on all 35, and the Changelog and the rule page both say so.

Issue #359, opened 2026-08-09
Decide whether the lookup cache pays for eviction statistics nothing reads


Crank-Git:
Found by #60 on 2026-08-08 while it measured the lookup cache. **#60 measured it, named the
design decision, and made none.**

## The measurement

**A full lookup cache holds about 47 MiB**, confirmed by two methods that agree to 0.1 MiB —
`tracemalloc` reports 47.06 MiB and `sys.getsizeof` reports 46.95 MiB.

**Roughly a third to a half of that is `StateTable._evicted_keys`**, holding 100000 keys that are
disjoint from the 100000 keys the cache holds. **So the table's bookkeeping approaches the size of
the table.**

## Why it is waste here specifically, and not in general

**That memory exists to compute `returned_connections`** — a statistic that tells an operator how
often an evicted connection came back, which is real evidence that a bound is set too low.

**Nothing reads it for this table.** `Processor.stats` collects the fingerprinters' tables, and
**`JA4DBClient` is no fingerprinter**. No code under `ja4plus/` calls `JA4DBClient._cache.stats()`.

**A fingerprint that was looked up and evicted and looked up again is not the same event as a
connection returning.** The statistic answers a question nobody asks of this table.

## What to decide, and it is a design decision

**`StateTable` tracks evicted keys for every table.** Making it optional touches a class that
Epic 14 bounded and that eight tables use. **Read those callers before proposing anything** —
a parameter that defaults wrong would silently drop a statistic the concurrency work relies on.

Two shapes, and **measure both rather than reasoning about them**:

1. **A constructor parameter** — `track_evictions=True` by default, `False` for the lookup cache.
   Every existing caller keeps its behaviour and only `JA4DBClient` opts out.
2. **A bound on `_evicted_keys` itself.** The set exists to answer "did this key come back", which
   a bounded structure answers approximately. **This is the more invasive reading and it changes
   the statistic for every table — if you propose it, say what it costs the concurrency contract.**

**Report the measured saving for whichever you propose.** A change that saves 15 MiB of 286 MiB
is worth a different decision than one that saves 23 MiB, and neither is worth guessing at.

## The context that decides whether this matters at all

**The user set the stated ceiling at 512 MiB for the one-million-packet case** (#279, decided
2026-08-08). **#60 measured the projection at 286.02 MiB with the cache included, which is 56% of
the ceiling.**

**So this is a saving, not a repair. Nothing is over the ceiling.** Read it that way and do not
[...]

PR #374, opened 2026-08-09
Let the lookup cache opt out of the memory of its evicted keys

Crank-Git:
Closes nothing on its own. This pull request builds #359, and the batch pull request closes the issue.

## What it does

`StateTable` remembers the key of every entry it evicts, and that memory buys `returned_connections`. `Processor.stats` collects the state tables of the fingerprinters, `JA4DBClient` is no fingerprinter, and no code under `ja4plus/` reads the count for the lookup cache.

`StateTable` and `BoundedStateTable` now take `track_evictions`. It defaults to True, so every existing caller keeps the statistic it had. `JA4DBClient` is the one caller that states False.

This is **shape 1** of the two shapes #359 names. **Shape 2, a bound on `_evicted_keys` itself, is unnecessary**, because shape 1 recovers the whole saving. Shape 2 would change the statistic for every table, and this pull request does not build it.

## The measurement

The saving is **16.06 MiB**, and the two methods #60 used agree on it to 0.00 MiB.

| Reading | Before | After | Saving |
|---|---|---|---|
| `tracemalloc` | 47.06 MiB | 31.00 MiB | 16.06 MiB |
| `sys.getsizeof` | 44.66 MiB | 28.60 MiB | 16.06 MiB |

The subject is a full lookup cache: a `JA4DBClient` whose remote lookup returns None, across 200000 distinct lookups. The first 100000 fill the entries and the second 100000 evict every one of them, so both structures are full. **The before figure reproduces the 47.06 MiB of #60 exactly**, which ties this measurement to that one.

Each run set `PYTHONDONTWRITEBYTECODE=1` and cleared every `__pycache__` outside `.venv/` first. A second script measured both settings **in one process**, so no difference in the run conditions reaches the difference between the two numbers, and it read the same 16.06 MiB by both methods.

## What it is measured against

**#279 measured the 512 MiB ceiling case without a lookup cache**, so this saving moves none of its four runs of 383.47, 388.25, 392.05 and 394.94 MiB. It moves the worst case of a program that runs a monitor and a full lookup cache together, from 442.00 MiB to 425.94 MiB.

## The callers

Eleven callers build a state table, and ten keep the statistic. Each one states no `track_evictions`, so each one costs nothing and reports what it reported before.

| Caller | file:line |
|---|---|
| `quic_fragment_table` | `ja4plus/fingerprinters/ja4.py:44` |
| The reassembly table | `ja4plus/fingerprinters/ja4h.py:98` |
| `connections`, `grouping_keys` | `ja4plus/fingerprinters/ja4l.py:100`, `:104` |
| `_quic_dcids` | `ja4plus/fingerprinters/ja4s.py:58` |
| `connections`, `_handshake_clients` | `ja4plus/fingerprinters/ja4ssh.py:80`, `:89` |
| `connections` | `ja4plus/fingerprinters/ja4t.py:51` |
| `times`, `prefixes` | `ja4plus/fingerprinters/ja4ts.py:107`, `:113` |
| `processed_certs`, `scan_offsets` | `ja4plus/fingerprinters/ja4x.py:140`, `:142` |
| `_connections` | `ja4plus/watch.py:445` |
| `TCPStreamReassembler` | `ja4plus/utils/tcp_stream.py:78` |
| `_cache`, the one caller that opts out | `ja4plus/ja4db.py:297` |

**The new parameter sits before `on_eviction` and breaks no caller**, because every caller of `on_eviction` states it by keyword: `ja4plus/fingerprinters/ja4ts.py:111`, `ja4plus/watch.py:449`, and three cases in `tests/test_state_table.py`.

## What does not move

The eviction count stands for the lookup cache, so FR-concurrency-safety-12 holds for it, and the invariant `inserts == entries + evictions + removals` holds. No entry-count bound and no age bound changes. No fingerprinter changes and no fingerprint moves. The register holds 135 keys against 135 xfailed, and `__all__` names 25 entries.

## How it is tested

Ten cases came before the change, and nine failed against it. The tenth, `test_the_lookup_cache_still_counts_the_entries_it_evicts`, is a control that passes on the base commit.

**Each new case is proven by its reversal.** With the `if not self.track_evictions: return` guard replaced by `if False:`, four cases fail:

```
FAILED tests/test_state_table.py::TestTheOptionalEvictionMemory::test_a_table_that_tracks_no_eviction_remembers_no_key
FAILED tests/test_state_table.py::TestTheOptionalEvictionMemory::test_a_table_that_tracks_no_eviction_counts_no_returned_connection
FAILED tests/test_state_table.py::TestTheOptionalEvictionMemory::test_the_age_bound_of_a_table_that_tracks_no_eviction_remembers_no_key
FAILED tests/test_ja4db.py::TestTheLookupCacheEvictionMemory::test_the_lookup_cache_remembers_no_evicted_key
```

**A first form of the new cases turned the resident-memory control of #279 red, and the production change is not implicated.** `test_a_smaller_entry_count_holds_less_resident_memory` failed 2 of 2 full runs under `--cov=ja4plus` on this branch and passed 3 of 3 on the base commit. A run that deselects the two new case classes passes, and a run that deselects the three `JA4DBClient` cases alone passes. Each of those cases loaded the whole mapping file into the session, and the control reads the resident memory of the session. The cases now patch `load_mapping_file` to an empty mapping, because each one reads the lookup cache and no mapping entry.

## Documentation

`docs/specs/spec.md` and `CHANGELOG.md` each carry a `## Changelog` row with the round as the literal `TBD`. `docs/api_reference.md`, `docs/specs/features/03-concurrency-safety.md` and `docs/specs/features/07-db-enrichment.md` state the parameter and the measurement.

Verified against: the repository's own code. This change reaches no external interface.


Crank-Git:
## Self-review

I reviewed the diff by four lenses. I found no blocker. Two findings changed the work, and both are already pushed.

### Correctness

`count_eviction` raises `evictions` **before** the guard returns, so a table that tracks no eviction still reports every eviction and FR-concurrency-safety-12 holds for it. `take_evicted_key` reads an empty `OrderedDict` and returns False for every key, so `count_insert` never raises `returned_connections`. `forget_evicted_keys` clears an empty mapping, and `clear` calls it safely. The invariant `inserts == entries + evictions + removals` holds, and `test_a_table_that_tracks_no_eviction_counts_every_eviction` asserts it.

`max_evicted_keys` goes unread when the table tracks no eviction. It stays on the object because `tests/test_ja4ts_prefix_bound.py:68` writes it on a tracking table.

### Caller safety

The new parameter sits **before** `on_eviction`, so I checked every construction site. Eleven callers build a state table under `ja4plus/`, and none passes an argument positionally past `max_connections`. Every caller of `on_eviction` states it by keyword: `ja4plus/fingerprinters/ja4ts.py:111` and `ja4plus/watch.py:449`, plus three cases in `tests/test_state_table.py`. No test or example builds `StateTable` directly.

### Test quality — every case is proven by a mutation

The project has been burned by tautological cases, so I mutated the code three ways and recorded which cases fail.

|---|---|
| The guard never fires, `if False: return` | `test_a_table_that_tracks_no_eviction_remembers_no_key`, `test_a_table_that_tracks_no_eviction_counts_no_returned_connection`, `test_the_age_bound_of_a_table_that_tracks_no_eviction_remembers_no_key`, `test_the_lookup_cache_remembers_no_evicted_key` |
| The guard precedes `self.evictions += 1` | `test_a_table_that_tracks_no_eviction_counts_every_eviction`, `test_the_lookup_cache_still_counts_the_entries_it_evicts` |
| The default flips to False | `test_the_default_table_tracks_its_evictions`, and 5 existing cases of `tests/test_processor_stats.py::TestTheReturnedConnections` |

The second mutation is the one that matters most: it shows the two cases I wrote as controls are **not** tautological, because they separate a guard that drops the memory from a guard that also drops the count.

`test_a_table_that_tracks_evictions_remembers_the_key_it_evicted` and `test_a_table_that_tracks_evictions_counts_the_connection_that_returns` state `track_evictions=True` explicitly, so the third mutation leaves them green. That is deliberate: they measure the explicit setting, and `test_the_default_table_tracks_its_evictions` measures the default.

### The finding that changed the work

**A first form of the `JA4DBClient` cases turned `test_a_smaller_entry_count_holds_less_resident_memory` red.** It failed 2 of 2 full runs under `--cov=ja4plus` on this branch and passed 3 of 3 on the base commit, which read as a defect in the change. It is not. A run that deselects the two new case classes passes, and a run that deselects the three `JA4DBClient` cases alone passes. Each of those cases loaded the whole mapping file into the session, and that control reads the resident memory of the session. The cases now patch `load_mapping_file` to an empty mapping. Commit `6ec798f` carries it, and the `## Changelog` rows record it.

### Scope

The change reaches `ja4plus/utils/state_table.py` and `ja4plus/ja4db.py` and nothing else under `ja4plus/`. No entry-count bound and no age bound moves. No fingerprinter changes. `__all__` names 25 entries, so the change adds no public name.

Issue #258, opened 2026-08-08
Repair the duplicate Changelog round 9 in the specification


Crank-Git:
Found while #211 counted the Changelog rounds of `docs/specs/spec.md`. #211 needed the
round count to write a note in `docs/specs/spec.html`, and the count of rows disagreed
with the highest round number.

## The measurement

`docs/specs/spec.md` holds 69 Changelog rows and 68 distinct round numbers. **Two rows
carry the number 9, and each row records different work.** The measurement runs against
`origin/batch/252-above-the-epics` at commit `07f5a7a`.

| Line | Row |
|---|---|
| 532 | `\| 9 \| 2026-08-06 \| Epic 1 batch 1 shipped: #29 ...` |
| 536 | `\| 9 \| 2026-08-06 \| #80 landed and JA4L conformance is measured for the first time. ...` |

Reproduce the count with one command, from the root of the repository.

```bash
awk -F'|' '/^\| [0-9]+ \| 2026-0[0-9]-[0-9][0-9] \|/ {gsub(/ /,"",$2); print $2}' \
  docs/specs/spec.md | sort -n | uniq -d
```

## Why this matters

A round number is a reference. Rounds 10 to 68 each cite work by round number, and a
reader who follows a citation to round 9 reaches two rows. A renumber moves every later
round, so the repair is a decision rather than an edit.

## The decision this needs

Three shapes repair it, and the user decides which one.

1. Renumber the second row to a new highest round, and leave rounds 10 to 68 alone.
2. Renumber every row after the first round 9, which moves 59 rounds and every citation
   that names one.
3. Merge the two rows into one round 9.

## Acceptance criteria

- [ ] `docs/specs/spec.md` holds one row for each Changelog round number.
- [ ] Every citation that names a round number reaches the round it named before.
- [ ] No file under `ja4plus/` changes and no fingerprint moves.

Crank-Git:
Triaged by the project manager on 2026-08-08. `status:needs-feedback`.

**This needs a decision the project manager cannot make.** `docs/specs/spec.md` is
append-only and every round is a citation target. Renumbering the later rounds moves every
citation that names one, across the specification package, the register causes and the
issue history. Leaving the two rows as they stand keeps every citation correct and leaves
the file holding two rows numbered 9.

**A third form exists:** keep both numbers and add a note beside the pair recording that
round 9 was written twice and that neither row moves.

The count is not load-bearing today. The project manager assigns the next round from the
[...]

Issue #80, opened 2026-08-06
Run the JA4L fingerprinter in the conformance harness


Crank-Git:
## Context

The deviation register that #23 builds holds 162 JA4L entries, which is 62% of the 263
known deviations. Every one of them fails in the same way:

```
badcurveball.pcap stream=0 172.16.225.48:55318 -> ... JA4L-C.1:
expected='149_64' produced=<none>, JA4L-C produced 0 value(s) on this stream
```

**No issue owns this failure.** #30 reads the propagation factor from the FoxIO
hop-count table, and #34 emits one JA4L client fingerprint per connection. Neither one
makes the conformance suite produce a JA4L value, and #30 cannot change a JA4L
fingerprint at all: the fingerprint is `<latency_us>_<ttl>`, and #30 changes
`calculate_distance`, which no fingerprint value reads.

## The cause

The conformance harness never runs the JA4L fingerprinter.

1. `tests/conformance_index.py:29` sets
   `FINGERPRINTER_METHODS = ("JA4", "JA4S", "JA4H", "JA4X", "JA4SSH")`. The comment on
   the line above says "JA4L is absent because ja4plus produces no JA4L value".
2. `index_produced` builds five fingerprinters, and `JA4LFingerprinter` is not one of
   them.

**The comment is wrong.** `ja4plus/fingerprinters/ja4l.py` produces JA4L values today.
Measured on the committed vectors:

| Vector | JA4L values produced |
|---|---|
| `ssh2.pcapng` | 610 |
| `v6.pcap` | 79 |
| `tls-handshake.pcapng` | 20 |
| `tls12.pcap` | 0 |

#12 already recorded that this claim is stale. It survived into the harness.

## The second defect

A JA4L entry carries a connection key of the form
`udp_172.16.224.1:53_172.16.225.48:60983`. `_connection_identity` in
`tests/conformance_index.py:205` reads the form `<address>:<port>-<address>:<port>` and
partitions on the hyphen. The JA4L key holds no hyphen, so the reader returns None and
every JA4L value lands under the unknown stream.

A JA4L fingerprint also carries the method name in the value, as `JA4L-S=21105_64`. The
reference holds `21105_64`. The harness must read the prefix off.

## What to build

- `index_produced` runs `JA4LFingerprinter` and reports `JA4L-C` and `JA4L-S`.
- `_connection_identity` reads the JA4L connection key form.
[...]

PR #375, opened 2026-08-09
Fold the duplicate Changelog round 9 in the specification

Crank-Git:
Repairs #258. The Changelog of `docs/specs/spec.md` carried the number 9 on two rows.

## What changed

**Fold, do not renumber.** The two rows are not two rounds. Line 569 records the shipment of Epic 1 batch 1 and names #80 as a member. Line 573 details #80 on the same date. The detail row is folded into the shipment row, and one row numbered 9 remains. **Every sentence of the folded row is preserved, and nothing is summarized away.**

**No row is renumbered.** A round number is an identifier and not an ordering, and the date column carries the order. Rounds 10 through 124 are all taken, so the repeated number has no free neighbour. Shifting the rows above it would invalidate every citation of a round number, and those citations are load-bearing: round 67 holds the #226 part e decision and is cited from #215 and #226.

`tests/test_specification_changelog.py` is new and holds the invariant.

## Three premises of the ruling were stale, and all three were re-measured

| Premise in the issue | Measurement on the base commit `39ca7e3` |
|---|---|
| The rows are at `spec.md:542` and `:546` | They were at `569` and `573` |
| Round 88 is the highest | The highest was `124`; the handoff said `125`, and that is also wrong |
| `grep -rn "[Rr]ound 9\b"` finds no citation of round 9 | The pattern is too narrow, and one citation exists |

**The citation the pattern misses.** Round 12 closed with the sentence `The rounds below hold two entries numbered 9.` It names the duplicate without the word "round", so `[Rr]ound 9\b` never reaches it. The fold makes that sentence false.

**One scope deviation, and it is deliberate.** The brief puts every other round row out of scope. Round 12 is edited anyway, because the fold is what falsifies its closing sentence — this cleans up the orphan the change itself creates, rather than adding scope. The edit is one sentence and it changes no other fact of the row:

```
-  ... The rounds below hold two entries numbered 9. |
+  ... The rounds below held two entries numbered 9, and #258 folded them into round 9. |
```

The sentence is rewritten rather than deleted, so the record that the duplicate existed survives, and it names no round number, so it needs no assignment at the gate.

## Test-driven development

The two cases were written first and both failed on `f4c30bf`:

```
AssertionError: these round numbers name more than one row: [9]
AssertionError: the Changelog holds 125 rows and its highest round is 124
```

**The cases discriminate.** Reintroducing a duplicate row 9 into the repaired file fails both again with the same two messages. A `MINIMUM_ROWS` floor of 124 fails a regular expression that stops matching the rows, because a parser that reads nothing would otherwise pass on an empty list. A row whose round is the literal `TBD` carries no number and reaches no case.

## How it was tested

| Check | Result |
|---|---|
| `pytest tests/ -m "not spec_validation"` | 2211 passed, 2 skipped, 8 xfailed, 96 subtests passed |
| `pytest tests/ -m spec_validation` | 1531 passed, 143 skipped, **135 xfailed** |
| `ruff format --check ja4plus/ tests/` | 164 files already formatted |
| Coverage | 93%, unchanged — no file under `ja4plus/` changes |
| Register invariant | `tests/foxio_deviations.json` holds 135 keys against 135 xfailed |

No file under `ja4plus/` changes, and no fingerprint moves. `git diff --name-only` against the base reports `CHANGELOG.md`, `docs/specs/spec.md` and `tests/test_specification_changelog.py` alone. `docs/specs/spec.html` is untouched, as the brief requires.

## Changelog

A row is added to **both** `docs/specs/spec.md` and `CHANGELOG.md`, with the round as the literal `TBD`. The project manager assigns the round at the batch gate, in both files.

No external interface is read, so this pull request cites no vendor documentation.


Crank-Git:
## Self-review

Reviewed by lens: fold correctness, table integrity, the new test's discrimination, scope, and the writing standard.

### Correctness of the fold — passes, and the proof is mechanical

The surviving row 9 equals the two base rows concatenated with one space, character for character:

```
merged == a + " " + b exactly       : True
len(merged) == len(a) + len(b) + 1  : True
```

Nothing is dropped, reworded or reordered.

### Only rounds 9 and 12 changed

```
round numbers whose row text changed: [9, 12]
round numbers present in base but absent now: []
round numbers new: []
```

### The regex is exact

Measured over the whole page, not the `## Changelog` section alone:

```
matched rows inside  ## Changelog : 124
matched rows OUTSIDE ## Changelog : 0 []
```

No row of the Terms table, the divergence register or the issue map reaches a case. The three `| TBD |` rows carry no number and reach no case, which is correct until the project manager assigns their rounds.

### Findings accepted and fixed

Six sentences of the new prose failed `.claude/rules/ste.md`. All are fixed in `f1c1a3a`.

| Rule | Text | Repair |
|---|---|---|
| 11, active voice | "Every sentence of the folded row **is preserved**." | "The fold preserves every sentence of the folded row." |
| 11, active voice | "round 67 ... **is cited from** #215 and #226" | "round 67 ..., and #215 and #226 cite it" |
| 1, 31 words | "The two rows are not two rounds: one records ... and the other details ..." | Split into three sentences |
| 1 and 14, 29 words with a gerund subject | "... and **shifting** the rows above it would invalidate ..." | "A renumber of the rows above it would break ..." |
| 1, over 40 words | The run-on that carried the `[Rr]ound 9\b` finding | Split into four sentences |
| 1, 28 words | The `CHANGELOG_ROW` comment | Split into two sentences |

Two further changes came out of the same pass. **"is not an ordering" was an `-ing` noun under rule 14**, and it now reads "it states no order". **"load-bearing" is a metaphor under rule 9**, and it now reads "carry weight".

### Findings rejected, with the reason

**"neighbour" as a British spelling.** The word is gone rather than re-spelled. The user's standing instruction is US English, and `docs/specs/spec.md:679` already holds "neighbouring". Re-spelling that row is out of scope for #258, and adding "neighbor" beside it would create a contradiction inside one file. The new sentence avoids the word: "no free number sits beside the number 9".

### One scope deviation, stated plainly

The brief puts every other round row out of scope. **Round 12 is edited anyway.** Its closing sentence read `The rounds below hold two entries numbered 9.`, and the fold is what makes that sentence false. This removes an orphan the change itself creates rather than widening scope. The sentence is rewritten and not deleted, so the record that the duplicate existed survives, and it names no round number, so it needs no assignment at the gate.

**The ruling's citation search missed this sentence** because `[Rr]ound 9\b` does not match "numbered 9". That is the third stale premise the pull-request body records.

PR #376, opened 2026-08-09
Rule on the six JA4L entries that #34 left open

Crank-Git:
Refs #272.

## What

Six JA4L entries of `tests/foxio_deviations.json` named the closed #34. #272 is now their live owner. Five are declined and recorded. The sixth is a defect of this project, and this pull request repairs it.

## Why the five are declined

**This project was never emitting more than the reference. The reference published nothing to compare.** The comparison is unreachable, not satisfied.

`python/ja4.py:339` runs `delete_keys(['JA4L-S','JA4L-C'], final)` when the generating run names another method, so three reference files publish no JA4L key at all. I re-read that line at the pinned commit rather than cite #200 unread.

A count of the JA4L keys of all 38 expected-output files corroborates the mechanism against the vectors:

```
CVE-2018-6794.pcap.json    NO JA4L KEY
https-connect.pcap.json    NO JA4L KEY
tls-handshake.pcapng.json  NO JA4L KEY
```

The five entries now carry `"decided": true` and `"capability": false`, so each records a value decline. No entry is marked `decided` to make a check pass, which is what #193 refused and #251 refused again.

## Why the sixth is repaired

`ssh2.pcapng/JA4L-S` compares two measured counts. The FoxIO Python file holds 10 JA4L-S values, and `ja4plus` produced 11. **The extra value is a duplicate and not a new reading.** Stream 15, the connection to `184.150.157.177:80`, produced `6252_58` twice where the reference holds it once.

`ja4plus/fingerprinters/ja4l.py` returned the server value on every SYN-ACK. A retransmitted SYN-ACK moves neither point `B` nor the server TTL, so the repeat described no second measurement. `.claude/rules/conformance.md` declines a value that describes nothing.

## The three stop conditions were each measured, and each is false

The ruling on #272 was conditional. I measured before changing any code.

| Condition | Result |
|---|---|
| The repair removes the only case that measures a JA4L-S value | **False.** 142 JA4L values stand after the repair, and `ssh2.pcapng` alone keeps 10. |
| The repair moves a JA4L value on a capture other than `ssh2.pcapng` | **False.** A replay of all 38 committed captures reports exactly 1 moved key. |
| The extra value is an artifact of the same `delete_keys` call | **False.** `ssh2.pcapng.json` publishes 10 JA4L-S keys, and the call deletes the key whole. |

The one moved fingerprint, by name:

```
MOVED ssh2.pcapng (('172.16.225.48', '57380'), ('184.150.157.177', '80'))|JA4L-S
      before= ['6252_58', '6252_58']   after= ['6252_58']
total moved JA4L keys = 1
```

## Test-driven development

`tests/test_ja4l_connection_state.py::test_a_retransmitted_syn_ack_reports_no_second_server_value` came first and failed against the base:

```
At index 1 diff: 'JA4L-S=500_64' != 'JA4L-C=2500_128'
Left contains one more item: 'JA4L-C=2500_128'
```

The case feeds a SYN-ACK that a second SYN-ACK follows, which is input the guard alone rejects. With the guard removed live, both the unit case and `test_the_produced_occurrence_keys_equal_the_reference[ssh2.pcapng-JA4L-S]` go red, so the guard is load-bearing in both directions.

## The register invariant

The key count of `tests/foxio_deviations.json` equals the xfailed count of the conformance suite, before and after.

| | Register keys | Conformance xfailed |
|---|---|---|
| Base | 135 | 135 |
| This branch | 134 | 134 |

The suite reports 1532 passed against 1531 before. The one case that moved from `xfail` to pass is the `ssh2.pcapng` JA4L-S occurrence comparison.

## How tested

Every run cleared each `__pycache__` outside `.venv/`, set `PYTHONDONTWRITEBYTECODE=1`, and confirmed the loaded source with `inspect.getsource`.

- `pytest tests/ -m "not spec_validation"` — 2214 passed, 2 skipped, 8 xfailed, 96 subtests passed
- `pytest tests/ -m spec_validation` — 1532 passed, 143 skipped, 134 xfailed
- `ruff format --check ja4plus/ tests/` — 164 files already formatted
- `mypy --strict ja4plus/` — Success, no issues in 31 source files
- Coverage — 93% before and after. 287 missed statements on both sides, and the total rises 4231 to 4232, so the added line is covered.

## Documentation

Two `Divergence register` rows in `docs/specs/spec.md`, and a `## Changelog` row in both `docs/specs/spec.md` and `CHANGELOG.md`. The round is the literal `TBD`.

Verified against: https://github.com/FoxIO-LLC/ja4/blob/main/python/ja4.py at pinned commit `27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8` (`python/ja4.py:339`, read 2026-08-09).


Crank-Git:
## Self-review: the mechanism corroborated against the vectors, not the source alone

`.claude/rules/conformance.md` states that a defect is proven and never asserted, and that a reading of the source alone is not proof. The `python/ja4.py:339` citation is a source reading, so here is the independent check against the expected-output files.

The method keys each of the three files publishes:

```
CVE-2018-6794.pcap.json    any JA4L key: False   methods: ['JA4H', 'JA4H_ro']
https-connect.pcap.json    any JA4L key: False   methods: ['JA4H', 'JA4H_ro']
tls-handshake.pcapng.json  any JA4L key: False   methods: ['JA4.1', 'JA4.2', 'JA4S', 'JA4S_r',
                                                           'JA4X.1', 'JA4X.2', 'JA4X.3',
                                                           'JA4_o.1', 'JA4_o.2', 'JA4_r.1',
                                                           'JA4_r.2', 'JA4_ro.1', 'JA4_ro.2']
```

Each set is what an `output_types` filter produces, and nothing else explains it. The first two files carry JA4H alone, so the generating run named `ja4h`. The third carries JA4, JA4S and JA4X and no JA4L, so its run named those three. `python/ja4.py:339` deletes `JA4L-S` and `JA4L-C` in each case.

The reading holds for a second reason: 22 of the 38 expected-output files publish a JA4L key. A gap in the reference reader would not spare 22 files and strike exactly these three.

Crank-Git:
## Self-review outcome

Two specialist passes over the diff: correctness, and documentation plus test quality.

The reviewer enumerated the four states of `(B set, A set)` entering the `if syn and ack:` branch and confirmed only the intended one changes.

| B set | A set | Before | After |
|---|---|---|---|
| No | No | sets `B`, returns None | same |
| No | Yes | sets `B`, returns `JA4L-S` | same |
| Yes | No | returns None | same |
| Yes | Yes | returns `JA4L-S`, repeating the first value | **returns None** |

`conn["isns"][source] = sequence` still runs before the guard, so connection reuse still restarts correctly. The reordered-capture path under `if syn and not ack:` is untouched.

**One case is worth recording.** Where a SYN-ACK for a new connection arrives while `A` and `B` are stale, because the capture missed the client SYN, the old code paired the stale `A` with the stale `B` and published a value for the wrong connection. The new code returns nothing. That case is no regression, and the guard improves it.

### Documentation and tests — one finding accepted, one rejected with reason

**Accepted: `docs/specs/foxio/JA4L.md` was stale.** D4 read `Open` and stated that `ssh2.pcapng/JA4L-S` stands unchanged under #34. The register holds that key no longer. D4 and the five-entry item now follow the convention D1 and D3 already use on the page. Commit `75b282d`.

**Accepted: two writing-standard repairs.** Two sentences move to active voice, and `wording` and `ruling` leave as nouns under STE rule 14.

**Rejected: the reading that `reading` is an `-ing` noun to remove.** `docs/specs/spec.md` uses `reading` 122 times and `readings` 29 times as the project term for a settled interpretation. STE rule 7 states one concept, one word, and it forbids synonym rotation. A replacement in one row alone would break that rule and part this row from every other row of the register. The word stays.

The reviewer also confirmed every number in the added prose against the repository: 134 register keys, 5 entries naming #272 with `"decided": true` and `"capability": false`, `ssh2.pcapng/JA4L-S` absent, zero JA4L keys in the three files, and 10 JA4L-S values in `ssh2.pcapng.json`.

### Re-verified after the fixes

- `pytest tests/ -m "not spec_validation"` — 2214 passed, 2 skipped, 8 xfailed
- `pytest tests/ -m spec_validation` — 1532 passed, 143 skipped, 134 xfailed
- `ruff check` clean, `ruff format --check` clean, `mypy --strict` clean
- Coverage 93%, and `ja4plus/fingerprinters/ja4l.py` holds 97% with the same 8 missed statements as the base

PR #377, opened 2026-08-09
Enable the ruff F401 rule and remove the 54 unused imports

Crank-Git:
Refs #297.

## What changed

`pyproject.toml` no longer holds `F401` in the `ignore` list of `[tool.ruff.lint]`. The
comment that named #297 now records the enabled rule. 54 unused imports are gone from 27
files.

## Why the numbers differ from the issue

**Every count the issue recorded is stale, and this branch measured each one again.**

| Measurement | The issue | This branch |
|---|---|---|
| F401 findings | 58 | **54** |
| Files that hold one | 28 | **27** |
| `ja4plus/utils/x509_utils.py` | 5 | **3** |
| I001 findings, per the `ignore` comment | 82 | **76** |
| Branch measured | `issue/47-py-typed` | `batch/362-hygiene-three` |
| `ruff` version | 0.14.5 | **0.16.2** |

`ruff>=0.6` is the pin, so the version floats. The `ignore` comment now states 76 for
I001.

## How each import was classified

**`ruff --fix` decides no removal here, because an unused import is not always unused.**
All 54 were classified before one was removed, against five shapes.

| Shape | Count |
|---|---|
| An import with a side effect | 0 |
| A re-export | 0 |
| A `TYPE_CHECKING` import | 0 |
| A test helper | 0 |
| Genuinely unused | **54** |

The three readings that produced that grouping:

1. **No re-export.** No consumer imports a flagged name from a flagged module. Every
   consumer takes a fingerprinter class or a `generate_*` function.
   `parse_http_request` is the one flagged name another module reads, and every reader
   takes it from `ja4plus/utils/http_utils.py` rather than from
   `ja4plus/fingerprinters/ja4h.py`. No file that holds a finding declares `__all__` or
   a `TYPE_CHECKING` block.
2. **No lost side effect.** No import statement under `ja4plus/` loses every name, so
   every `scapy.all` line keeps `Packet`. `tests/test_ja4x.py` is the one file whose
   whole `scapy.all` statement goes, and `ja4plus.fingerprinters.ja4x` still imports
   `scapy.all` for it. `ja4plus/utils/ssh_utils.py` holds constants and functions alone,
   with no registration, so the removed block in `tests/test_ja4ssh_deep.py` is inert.
3. **No indirect read.** Each flagged name was searched across its own file. Every other
   occurrence is prose in a comment or docstring, an attribute access
   (`self.fp.get_raw_fingerprint`), or a `patch("sys.argv", ...)` string that resolves
   the module by name and needs no local `import sys`.

## The failing gate, then the passing gate

`ruff check ja4plus/ tests/` with `F401` absent from the `ignore` list, before the
removals:

```
tests/test_watch_statistics.py:19:8: F401 [*] `sys` imported but unused
Found 54 errors.
[*] 54 fixable with the `--fix` option.
```

After the removals:

```
All checks passed!
```

## Verification

Every run cleared every `__pycache__` outside `.venv/` and set
`PYTHONDONTWRITEBYTECODE=1`, because a stale `.pyc` lets a removed import keep working.

| Gate | Base `batch/362-hygiene-three` | This branch |
|---|---|---|
| `pytest tests/ -m "not spec_validation"` | 2212 passed, 2 skipped, 1809 deselected, 8 xfailed, 96 subtests | **identical** |
| `pytest tests/ -m spec_validation` | 1531 passed, 143 skipped, 2222 deselected, 135 xfailed | **identical** |
| `ruff check ja4plus/ tests/` | passes, `F401` ignored | passes, **`F401` enforced** |
| `ruff format --check ja4plus/ tests/` | 164 files already formatted | 164 files already formatted |
| `mypy --strict ja4plus/` | no issues in 31 source files | no issues in 31 source files |
| `coverage report --format=total` | 93 | 93 |
| Register keys against xfailed | 135 / 135 | 135 / 135 |

**No fingerprint moves.** 15 of the 54 findings sit under `ja4plus/fingerprinters/` and
`ja4plus/utils/`, across 7 files. The conformance suite reports identical counts before
the change and after it, and that is the proof each of those removals is inert.

Coverage holds at 93. The missed line count stays at 287, and the statement count falls
from 4231 to 4226, because five removed import lines were covered statements.

## Documentation

`docs/specs/spec.md` and `CHANGELOG.md` each carry a new row, with the round as the
literal `TBD`. `tests/test_specification_changelog.py` ignores a `TBD` row, and both of
its cases pass.


Crank-Git:
## Self-review — correctness lens

An independent reviewer read the whole diff against four checks, and it found nothing.
Its readings corroborate the classification in the pull-request body, and two of them go
past it.

1. **No removed name survives in its own file.** Every other occurrence is a comment, a
   docstring, or a definition elsewhere. `unittest.mock.patch("sys.argv", ...)` in
   `tests/test_cli.py` and `tests/test_watch_statistics.py` keeps working after
   `import sys` goes, because `mock.patch` resolves the target string through its own
   `importlib.import_module` and reads no local import.
2. **No side-effect import disappears.** Every `scapy.all` statement in a touched file
   keeps `Packet`. `tests/test_ja4x.py` is the one file whose whole `scapy.all`
   statement goes, and `ja4plus/fingerprinters/ja4x.py` still imports `scapy.all` for it.
3. **No re-export breaks.** Every consumer of `parse_http_request`, `DHCP_SKIP_OPTIONS`,
   `ExtensionOID`, `is_ssh_packet`, `parse_ssh_packet` and `extract_hassh` imports the
   name from its owning module, and none imports it from a touched file.
4. **No runtime annotation read.** The repository holds zero calls to `get_type_hints`
   and zero to `get_args`, so removing `typing.Any` from `ja4plus/fingerprinters/ja4ts.py`
   is safe under `from __future__ import annotations`.

## Two further readings

**The conformance proof is two-sided.** `tests/foxio_deviations.py` marks each registered
case `xfail(strict=True)`. A fingerprint that moves away from its expected value fails one
of the 1531 passing cases. A fingerprint that moves into a match turns one of the 135
strict xfails into an XPASS, and that fails too. Both directions are measured.

**Every module imports clean from a cold cache.** A run that cleared every `__pycache__`
outside `.venv/` and set `PYTHONDONTWRITEBYTECODE=1` imported all 30 submodules of
`ja4plus` with no error. `ja4plus analyze tests/foxio_vectors/tls12.pcap` then produced
`t13d1715h2_5b57614c22b0_3d5424432f57`.

**The `sys.path.insert` invariant of the `I001` comment holds.** `tests/test_ja4.py` keeps
`sys` and `os` above the path insertion at line 6, and every project import stays below
it.

No finding needs a fix.

Issue #345, opened 2026-08-08
Correct the decided-entry count in the marker rule


Crank-Git:
Found by #341 while it read `tests/foxio_deviations.py`. #341 changed neither line, because
the count sits in a different section of the docstring and no part of #341 touches the
marker rule.

## The measurement

**The `## The marker rule` section of `tests/foxio_deviations.py` states a decided-entry
count that is one too low.** It reads:

> The rule runs in one direction. An entry that cites no decision may still carry the
> marker, which 38 of the 128 decided entries do.

The register holds **129** decided entries today, of 135 keys.

```
decided entries: 129
decided entries that cite no decision: 38
```

**The 38 is correct**, and so is the sentence that follows it: the 38 are the 34 entries
of #138 and 4 of the 5 entries of #151. The denominator alone is stale.

## Why this is worth a line

A count in prose that nothing measures goes stale on the next entry, and this one already
has. The register gained an entry after the sentence was written and no case caught it.

**Consider a case rather than a corrected number.** `tests/test_foxio_deviations.py`
already holds `test_the_register_holds_six_undecided_entries`, which is the same shape of
state gate. A case that reads the count out of the register makes the next drift fail
instead of sitting in a docstring.

## Acceptance criteria

- [ ] The `## The marker rule` section of `tests/foxio_deviations.py` states the count the
      register holds.
- [ ] A case fails when the decided-entry count moves and the prose does not.
- [ ] No fingerprint moves and no file under `ja4plus/` changes.

## Links

- Found by: #341
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)

Crank-Git:
Sub-merged into `batch/362-hygiene-three` as `60b179c`. Labelled `status:batched`. The issue closes when the batch pull request merges into `dev`.

**The issue asked for a corrected number and got a case instead, which is what it recommended.** `TestTheMarkerRuleCounts` reads both counts out of the register, so the next drift fails in the suite rather than sitting in a docstring.

## Every number in this issue was stale, including the one the brief supplied

| Measurement | The issue recorded | Measured on the base |
|---|---|---|
| Register keys | 135 | **134** |
| Decided entries | 129 | **134** |
| Undecided entries | 6 | **0** |
| Entries carrying the marker with no citation | 38 | **38** |

**The numerator was right and the denominator was stale**, which is exactly what the issue predicted would happen again.

[...]

PR #379, opened 2026-08-09
Hold the marker rule counts in a case

Crank-Git:
Closes nothing on its own. This is a batch member of #361, and it references #345.

## What changed

The `## The marker rule` section of `tests/foxio_deviations.py` read `38 of the 128
decided entries`. The register holds **134 decided entries of 134 keys** on the base
commit `f238c15`. The denominator went stale twice, because nothing measured it.

`TestTheMarkerRuleCounts` in `tests/test_foxio_deviations.py` now reads the two counts
out of the prose with the pattern `which (\d+) of the (\d+) decided entries do` and
measures them against the register. `measured_marker_rule_counts` reads a citation with
`DECISION_CITATION`, the same reader `unmarked_decisions` uses, so one definition of a
citation serves both gates.

## Why

A count in prose that nothing measures goes stale on the next entry. The correction of
the number is the smaller half. The rule is now a condition a case tests.

## Re-measured, and the issue body was stale

Every number was measured on the base commit and none was carried from the issue body.

| Fact | Issue body | Measured on `f238c15` |
|---|---|---|
| Register keys | 135 | 134 |
| Decided entries | 129 | 134 |
| Decided entries that cite no decision | 38 | 38 |
| Undecided entries | 6 | 0 |

#272 sub-merged into the base branch and moved the register: it removed
`ssh2.pcapng/JA4L-S` and it decided the five method filter entries. The numerator of 38
was correct, and the 38 are still the 34 entries of #138 and 4 of the 5 entries of #151.

## A finding, reported and not repaired

`test_the_register_holds_six_undecided_entries` no longer exists. **#272 already treated
it**: it reads `test_the_register_holds_no_undecided_entry` today and asserts zero. The
case is near vacuous now and it is not vacuous, because a new undecided entry fails it.
This branch changes no case of #272. The treatment of a state gate that asserts an empty
set is a question for the user.

## How this was tested

The case came first and it failed on the unchanged prose with `assert 128 == 134`.

Three live mutations discriminate in both directions. Each one was reverted, and the
eight cases pass clean before and after each one.

| Mutation | Result |
|---|---|
| The register gains a decided entry whose cause names `Changelog round 66` | `assert 134 == 135`, 1 failed |
| The register gains a decided entry whose cause cites no decision | `assert 134 == 135` and `assert 38 == 39`, 2 failed |
| The prose alone moves to `133` | `assert 133 == 134`, 1 failed |

`git status` reports `tests/foxio_deviations.json` unchanged.

## The floor

- `pytest tests/ -m "not spec_validation"`: 2223 passed, 2 skipped, 8 xfailed, 96 subtests passed
- `pytest tests/ -m spec_validation`: 1532 passed, 143 skipped, **134 xfailed** against the 134 keys of `tests/foxio_deviations.json`
- `ruff check ja4plus/ tests/`: All checks passed
- `ruff format --check ja4plus/ tests/`: 164 files already formatted
- `mypy --strict ja4plus/`: no issues in 31 source files

No file under `ja4plus/` changes, so no fingerprint moves. This branch reads no FoxIO
material and adopts no vector.

## Documentation

`docs/specs/spec.md` and `CHANGELOG.md` each gain one row with the round as the literal
`TBD`. The project manager assigns the round at the batch gate.


Crank-Git:
## Self-review

Two Sonnet reviewers read the diff: one for correctness, one for the writing standard and scope.

### Correctness — one finding, accepted

`test_the_measurement_rises_when_the_register_gains_a_decided_entry` asserted a rise of one. A rise of one reads the same on a measurement that counts every entry and ignores the marker, so the case could not discriminate a dropped `decided` filter. The case now states both counts against a register that holds an undecided entry, and its docstring records the reason.

The reviewer independently reproduced the three drift points from git history: `eb7a2fe` holds 128 decided of 137 keys, #341 holds 129 of 135, and the base holds 134 of 134. The claim "the denominator went stale twice" is therefore measured, not asserted.

No tautological or vacuously passing case was found. `stated_marker_rule_counts` reads `tests/foxio_deviations.py` and `measured_marker_rule_counts` reads `tests/foxio_deviations.json`, so the two sides of every comparison come from different files.

### Writing standard — findings accepted

The `docs/specs/spec.md` round row carried passive voice and six sentences over the 25-word cap. Each one is split or restated in the active voice. `state gate` was a word the project does not use, and it now reads `gate`. The `CHANGELOG.md` entry lost its one long sentence and its `-ing` noun.

### Scope — clean

The diff touches `tests/foxio_deviations.py`, `tests/test_foxio_deviations.py`, `docs/specs/spec.md` and `CHANGELOG.md`. `tests/foxio_deviations.json` is unchanged and no file under `ja4plus/` changed.

## The base branch moved, and this branch carries it

#377 landed on `batch/362-hygiene-three` after this branch forked. It is merged in here. `docs/specs/spec.md` conflicted, because both rows append to the Changelog table; the resolution keeps both rows and renumbers neither.

**#377 turns the `F401` rule on.** The two changed test files pass `ruff check --select F401`.

## The floor, re-run on the merged tree

- `pytest tests/ -m "not spec_validation"`: 2223 passed, 2 skipped, 8 xfailed, 96 subtests passed
- `pytest tests/ -m spec_validation`: 1532 passed, 143 skipped, **134 xfailed** against 134 register keys
- `ruff check`: All checks passed. `ruff format --check`: 164 files already formatted
- `mypy --strict ja4plus/`: no issues in 31 source files
- Coverage: **93%**, 287 missed of 4227 statements. It matches the 93% and the 287 that #297 records on the base.

The discrimination proof was re-run on the merged tree: a decided entry that cites no decision fails both state cases with `assert 134 == 135` and `assert 38 == 39`, and the restore returns 8 passed.

Issue #316, opened 2026-08-08
Name the errors the two X.509 byte reader handlers expect


Crank-Git:
Found by the worker of #314 while it removed `extract_certificate_info`. This issue is
untriaged, and the project manager owns the triage.

## What remains

`extract_certificate_from_bytes` holds two plain `except Exception` handlers.

- `ja4plus/utils/x509_utils.py:148` holds the ASN.1 parse.
- `ja4plus/utils/x509_utils.py:154` holds the whole body of the function.

`CLAUDE.md` states the rule: a fingerprinter catches the parse errors it expects, and it
does not catch bare `Exception`.

## Why neither is repaired yet

#294 targeted the deceptive `(ValueError, TypeError, Exception)` form, and each of these
wrote the plain form, so #294 narrowed neither. #314 removed the two plain handlers that
sat inside `extract_certificate_info`, and these two are the last two in the module.

The two carry different cost.

1. **The inner handler is small work.** It holds one call,
   `x509.load_der_x509_certificate(candidate, default_backend())`. The `cryptography`
   documentation states that the call raises `ValueError`. No later call reads the
   certificate, so the three `cryptography` errors that `Certificate.extensions` and
   `Certificate.version` raise cannot reach it.
   `test_the_asn1_reader_returns_nothing_for_a_sequence_that_is_no_certificate` already
   reaches this handler with a real `ValueError`.
2. **The outer handler needs a measurement first.** It holds the whole body, and
   `CLAUDE.md` rule 2 states that a parser which cannot read a packet returns nothing and
   does not raise. A narrow list that omits an error the body raises turns a silent miss
   into a raise that reaches a caller. Name every error the body raises before you narrow
   it, and write the case that proves each one.

## Acceptance criteria

- [ ] `ja4plus/utils/x509_utils.py` holds no plain `except Exception`.
- [ ] Each narrow list names only errors the `cryptography` documentation states for the
      calls inside its block, and the issue cites the documentation URL and the version.
- [ ] One case reaches each handler with a named error, and the reader returns nothing.
- [ ] One case sends each handler an error the narrow list omits, and the handler holds
      it no longer. That case fails on the wide form.
- [ ] The five hostile inputs of `tests/test_ja4x_named_exceptions.py` still return
      nothing and raise nothing.
- [ ] The conformance suite is unchanged and the register invariant holds.

## Links

- Found by: #314
- Earlier work on the same shape: #294
[...]

PR #381, opened 2026-08-09
Narrow the four parser handlers and document the ten wide ones

Crank-Git:
Closes nothing on its own. This pull request builds #319 for batch #361, and it is rebased on `60b179c`.

It also repairs #382, a pre-existing defect the self-review found. The project manager ruled that #319 absorbs it.

## What this changes

`grep -rn "except Exception" ja4plus/` reads **14 sites across five files** on the base
commit. **#319 is a reading of all fourteen, and the narrowing is what the reading
justifies.** The reading places every site in one of the three groups the issue names.

| Group | Sites | What happens |
|---|---|---|
| 1 — a parser reading hostile input | 4 | Narrowed, plus the #382 repair. |
| 2 — the caller receives the error | 3 | Documented at the site. |
| 3 — top-level reporting | 7 | Documented at the site. |

### Group 1 — narrowed

| Site | Names |
|---|---|
| `ja4plus/utils/quic_utils.py` `decrypt_quic_initial_crypto` | `IndexError`, `ValueError`, `InvalidTag` |
| `ja4plus/utils/quic_utils.py` `decrypt_quic_server_initial_crypto` | `IndexError`, `ValueError`, `InvalidTag` |
| `ja4plus/utils/quic_utils.py` `parse_quic_initial` | `IndexError`, `ValueError`, `InvalidTag` |
| `ja4plus/__init__.py` `compute_ja4x_from_pem` | `ValueError`, `x509.InvalidVersion` |

Each one returns nothing for hostile input, exactly as it did before. **No `return None`
became a raise.**

**The issue put `ja4plus/__init__.py` in group 3, and the reading moves it to group 1.**
That function reports nothing to an operator and it exits nothing. It reads a
caller-supplied certificate and returns nothing, which is the shape `CLAUDE.md` binds.

**Group 1 holds four sites, which meets the sizing bound of the issue.** This pull
request narrows those four and stops.

### Groups 2 and 3 — documented, not narrowed

`ja4plus/processor.py` holds three sites. #45 made the processor hand each parse failure
to its caller, so a clause that named a list would drop the failure the method exists to
report. `ja4plus/cli.py` holds five and `ja4plus/ja4db.py` holds two. Each one reports a
failure and returns, and a traceback would tell the operator less. **Each of the ten now
carries a comment that states why the wide catch is the design.**

## Why these names, and not others

**The narrowing rests on measurement, not on documentation.**

- A fuzz of 90000 datagrams through the three QUIC try-bodies returned 0 times. It raised
  `IndexError` 81153 times, `InvalidTag` 8598 times and `ValueError` 249 times.
- **`InvalidTag` inherits `Exception` and not `ValueError`.** That is the shape #316 found
  in `InvalidVersion`, so a list of `ValueError` alone drops the common case.
- **The random fuzz never decrypts, so it never reaches the read behind the decryption.**
  A second fuzz of 40005 attacker-controlled payloads measured that part, and it raised
  `IndexError` 15 times and nothing else.
- **`struct.error` is unreachable at `parse_quic_initial`, and a bound proves it.**
  `MAXIMUM_CRYPTO_BUFFER_BYTES` is 16384, so the reassembled ClientHello never reaches
  the 65536 that the `!H` length field would refuse.
- **The PEM loader raises `InvalidVersion` too, measured rather than inferred.** A
  certificate whose version field carries 18 raises `18 is not a valid X509 version`. A
  fuzz of 8003 malformed PEM inputs raised `ValueError` alone, so only the constructed
  case reaches the second name.

## How this was tested

`tests/test_bare_exception_reading.py` holds 18 cases. The four source cases came first
and failed against the base commit.

- **Each named error discriminates.** Eleven drops of one name from one clause each turn
  at least one case red. Every reversal ran live, and the file returned to its narrowed
  form after each drop.
- **Each narrowed caller returns nothing for hostile input, over more than one input.**
  6000 datagrams reach the three QUIC readers and 200 inputs reach the PEM reader.
- **The rule is a condition a case tests.** `test_a_group_one_reader_names_no_bare_exception`
  reads the loaded source of all four readers, so a stale `.pyc` turns it red.
- **Four group 2 and group 3 cases assert that the caller returns rather than raises.**

**`ja4plus/cli.py:619` carries no case, and the comment at the site records why.** This
host grants no capture socket, so no case reaches the report path of a live capture.

**1734 real UDP payloads of the committed captures run through the three narrowed readers,
0 escape, and 51 genuine QUIC Initials still produce a result.**

## #382 absorbed — the escape is repaired here

The self-review found that **`decrypt_quic_initial_crypto` and
`decrypt_quic_server_initial_crypto` call `parse_crypto_frames` outside the handler**, and
that the CRYPTO branch of that reader guarded no index. A 76-byte datagram that
authenticates, and whose plaintext ends on a CRYPTO frame type byte, made the first reader
raise `IndexError`.

**That is the failure `CLAUDE.md` rule 2 forbids, and it is remotely triggerable by
construction.** RFC 9001 Section 5.2 derives the Initial keys from the destination
connection ID, and the Initial packet carries that value in the clear, so any sender
builds a datagram that decrypts. I filed it as #382 and reported the acceptance criterion
unmet; **the project manager ruled that #319 absorbs the repair**, and this branch carries
it.

### The reading decided the shape, not a preference

`parse_crypto_frames` has **three callers**: `extract_crypto_frames` (protected, because
`parse_quic_initial` calls it inside its own try), `decrypt_quic_initial_crypto`
(unprotected) and `decrypt_quic_server_initial_crypto` (unprotected). **A guard inside the
reader repairs all three; moving one `return` repairs one.** The guard applied is the one
the ACK branch of the same function already holds, so the repair adds no new idiom.

### What the repair proves

- **The exact 76-byte datagram is a committed case.** `build_authenticating_initial` in
  `tests/test_bare_exception_reading.py` constructs it, and a comment names the
  construction, because no fuzz will find that datagram again.
- **The reversal is proven live.** With the guard removed the two new cases fail
  (`2 failed, 18 passed`); with it restored all 20 pass. `inspect.getsource` reports the
  guard in the loaded CRYPTO branch, with `PYTHONDONTWRITEBYTECODE=1` and every
  `__pycache__` outside `.venv/` cleared.
- **Both fuzz campaigns re-run clean.** The 90000 datagrams still return 0 times and raise
  only the three named errors. **The 40005 decrypted payloads fall from 15 escaping
  `IndexError` to 0** — the guard closed real escapes beyond the one reported datagram.
- **The repair silences no packet the package reads today.** The 1734 real UDP payloads of
  the committed captures still produce **51 genuine QUIC Initials** through both the client
  reader and the parser, and 0 escape.
- A second case asserts the reader **keeps every complete frame that precedes a truncated
  one**, so the guard discards nothing.

## Verification

| Check | Base | This branch |
|---|---|---|
| `pytest -m "not spec_validation"` | 2223 passed, 2 skipped, 8 xfailed, 96 subtests | **2243 passed**, 2 skipped, 8 xfailed, 96 subtests |
| `pytest -m spec_validation` | 1532 passed, 143 skipped, **134 xfailed** | 1532 passed, 143 skipped, **134 xfailed** |
| Register keys | 134 | 134 |
| `ruff format --check` | clean | clean |
| `mypy --strict ja4plus/` | clean | clean |
| Coverage | 93% | **94%**, missed falls 287 to 274 |

**No fingerprint moves.** The conformance cases are `xfail(strict=True)`, so a fingerprint
that moved in either direction would fail the run. Identical counts on both sides is the
proof that none did.

Verified against: https://cryptography.io/en/latest/exceptions/ (`cryptography` 46, retrieved 2026-08-09), corroborated by measuring `__mro__` on the installed package.

## Changelog

A `## Changelog` row is added to both `docs/specs/spec.md` and `CHANGELOG.md`, with the
round as the literal `TBD`. The project manager assigns the round at the batch gate.


Crank-Git:
## Self-review

Two Sonnet lenses read the diff: correctness, and the writing standard. Both reported.

### Correctness — one real finding, filed as #382 and not repaired here

**`decrypt_quic_initial_crypto` and `decrypt_quic_server_initial_crypto` call
`parse_crypto_frames` outside the handler**, and that call guards no index in its CRYPTO
branch. A plaintext that ends on the frame type byte `0x06` leaves no byte for the varint,
and `_decode_varint` raises `IndexError` past the handler.

**I reproduced it rather than accepting the report.** The QUIC Initial keys derive from the
connection ID the datagram carries in the clear, so any sender can build a datagram that
authenticates. A 76-byte datagram whose plaintext is 40 PADDING frames followed by `0x06`
gives:

```
decrypt_quic_initial_crypto: RAISED builtins.IndexError: index out of range
parse_quic_initial: returned None
```

**The defect predates this pull request, and I measured that too.** Loading
`ja4plus/utils/quic_utils.py` from the base commit `60b179c` reports the clause
`except Exception as e:` and raises the same `IndexError` for the same datagram. The wide
catch never covered that call, because the call sits outside the try. **This change
therefore causes no regression.**

**Why no fuzz found it.** A random datagram never authenticates, so neither the 90000
fuzzed datagrams nor the 1734 real payloads of the committed captures reach the code
behind the decryption. A case must build a datagram that decrypts.

**I did not repair it.** The repair moves a line out of the function's current shape, and
#319 scopes this work to "Narrow the handler. Do not tidy the function." The
out-of-scope rule says file, do not fix. #382 records the defect, the builder, and the
acceptance criteria.

**I corrected the claim this pull request made about it.** The module docstring of
`tests/test_bare_exception_reading.py` read "returns nothing and raises nothing for
hostile input". It now scopes the claim to input that reaches the handler, and it carries
a warning that names #382.

The lens found no other issue: no type escapes any of the four narrowed clauses, no
narrowed clause changed a return value or the control flow, and no assertion in the new
case file is vacuous.

### Writing standard — four corrections applied

- `narrowing` used as a noun three times in the `docs/specs/spec.md` row. Rule 14 bars an
  `-ing` form as a noun, `narrowing` is not in the `## Terms` table, and all three uses
  were mine. Reworded.
- `**Top-level reporting.**` as a list heading in the test module docstring. Reworded to
  `**A top-level report.**`
- One 32-word sentence in `CHANGELOG.md` and one 26-word sentence in `docs/specs/spec.md`,
  both over the limit. Split.
- One `-ing`-as-noun in a case docstring. Reworded.

The lens found the test names, the docstrings, the code comments, and every numeric claim
clean. It independently checked the pre-change line numbers, the 134-key register count,
`MAXIMUM_CRYPTO_BUFFER_BYTES = 16384`, and that the fuzz tallies sum to 90000.

### Rejected nothing

Both lenses' findings were real. I acted on all of them.

Crank-Git:
## #382 absorbed, per the project manager's ruling

The scope of #319 is widened for this one defect, on the record. Commit `4902746` carries
the repair.

### The shape, decided by the reading and not by preference

`parse_crypto_frames` has **three callers**:

| Caller | Calls it |
|---|---|
| `extract_crypto_frames` (`quic_utils.py:231`) | inside `parse_quic_initial`'s try — protected |
| `decrypt_quic_initial_crypto` (`quic_utils.py:391`) | outside the handler — unprotected |
| `decrypt_quic_server_initial_crypto` (`quic_utils.py:483`) | outside the handler — unprotected |

**A guard inside the reader repairs all three. Moving one `return` repairs one.** So the
guard goes in `parse_crypto_frames`, in its CRYPTO branch, and it is **the same
`except (IndexError, ValueError): break` that the ACK branch of the same function already
holds.** The repair adds no idiom the module did not already use, and it touches nothing
else.

### The five proofs the ruling asked for

**1. The exact 76-byte datagram is a committed case.** `build_authenticating_initial`
constructs it and a comment names the construction: 40 PADDING frames, then the single
byte `0x06`, encrypted under keys derived from an 8-byte connection ID. Before the repair:

```
decrypt_quic_initial_crypto: RAISED builtins.IndexError: index out of range
```

After:

```
decrypt_quic_initial_crypto: returned ([], b'\x01\x02\x03\x04\x05\x06\x07\x08')
```

**2. The reversal, live.** Guard present: `20 passed`. Guard removed: `2 failed, 18
passed`. Guard restored: `20 passed`. `inspect.getsource` on the loaded
`parse_crypto_frames` reports `CRYPTO branch holds a guard: True`, with
`PYTHONDONTWRITEBYTECODE=1` set and every `__pycache__` outside `.venv/` cleared before
each run.

**3. Both fuzz campaigns re-run.** The 90000 datagrams: `returned: 0`, `IndexError 81153`,
`InvalidTag 8598`, `ValueError 249` — unchanged, nothing escapes. **The 40005 decrypted
payloads: `returned: 40005`, and the 15 escaping `IndexError` of the earlier run are gone.**
The guard closed real escapes beyond the single datagram #382 reported.

**4. The 1734 real payloads re-run.** `escaped: 0`, and
`produced a result: {'client': 51, 'parse': 51}` — **the 51 genuine QUIC Initials still
produce their results. The repair silences no packet the package reads today.**

**5. The conformance suite is unchanged.** 1532 passed, 143 skipped, **134 xfailed**
against **134** register keys. No fingerprint moved.

### Also in this commit

- A second case, `test_the_frame_reader_returns_the_frames_it_read_before_a_truncated_one`,
  asserts the reader **keeps every complete frame that precedes a truncated one**. The
  guard stops the read; it discards nothing already held.
- **The test module docstring is widened back.** It states the full claim — every group 1
  reader returns nothing and raises nothing for hostile input — and the #382 caveat is
  gone, because the escape is gone.
- Both Changelog rows extended, round still the literal `TBD`.

### Counts

| Check | Base | This branch |
|---|---|---|
| Unit | 2223 passed, 8 xfailed, 96 subtests | **2243 passed**, 8 xfailed, 96 subtests |
| Conformance | 1532 passed, 143 skipped, 134 xfailed | identical |
| Register keys | 134 | 134 |
| Coverage | 93% | **94%**, missed falls 287 to 274 |
| ruff / format / mypy --strict | clean | clean |

**Acceptance criterion 4 of #319 is now met**, and it is the criterion this repair exists
to satisfy.

Issue #62, opened 2026-08-06
Correct the README coverage claim and state the contracts


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/08-documentation.md`](docs/specs/features/08-documentation.md):

> FR-documentation-1 — The README states which FoxIO methods the project implements and which it does not.

> FR-documentation-2 — The README states the concurrency contract.

> FR-documentation-3 — The README states the memory bounds and their defaults.

> FR-documentation-14 — The documentation states that FoxIO owns the JA4+ standard and that this project is an independent implementation.

## What to build

The README says the project implements "all ten JA4+ methods". FoxIO documents twelve: JA4, JA4S, JA4H, JA4L, JA4LS, JA4X, JA4SSH, JA4T, JA4TS, JA4D, JA4D6 and JA4TScan. This project implements ten and does not implement JA4TScan, which is a permanent non-goal because it sends crafted packets rather than reading traffic.

The README today says nothing about thread safety and nothing about memory growth. An integrator who shares a processor between threads gets no warning.

Verified against https://github.com/FoxIO-LLC/ja4 (retrieved 2026-08-06).

## Acceptance criteria

- [ ] The README method table lists twelve FoxIO methods and marks JA4TScan as not implemented.
- [ ] The README no longer claims that the project implements all JA4+ methods.
- [ ] The README states, in one paragraph, whether a processor may be shared between threads.
- [ ] The README states the default maximum connection count and the default maximum connection age.
- [ ] The README states that FoxIO owns the JA4+ standard and that this project is an independent implementation.

## Links

- Feature: [`docs/specs/features/08-documentation.md`](docs/specs/features/08-documentation.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #19

Crank-Git:
Sub-merged into `epic/19-documentation` as `2eba62b`. Labelled `status:batched`. The issue closes when the Epic 8 pull request merges into `dev`.

**Thirty-seven cases hold the README's claims against the code, and the strongest of them reads the code rather than the README twice.**

## The count in this issue is wrong, and the worker built to the right one

**The issue asks for ten of twelve. The number is eleven.**

[...]

Issue #19, opened 2026-08-06
Epic 8: Documentation


Crank-Git:
## Goal

Make every published statement about `ja4plus` true and testable.

## Exit criteria

The documentation site builds and publishes. Every code sample runs in continuous integration. The README states the concurrency contract and the real method coverage.

## Spec

- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Feature: [`docs/specs/features/08-documentation.md`](docs/specs/features/08-documentation.md)

Depends on: #17

## Sub-issues

- [x] #62
- [x] #63
- [x] #64
- [x] #65
- [x] #66

Crank-Git:
**Epic 8 is complete and merged into `dev` as `2702d10`. All nine of its issues are closed, and the run was green on the first attempt across all eleven jobs**, including the two the epic itself added: `build`, which installs the committed documentation pins into an empty environment and builds the site, and `samples`, which runs every code sample.

`dev` now holds Changelog **rounds 1 to 144, contiguous, no duplicate, and no `TBD` in either file.**

| Round | Issue |
|---|---|
| 136 | #62 the README states the coverage, the contracts and the bounds |
| 137 | #64 the site builds, with an API reference from the docstrings |
| 138 | #63 every sample and every example script runs |
| 139 | #65 eleven method pages, the schema and the migration page |
| 140 | #391 the pins build the site, proven from a clean environment |
| 141 | #395 the two unrecorded breaking changes |
| 142 | #66 the publish workflow and the 1.0.0 release notes |
| 143 | #399, #401 and #403 the migration page answers to the record |
| 144 | #389 the memory control reads what the host cannot move |

## Four issues were adopted, and every one was a defect this epic found in itself

**#391, #399, #401 and #403 were defects in the epic's own output**, each found by a later member reading an earlier one. **#389 was adopted because a control that fails under memory pressure turns a single continuous integration run into a coin toss.**

**The five members alone would have shipped a site that does not build from its own pins, a sample harness that silently skipped whole pages, a migration page citing the wrong round, and release notes stating something false.** Every one of them was caught by a worker reading, not by a suite.

## The three lessons this epic paid for

[...]

PR #386, opened 2026-08-09
Correct the README coverage claim and state the contracts

Crank-Git:
Closes nothing on its own. Refs #62. Part of #19.

## What

The README claimed **"all ten JA4+ methods"**. FoxIO publishes twelve, and this project
implements **eleven** of them.

The claim was false in two directions at once:

- The project builds no **JA4TScan**, so the coverage claim overstated.
- The project builds **JA4LS**, and the table never named it, so the table understated.

## Why the count of ten is wrong, and where it comes from

`JA4LFingerprinter` writes both `JA4L-C=` (`ja4plus/fingerprinters/ja4l.py:482`) and
`JA4L-S=` (`ja4plus/fingerprinters/ja4l.py:446`). Ten fingerprinter classes therefore
carry eleven FoxIO methods, and **the count of ten is a count of classes read as a count
of methods.**

`docs/specs/spec.md:42` already states the correct number:

> FoxIO publishes twelve methods. **`ja4plus` implements eleven of them.**

**The issue body of #62 and `docs/specs/features/08-documentation.md:14-16` each state
"ten of the twelve", and that number is wrong** — twelve less one decline is eleven.
`docs/specs/spec.html:126` carries the same stale count. This pull request changes none
of the three, because each sits outside the scope of #62. The project manager decides
whether to correct them.

## What changed

`README.md`:

- The method table holds **twelve rows** and an `Implemented` column. JA4TScan alone
  reads `No`, and a paragraph states why the project declines it.
- The opening claim reads "FoxIO publishes twelve JA4+ methods, and this project
  implements eleven of them."
- A new `#### The default bounds` table states the four defaults.
- The memory ceiling moved into its own `#### The memory ceiling` heading. Its number is
  unchanged.
- The opening states "FoxIO owns the JA4+ standard".

`tests/test_readme_contracts.py` is new. `CHANGELOG.md` and `docs/specs/spec.md` each
carry one `## Changelog` row reading the literal `TBD`.

## Every number was measured today

| Claim | Measured from | Value |
|---|---|---|
| The FoxIO method count | `technical_details/README.md:5-16` at the pinned commit | 12 |
| The methods this project builds | `ja4plus/fingerprinters/`, `ja4plus/__init__.py` | 11, across 10 classes |
| `__all__` entries | `.venv/bin/python -c "import ja4plus; print(len(ja4plus.__all__))"` | 25 |
| The maximum entry count of a state table | `ja4plus/utils/state_table.py:46` | 10000 |
| The maximum age of a state table entry | `ja4plus/utils/state_table.py:52` | 600 seconds |
| The maximum connection count of the monitor | `ja4plus/watch.py:77` | 10000 |
| The maximum connection age of the monitor | `ja4plus/watch.py:83` | 300.0 seconds |
| The memory ceiling | `tests/test_memory_bounds.py:83` | 512.0 MiB |
| Register keys | `tests/foxio_deviations.json` | 134 |
| Coverage | `pytest --cov=ja4plus` on the base and on this branch | 94% both, 274 missed both |

**The 512 MiB ceiling is quoted and not re-measured.** #279 read 383.47, 388.25, 392.05
and 394.94 MiB across four runs. No earlier figure appears anywhere in this change.

## A claim is a condition, not a sentence

`tests/test_readme_contracts.py` holds fifteen case functions, 35 collected cases. **Each
one reads its number out of `ja4plus/` or out of the FoxIO method tuple of
`tests/test_documentation_image_count.py`.** A case that restated the number would pass
against a README that contradicts the code.

Four mutations prove they discriminate. Each was applied, measured, and reverted.

| Mutation | Cases that failed |
|---|---|
| README age 600 → 300 seconds | `test_the_readme_states_the_default_maximum_age_of_a_state_table` |
| README JA4TScan `No` → `Yes` | `test_the_readme_method_table_marks_ja4tscan_as_not_implemented` |
| README "FoxIO owns the JA4+ standard" → "FoxIO made JA4+" | `test_the_readme_names_foxio_as_the_owner_of_the_standard` |
| **Code** `DEFAULT_MAX_CONNECTION_AGE` 600 → 900 | `test_the_readme_states_the_default_maximum_age_of_a_state_table` |

**The fourth mutation is the one that matters.** It changes no README text, and the case
still fails, which proves the bound cases read `ja4plus/utils/state_table.py` rather than
reading the README twice.

Before the README change, 31 of the 35 cases failed. The 4 that passed record claims the
README already met: the import block, the 512 MiB ceiling, the concurrency paragraph, and
the "independent implementation" sentence.

## Two findings this pull request does not fix

1. **The README license list is incomplete.** It names JA4S, JA4H, JA4T, JA4TS, JA4L,
   JA4X and JA4SSH under the FoxIO License 1.1. FoxIO's `License FAQ.md:5` also names
   **JA4LS, JA4D and JA4D6**. A license statement is a decision for the user, so this
   change leaves it.
2. **A citation reads oddly.** `CLAUDE.md` and `.claude/rules/conformance.md` each state
   that "#179 records six unbounded state tables". #179 is titled *Correct the maximum-age
   default in the concurrency-safety feature file*. Its Changelog row 43 does record the
   six, so the claim holds, but the citation points at a document about a different
   subject.

## Verification

`ci: skip`, so the full local suite replaces provider CI. `PYTHONDONTWRITEBYTECODE=1`
was set and every `__pycache__` outside `.venv/` was cleared before each measurement.

- `pytest tests/ -m "not spec_validation"` — **2286 passed**, 2 skipped, 8 xfailed, 98
  subtests. The base reads 2251 passed, so the 35 new cases are the whole difference.
- `pytest tests/ -m spec_validation` — 1532 passed, 143 skipped, **134 xfailed**, against
  **134** register keys. The invariant holds.
- `ruff format --check ja4plus/ tests/` — 167 files already formatted.
- `mypy --strict ja4plus/` — Success, no issues in 31 source files.
- Coverage — **94% on the base and 94% on this branch**, 274 missed lines both.

**One flake, named.** On the first full run,
`TestTheStatedMemoryCeiling::test_a_smaller_entry_count_holds_less_resident_memory`
failed. It is a resident-memory ratio control, it runs in a subprocess, and **no file
under `ja4plus/` changes in this pull request**, so nothing here can reach it. It passes
in isolation and it passed on the re-run of the full suite on a quiet host.

**No fingerprint moves. No file under `ja4plus/` changes.**

Verified against: https://github.com/FoxIO-LLC/ja4/blob/main/technical_details/README.md
(FoxIO JA4+, pinned commit `27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8`, read from the
local checkout on 2026-08-09)


Crank-Git:
## Self-review: three cases were evaded, and all three are closed

I ran two specialist reviews over the diff — one for the writing standard, one asking the
question `.claude/rules/conformance.md` § "Ask whether a case can fail" asks. The second
one earned its keep.

**Three cases passed against a README that was wrong.** Each evasion below was applied to
`README.md`, measured, and reversed.

| Evasion | The first form | Now |
|---|---|---|
| Delete `JA4TFingerprinter,` from the import block and leave `# NOTE: JA4TFingerprinter intentionally omitted here` | passed | fails |
| Replace the opening claim with "**This project provides full coverage of every JA4+ method FoxIO publishes.**" | passed | fails |
| Rewrite the independence sentence as "This library copies FoxIO's design wholesale and calls itself an independent implementation in name only." | passed | fails |

All three are the same fault in three costumes: **a substring search reads words, and a
claim is not its words.**

### The repairs

- `test_the_readme_import_block_imports_every_exported_fingerprinter` reads the import
  lines with `^\s+(JA4\w*Fingerprinter)\s*,` instead of testing `name in block`. A name
  inside a comment imports nothing.
- `test_the_readme_claims_no_full_foxio_method_coverage` matches four patterns rather than
  three fixed strings. The first form held a literal list, which is exactly the failure the
  module docstring warns about.
- `test_the_readme_names_foxio_as_the_owner_of_the_standard` and
  `..._names_this_project_an_independent_implementation` each read one whole sentence out
  of the first 20 lines, with whitespace normalized so the 90-column wrap does not hide it.

### Two cases are new

- `test_the_readme_states_the_count_of_methods_its_own_table_marks` — the count word in the
  prose must equal the rows the table marks `Yes`. **The prose and the table can no longer
  drift apart**, and no fixed number appears in the case.
- `test_the_readme_states_the_count_of_methods_foxio_publishes` — the published count comes
  from the FoxIO method tuple.

The three evasions now fail **five** cases between them.

### One finding I accepted but did not act on

`MEMORY_CEILING_MIB = 512.0` lives in `tests/test_memory_bounds.py`, not under `ja4plus/`.
The reviewer is right that this is a policy constant and not a number the code computes —
its own comment says "no case here may derive the number". The case is still sound: it
reads the one constant the suite measures the ceiling against, and unbolding `**512 MiB**`
fails it. I am naming the nuance rather than papering over it.

### Writing standard

Nine STE findings, all applied: an `-ing` noun in the opening sentence, `shed` rotated
against the Terms-table verb `evict` in two files, four sentences over the 25-word
description limit, a four-condition sentence that rule 16 wants as separate sentences, and
one test name written as a double negative
(`..._names_no_method_foxio_does_not_publish` → `..._names_only_methods_foxio_publishes`).

### Verification after the repair

- `pytest tests/ -m "not spec_validation"` — **2288 passed**, 2 skipped, 8 xfailed.
- `pytest tests/ -m spec_validation` — 1532 passed, 143 skipped, **134 xfailed** against
  **134** register keys.
- `ruff check` / `ruff format --check` / `mypy --strict ja4plus/` — all clean.
- Coverage — **94%**, 274 missed. Unchanged from the base.

No file under `ja4plus/` changes. No fingerprint moves.

Issue #64, opened 2026-08-06
Build the documentation site with an API reference


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/08-documentation.md`](docs/specs/features/08-documentation.md):

> FR-documentation-7 — The documentation site builds from the Markdown files that already exist.

> FR-documentation-9 — The documentation site carries an API reference generated from the docstrings.

> FR-documentation-15 — A broken internal link fails the documentation build.

## What to build

New file `mkdocs.yml`. The site builds from the Markdown files that already exist under `docs/`.

| Tool | Version | Documentation |
|---|---|---|
| `mkdocs-material` | 9.x | https://squidfunk.github.io/mkdocs-material/ |
| `mkdocstrings-python` | 1.x | https://mkdocstrings.github.io/python/ |

The API reference is generated from the docstrings, which is why Epic 4's annotations land first. `mkdocs build --strict` fails on a broken link, which is the requirement.

Verified against the pages above, retrieved 2026-08-06.

## Acceptance criteria

- [ ] `mkdocs build --strict` succeeds.
- [ ] The site serves an API reference page for `Processor`.
- [ ] A broken internal link fails the build.

## Links

- Feature: [`docs/specs/features/08-documentation.md`](docs/specs/features/08-documentation.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #19
Depends on #62

Crank-Git:
Sub-merged into `epic/19-documentation` as `da2338d`. Labelled `status:batched`. The issue closes when the Epic 8 pull request merges into `dev`.

**The site builds strict and a dead link fails it.** MkDocs 1.6.1 with Material 9.7.7, `docs_dir: docs` so no existing page moved, `docs/specs/` excluded, and the API reference generated by `mkdocstrings-python` 1.15.0 into `docs/reference/`.

## FR-documentation-15 was proven, not asserted

**Both settings are required and the worker found that out by measuring.** `strict: true` alone misses a dead anchor, which MkDocs reports at info level; `validation.links.anchors: warn` raises it. Two reversals, each restored:

- a dead page link — `Aborted with 1 warnings in strict mode!`, exit 1
[...]

Issue #63, opened 2026-08-06
Run every code sample and example in continuous integration


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/08-documentation.md`](docs/specs/features/08-documentation.md):

> FR-documentation-4 — Every code sample in the README runs in continuous integration.

> FR-documentation-5 — Every code sample in `docs/` runs in continuous integration.

> FR-documentation-6 — Every script in `examples/` runs in continuous integration.

## What to build

Nothing tests any code sample in the README, in `docs/`, or in `examples/` today, so a sample that stops working stops working silently.

An example script runs against a committed capture, not against a live interface. `examples/live_traffic_fingerprinting.py` and `examples/tcpdump_workflows.py` need either a capture-file mode or an exclusion marker that names the reason.

`examples/monitoring_daemon.py` was removed in Epic 6; the supported `ja4plus watch` command replaces it.

New file `tests/test_examples.py`.

## Acceptance criteria

- [ ] `pytest --doctest-glob="*.md" README.md docs/` passes.
- [ ] `pytest tests/test_examples.py` runs every script in `examples/` and passes.
- [ ] A sample that cannot run in continuous integration is marked, and the marker names the reason.

## Links

- Feature: [`docs/specs/features/08-documentation.md`](docs/specs/features/08-documentation.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #19
Depends on #62

Crank-Git:
Sub-merged into `epic/19-documentation` as `bf3d5ec`. Labelled `status:batched`. The issue closes when the Epic 8 pull request merges into `dev`.

**45 samples of the README and `docs/` run, five example scripts run under six command lines, and 14 samples carry a marker naming why they cannot.**

## The merge turned a case red, and the obvious fix would have hidden the real defect

#64 added five `docs/reference/` pages after this branch forked, and the census case went red. **Adding the five pages to the hand-written table was the minimal repair, and it would have been the wrong one.**

**`PYTHON_SAMPLE_FILES` read that table's keys.** So a page absent from the table was **read, counted, classified `run`, and never executed.** The worker proved it: a probe page holding a broken import produced `docs/reference/probe.md:3 python run` from the reader and `3 passed` from the runner.

[...]

PR #390, opened 2026-08-09
Run every code sample and example in continuous integration

Crank-Git:
Part of #19. Builds #63.

FR-documentation-4, FR-documentation-5 and FR-documentation-6.

## What changed

Nothing ran a code sample before this change, so a sample that stopped working stopped working in silence. This project records seventeen instances of a comparison that is never made reading as a comparison that passes, and an unrun sample is that shape.

- New `tests/documentation_samples.py` reads every fenced block of `README.md` and of `docs/`, and it states one disposition for each block.
- New `tests/test_documentation_samples.py` carries the floors, the census cases and the execution cases.
- New `tests/test_examples.py` runs every script of `examples/`.
- New `samples` job in `.github/workflows/test.yml`. The `test` job already reads `tests/`, so the harness also runs on every version of the matrix and on macOS.

## How a block declares itself runnable

A `python` block and a `bash` block **run by default**. An explicit marker on the line before the fence declares the exception:

- `<!-- sample: skip <reason> -->` states why the block runs never.
- `<!-- sample: raises <error> <reason> -->` states the error the block raises on purpose.

The default runs the sample, so a new block that cannot run **fails** until somebody writes the reason. A marker that names no known action raises. A skip reason shorter than four words fails a case, so no skip is silent. A block whose info string names another language, or no language, carries output; `test_no_block_the_reader_calls_output_holds_a_code_sample` proves that classification.

## Both counts

| Reader | Blocks across `README.md` and `docs/` |
|---|---|
| `tests/documentation_samples.py`, a fence grammar that tracks the fence width | 157 |
| `_second_reader_block_count`, string operations and a toggle | 157 |

**The two readers disagreed on the first run, and the disagreement was a real defect.** The second reader counted one extra block in each of `docs/specs/foxio/JA4D.md`, `docs/specs/foxio/JA4SSH.md`, `docs/specs/foxio/JA4T.md` and `docs/specs/foxio/JA4X.md`. Each file holds an inline code span that opens a line with three backticks, such as `docs/specs/foxio/JA4D.md:137`. The second reader was corrected, and **the floor was set from the corrected reader** and not from the first draft. #302 set its floor from the count its own broken parser read.

## The census

| Action | Count |
|---|---|
| Runs | 43 |
| Raises a named error | 2 |
| Skipped, with a reason | 94 |
| Output, a schema line or a table | 18 |
| **Total** | **157** |

`MINIMUM_BLOCKS` reads 157 and `MINIMUM_RUNNABLE_SAMPLES` reads 45. `BLOCKS_PER_SAMPLE_FILE` records the count of each user documentation page, and `test_the_table_names_every_markdown_file_outside_the_specification_package` fails on a new page nobody classifies.

## Every skip, and why

Fourteen samples carry a marker.

| Where | Reason |
|---|---|
| `README.md:57`, `README.md:66` | The command reaches the Python Package Index. |
| `README.md:100` | The command opens a capture socket, and continuous integration holds no capture privilege. |
| `README.md:119`, `README.md:240` | The command reaches the lookup service at `https://ja4db.com`. |
| `README.md:465` | The first command downloads the vectors from the FoxIO repository. |
| `README.md:473` | The first command clones the repository from GitHub. |
| `docs/usage.md:379`, `docs/usage.md:412`, `docs/usage.md:452`, `docs/usage.md:511` | The command opens a capture socket, and continuous integration holds no capture privilege. |
| `docs/usage.md:490` | The command names a process identity that no host holds. |
| `docs/api_reference.md:640` | The block names a placeholder argument, and `db update` reaches the FoxIO repository. |
| `docs/implementation_notes.md:1477` | The block quotes the FoxIO reference implementation, which this project runs never. |

Eighty further blocks live under `docs/specs/`, which holds the specification package and the verbatim FoxIO transcription. `.claude/rules/ste.md` forbids a rewrite of text copied from the FoxIO material, and a runnable sample needs a rewrite. The census counts all eighty and reports them; none is silent.

**No sample reaches the network.** A fixture refuses `socket.socket.connect` and `socket.create_connection` while a Python sample runs, and every sample that would reach `https://ja4db.com` carries a skip marker. #57 made the remote lookup opt-in.

## The examples

Five scripts run under six command lines. `EXAMPLE_COMMANDS` names every script, so a new script fails a case until somebody records how to run it.

| Script | Command lines |
|---|---|
| `demo_fingerprints.py` | the built-in demonstration |
| `pcap_analysis.py` | a capture file |
| `threat_detection.py` | a capture file |
| `tcpdump_workflows.py` | the analyze mode, the filter recipes |
| `live_traffic_fingerprinting.py` | the capture-file mode |

`examples/live_traffic_fingerprinting.py` gained a `--pcap` option that passes `offline` to `sniff`. No script is marked unrunnable.

## The four discrimination proofs

Each path reaches different code, so each path was broken on its own and then restored.

| Path | Mutation | Result |
|---|---|---|
| README, Python | `fp = JA4Fingerprinterr()` | `Failed: the sample at README.md:272 failed: NameError: name 'JA4Fingerprinterr' is not defined` |
| README, shell | `ja4plus analyse capture.pcap` | `AssertionError: the sample at README.md:74 ended with the status 2` |
| `docs/`, Python | `fp.get_raw_fingerprintt(packet)` | `Failed: the sample at docs/usage.md:47 failed: AttributeError: 'JA4Fingerprinter' object has no attribute 'get_raw_fingerprintt'` |
| `examples/` | `from scapy.all import rdpcapp` | `AssertionError: pcap_analysis.py under a capture file ended with the status 1` |

After each restore the harness reported `29 passed`.

## The defect the harness found

**`ja4plus.generate_ja4` reads the `tls_info` dictionary that `extract_tls_info` returns, and it reads no packet.** `README.md` and `docs/api_reference.md` each wrote `generate_ja4(packet)`, and that call raises `AttributeError: get`. The nine other one-shot functions read a packet, and `generate_ja4x` reads a dictionary that `docs/api_reference.md` already recorded.

This change repairs the two documents. **It changes no signature**, because #63 alters no behaviour under `ja4plus/`.

## Verification

Run on macOS with `PYTHONDONTWRITEBYTECODE=1` and every `__pycache__` outside `.venv/` cleared.

- `pytest tests/ -m "not spec_validation"`: **2320 passed, 2 skipped, 8 xfailed, 98 subtests passed**.
- `pytest tests/ -m spec_validation`: **1532 passed, 143 skipped, 134 xfailed**.
- The register invariant holds: `tests/foxio_deviations.json` holds **134 keys against 134 xfailed**.
- `ruff check ja4plus/ tests/`: clean.
- `ruff format --check ja4plus/ tests/`: clean.
- `mypy --strict ja4plus/`: **Success: no issues found in 31 source files**.
- Line coverage: **94%**, 273 of 4232 statements missed. Without the two new files the same suite misses 274, so coverage rises by one line and does not fall.

`TestTheStatedMemoryCeiling::test_a_smaller_entry_count_holds_less_resident_memory` failed once on a loaded host and passed on a quiet one, and it passed on the final full run. This change repairs it never.

## Notes for the sub-merge

#64 builds the documentation site on this integration branch. It will meet this change in `.github/workflows/test.yml`, where this change adds one `samples` job at the end. It will also meet `BLOCKS_PER_SAMPLE_FILE` in `tests/test_documentation_samples.py`: a new page under `docs/` fails the census until somebody records its block count, which is the intended behaviour. This change adds no documentation dependency and builds no second documentation toolchain.

Verified against: the `scapy.sendrecv.sniff` docstring of the installed release, which states `offline: PCAP file (or list of PCAP files) to read packets from` (retrieved 2026-08-09).

---

## Two blocks that neither reader finds

A fence inside a blockquote reaches neither reader, so the two counts agree on a block that neither one found. **Two such fences exist**, at `docs/specs/features/00-foundation.md:49` and `:53`. They open and close one `.gitignore` fragment, which is data rather than a code sample, and they sit inside the specification package that the harness excludes whole. `KNOWN_BLOCKQUOTE_FENCES` records both, so the miss is counted rather than silent, and a third one fails a case. A blockquote fence in the user documentation fails a case outright.

The self-review found this, and the two review comments below record every finding, every repair and every declined finding with its reason.


Crank-Git:
## Self-review — harness soundness

Two specialist passes over the diff. The findings and what each one produced.

### Accepted and repaired in 7199eb2

1. **A marker that a blank line parts from its fence was dropped.** `read_fenced_blocks` read `lines[index - 1]` alone, so a Markdown tidy-up would have removed the marker and the block would then run with no reason recorded. The reader now passes over the blank lines above the fence.
2. **A marker on a block of another language vanished.** `disposition` reads the language before it reads the marker, so a marker on a JSON block or on an output block stated a reason that nothing acted on. `test_every_marker_sits_on_a_block_the_harness_runs` now reports the misplaced marker.
3. **A Python sample ran with no time limit.** A shell sample already held one through the subprocess. Each Python sample now runs under a 300 second limit, through `signal.setitimer` where the platform carries `SIGALRM`.
4. **A fence inside a blockquote reaches neither reader.** The primary reader rejects the `>` prefix, and the second reader strips the line and then rejects it too. The two readers therefore agree on a block that neither one found, so the miss is silent. **Two such fences exist**, at `docs/specs/features/00-foundation.md:49` and `:53`, and they open and close one `.gitignore` fragment. `KNOWN_BLOCKQUOTE_FENCES` records both, so the miss is counted and a third one fails a case. A blockquote fence in the user documentation fails a case outright, because it would hide a runnable sample.

The fourth finding came from a case this review added, which is the point of adding it.

### Declined, with the reason

- **The second reader knows no tilde fence.** Correct, and it stays that way. A tilde fence in the documentation would make the two readers disagree, and `test_two_readers_count_the_same_fenced_blocks` fails loudly on a disagreement. Teaching the second reader the same grammar as the first would reduce the independence that makes the comparison worth making. No tilde fence exists today.
- **The info-string group forbids a backtick in a tilde fence, which CommonMark permits.** True and dormant. No tilde fence exists.
- **A fence indented by four spaces or more is invisible to the primary reader.** CommonMark reads four leading spaces as an indented code block rather than a fence, so the reader is right. A fence indented inside a list item would make the two readers disagree, because the second reader strips the line, so the shape fails loudly rather than passing silently.

### Confirmed by measurement, not by inspection

- `all_blocks()` finds 157 blocks and `census()` reports `{'run': 43, 'raises': 2, 'skip': 94, 'output': 18}`. `MINIMUM_BLOCKS` reads 157 and `MINIMUM_RUNNABLE_SAMPLES` reads 45, so both floors sit at the measured count rather than below it.
- The `test` job globs `tests/`, and neither new file carries the `spec_validation` marker, so the harness runs on Python 3.9 through 3.13 on Ubuntu and on 3.12 on macOS as well as in the new `samples` job.
- No block that runs today calls `sniff`, `socket.socket`, `time.sleep`, `input`, `recv` or `while True`.
- The three new files carry no PEP 604 runtime union, no `match`, and no `removeprefix`, so Python 3.9 reads them.

Crank-Git:
## Self-review — documentation and the writing standard

### Accepted and repaired in 7199eb2

1. **A sentence contradicted itself about `generate_ja4x`.** Both Changelog rows read "The nine other one-shot functions read a packet, and `generate_ja4x` reads a dictionary", and `generate_ja4x` is one of those nine. The rows now read "Eight of the nine other one-shot functions read a packet. `generate_ja4x` reads a dictionary". `docs/api_reference.md` already carried the correct form.
2. **The count of samples that open a capture socket read four, and it is five.** `README.md:100`, `docs/usage.md:379`, `:412`, `:452` and `:511`. The specification row now states five, and the eight categories now sum to fourteen without the compensating error.
3. **The rows reworded the project idiom into the passive voice.** They read "a comparison that is never made reads as a comparison that passes", where round 134 and `CHANGELOG.md` carry "a comparison that never runs reads as a comparison that passes". Rule 11 asks for the active voice and rule 7 bars a synonym rotation on a fixed phrase. Both rows now carry the active form.
4. **One sentence carried eight list items across about 50 words**, against rule 1 and rule 16. It is now eight short sentences.
5. **"the documentation said a packet" rotated a synonym** against "states", which the file uses everywhere else. It now reads "stated".

### Confirmed against the code

- `ja4plus/fingerprinters/ja4.py:110` reads `def generate_ja4(tls_info: dict[str, Any] | None, ...)`, so the function takes a dictionary and not a packet.
- `ja4plus/utils/tls_utils.py` reads `extract_tls_info(packet: Packet) -> dict[str, Any] | None`, which is the call the two repaired samples make.
- Of the ten one-shot functions, `generate_ja4` and `generate_ja4x` read a dictionary and the other eight read a packet.

### Confirmed unbroken by the fence splits

- The anchor `#the-lookup-makes-no-network-request-unless-you-ask-for-one` still resolves. No heading moved.
- No sentence near a split block refers to a block the reader no longer sees at that place.
- No file under `docs/specs/foxio/` changed, so no verbatim FoxIO material moved. The harness references `docs/specs/foxio/JA4D.md:137` by line alone, and that line does open with three backticks and close the span on the same line.

Crank-Git:
## A question for the project manager — acceptance criterion 1

The first acceptance criterion of #63 reads:

> - [ ] `pytest --doctest-glob="*.md" README.md docs/` passes.

**That command collects nothing today and ends with the status 5.**

```
$ pytest --doctest-glob="*.md" README.md docs/
no tests ran in 0.06s
exit=5
```

`grep -rn '>>>' README.md docs/*.md` reads **0**. Every sample of this documentation is a fenced block, and `--doctest-glob` collects a `>>>` session and nothing else. The criterion therefore names a command that reports green over zero samples, which is the exact defect #63 exists to remove.

`docs/specs/features/08-documentation.md` states the same mechanism under Behaviour rules: "A code sample is tested by `pytest --doctest-glob`."

**This change satisfies the requirements the criterion serves, through another mechanism.** FR-documentation-4 and FR-documentation-5 both hold: 45 samples of `README.md` and `docs/` run in continuous integration, and every one that runs never carries a marker that names the reason.

**Two ways to close the criterion, and the choice is yours.**

1. **Amend the criterion and the behaviour rule** to name `pytest tests/test_documentation_samples.py`, and record the reason: a fenced block carries no `>>>` session, and a documentation set with no doctest makes `--doctest-glob` report green over nothing.
2. **Rewrite the samples into doctest sessions.** That reaches every fenced block of the user documentation, and it asks each one for exact expected output. Several samples produce output that changes between runs or produce none, so they would become unrunnable under the rewrite. #63 puts "rewriting a sample to be nicer" out of scope, so this needs your approval before I do it.

**I have declined the third way**, which is to add one token doctest so the command collects something. That would make the criterion green and prove nothing, and it is the shape this project records seventeen times.

I have not marked the criterion met. The other two criteria are met with evidence in the pull-request body.

Crank-Git:
## The merge with #64 — a hole, not a stale number

### What the red case actually found

The failing case was working, and the repair it pointed at was the wrong one. **Adding the five pages to the hand-written table would have fixed the symptom and left the cause.**

`PYTHON_SAMPLE_FILES` read the keys of `BLOCKS_PER_SAMPLE_FILE`. A page absent from that table was read by the reader, counted by the census and classified as runnable — **and then never run.** Proved with a probe page:

```
$ cat docs/reference/probe.md
# Probe

```python
from ja4plus import ThisNameDoesNotExist
```

--- the reader sees the block ---
  docs/reference/probe.md:3 python run
--- does any execution case run it? ---
3 passed, 20 deselected
```

**The reader called it runnable and the runner reported green.** That is the exact shape #63 exists to remove, and it sat inside the harness built to remove it. It would have reached any page #65 adds as well.

After the repair, the same probe page fails for the right reason:

```
E  Failed: the sample at docs/reference/probe.md:3 failed: ImportError: cannot import name 'ThisNameDoesNotExist' from 'ja4plus'
```

### Answer to "should the census have caught this earlier" — derive the set, measure the counts

**Derived, because the filesystem is the one record of which pages exist.** A hand-written copy of that fact is a second record, and it went stale within one sub-merge of being written. `user_documentation_files()` now reads the filesystem, and `PYTHON_SAMPLE_FILES` reads the blocks. `test_two_readers_count_the_same_fenced_blocks` and the census already read the filesystem, which is why they covered the five new pages with no edit.

**Hand-measured, because the only tool that could derive a block count is the reader these cases exist to check.** A floor taken from the reader reports nothing about the reader. That is #302 exactly, and `MINIMUM_BLOCKS`, `MINIMUM_RUNNABLE_SAMPLES` and `MINIMUM_BLOCKS_PER_PAGE` stay measured by hand.

**`MINIMUM_BLOCKS_PER_PAGE` is now a floor for each page and no longer an exact count.** An exact count breaks on every page that gains a block, which is the staleness you named. A floor keeps the one regression the totals miss — a page that loses a sample while another page gains more — and tolerates growth.

Two cases replace the one I removed:

- `test_the_execution_cases_reach_every_sample_the_reader_calls_runnable` compares the reader against the runner, so the two lists cannot drift apart again. This is the case that would have caught the hole on day one.
- `test_the_floor_table_names_no_page_the_repository_lacks` fails when a floor entry names a deleted page, so that a missing page never reports as a missing sample.

### Both counts, re-measured on the merged tree

| Page | Reader | Toggle reader | Census |
|---|---|---|---|
| `README.md` | 21 | 21 | 10 run, 7 skip, 4 output |
| `docs/README.md` | 0 | 0 | — |
| `docs/api_reference.md` | 22 | 22 | 18 run, 2 raises, 1 skip, 1 output |
| `docs/implementation_notes.md` | 9 | 9 | 1 skip, 8 output |
| `docs/mutation_sweep.md` | 0 | 0 | — |
| `docs/output-schema.md` | 4 | 4 | 4 output |
| `docs/reference/fingerprinters.md` | 0 | 0 | — |
| `docs/reference/index.md` | 0 | 0 | — |
| `docs/reference/ja4db.md` | 0 | 0 | — |
| `docs/reference/processor.md` | 0 | 0 | — |
| `docs/reference/types.md` | 0 | 0 | — |
| `docs/usage.md` | 21 | 21 | 15 run, 5 skip, 1 output |
| **All files, including `docs/specs/`** | **157** | **157** | **43 run, 2 raises, 94 skip, 18 output** |

**The five new pages hold `:::` directives and no fenced block, so no floor moved.** `MINIMUM_BLOCKS` stays 157 and `MINIMUM_RUNNABLE_SAMPLES` stays 45, and both remain the measured numbers rather than numbers carried over. The user documentation grew from 7 pages to 12.

### The four discrimination proofs still discriminate

Re-run on the merged tree. README Python: `NameError`. README shell: status 2. `docs/usage.md`: `AttributeError`. `examples/pcap_analysis.py`: status 1. Each file restored, and `31 passed` after.

### `mkdocs build --strict` fails, and the cause is not in this pull request

Filed as #394. `mkdocstrings==1.0.6` ships no `mkdocstrings.handlers` package and `mkdocstrings-python==1.15.0` imports `mkdocstrings.handlers.base`, so every `:::` directive fails the build. It reproduces on macOS under Python 3.14 and on Ubuntu 24.10 under Python 3.12, at the pinned versions exactly as `pyproject.toml` writes them.

`git diff origin/epic/19-documentation...HEAD -- pyproject.toml mkdocs.yml docs/reference/` is empty, so this pull request changes no part of that toolchain. **No workflow builds the site**, so no gate reads the failure today. I have not repaired it.

### Gates on the merged tree

- `pytest tests/ -m "not spec_validation"`: **2331 passed**, 2 skipped, 8 xfailed, 98 subtests passed.
- `pytest tests/ -m spec_validation`: **1532 passed**, 143 skipped, **134 xfailed**; register holds **134 keys**.
- `ruff check`, `ruff format --check` (171 files), `mypy --strict` (31 source files): all clean.
- Line coverage **94%**, 273 of 4232 statements missed, unchanged.
- Linux, Ubuntu 24.10, Python 3.12: **31 passed**.
- `mkdocs build --strict`: **fails, #394**.

`TestTheStatedMemoryCeiling::test_a_smaller_entry_count_holds_less_resident_memory` passed on every run of this round. #389 owns it and I repaired it never.

Issue #391, opened 2026-08-09
Correct the mkdocstrings pin that stops the documentation build


Crank-Git:
Found by #65 on 2026-08-09 while it wrote the method, schema and migration pages. **#65 built to a working pin locally, changed no file of the site configuration, and reports it.** The site configuration belongs to #64.

## The measurement

**`mkdocs build --strict` fails on `epic/19-documentation` at `da2338d`, with the `docs` extra installed exactly as `pyproject.toml` pins it.** The failure predates #65: the same command fails on a clean checkout of the base commit with no change applied.

```
ModuleNotFoundError: No module named 'mkdocstrings.handlers'
```

The traceback ends at `mkdocstrings_handlers/python/handler.py:25`:

```
from mkdocstrings.handlers.base import BaseHandler, CollectionError, CollectorItem, HandlerOptions
```

## The cause

The `docs` extra pins two packages that cannot work together.

| Package | Pin | What it does |
|---|---|---|
| `mkdocstrings` | `1.0.6` | Moved every handler interface under `mkdocstrings/_internal/`. The distribution ships `mkdocstrings/__init__.py` and `mkdocstrings/_internal/` and **no `mkdocstrings/handlers/`**. |
| `mkdocstrings-python` | `1.15.0` | Imports `mkdocstrings.handlers.base`, which the 1.x line of `mkdocstrings` removed. |

Reproduce the package layout with one command:

```bash
ls .venv/lib/python3.*/site-packages/mkdocstrings/
# _internal/  __init__.py  py.typed
```

**#64 met this exact error and recorded it.** Its Changelog entry states that a floating `mkdocstrings>=0.28` resolved to 1.0.6 and killed the build with this message. **The repair pinned the version that produces the error.** The pin needed to hold `mkdocstrings` on the 0.x line, which is the line `mkdocstrings-python` 1.15.0 imports from.

## The measured repair

`mkdocstrings==0.30.1` is the newest release that still ships `mkdocstrings.handlers`. With that one version changed and nothing else, the build is clean:

```
$ .venv/bin/mkdocs build --strict
INFO    -  Documentation built in 1.51 seconds
$ echo $?
0
```

The build writes two `DeprecationWarning` lines at the information level, from `mkdocstrings.loggers`. Strict mode fails on a warning of MkDocs and not on a Python warning, so neither one fails the build.

[...]

Issue #66, opened 2026-08-06
Publish the documentation site and write the release notes


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/08-documentation.md`](docs/specs/features/08-documentation.md):

> FR-documentation-8 — The documentation site publishes to GitHub Pages on a push to the live branch.

> FR-documentation-13 — The `CHANGELOG.md` file records every breaking change in this release.

## What to build

New file `.github/workflows/docs.yml`. The workflow needs `pages: write` and `id-token: write` permissions, and the repository must have GitHub Pages set to build from GitHub Actions. If Pages is not enabled, the workflow fails with a message naming the setting.

| What | Version | Documentation |
|---|---|---|
| GitHub Pages deployment | `actions/deploy-pages` v4 | https://github.com/actions/deploy-pages |

The live branch is `master` under the dev-and-live model.

Verified against the page above, retrieved 2026-08-06.

## Acceptance criteria

- [ ] The documentation workflow publishes the site on a push to the live branch.
- [ ] `CHANGELOG.md` holds a `1.0.0` section that names each breaking change.

## Links

- Feature: [`docs/specs/features/08-documentation.md`](docs/specs/features/08-documentation.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #19
Depends on #64

Crank-Git:
Blocked on #64, which builds the documentation site this issue publishes.

**A second reading the project manager records now, so it is not discovered late.** `gh api repos/Crank-Git/ja4plus` reports `has_pages: false`, and the Pages endpoint returns 404. **GitHub Pages is not enabled and the project manager will not enable it** — publishing a public website is the user's decision, not an autonomous one.

**That does not block building this issue.** The workflow fires on a push to the live branch, `master`, and a promotion from `dev` to `master` is already a separate step the user approves. So the workflow lands unfired, and the issue body already states the behaviour: if Pages is not enabled, the workflow fails with a message naming the setting.

**What the user decides, when they choose to:** whether to enable Pages with the source set to GitHub Actions. Until then the site builds in continuous integration and publishes nowhere.

Crank-Git:
**#64 has landed, so the site exists. This issue is now blocked on two other things, both measured.**

[...]

Issue #65, opened 2026-08-06
Write the method, schema and migration pages


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/08-documentation.md`](docs/specs/features/08-documentation.md):

> FR-documentation-10 — The documentation site carries a page per method.

> FR-documentation-11 — The documentation site carries the output schema.

> FR-documentation-12 — The documentation site carries a migration page for the move from version 0.6.0 to version 1.0.0.

## What to build

New files under `docs/methods/`, one per implemented method, plus `docs/migration-0.6-to-1.0.md` and `docs/concurrency.md`. `docs/output-schema.md` came from Epic 5.

The migration page lists each breaking change with the old form, the new form, and the reason. The breaking changes in this release are:

- `Processor.process_packet` returns `FingerprintResult` objects rather than dictionaries.
- `ja4plus.collector` is removed.
- JA4SSH emits every 200 SSH packets rather than every 10.
- The remote fingerprint lookup is opt-in.
- The Python floor is 3.9.

Prose follows `.claude/rules/ste.md` and the `## Terms` table in `docs/specs/spec.md`.

## Acceptance criteria

- [ ] The site serves one page per implemented method.
- [ ] Every method page cites its FoxIO source.
- [ ] The site serves the output schema page.
- [ ] The migration page lists every breaking change this release makes.

## Links

- Feature: [`docs/specs/features/08-documentation.md`](docs/specs/features/08-documentation.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #19
Depends on #64

Crank-Git:
Blocked on #64, which builds the documentation site this issue writes pages onto. #64 is in flight on `epic/19-documentation`. This issue starts when #64 sub-merges, branching from the updated integration branch.

Crank-Git:
Unblocked. #64 sub-merged into `epic/19-documentation` as `da2338d`.

[...]

PR #396, opened 2026-08-09
Correct the mkdocstrings pin that stops the documentation build

Crank-Git:
## What

`mkdocstrings==1.0.6` is an exact pin and an incompatible one. The `docs` extra now pins
`mkdocstrings==0.30.1`, and no other version moves. `.github/workflows/docs-build.yml`
installs the `docs` extra alone into an empty environment and builds the site, so the next
wrong pin fails a check instead of a reader.

Refs #391.

## Why

`mkdocstrings_handlers/python/handler.py` of the 1.x line imports `mkdocstrings.handlers`,
and `mkdocstrings` 1.0.0 removed that public module. `mkdocstrings-python` 1.15.0 declares
`mkdocstrings>=0.28`, which 1.0.6 satisfies, so the upper bound has to live in
`pyproject.toml`.

**Two guards existed and neither could catch it.**
`test_every_documentation_dependency_pins_one_version` reads the pins, and an exact pin can
still be an incompatible pin. #64 then built the site inside a virtual environment that an
earlier resolution had already filled, and the gate repeated that measurement in the same
environment. Nobody installed the committed set from scratch.

**A unit case cannot close the gap.** The suite runs inside an environment that another
resolution already filled, so a case that imports `mkdocs` reports that environment and not
`pyproject.toml`. The instrument is a job that starts from an empty runner.

## What I read

`mkdocstrings` 1.0.0 lists `mkdocstrings.handlers` under `### Breaking Changes`, as
`mkdocstrings.plugin`, `mkdocstrings.loggers` and `mkdocstrings.inventory` beside it, each
with the note "import from `mkdocstrings` directly". 0.30.1 still ships the module as a
deprecation shim, and the green build below prints the matching `DeprecationWarning`.

The metadata states the second constraint: `mkdocstrings` 1.0.0 requires Python 3.10, and
`griffe` 2.1.0 requires 3.10 too. **The `docs` extra therefore needs Python 3.10 or later,
and the package floor of Python 3.9 stays**, because no user builds the site. The comment
in `pyproject.toml` claimed the 3.9 floor held the whole extra, and it is corrected.

```
Verified against: https://github.com/mkdocstrings/mkdocstrings/blob/main/CHANGELOG.md (1.0.0, retrieved 2026-08-09)
Verified against: https://pypi.org/pypi/mkdocstrings-python/1.15.0/json (requires_dist, retrieved 2026-08-09)
Verified against: https://pypi.org/pypi/mkdocstrings/0.30.1/json (requires_python >=3.9, retrieved 2026-08-09)
Verified against: https://pypi.org/pypi/griffe/2.1.0/json (requires_python >=3.10, retrieved 2026-08-09)
```

## The guard, shown failing and passing

The steps of `.github/workflows/docs-build.yml` ran against two clean clones on Linux,
Ubuntu 24.10, Python 3.12, which is the platform continuous integration uses.

```
python3.12 -m venv .venv-docs
.venv-docs/bin/pip install -e ".[docs]"
.venv-docs/bin/mkdocs build --strict
```

**Red, on `da2338d`:**

```
mkdocstrings 1.0.6 · mkdocstrings-python 1.15.0 · griffe 2.1.0 · mkdocs 1.6.1 · mkdocs-material 9.7.7

  File ".../mkdocstrings_handlers/python/handler.py", line 25, in <module>
    from mkdocstrings.handlers.base import BaseHandler, CollectionError, CollectorItem, HandlerOptions
ModuleNotFoundError: No module named 'mkdocstrings.handlers'
EXIT=1
```

**Green, on this branch:**

```
mkdocstrings 0.30.1 · mkdocstrings-python 1.15.0 · griffe 2.1.0 · mkdocs 1.6.1 · mkdocs-material 9.7.7

INFO    -  Documentation built in 2.38 seconds
EXIT=0
Processor docstring: present
one-shot functions: all ten rendered
```

macOS with Python 3.11 gives the same two results, and the green build there reports
`Documentation built in 0.79 seconds`.

## How the job is built

- It runs on a change to `pyproject.toml`, to `mkdocs.yml`, to `docs/` or to itself.
- It installs `.[docs]` and never `.[dev]`. The `dev` extra brings a second resolution that
  the release never installs.
- After the build it reads two results out of the generated HTML: the `Processor` docstring
  sentence on `site/reference/processor/index.html`, and the ten `generate_ja4*` one-shot
  functions on `site/reference/fingerprinters/index.html`. A build that produces empty pages
  exits 0.
- **The name `docs.yml` stays free for #66**, which owns the publication. This job publishes
  nothing.

## Tests

Three cases in `tests/test_documentation_site.py`, each measured red before the repair:

- `test_the_mkdocstrings_pin_holds_the_handler_module_its_handler_imports` states the upper
  bound as text, so a reader of `pyproject.toml` alone sees it.
- `test_a_continuous_integration_job_installs_the_documentation_pins_and_builds`
- `test_the_documentation_job_runs_when_a_pin_or_a_page_changes`

```
FAILED tests/test_documentation_site.py::test_the_mkdocstrings_pin_holds_the_handler_module_its_handler_imports
FAILED tests/test_documentation_site.py::test_a_continuous_integration_job_installs_the_documentation_pins_and_builds
FAILED tests/test_documentation_site.py::test_the_documentation_job_runs_when_a_pin_or_a_page_changes
3 failed, 12 deselected
```

## The local suite

```
pytest tests/ -m "not spec_validation"   2302 passed, 3 skipped, 8 xfailed, coverage 94%
register keys                            134
ruff format --check ja4plus/ tests/      168 files already formatted
```


Crank-Git:
Self-review, correctness lens. `pyproject.toml:59` named `.github/workflows/docs.yml`, and the job is `.github/workflows/docs-build.yml`. The name changed after the comment was written, so a reader of the extra reached a file that does not exist. Repaired.

Crank-Git:
Self-review, all lenses, done and pushed as `68bfc0c`.

**Correctness lens, two real defects.**

1. `pyproject.toml:59` and the docstring of `test_the_mkdocstrings_pin_holds_the_handler_module_its_handler_imports` named `.github/workflows/docs.yml`. The job is `docs-build.yml`, and `docs.yml` stays free for #66. Repaired.
2. **The bound sat behind `if handler.startswith("1.")`, so a bump of `mkdocstrings-python` past the 1.x line would switch the guard off in silence.** That is the shape #391 exists to bar: a comparison that stops running reads as a comparison that passes. The case now reads the line of the handler, states the bound for each line it has read, and **fails on a line it has not read**. Two mutations prove it: `mkdocstrings==1.0.6` turns it red, and `mkdocstrings-python==3.0.0` turns it red.

Two nits, both taken:

- `test_the_documentation_job_runs_when_a_pin_or_a_page_changes` read the path filter from the whole file, so a filter that covered `push` alone would have passed while every pull request went unguarded. It now reads each trigger block. Removing the filter from `pull_request` alone turns it red.
- The unguarded `grep -q` on the `Processor` page failed the step with no annotation. It now writes `::error::`.

The reviewer confirmed the rest: `_pinned_version("mkdocstrings")` does not match `mkdocstrings-python==1.15.0`, `grep -q ... || missing=...` does not trip `errexit`, the `$GITHUB_STEP_SUMMARY` form is right, and the action SHAs match the three existing workflows.

**Writing lens, five real defects.** The British spelling `artefact` in four places, three passive sentences, one sentence of 26 words, and a specification row that created `.venv` and then called `.venv-docs/bin/pip`. All repaired.

**Five mutations, each measured red, and every file restored:**

```
RED   the pin returns to the incompatible one
RED   the handler moves to an unread line
RED   the job installs the dev extra too
RED   the job runs no strict build
RED   the pull request trigger loses its path filter
```

**The render check is proven by its own reversal.** The body of the step ran under `bash -eo pipefail` against the built site. It exits 0 on the real site. An emptied `site/reference/processor/index.html` gives `::error::The Processor page carries no docstring text from the source.` and exit 1. A renamed `ja4x` anchor gives `::error::The API reference renders no one-shot function for: ja4x` and exit 1.

Crank-Git:
#63 and #65 landed on `epic/19-documentation` while this branch was open, so `c632afc` is merged in. **The one conflict is the Changelog table of `docs/specs/spec.md`, where three members each appended a row.** The rows keep the order the work landed: #63, then #65, then #391. No other file conflicts, because this change touches `pyproject.toml`, one workflow, one test module and the two Changelog files, and it edits no page and no nav entry.

**The merged tree was measured again, and the thirteen new pages of #65 build.**

```
python3.11 -m venv .venv          (a directory of its own, not the worktree venv)
.venv/bin/pip install -e ".[docs]"
.venv/bin/mkdocs build --strict   INFO - Documentation built in 0.92 seconds, exit 0
the render step                   The API reference renders the ten one-shot functions, exit 0
```

The local suite on the merged tree:

```
pytest tests/ -m "not spec_validation"   2477 passed, 3 skipped, 8 xfailed, coverage 94%
register keys                            134
ruff format --check ja4plus/ tests/      172 files already formatted
```

The unit suite rises from 2302 to 2477 because #63 and #65 bring their own cases. Coverage holds at 94.

Issue #401, opened 2026-08-09
Part round 122 from round 123 on the migration page


Crank-Git:
## What is wrong

`docs/migration-0.6-to-1.0.md:20` cites one change under two rounds:

> | The result of a lookup | ... | Round 122, #59 and #364 |

The Changelog table of `docs/specs/spec.md` holds two rows, and each one records a
different change.

- `| 122 | 2026-08-08 | #59 landed and it closes Epic 7. ... |`
- `| 123 | 2026-08-08 | #364 landed and a `LookupResult` reads by field name, ... |`

Row 122 never mentions #364, and row 123 is the row that records it. A reader who follows
`Round 122` for the item-access rule reaches a row about `lookup_many` and the source
field.

`CHANGELOG.md:605` states `Round 123.` for #364, so `CHANGELOG.md` and `docs/specs/spec.md`
already agree. The migration page is the one file that does not.

## Evidence

- `docs/specs/spec.md`, the Changelog rows 122 and 123.
- `CHANGELOG.md:605`, the entry of #364, which reads `Round 123.`
- `CHANGELOG.md:643`, the entry of #59, which reads `Round 122.`

## What to build

Part the cell of `docs/migration-0.6-to-1.0.md:20` into the two records, or split the row
into two rows. The release notes of version 1.0.0 in `CHANGELOG.md` already hold two rows:
`Round 122, #59` for the typed result, and `Round 123, #364` for the item access.

`tests/test_release_notes.py::test_every_citation_of_the_release_notes_names_the_row_that_records_it`
holds every citation of the release notes against the Changelog table. It reads the
migration page for issue numbers alone, so the repair of this page needs no change to the
notes.

## Who found it

#66 found the conflation while it wrote the release notes. The self-review of pull request
#400 reports the measurement. #65 owns the page, so #66 recorded the gap and edited no page
of another issue.

Crank-Git:
## Adopted into Epic 8 by the project manager, 2026-08-09

**This is a defect in a page Epic 8 itself created.** `docs/migration-0.6-to-1.0.md` shipped with #65 in this same epic, and #66 found this while writing the release notes against it.

**Shipping the epic with a known defect in its own new page would be the wrong call**, so this runs before the Epic 8 gate rather than after it.

**#399 and #401 are worked together by one worker**, because both change the same file and two workers on one file would collide for no gain. The Epic 8 pull request closes them both.

[...]

Issue #399, opened 2026-08-09
Record the narrowed certificate readers on the migration page


Crank-Git:
## What is wrong

`docs/migration-0.6-to-1.0.md` holds eleven rows under `## The breaking changes`, and none
of them records #319.

#319 narrowed the handlers of `compute_ja4x_from_pem` and `compute_ja4x_from_der`. An input
that returned `None` in version 0.6.0 can now raise. `CHANGELOG.md` records the change
under round 133, and the sweep of #395 named it in the table of pull request #397 as
"Recorded. An input that returned `None` can now raise."

`docs/specs/features/08-documentation.md` asks the page to list every breaking change this
release makes:

> The migration page lists every breaking change this release makes.

The page therefore misses one row today.

## Evidence

- `CHANGELOG.md`, the round 133 entry of #319.
- Pull request #397, the row `compute_ja4x_from_pem and compute_ja4x_from_der narrow except Exception`.
- `docs/migration-0.6-to-1.0.md:17-29`, the whole breaking-change table.
- `tests/test_release_notes.py::test_the_release_notes_name_the_narrowed_certificate_readers`
  records the gap and its reason.

## What to build

Add one row to `## The breaking changes` of `docs/migration-0.6-to-1.0.md`. State the old
form, the new form, the reason and the record `Round 133, #319`.

`tests/test_release_notes.py::test_the_release_notes_repeat_the_record_cell_of_the_migration_page`
already holds the release notes of `CHANGELOG.md` against that table, so the new row needs
no change to the notes.

## Who found it

#66 found the gap while it wrote the release notes of version 1.0.0. #65 owns the page, so
#66 recorded the gap and edited no page of another issue.

Crank-Git:
## Adopted into Epic 8 by the project manager, 2026-08-09

**This is a defect in a page Epic 8 itself created.** `docs/migration-0.6-to-1.0.md` shipped with #65 in this same epic, and #66 found this while writing the release notes against it.

**Shipping the epic with a known defect in its own new page would be the wrong call**, so this runs before the Epic 8 gate rather than after it.

**#399 and #401 are worked together by one worker**, because both change the same file and two workers on one file would collide for no gain. The Epic 8 pull request closes them both.

**Neither is #66's to repair and #66 was right to file them.** The page belongs to #65.

Crank-Git:
[...]

Issue #389, opened 2026-08-09
Stop the memory ceiling control from failing under load


Crank-Git:
Observed by two workers independently on 2026-08-09, on the same day the case was repaired at the batch #361 gate. **Neither worker changed it and neither owns it.** The project manager files it.

## The observations

`tests/test_memory_bounds.py::TestTheStatedMemoryCeiling::test_a_smaller_entry_count_holds_less_resident_memory` fails and then passes on an unchanged tree.

| Worker | Result | Load |
|---|---|---|
| #62 | failed once on the first full run, passed on two later runs | busy, then quiet |
| #64 | failed on three full runs, passed on two | **every failure at load average above 9 on a 10-core host; both passes below it** |
| the project manager | passed | load average 8.12 |

**#64 disproved its own change as the cause**, which is why this is a case defect and not a regression: the suite without its new file reads 2251 passed, its file with the memory module reads 39 passed, and `measure_resident_memory` runs in a separate interpreter at `tests/test_memory_bounds.py:367` that imports only `ja4plus` and `scapy`. **No file under `ja4plus/` changed in either pull request.**

## What the case measures, and why load moves it

The case runs two subprocesses and compares the resident growth of a shipped run against a run whose state tables carry a bound of 100. **The comparison is a ratio, and it asserts the control grows less than `CONTROL_GROWTH_RATIO`, which is 0.85, of the shipped growth.**

**Both readings are taken on a host that is doing other things.** A loaded host changes allocator behaviour, page reclaim and scheduling, and it moves the two readings by different amounts. **The ratio is then a measurement of the host as much as of the entry count bound.**

**The measured margin is not thin, which makes the flake more interesting rather than less.** On Linux the ratio reads 0.424 against the 0.85 threshold and on macOS it reads 0.732 to 0.769. **A control failing at load while sitting that far from its threshold means load moves the readings a long way, not that the threshold is badly chosen.**

## Why it matters more than an ordinary flake

**This case is a control. It exists to prove that another assertion is not vacuous** — that the ceiling case measures the traffic rather than the interpreter.

**A control that fails when nothing is wrong teaches a reader to disregard it, and a disregarded control is worse than no control**, because the suite still reports that the comparison is made. **This project has recorded seventeen instances of a comparison that is never made reading as a comparison that passes. A control nobody believes is the eighteenth in waiting.**

[...]

PR #402, opened 2026-08-09
Hold the migration page to the record for #399 and #401

Crank-Git:
Repairs #399, #401 and #403. All three change `docs/migration-0.6-to-1.0.md` or the record
of it, so one worker carried them and the commits stay separate.

## What was wrong

**#401 — one citation named two rounds and only one records the change.** The page cited
one row under `Round 122, #59 and #364`. The Changelog table of `docs/specs/spec.md` holds
round 122 for #59, which records `lookup_many` and the source field, and round 123 for
#364, which records the item access of a `LookupResult`. A reader who followed `Round 122`
for the item-access rule reached the wrong row.

**#399 — the page held eleven breaking changes and none recorded #319.** That round
narrowed the handlers of `compute_ja4x_from_pem` and `compute_ja4x_from_der`, so an input
that returned `None` in version 0.6.0 can now raise. `CHANGELOG.md` records it under round
133 and the release notes carry it. The page did not.

## What the repair is

The page holds thirteen rows. The conflated row parts into two, each naming the round that
records it, and the certificate readers gain one row with the old form, the new form, the
reason and `Round 133, #319`.

**The two edits are the smaller half.** Both defects sit in a page this epic created, and
#66 found them only because it wrote the release notes against the page and compared. The
question is why the page's own cases did not.

**#66's first citation case compared two copies of one claim.** It read the `Record` cell
of the release notes against the `Record` cell of the migration page. Both files held the
same wrong round, so the two agreed and the case passed on the very error #401 records.
That is the eighteenth recorded instance of a comparison that never runs.

Three new cases in `tests/test_release_notes.py` make the page, and the notes' claim about
the page, answerable to the record.

1. `test_every_citation_of_the_migration_page_names_the_row_that_records_it` reads every
   `Round N, #M` citation of the page against the Changelog table of `docs/specs/spec.md`,
   and it opens no second file. `_specification_rows`, `_citations` and
   `_wrong_citations` now serve the notes case and the page case, so one reader holds both
   documents to one table.
2. `test_every_breaking_change_of_the_record_reaches_the_migration_page` reads the
   direction #66 left open. `test_every_breaking_change_of_the_migration_page_reaches_the_release_notes`
   reads the page into the notes, and a page that holds fewer changes than the record
   passes it. That is why the gap of #319 survived a green suite. The record is the two
   release-notes tables of `CHANGELOG.md` and every entry that opens with the `**BREAKING`
   marker, which is 20 issues. **The marker alone would not have caught #319**: eight
   entries carry it and all eight already reached the page.

**#403 — the sentence this repair falsified.** `CHANGELOG.md` read *"One breaking change
reaches this record and reaches no row of the migration page"*. The row added under #399
made that untrue. The worker filed #403 and declined the edit, because #66 owned the
release notes; **#66 sub-merged as `2a1fb33` before the worker finished**, and the project
manager widened the scope of this round on the record.

The paragraph now states that the record and the page name the same breaking changes, and
it keeps the history: #66 found the gap, #399 added the row, #403 corrected the paragraph.
It cites no round, because the interface table above already carries `Round 133, #319`.

3. `test_the_release_notes_report_the_gap_the_record_shows_and_no_other` connects the prose
   to the measurement. **No case read that paragraph, which is why it went stale silently.**
   The prose asserted the same thing as case 2, in the opposite direction, and nothing
   connected them. This case reads a present-tense gap claim of the notes against the set
   difference of the record and the page — never against a second copy of the claim. It
   fails in both directions. `GAP_CLAIM` reads a present-tense claim alone, and that is
   deliberate: a past-tense sentence records a gap an issue closed and stays true, so the
   paragraph keeps its history. The docstring states the asymmetry.

## Six mutations prove the cases discriminate, and each was restored

| Mutation | Result |
|---|---|
| `Round 123, #364` back to `Round 122, #364` | The citation case fails alone: `round 122 records no #364` |
| `Round 999, #364` | The citation case fails: `the migration page cites rounds the specification does not hold: [999]` |
| The `Record` cell of the new row from `Round 133, #319` to `Round 135, #76` | The row count holds at thirteen, and the completeness case fails alone: `the record names these breaking changes and the migration page omits them: [319]` |
| The removal of the whole `#319` row | Five cases fail, the row floor among them |
| The stale gap sentence restored, with the page intact | The gap case fails alone: `the release notes claim the migration page misses a breaking change, and the record shows none: 'reaches no row of the migration page'` |
| `Round 135, #76` again, with the corrected paragraph | The gap case fails beside the completeness case: `the record shows these breaking changes the migration page omits, and the release notes report no gap: [319]` |

The floors rise with the page. `MINIMUM_BREAKING_CHANGES` and `MINIMUM_MIGRATION_ROWS`
read 13, `MINIMUM_MIGRATION_CITATIONS` reads 14, and `MINIMUM_RECORD_ISSUES` reads 20.

## The record, re-measured on the base commit `2a1fb33`

The Changelog table of `docs/specs/spec.md` holds 135 assigned rounds, numbered 1 through
135 with no gap and no duplicate, and 7 rows reading `TBD`.

## Two findings, reported and not repaired

**`_citations` is over-strict on wrapped prose.** It reads the first `Round` of a line and
every issue behind it, which is right for a table row and wrong for a sentence that wraps.
It failed on this round's own corrected paragraph, where `Round 133, #319. #66 found` read
as `round 133 records no #66`. **It fails loud and never passes silently**, so the
paragraph drops the round rather than the reader loosening. A tighter pattern that reads
the citation form is a later round's call, not this one's.

**The intro of the release notes under-describes its own guard.** `CHANGELOG.md:23-26`
states that a change the migration page records reaches a row below or fails a case, and
it names the new record-to-page direction not at all. The sentence is true and incomplete,
so this branch leaves it: the ruling was to fix only what this change falsified.

## How it was tested

Environment: `python3 -m venv .venv && .venv/bin/pip install -e ".[dev,docs]"`,
`PYTHONDONTWRITEBYTECODE=1`, every `__pycache__` outside `.venv/` cleared.

- `pytest tests/ -m "not spec_validation"` — **2502 passed**, 2 skipped, 8 xfailed, 98
  subtests. The base commit reads 2499, and the three new cases are the difference.
- `pytest tests/ -m spec_validation` — **1532 passed**, 143 skipped, **134 xfailed**
  against the **134 keys** of `tests/foxio_deviations.json`. The register invariant holds.
- `ruff check ja4plus/ tests/` — clean. `ruff format --check` — 175 files formatted.
- `mypy --strict ja4plus/` — no issue in 31 source files.
- **Coverage holds at 94**, 273 missed of 4232 statements. No file under `ja4plus/`
  changes, so no line of the package moves.
- `mkdocs build --strict` — exit 0. Both new citations render on
  `site/migration-0.6-to-1.0/index.html`.
- `TestTheStatedMemoryCeiling::test_a_smaller_entry_count_holds_less_resident_memory`
  passed on this quiet host. #389 owns it and this branch does not touch it.

Refs #399, #401, #403.


Crank-Git:
Filed the out-of-scope finding as #403: the release notes still state that one breaking change reaches no row of the migration page, and #399 closed that gap. The release notes are out of scope here and #66 owns them, so this branch reports the sentence and edits it never.

Crank-Git:
## Self-review, two lenses in parallel

**Correctness — clean.** The reviewer read the diff, checked every factual claim against the source, and reproduced all four mutations independently. It confirms: the `0.6.0` form of both certificate readers was a bare `Exception` catch (`git show v0.6.0:ja4plus/__init__.py:65` and `:ja4plus/fingerprinters/ja4x.py:281`, where `(ValueError, TypeError, Exception)` is a bare catch in effect); both readers name a narrowed list today; the parted rows match Changelog rows 122 and 123; the refactor of `test_every_citation_of_the_release_notes_names_the_row_that_records_it` into `_wrong_citations` loses no detection; and no vacuous-pass path exists, because `_migration_breaking_rows` and `_record_issues` each assert a floor before they return. It found no correction to make.

**Writing standard — six findings, four taken.**

Taken:
- Synonym rotation, rule 7. The added prose wrote `document` where the surrounding prose already writes `file`. Every added use now reads `file`.
- Sentence length, rule 1. Three sentences of the `CHANGELOG.md` entry and two of the Changelog row ran past 25 words. Each parts at its join.
- The entry title named the repaired state, where the adjacent entry of #63 names the defect. Retitled.
- A 33-word docstring sentence in `_wrong_citations` parts into two.

Declined, with the reason:
- **`finding` as an `-ing` noun.** `docs/specs/spec.md` uses the word 15 times and `CHANGELOG.md` 8 times. It is established project vocabulary, and changing it here would rotate a term rather than repair one.
- **The remaining long sentences of the new Changelog row.** Every adjacent row of that table runs longer, and the reviewer records the precedent itself. A row that alone read short would part from the table it sits in.

Pushed as `3cd7021`. The gate is unchanged and green.

Crank-Git:
## #403 absorbed, on the project manager's ruling

The boundary I drew was right when I drew it and wrong once #66 sub-merged as `2a1fb33`. The paragraph belonged to nobody, and my own row is what made it false. Corrected rather than deleted, so the record that the gap existed and that #399 closed it survives.

**I re-read the whole block, not the one sentence.** Three neighbouring claims lean on nearby premises:

- `CHANGELOG.md:20` — "That table holds eight rows. Seven of them move a value..." — **unaffected**, the fingerprint table is untouched.
- `CHANGELOG.md:61-64` — the two changes #395 judged marginal, #47 and #94 — **unaffected**, and both citations still resolve.
- `CHANGELOG.md:23-26` — states the guard set as page-to-notes only, and names the new record-to-page direction not at all. **True but incomplete.** Reported, not fixed: the ruling was to fix only what this change falsified.

## The case

`test_the_release_notes_report_the_gap_the_record_shows_and_no_other`. It reads a present-tense gap claim of the notes against **the measured set difference of the record and the page** — not against the migration page's copy of the claim. It fails both ways, and two mutations prove it:

- the stale sentence restored, page intact, fails it **alone**;
- the `Round 135, #76` mutation, row count held at thirteen, fails it **beside** the completeness case, demanding the notes report the real gap.

`GAP_CLAIM` matches present tense only. A past-tense sentence records a closed gap and stays true, which is what lets the corrected paragraph keep its history. The docstring states that asymmetry rather than hide it.

## One thing worth your attention

**My first draft of the corrected paragraph failed my own citation case.** `_citations` reads the first `Round` of a line and every issue behind it, so the wrapped `Round 133, #319. #66 found` read as `round 133 records no #66`. Right for a table row, over-strict for prose.

It **fails loud and never passes silently**, so I dropped the round from the paragraph rather than loosen the reader — the interface table above already carries `Round 133, #319`. A tighter pattern that reads the citation form is a later round's call. Recorded in the Changelog row.

Pushed as `c9c54d3`. Gate: 2502 passed (2499 on base, three new cases), 134 xfailed against 134 register keys, coverage 94, `ruff`/`mypy --strict`/`mkdocs build --strict` clean.

PR #404, opened 2026-08-09
Stop the memory ceiling control from failing under load

Crank-Git:
Refs #389. The Epic 8 pull request closes it.

## What this changes

The third control of `TestTheStatedMemoryCeiling` compared the **resident growth** of two
runs. It now compares the **count of memory blocks** each run holds, and it carries the
name `test_a_smaller_entry_count_holds_fewer_memory_blocks`.

**No file under `ja4plus/` changes and no fingerprint moves.** Five files change:
`tests/memory_ceiling_run.py`, `tests/test_memory_bounds.py`,
`docs/specs/features/03-concurrency-safety.md`, `docs/specs/spec.md` and `CHANGELOG.md`.

## Why the case failed, measured rather than reasoned

**A resident reading states what the host left in memory. A block count states what the
program holds.** A host under memory pressure reclaims the pages of a running process,
and it reaches the run that holds the most memory first. The reclaim moves the two runs
by different amounts, and it moves the growth ratio toward one.

**The reproduction ran on `bigboy`**, Ubuntu 24.10, 56 cores, 137 GiB, `python3.12`, at
30000 packets. Each measurement run ran under a memory limit of its own:

```
systemd-run --user --scope -q -p MemoryMax=<limit> .venv/bin/python \
    tests/memory_ceiling_run.py --packets 30000 --connections 3000 --bound <0|100>
```

A limit confined to one run reclaims that run's pages and leaves the host untouched, so
nothing else on the box was disturbed. `uptime` was read first, every load was bounded by
`timeout`, and every clone and process was removed afterwards.

| Memory limit | Resident growth, shipped and control | Resident ratio | Block ratio |
|---|---|---|---|
| none | 29.34 MiB, 12.72 MiB | 0.434 | 0.392 |
| 80 MiB | 10.98 MiB, 10.56 MiB | **0.962** | 0.392 |
| 75 MiB | 6.13 MiB, 4.82 MiB | 0.786, floor fails | 0.392 |
| 70 MiB | 0.00 MiB, 0.00 MiB | no reading | 0.392 |

The base-commit case fails at 80 MiB on the ratio, and at 75 MiB and 70 MiB on the floor.
One base-commit run read `the shipped run added 0.87 MiB, which is too little to read`.
**The block growth of the shipped run held between 376952 and 376965 across all four**, so
the four readings differ by 13 blocks.

**A deliberate processor load reproduced nothing, and that reading is what names the
mechanism.** 56 processes that hold a processor busy held the load average at 58 on 56
cores for three rounds, which matches the ratio of load to cores at every reported
failure. The runs took 59 to 67 seconds against 27 quiet, and the resident ratio read
0.361, 0.445 and 0.407. **Memory pressure moves this reading and processor contention
does not.**

**Why the margin was thin on macOS and wide on Linux.** Each growth reading holds a term
the entry count bound governs and a term it does not, because `BaseFingerprinter.fingerprints`
grows with the packet count in both runs. The governed term is 57 percent of the Linux
reading and 25 percent of the macOS reading, so the same perturbation crosses the
threshold on macOS first. That is why the issue saw 0.424 on Linux and 0.732 to 0.769 on
macOS.

## The shape chosen, and the one rejected

**Shape 3, a quantity the host does not move.** `sys.getallocatedblocks` costs nothing,
needs no extra run, and read 0.392 at all four memory limits.

**Shape 2, repeat and require agreement, is rejected by measurement.** Memory pressure
lasts as long as the host holds it, so every round reads the same clipped pair. Three
rounds at the 80 MiB limit read **1.047, 0.885 and 1.045**, so the median fails and
agreement across three rounds fails. It also costs three times the wall clock, at 57
seconds a round on that host.

**The run collects the cyclic garbage before it counts the blocks.** A count taken between
two collections holds the garbage the collector has not reached, and the point it reaches
moves with the allocation history. Without the collection, two macOS readings gave a
control growth of 230616 and 247991, at a ratio of 0.612 and 0.664. With the collection
three rounds read 147562, 147566 and 147569 at a ratio of 0.398 each. **The collection
runs after the resident reading at both points, so it moves no part of the ceiling
reading.**

## The two thresholds

**`CONTROL_GROWTH_RATIO` stays 0.85.** The true reading is 0.392 on Linux and 0.398 on
macOS, and the mutation reads 1.361 and 1.367. The threshold sits between the two with a
wider margin on both sides than it had.

**`CONTROL_FLOOR_MIB` is retired, and the floor now reads four blocks for each packet.** A
MiB floor is itself the reading the host moves: under the 70 MiB limit the shipped run
added 0.00 MiB and 376954 blocks. The measured rate is 12.37 blocks for each packet at
30000 packets, 13.99 at 5000 and 15.08 at 1000, so **the reading sits 3.1 times above the
floor where the MiB reading sat 2.9 times above the MiB floor.** The guard holds the
strength it had. `CONTROL_FLOOR_MIB` had no other reader. **A first draft of this change
set the rate to 1, which is a looser guard; the self-review caught it and `e164d1b`
repaired it.**

## Proof, both ways

**Test-driven.** `a7edf8d` adds the two cases and fails with
`NameError: name 'traffic_growth_blocks' is not defined`. `1d792fe` makes them pass.

**The repair passes under the pressure that reproduced the failure.** The case itself ran
against readings taken at each of the four limits, including the 70 MiB limit where the
resident growth reads 0.00 MiB. It passes at all four.

**The case still fails when the entry count bound stops holding.** With `CONTROL_BOUND`
raised from 100 to the shipped 10000 the case fails and reads:

```
a bound of 10000 added 506935 blocks and the shipped bounds added 370945 blocks, a ratio of 1.367, so the entry count changed nothing
```

The ratio reads 1.361 on Linux under the same mutation. **The file was restored after each
mutation**, and `git status` is clean.

**The ceiling case and the two other controls are unchanged, and all three pass under the
mutation run.** The ceiling reads a high-water mark and asserts an upper bound, so memory
pressure only lowers that reading and never turns it red.

## Local verification

- `pytest tests/ -m "not spec_validation"` — **2485 passed**, 3 skipped, 8 xfailed, 98
  subtests, coverage **94% and 273 missed**. The same run on the base test files reads
  2483 passed and **94% and 273 missed**, so coverage does not fall and the two new cases
  are the whole difference.
- `pytest tests/ -m spec_validation` — 1532 passed, 143 skipped, **134 xfailed** against
  **134 register keys**.
- `ruff check ja4plus/ tests/` clean, F401 enforced. `ruff format --check` clean on 173
  files. `mypy --strict ja4plus/` clean on 31 source files.
- `tests/test_changelog_round_agreement.py`, `tests/test_specification_changelog.py`,
  `tests/test_breaking_change_record.py` and `tests/test_documentation_samples.py` pass
  together at 33 cases.
- Every run set `PYTHONDONTWRITEBYTECODE=1`, and every `__pycache__` outside `.venv/` was
  cleared first.

## Two observations, not repaired here

**The second control reads the resident growth too**, and it asserts `growth > 0.0`. Under
the 70 MiB limit that reading is 0.00 MiB, so the second control fails there as well. It
needs far more pressure than the third control did, no worker reported it, and #389 scopes
this change to the third control. Recorded rather than changed.

**Five workers reported the flake, where issue #389's body names two.** The project
manager's comment on #389 records the other three, and the Changelog rows follow that
comment.

Verified against: https://man7.org/linux/man-pages/man5/proc_pid_statm.5.html (retrieved 2026-08-09)
Verified against: https://docs.python.org/3.12/library/sys.html#sys.getallocatedblocks (retrieved 2026-08-09)
Verified against: https://www.freedesktop.org/software/systemd/man/latest/systemd.resource-control.html (MemoryMax, retrieved 2026-08-09)


Crank-Git:
## Self-review, three lenses

Three reviewers read the diff: correctness, the writing standard, and an audit of every
number against the three records that carry it. **Nine findings were repaired in
`e164d1b` and two are declined with a reason.**

### The finding that changed the code

**The new floor was weaker than the one it replaced, and the pull request called it
stricter.** The correctness lens read the numbers the other way round and it was right. A
floor of one block for each packet puts the reading 12.6 times above the floor, where the
retired 10.0 MiB floor sat 2.9 times below its reading. **A floor further below the
reading is a looser guard, not a tighter one.**

`CONTROL_FLOOR_BLOCKS_PER_PACKET` now reads **4**. The measured rate is 12.37 blocks for
each packet at 30000 packets, 13.99 at 5000 and 15.08 at 1000, so the reading sits **3.1
times above the floor** at the default size and the guard holds the strength the MiB
floor held. The rate was measured at all three sizes rather than assumed from one.

### The eight prose and record repairs

- `measure_resident_memory` documented a return shape that omitted `idle_blocks` and
  `final_blocks`, which callers now read.
- `test_the_recorded_resident_pair_fails_the_threshold_the_block_pair_passes` named two
  behaviours. It is now
  `test_the_block_pair_holds_where_the_recorded_resident_pair_fails`, which states one
  behaviour with its condition first.
- Four sentences in `CHANGELOG.md` ran past the 25-word limit of the writing standard.
  Each is now two or three sentences.
- **"a spread of 13 blocks in 376955" read as arithmetic and was not.** 376955 is neither
  endpoint nor the midpoint. It now reads "the four readings differ by 13 blocks".
- **"three macOS rounds read a control growth of 230616, 247991" named three rounds and
  gave two numbers.** There were two readings without the collection, and three with it.
  All three records now say so.
- `spinners` was undefined jargon. It now reads "56 processes that hold a processor busy".
- One comment carried a condition after its action, and one carried "an order of
  magnitude" where every sibling record gave the exact multiple.
- The `## Terms` table gains `growth ratio`, because `resident ratio` and `block ratio`
  carry the argument of this change and neither was in the controlled vocabulary. The
  table already gained `memory block` and `memory pressure` in `63763ec`.

### Two findings declined

**The specification Changelog row is one long table cell.** The writing standard caps a
description sentence at 25 words, and several sentences in that row run past it. **Every
existing round row of that table is written the same way**, so this is the table's format
and not this change. Reformatting the record of 135 rounds is not #389's work. Recorded
here for the user to decide.

**`cyclic garbage` is not in the `## Terms` table.** It names a mechanism of the
interpreter rather than a concept of this project, in the same way `high-water mark` and
`allocator` do, and none of those hold a row. Adding a row for each interpreter mechanism
would grow the controlled vocabulary without settling any ambiguity.

### What the audit confirmed

Every number that appears in more than one of the pull-request body, the code comments,
`CHANGELOG.md` and the `docs/specs/spec.md` row agrees, and the arithmetic of every ratio
follows from the pair of numbers stated beside it. The two recorded readings in
`TestTheBlockReading` imply a resident ratio of 0.962 and a block ratio of 0.392, which
is the split the case asserts. The diff touches no file under `ja4plus/`.

### Gates re-run after the repairs

Unit **2485 passed** at **94% and 273 missed**, conformance **1532 passed and 134 xfailed**
against **134 register keys**, the four guard files **33 passed**, `ruff` clean with F401
enforced, `ruff format --check` clean on 173 files, `mypy --strict` clean on 31 source
files. **The mutation still kills the case** at a ratio of 1.367, and the file was
restored.

Issue #410, opened 2026-08-09
Run the five gates on the granted Linux host


Crank-Git:
## What to build

**Warning: the host is shared. It runs an Elasticsearch node and continuous-integration
runners. Read `uptime` before each run, and start no load generator.** A deliberate
processor load is barred, and #389 already records that a processor load reproduces
nothing.

`.issue-flow.json` records the host: `ssh e@bigboy`, Ubuntu 24.10, kernel 6.11, 56
cores, 137 GiB, `python3.12` only. **It records no checkout path, and no run has used
the host yet.** The first step confirms whether a checkout exists and states where it
is.

Run the gates on that host against the head of the branch under test, and record the
verbatim output of each command in a comment on this issue.

1. `ssh e@bigboy uptime`, before the first gate and after the last one.
2. `python3.12 -m venv` and `pip install -e ".[dev]"` in a checkout on that host.
3. `python -m pytest tests/ -m "not spec_validation" --cov=ja4plus --cov-report=term`.
   **Use the directory form of `--cov`. Never pass a dotted module name**: it makes 48
   JA4L and QUIC cases fail, and `.claude/rules/conformance.md` records the cause.
4. `python -m pytest tests/ -m spec_validation`.
5. `python -m pytest tests/fuzz/`.
6. `ruff check ja4plus/ tests/`, `ruff format --check ja4plus/ tests/`,
   `mypy --strict ja4plus/`.
7. `python -m pytest tests/test_memory_bounds.py -k TestTheStatedMemoryCeiling`. This is
   the Linux reading of the memory ceiling, and #389 measured the mechanism that moves
   it.

**The caveat is recorded beside every reading.** This host verifies a Linux code path.
It predicts no `ubuntu-latest` result, because the runner holds another kernel, another
processor count, another memory size and another Python version. Write the caveat in the
issue comment and in the feature document.

**The host runs `python3.12` only.** The 3.9, 3.10, 3.11 and 3.13 rows of the matrix get
no reading here. State that gap in the comment rather than leaving it unstated.

Add no continuous-integration job. This run is manual and it duplicates none of the
eleven jobs.

## Acceptance criteria

- [ ] The issue holds the verbatim output of `ssh e@bigboy uptime` from before the first
      gate and from after the last gate.
- [ ] The issue states the checkout path used on the host, and the commit that checkout
      held, and that commit equals the head of the branch under test.
- [ ] The issue holds the verbatim summary line of
[...]

Issue #409, opened 2026-08-09
Fingerprint a capture from the installed source distribution


Crank-Git:
## What to build

**The user added this scope on 2026-08-09**, after the decomposition named it as a gap
it would not fill without a ruling.

`#408` installs the wheel. **Nothing installs the source distribution.** A
`pip install ja4plus` falls back to the source distribution where no wheel matches the
host, and that path runs the build backend on the user's machine. It can fail where the
wheel succeeds: a missing file in `MANIFEST.in`, a build dependency the wheel does not
need, a package-data rule that `setuptools` applies to one artefact and not the other.

Extend `tests/test_installed_wheel.py` with the source-distribution path, or add
`tests/test_installed_sdist.py` beside it and share the fixture. The marker
`installed_wheel` covers both, so the job of `#408` runs both.

The run repeats the three paths of `#408` against the source distribution:
`ja4plus --version`, `ja4plus analyze <capture> --format json`, and one script that
imports `ja4plus` and builds a `Processor`.

**Install the source distribution and not the wheel.** `pip install <the sdist>` will
build a wheel from it and install that, which is the path a user takes. Pass
`--no-binary ja4plus` so that `pip` cannot resolve a wheel from the local directory or
from an index instead of the artefact under test, and assert on the recorded install
output that the source distribution was the input.

**The data file is the likeliest failure and it needs its own reading.**
`ja4plus/data/ja4plus-mapping.csv` reaches the wheel through a package-data rule. A
source distribution that omits it produces a package that imports and then reports an
empty database, which reads as a working install to any case that only imports.

## Acceptance criteria

- [ ] `python -m build` writes a source distribution, and the case installs that file
      with `pip install --no-binary ja4plus <the sdist>`.
- [ ] The case asserts, from the recorded `pip` output, that the installed artefact was
      built from the source distribution and not resolved as a wheel.
- [ ] The case asserts that `ja4plus.__file__` of the created environment starts with
      the `site-packages` path of that environment, and holds no part of the repository
      root.
- [ ] The case asserts that the JSON output lines of `ja4plus analyze` from the
      source-distribution environment equal the output lines of the same command from
      the source tree, byte for byte.
- [ ] The case asserts that `ja4plus/data/ja4plus-mapping.csv` exists inside the
      installed package, and that its byte count equals the byte count of the file in
      the repository.
- [ ] The case asserts that a lookup through the installed package returns a record for
      one named fingerprint, so the data file is read and not merely present.
[...]

Issue #408, opened 2026-08-09
Fingerprint a capture from the installed wheel


Crank-Git:
## What to build

New file `tests/test_installed_wheel.py`, and one marker `installed_wheel` in the
`[tool.pytest.ini_options]` block of `pyproject.toml`.

The case builds the package, installs the wheel, and runs the package from outside the
source tree.

1. `python -m build` writes the source distribution and the wheel into a temporary
   directory.
2. `python -m venv` creates the environment. `pip install <the wheel>` installs it. The
   environment installs no development extra, so the run reads the shipped dependency
   list.
3. Every command below runs with the working directory set to the temporary directory,
   so the repository root is not on `sys.path`.

The run exercises three paths.

- `ja4plus --version` writes the version of `pyproject.toml`.
- `ja4plus analyze <capture> --format json` reads a copy of
  `tests/foxio_vectors/tls12.pcap` and writes output lines.
- One script imports `ja4plus`, builds a `Processor`, reads the same capture with
  `scapy`, and writes the fingerprints it collects.

The comparison is between the installed run and the source-tree run of the same command
on the same capture. The two outputs are equal byte for byte.

**A run that resolves the working copy proves nothing.** The case reads
`ja4plus.__file__` from the installed interpreter and fails when that path is not below
the `site-packages` directory of the created environment.

**This case duplicates no continuous-integration job.** The eleven jobs of
`.github/workflows/test.yml` install with `pip install -e ".[dev]"`, which resolves the
working copy. Add one job named `installed-wheel` to that file. A member pull request
carries `[skip ci]`, so this job fires at the batch gate and on a push to `dev` or
`master`, and it adds no cost to a member push.

#68 installs the wheel inside `.github/workflows/publish.yml`. **#68 imports this test
file rather than writing the check again**, and this issue states that in a comment at
the head of the file.

## Acceptance criteria

- [ ] `pytest tests/test_installed_wheel.py -m installed_wheel` passes on macOS and on
      Linux.
- [ ] The case asserts that `ja4plus.__file__` of the created environment starts with
      the `site-packages` path of that environment, and holds no part of the repository
      root.
- [ ] The case asserts that the JSON output lines of `ja4plus analyze` from the
      installed environment equal the output lines of the same command from the source
      tree, byte for byte.
- [ ] The case asserts that the fingerprint list the imported `Processor` produces in
      the installed environment equals the list the source tree produces, in order.
[...]

Issue #406, opened 2026-08-09
Epic 11: Pre-release validation


Crank-Git:
## Goal

**Produce the evidence that version 1.0.0 is fit to publish, before Epic 9 publishes it.**

The gates of this project prove that the source tree passes its own suite on
`ubuntu-latest` and on `macos-latest`. Four statements a user relies on hold no
measurement.

1. **Nothing runs the shipped package.** #68 and #69 read the contents of the wheel.
   No case installs the wheel and then produces a fingerprint through the public
   interface and through `ja4plus analyze`. No case installs the source distribution
   at all.
2. **Every Linux reading comes from a runner.** The development host is macOS. Batch
   #361 sub-merged with macOS proof and failed four `ubuntu-latest` jobs at the gate.
   The user granted one Linux host on 2026-08-09, and no run has used it yet.
3. **The largest hole in the evidence is open.** #206 names 976 candidate cases of the
   mutation sweep, and two of them are settled. A sweep of an earlier session applied
   eleven mutations to the QUIC guards, and ten of them left the suite green.
4. **The package states no throughput.** `Non-goals` of `docs/specs/spec.md` states
   that this project measures throughput and reports it.
   `docs/specs/features/03-concurrency-safety.md` states 512 MiB for the
   one-million-packet case, and that is the only performance number this project holds.

## Exit criteria

**The exit criterion of this epic is evidence, not a release. This epic publishes
nothing and moves no fingerprint.**

- One case installs the built wheel into a clean environment, produces a fingerprint
  through the public interface and through the command-line program, and compares the
  output against the output of the source tree. One case does the same for the source
  distribution.
- The five gates run on the granted Linux host, and the issue carries the verbatim
  output of each one. The recorded caveat stands: the host verifies a Linux code path
  and predicts no `ubuntu-latest` result.
- Every module that the sweep reaches is swept whole, and every candidate of those
  modules is settled: repaired and proved, or recorded as correct with the reason the
  mutation cannot reach it. The count of unsettled candidates is stated for each
  module, and one case holds each count.
- One recorded measurement states the packet throughput of the processor for a named
  packet run on a named host. **This epic sets no throughput target and adds no gate
  on a rate.**
- The conformance suite reports 1532 passed, 143 skipped and 134 xfailed against 134
  keys of `tests/foxio_deviations.json`. No fingerprint moves in this epic.

## Scope decided by the user, 2026-08-09

The user selected all four gaps, and added the source distribution as a fifth after the
[...]

PR #417, opened 2026-08-09
Fingerprint a capture from the installed wheel

Crank-Git:
## What this changes

`FR-pre-release-validation-1` to `FR-pre-release-validation-7`. One case builds the
package, installs the wheel into a clean environment, and produces a fingerprint from
that environment.

- New file `tests/test_installed_wheel.py`, seven cases under the `installed_wheel`
  marker.
- New file `tests/conftest.py`, which deselects that marker from a run that does not name
  it. The `installed-wheel` job is the only runner of the marker, and the `conformance`
  job holds the same relation to `spec_validation`.
- New file `tests/test_installed_wheel_selection.py`, four cases that hold the
  deselection rule against three commands and against the job name.
- `pyproject.toml` gains the `installed_wheel` marker, and `build` joins the `dev` extra.
- `.github/workflows/test.yml` gains the `installed-wheel` job.
- `CHANGELOG.md` and `docs/specs/spec.md` each gain one `Round TBD` entry.

**No file under `ja4plus/` changes, and no fingerprint moves.**

## Why the check discriminates

Every other job of `.github/workflows/test.yml` installs with `pip install -e ".[dev]"`,
which puts the source tree on the import path. A case that only imports `ja4plus`
therefore measures the checkout. Three facts hold the case to the wheel.

1. Every command runs with the working directory set to a temporary directory, so the
   repository root reaches no `sys.path`.
2. The case reads `ja4plus.__file__` of the created environment. It fails when that path
   is not below the `site-packages` directory of the environment, or when it holds any
   part of the repository root.
3. `test_the_import_check_fails_for_the_source_tree` holds the same predicate against the
   interpreter that runs the suite, which the editable install resolves to the checkout,
   and it asserts that the predicate reports False.

**The assertion was proved live before it was trusted.** One mutation moved the working
directory of the clean interpreter to the repository root, and the case failed:

```
E       AssertionError: the clean environment imported /Users/christiancrank/Development/Personal/ja4plus/.claude/worktrees/agent-aa66fe6fa6a3c0f54/ja4plus/__init__.py, which is not below /private/var/folders/pj/8q0k6gh973ng3vsz__flc6n80000gn/T/pytest-of-christiancrank/pytest-1637/wheel-env0/env/lib/python3.14/site-packages
E       assert False
```

The mutation is reverted. `test_the_import_check_fails_for_the_source_tree` keeps the
same proof in the suite.

## The control case

Without a control the byte-for-byte comparison passes on any wheel that imports.
`test_the_lookup_reports_an_empty_database_without_the_mapping_file` installs the wheel
into a second clean environment and reads one probe twice. The probe writes the entry
count of the mapping and the answer for one named fingerprint.

| State | The probe wrote |
|---|---|
| Before the removal | `70 True` |
| After the removal of `ja4plus/data/ja4plus-mapping.csv` | `0 False` |

## The measured duration of the `installed-wheel` job

The issue asks for a number rather than an impression, so this section states four runs
rather than one.

| Run | Duration |
|---|---|
| 1 | 40.39 s |
| 2 | 24.25 s |
| 3 | 24.40 s |
| 4 | 22.22 s |

**The settled figure on this host is about 23 s.** The command is
`pytest tests/test_installed_wheel.py -m installed_wheel`, the host is macOS 26.0 on
arm64, and the interpreter is Python 3.14.3. The `pip` cache is warm in every run above.
The three parts measured alone are `python -m build` at 3.04 s, `python -m venv` at
1.28 s and `pip install <wheel>` at 1.39 s. The case creates two clean environments, so
it runs `venv` twice and `pip install` twice.

**An earlier figure of 12.88 s is withdrawn.** It came from one run directly after
another run, and three further runs did not reproduce it.

**A `ubuntu-latest` runner starts with a cold `pip` cache**, so it downloads `scapy`,
`cryptography`, `cffi` and `pycparser`. The job there takes longer than 23 s. This table
predicts no runner result, and the first run at the batch gate reports that number.

## Four findings

**0. `pip install` reports success when it installs the dependencies alone.** The
self-review found that `PYTHONPATH` reached the clean environment, and a measurement then
showed the effect. Where `PYTHONPATH` names the checkout, `pip` reads the requirement as
satisfied, installs `scapy`, `cryptography`, `cffi` and `pycparser`, installs no
`ja4plus`, and writes `Successfully installed` without `ja4plus` in the list. **A `pip`
report of `Successfully installed` therefore does not state that the package under test
is installed.** `_scrubbed_environment` now removes `PYTHONPATH` and sets
`PYTHONNOUSERSITE=1` for the build, for `venv`, for `pip install` and for every probe.
`create_clean_environment` then asserts that `<site-packages>/ja4plus` exists. The review
comment on this pull request holds the transcript of both directions. **#409 and #68
inherit this**, because both call the same two functions.

**1. A premise of #408 is contradicted.** The issue states "The eleven jobs of
`.github/workflows/test.yml` install with `pip install -e ".[dev]"`". The file held five
job definitions before this change: `lint`, `test`, `fuzz`, `samples` and `conformance`.
The `test` matrix expands to six job runs (five Python versions on `ubuntu-latest`, plus
`macos-latest` on 3.12), so the file held **ten job runs, not eleven**. The statement the
premise carries is still true of every one of them, so the case is unaffected.
`tests/test_installed_wheel.py` states the fact without the count.

**2. `build` joins the `dev` extra, and the plan said `pyproject.toml` changes the
markers list alone.** The plan's file list is a premise, and this is the reason it does
not hold. The `installed-wheel` job and a local run each need `python -m build`
importable, and `build` is in no extra today. The project manager reviewed this deviation
and adopted it. The `installed-wheel` job needs no extra `pip` step as a result.

**3. `addopts` cannot deselect a marker that a gate command overrides.** This repository
has no `addopts` key, and `spec_validation` is deselected by the `-m` expression each job
passes. `pytest` reads `-m` as one value, so a command-line expression replaces the one in
`addopts`. Measured:

```
$ pytest tests/test_installed_wheel.py -o addopts="-m 'not installed_wheel'" -m "not spec_validation" --collect-only -q
7 tests collected in 0.03s
```

**An `addopts` entry would read as a deselection and do nothing.** `tests/conftest.py`
carries the rule as a collection hook instead, and
`tests/test_installed_wheel_selection.py` holds the rule against three commands.

**One further conflict is for the epic, not for this pull request.**
`FR-pre-release-validation-26` states "No case of this feature set opens a network
connection." `FR-pre-release-validation-1` requires the install, and a behaviour rule
states "The clean environment installs no development extra, so the run reads the shipped
dependency list". `pip` therefore resolves `scapy` and `cryptography` from the index, and
that opens a network connection. The two requirements cannot both hold for this case. The
acceptance criteria of #408 name no network rule, so this pull request follows them and
reports the conflict.

## How this was tested

| Gate | Result |
|---|---|
| `pytest tests/test_installed_wheel.py -m installed_wheel` | 7 passed in 23.00s |
| The same command with `PYTHONPATH` naming the checkout | 7 passed in 37.61s |
| `pytest tests/ -m "not spec_validation"` | 2509 passed, 3 skipped, 8 xfailed, 1816 deselected, 98 subtests passed |
| `len(load_register())` | 134 keys, which equals the xfailed count |
| `ruff format --check ja4plus/ tests/` | 179 files already formatted |
| Line coverage, `--cov=ja4plus` | 94% (4232 statements, 273 missed), the same figure the base reports |

**The unit gate count fell by seven and rose by four**, from 2512 to 2509. The seven
`installed_wheel` cases left it under the deselection, and the four selection cases
joined it. The deselected count rose from 1809 to 1816 by the same seven.

`PYTHONDONTWRITEBYTECODE=1` was set and every `__pycache__` directory was removed before
each run above.

## What #409 and #68 reuse

**#409 owns the source-distribution path, and this pull request builds none of it.** The
build, the environment creation and the comparison each take the artefact as a parameter:
`build_artifacts(destination)` returns both the `wheel` key and the `sdist` key, and
`create_clean_environment(root, artefact, pip_arguments)` accepts the
`--no-binary ja4plus` arguments that `FR-pre-release-validation-8` states.
`CleanEnvironment.pip_output` holds the recorded `pip` output that
`FR-pre-release-validation-9` reads.

**#68 owns `.github/workflows/publish.yml`, and this pull request does not touch it.** The
head of `tests/test_installed_wheel.py` states that #68 imports this file rather than
writing the check again, so `build_artifacts`, `create_clean_environment`,
`package_file_of`, `resolves_outside_the_repository`, `run_probe` and `declared_version`
are public.

## External interfaces

```
Verified against: https://build.pypa.io/en/stable/reference/cli.html (retrieved 2026-08-09)
Verified against: https://pip.pypa.io/en/stable/cli/pip_install/ (retrieved 2026-08-09)
Verified against: https://docs.python.org/3/library/venv.html (retrieved 2026-08-09)
```

Four readings come from those pages and no recollection reached the code.

1. `python -m build` "By default, a source distribution (sdist) is built from the project
   root and a binary distribution (wheel) is built from the sdist." The case therefore
   gets both artefacts from one run.
2. `--outdir` / `-o` names the "output directory (defaults to `{srcdir}`/dist)". The case
   passes a temporary directory, because the default writes into the checkout.
3. `pip install <path-to-wheel>` installs the local file and resolves the declared
   dependencies from the index. `--no-deps` and `--no-index` are the options that stop
   that, and the case passes neither, because the clean environment must read the shipped
   dependency list.
4. `venv` bootstraps `pip` "Unless the `--without-pip` option is given", and it creates a
   `bin` directory on POSIX and a `Scripts` directory on Windows. The case reads the
   `site-packages` path from `sysconfig.get_paths()['purelib']` of the created
   interpreter rather than assembling `lib/pythonX.Y/site-packages` by hand.

`--disable-pip-version-check` appears in no option list of the `pip install` page, so the
case does not pass it.

## Not verifiable from inside the worktree

`pytest tests/test_installed_wheel.py -m installed_wheel` passes on macOS, and the
transcript is above. **No Linux result exists yet.** This pull request is a batch member
and carries `[skip ci]`, so no runner has read it. The `installed-wheel` job produces the
Linux result at the batch gate, and #410 owns the run on the granted Linux host.

Part of #406. Refs #408.


Crank-Git:
## Self-review, by lens

Two reviews read the diff: one for correctness and portability, one for
`.claude/rules/ste.md` and the `## Terms` table. Seven findings are accepted and fixed.
One is rejected, with the reason.

### Accepted — the isolation was incomplete, and the failure was worse than reported

**The review found that `PYTHONPATH` reached the clean environment.** A measurement then
showed a second effect that neither the issue nor the plan predicted.

`pip install <wheel>` reads the requirement as satisfied where the package already
imports. With `PYTHONPATH` naming the checkout, `pip` installed the four dependencies,
installed no `ja4plus`, and reported success:

```
Installing collected packages: scapy, pycparser, cffi, cryptography

Successfully installed cffi-2.1.1 cryptography-50.0.0 pycparser-3.0 scapy-2.7.0
```

`ja4plus` is absent from that list, and `<env>/bin/ja4plus` does not exist. **A `pip`
report of `Successfully installed` therefore does not state that the package under test
is installed.**

Three changes follow.

1. `_scrubbed_environment` removes `PYTHONPATH` and sets `PYTHONNOUSERSITE=1`. The
   build, `venv`, `pip install` and every probe read it. Every other variable stays,
   because `pip` reads the index settings and the proxy settings of the host.
2. `create_clean_environment` asserts that `<site-packages>/ja4plus` exists after the
   install. The check names the state where it happens.
3. The docstring records the measurement, so a later reader does not repeat it.

**Both directions are proved.** With the scrub removed and `PYTHONPATH` naming the
checkout, the guard fails:

```
E       AssertionError: pip wrote no /private/var/folders/.../wheel-env0/env/lib/python3.14/site-packages/ja4plus, and it reported: Processing /private/var/folders/.../dist0/ja4plus-0.6.0-py3-none-any.whl
E         Collecting scapy>=2.4.0 (from ja4plus==0.6.0)
```

With the scrub in place and the same hostile `PYTHONPATH` set, all seven cases pass.

### Accepted — three smaller correctness findings

- **The control case shared its cache directory with the comparison cases.** No probe
  writes a cache file today, so the isolation was accidental rather than stated. The
  control case now reads a directory beside its own environment, and a comment names the
  reason.
- **The two probes took their path in two styles.** `lookup_probe.py` was relative and
  `processor_probe.py` was absolute, and both worked only because `cwd` was passed. Both
  are absolute now.
- **The asymmetry between the install calls and the probe calls carried no comment.** The
  scrub above supersedes the finding, and both now read one function.

### Accepted — three writing findings

- **`created environment` was a fourth synonym for `clean environment`**, which the
  `## Terms` table defines and which lists `fresh environment`, `sandbox` and
  `virtualenv` as forbidden forms. `.github/workflows/test.yml` now says
  `clean environment`.
- **`artefact` collided with the function name `build_artifacts`**, and the rest of this
  repository uses `artifact`. Eighteen occurrences across
  `tests/test_installed_wheel.py`, `CHANGELOG.md` and `docs/specs/spec.md` now read
  `artifact`.
- **Four sentences passed the 25-word limit of STE rule 1**, in `CHANGELOG.md`,
  `docs/specs/spec.md` and one docstring. Each is split.

### Rejected — the Windows cross-drive case

The review reported that `os.path.commonpath` raises `ValueError` where two paths sit on
different Windows drives, and it noted that `_binary_directory` already branches on
`os.name == "nt"`.

**This finding is declined, and the reason is the failure mode.** A raised `ValueError`
fails the case. It does not pass it, so it produces no false report of a proven wheel. No
job of `.github/workflows/test.yml` runs Windows, so a handler here would guard a state
that nothing measures, and `CLAUDE.md` asks for no error handling of that shape. Where
this project adds a Windows job, the finding becomes real and the repair is one
`try`/`except` that fails with a named message.

### One observation outside the scope of #408

`docs/specs/features/11-pre-release-validation.md:248` and `:254` spell the word
`artefact`. #407 wrote that file and it is on the base, so this pull request changes no
line of it. The two occurrences are the only ones left in the repository after this
change.

### The corrected duration

The first measurement in the body of this pull request read 12.88 s, and it came from a
run with a fully warm `pip` cache directly after another run. **That number is
withdrawn.** Four consecutive runs measured 40.39 s, 24.25 s, 24.40 s and 22.22 s, so the
settled figure on this host is about 23 s. The body carries the corrected table.

Crank-Git:
## The deselection is in, and the mechanism is not `addopts`

**The change is made and every gate is green.** One part of the instruction could not be
followed as written, and this comment states why, with the measurement.

### `addopts` cannot hold this rule

**This repository has no `addopts` key.** `[tool.pytest.ini_options]` of `pyproject.toml`
held the `markers` list alone. `spec_validation` is not deselected by configuration
today: the `test` job passes `-m "not spec_validation"` and the `conformance` job passes
`-m spec_validation`, so the precedent is a command-line expression in each job.

**An `addopts` entry would not have worked, and the reason is the option semantics.**
`pytest` reads `-m` as one value, so a `-m` expression on the command line replaces the
one in `addopts`. The gate command names `-m "not spec_validation"`, which would replace
`-m "not installed_wheel"` and select the seven cases again.

Measured rather than recalled:

```
$ pytest tests/test_installed_wheel.py -o addopts="-m 'not installed_wheel'" -m "not spec_validation" --collect-only -q
7 tests collected in 0.03s
```

**All seven cases were still collected.** An `addopts` entry would have read as a
deselection and done nothing, which is the exact defect class this issue exists to
repair.

### What carries the rule instead

New file `tests/conftest.py`, with a `pytest_collection_modifyitems` hook. The rule is
one sentence: **where the `-m` expression does not name `installed_wheel`, the run
deselects every case that carries the marker.** A deselected case appears in the summary
line, so the count stays visible exactly as the 1809 `spec_validation` cases do.

| Command | Result |
|---|---|
| `pytest tests/ -m "not spec_validation"` | 7 deselected, no wheel build |
| `pytest tests/test_installed_wheel.py -m installed_wheel` | 7 collected, 7 passed |
| `pytest tests/` | 7 deselected |

**A rule a reader believes is not a rule a check tests**, so new file
`tests/test_installed_wheel_selection.py` holds it against all three commands and against
the job name. A rename of the marker, a lost `conftest.py`, or a job that stops naming
the marker each produce a green gate that builds no wheel. Those four cases carry no
marker, so the unit gate runs them, and they cost 0.82 s because each one runs
`--collect-only`.

### The gate counts, before and after

| Gate | Before the deselection | After |
|---|---|---|
| `pytest tests/ -m "not spec_validation"` | 2512 passed, 1809 deselected | **2509 passed, 1816 deselected** |
| Line coverage, `--cov=ja4plus` | 94% (4232 statements, 273 missed) | **94% (4232 statements, 273 missed)** |
| `pytest tests/ -m spec_validation` | 1532 passed, 143 skipped, 134 xfailed | **unchanged** |
| Register keys | 134 | **134** |
| `pytest tests/test_installed_wheel.py -m installed_wheel` | 7 passed | **7 passed in 23.00s** |

**2509 = 2512 − 7 + 4.** Seven wheel cases left the gate and four selection cases joined
it. `ruff check`, `ruff format --check` and `mypy --strict` report no finding.

### The other three items

- **`build` stays in the `dev` extra**, as instructed.
- **Criterion 1 is attested as met on macOS, with the Linux half named and routed.** It
  is not marked plainly true.
- **No requirement text changed.** The `FR-pre-release-validation-26` conflict stays
  reported in the body of this pull request as evidence, and this pull request resolves
  none of it.

Head commit `d616731`. Every commit carries `[skip ci]`, and no run has fired on this
branch.

Issue #414, opened 2026-08-09
Settle the sweep candidates of the interface modules


Crank-Git:
## What to build

Settle every candidate of `ja4plus/cli.py`, `ja4plus/watch.py`, `ja4plus/output.py`,
`ja4plus/processor.py`, `ja4plus/ja4db.py`, `ja4plus/types.py` and `ja4plus/__init__.py`.
Confirm that list against `git ls-files 'ja4plus/*.py'` before you start.

**#206 deferred three of these, and each deferral has expired.** It read that Epic 5 and
Epic 6 rebuild `cli.py`, and that Epic 7 rebuilds `ja4db.py`. #16, #17 and #18 are
closed. It read that #48 removes `collector.py`; #48 is closed and the file is gone. The
sample of #206 read 10 survived of 12 for `cli.py` and 10 of 12 for `ja4db.py`.

The procedure of `#412` applies without change.

Two hazards belong to these modules.

1. **An aggregate that flattens a structure and then counts inside it measures the
   flattening.** `ja4plus/output.py` writes one output line for each result, and
   `ja4plus/processor.py` collects results into a list. A case that joins the output
   lines into one string and then counts a substring measures the join. Read the list of
   results, or parse each output line back into an object, and compare field by field.
2. **A rejection case needs input that would succeed if the rejection were removed.**
   `ja4plus/ja4db.py` refuses a remote lookup until the caller opts in, and
   `ja4plus/cli.py` writes the disclosure notice. A case that asserts no request happened
   must use a call that would send a request with the opt-in check removed.

Write the settlement record to new file `docs/mutation_settlements/interface.md`, and the
count table to the file this module group owns under the layout `#411` created.

## Acceptance criteria

- [ ] Every module named above holds a run with `--max-per-module 0`, and the report of
      each run is attached to this issue.
- [ ] `docs/mutation_settlements/interface.md` holds one row for each candidate, with the
      verdict `repaired` and the case name, or `correct` and the reason.
- [ ] The row count equals the candidate count `tests/mutation_census.py` reports for
      those modules.
- [ ] For each repaired case the pull request holds the failure with the mutation live
      and the pass after the revert, with the `inspect.getsource` line for both.
- [ ] Every repaired output case reads a parsed output line or a result object, and
      asserts on a named field. A case asserts that the repaired cases call no
      `"".join` over the output lines before they compare.
- [ ] Every repaired lookup case sends a call that reaches the network when the opt-in
[...]

Issue #413, opened 2026-08-09
Settle the sweep candidates of the eleven fingerprinters


Crank-Git:
## What to build

Settle every candidate of `ja4plus/fingerprinters/`: `base.py`, `ja4.py`, `ja4d.py`,
`ja4d6.py`, `ja4h.py`, `ja4l.py`, `ja4s.py`, `ja4ssh.py`, `ja4t.py`, `ja4ts.py`,
`ja4x.py` and `__init__.py`. Confirm that list against
`git ls-files 'ja4plus/fingerprinters/*.py'` before you start.

The procedure of `#412` applies without change: sweep each module whole with
`--max-per-module 0`, name every test file that reads it, and settle each candidate with
the mutation proven live in both directions by `inspect.getsource`, with
`PYTHONDONTWRITEBYTECODE=1` set and `__pycache__` cleared.

Three readings from #206 apply here.

1. `ja4plus/fingerprinters/__init__.py` reported 9 survived of 9. It holds exports and
   names, and a survived string mutation there may be correct. **Record the reason rather
   than inventing a case for a name.**
2. A candidate the baseline reports as xfailed cannot fail. 134 conformance cases carry
   `xfail(strict=True)` from `tests/foxio_deviations.json`. Separate those from the rest
   and count them.
3. `ja4plus/collector.py` is gone. #48 removed it, so that row of #206 is closed.

**No repair moves a fingerprint.** A repaired case makes a comparison sharper. Where a
candidate reveals that a fingerprinter produces a wrong value, **this issue opens a
decision issue and changes no fingerprinter.** `CLAUDE.md` rule 1 binds every change
under `ja4plus/fingerprinters/`: the specification decides intent and schema, and the
vectors decide the exact bytes.

Write the settlement record to new file `docs/mutation_settlements/fingerprinters.md`,
and the count table to the file this module group owns under the layout `#411` created.

## Acceptance criteria

- [ ] Every module named above holds a run with `--max-per-module 0`, and the report of
      each run is attached to this issue.
- [ ] `docs/mutation_settlements/fingerprinters.md` holds one row for each candidate,
      with the verdict `repaired` and the case name, or `correct` and the reason.
- [ ] The row count equals the candidate count `tests/mutation_census.py` reports for
      those modules.
- [ ] For each repaired case the pull request holds the failure with the mutation live
      and the pass after the revert, with the `inspect.getsource` line for both.
- [ ] The pull request states, for each module, the count of candidates the baseline
[...]

Issue #412, opened 2026-08-09
Settle the sweep candidates of the protocol parsing modules


Crank-Git:
## What to build

Settle every candidate of the eleven modules under `ja4plus/utils/`: `quic_utils.py`,
`tls_utils.py`, `http_utils.py`, `ssh_utils.py`, `x509_utils.py`, `tcp_stream.py`,
`tcp_options.py`, `packet_utils.py`, `state_table.py`, `tunnels.py`, `loopback.py`.
Confirm that list against `git ls-files 'ja4plus/utils/*.py'` before you start; the
module set of the day is the module set, not this list.

**Start with `ja4plus/utils/quic_utils.py`.** A sweep of an earlier session applied
eleven mutations to its guards, and ten of them left the suite green. **That record is
not in this repository**, so this issue sweeps the module whole rather than naming the
eleven.

For each module:

```bash
export PYTHONDONTWRITEBYTECODE=1
find . -name __pycache__ -type d -prune -exec rm -rf {} +
python tests/mutation_sweep.py --module "ja4plus/utils/quic_utils.py" \
  --max-per-module 0 --tests tests/test_quic_utils.py \
  --tests tests/test_quic_multipacket.py --tests tests/test_quic_server_initial.py \
  --report quic_sweep.json
```

**Name every test file that reads the module.** A mutation the sweep applies to code that
the named files never reach reports nothing about those files.

Read each candidate in this order. **A case may be correct and the mutation wrong.**

1. Read the case.
2. Apply the mutation by hand, confirm it is live with `inspect.getsource`, and run the
   case. Record the failure.
3. Revert the mutation, confirm the revert with `inspect.getsource`, and run the case
   again. Record the pass. **Both directions, or the reading proves nothing.**
4. Repair the case, or record it as correct with the reason the mutation cannot reach
   it.

**A rejection case needs input that would succeed if the rejection were removed.** Most
candidates of `quic_utils.py` sit on a guard that rejects a packet. A case that feeds a
packet the parser rejects for a second reason passes with the guard removed, so it
measures nothing. Build the input so that the parser produces a fingerprint when the
guard alone is removed, and record that reading beside the case.

Write the settlement record to new file `docs/mutation_settlements/utils.md`, and the
count table to the file this module group owns under the layout `#411` created. **Write
no line of `docs/mutation_sweep.md` or `mutation_sweep.json`; `#411` owns those.**

## Acceptance criteria

[...]

Issue #411, opened 2026-08-09
Re-baseline the mutation sweep and publish a candidate census


Crank-Git:
## What to build

**The report `docs/mutation_sweep.md` was written on 2026-08-07 against a commit that no
longer exists on `dev`.** `.claude/rules/conformance.md` states that a checkpoint
belongs to one commit, because it keys each result on the position of the expression in
the file. Delete `mutation_sweep_checkpoint.jsonl` before the first run of this issue.

**Warning: the sweep writes to a module file. Commit or stash your work before you start
it.** The sweep refuses to start when `ja4plus/` holds an uncommitted change. If a killed
sweep leaves a change behind, run `git checkout -- ja4plus/`.

**Warning: stale bytecode makes a survived mutation indistinguishable from a reached
one.** Before every sweep run and every hand-applied mutation:

```bash
export PYTHONDONTWRITEBYTECODE=1
find . -name __pycache__ -type d -prune -exec rm -rf {} +
```

Then confirm the mutation is live, and confirm it in both directions:

```bash
python -c "import inspect, ja4plus.utils.quic_utils as m; print(inspect.getsource(m))" | grep -n "<the mutated line>"
```

**Measure the run before you start it.** The mutation count of the whole package at
`--max-per-module 0` is recorded nowhere in this repository.
`.claude/rules/conformance.md` states about 17 seconds for one suite run and about 80
minutes for 12 mutations for each module, so the product decides whether one run is
possible. Run `python tests/mutation_sweep.py --dry-run --max-per-module 0` first, state
the count in this issue, and split the run across checkpoints when the product is too
large. **Do not estimate the count. Read it.**

Then run the sweep against the current head:

```bash
python tests/mutation_sweep.py --max-per-module 0 \
  --report mutation_sweep.json --markdown docs/mutation_sweep.md
```

Then publish the census. New file `tests/mutation_census.py` reads
`mutation_sweep.json` and returns, for each module, the count of candidates and the
count the settlement records mark settled. New file `tests/test_mutation_census.py`
compares those counts against a committed table, so a count cannot silently become zero.
#206 asks for exactly that.

**The committed table lives in one file for each module group**, and not in one table
that three issues edit. `#412`, `#413` and `#414` run in parallel, and a shared table
makes each of them conflict with the other two. The census reads the directory.

**Read the report structure and never a flattened form.** `docs/mutation_sweep.md` is
one Markdown page, and a count taken from its lines counts the page layout rather than
[...]

PR #418, opened 2026-08-09
Partition the mutation sweep and publish the candidate census

Crank-Git:
## What this changes

Part of #411.

**This pull request runs no sweep and settles no candidate.** #411 measured the
whole-package sweep at 71.6 hours, and the project manager partitioned it across #412,
#413 and #414 on 2026-08-09. This branch carries the measurement, the three corrections it
produced, and the tooling the three sweeps read.

1. `tests/mutation_sweep.py` writes a `commit` key and a `Commit` row, and it exposes
   `DEFAULT_MODULE_PATTERNS`.
2. New `tests/mutation_census.py` reads a directory of sweep reports and a directory of
   settlement records.
3. New `tests/test_mutation_sweep_module_list.py` fails when the module list one sweep
   reads stops matching the tracked Python files of the package.
4. `FR-pre-release-validation-16` loses the defective pathspec, and
   `FR-pre-release-validation-16a` states the new rule.
5. `.claude/rules/conformance.md` carries the measured cost of one sweep and its date.

**No file under `ja4plus/` changes.** `git status --porcelain ja4plus/` reports nothing.
No fingerprint moves.

### One whole-package sweep costs 71.6 hours

```
$ python tests/mutation_sweep.py --dry-run --max-per-module 0
...
total 3545 runs

$ python -m pytest tests/ -q --tb=no -rfE -p no:cacheprovider
4037 passed, 146 skipped, 142 xfailed, 16 warnings, 98 subtests passed in 72.75s (0:01:12)
```

3545 runs at 72.75 seconds is 257,899 seconds. `.claude/rules/conformance.md:155` stated
"about 17 seconds" for one suite run, which is **4.3 times low**. The suite grew from the
tree #172 measured, and the stale figure is what let this issue plan an impossible run. The
rule now carries the measured number, its date, and the instruction to measure both again.

A checkpoint makes such a run resumable and it removes no work. The 15 slowest cases hold
37 seconds of the 73, so no single case dominates and no reduction of the suite is
available.

### The partitioned sweep is the method, and not a cheaper approximation

`tests/mutation_sweep.py:482` reads `tests = options.tests or ["tests/"]`, so a sweep runs
the whole suite for each mutation when `--tests` names nothing. The 72.75 seconds above is
that whole suite. A sweep scoped to one module group and to the test files that read it
runs a small suite for each mutation, and `tests/mutation_sweep.py:340` already directs
that method in the preamble of its own report.

**The union of the per-module sweeps is larger than the candidate set of one whole-package
sweep, and the union is the correct input for settlement.** A whole-package sweep names the
cases that no mutation of any module kills. A sweep of module X names the cases that no
mutation of X kills, and a case that module Y kills still reaches that list. A reader who
compares the two counts reads a regression that did not happen.
`.claude/rules/conformance.md` and `docs/specs/features/11-pre-release-validation.md` both
state it now.

### `git ls-files 'ja4plus/**/*.py'` lists 24 files and the package holds 31

```
$ git ls-files 'ja4plus/**/*.py' | wc -l
24
$ git ls-files 'ja4plus/*.py' 'ja4plus/*/*.py' | wc -l
31
```

Git reads `**` in a pathspec as one or more directories, so the pattern matches no file of
the top directory of the package. It drops `ja4plus/__init__.py`, `ja4plus/cli.py`,
`ja4plus/ja4db.py`, `ja4plus/output.py`, `ja4plus/processor.py`, `ja4plus/types.py` and
`ja4plus/watch.py`, which hold 556 of the 3545 mutations.

`FR-pre-release-validation-16` now names the working pair, and
`tests/test_mutation_sweep_module_list.py` compares `DEFAULT_MODULE_PATTERNS` against every
tracked Python file of the package. **The front matter of `.claude/rules/ste.md` and of
`.claude/rules/external-apis.md` holds the same `**` string, that glob follows the
gitignore rules, it matches every module, and this branch repairs neither.**

## Why the census groups by test file and not by module

`tests/mutation_sweep.py:593` builds the candidate set as one flat list over every module
the sweep read, so a candidate belongs to the sweep that named it and no module owns one.

A candidate is therefore keyed by the sweep that named it and by the case. Two sweeps may
name the same case, and that is one candidate for each sweep and not a duplicate claim.

## How this was tested

**Test-driven development.** Each case was written first and each one failed first.

The commit field, before:

```
$ python -m pytest tests/test_mutation_sweep_commit.py -q --tb=line
tests/test_mutation_sweep_commit.py:54: AttributeError: module 'tests.mutation_sweep' has no attribute 'head_commit'
4 failed in 0.02s
```

The module list, before:

```
$ python -m pytest tests/test_mutation_sweep_module_list.py -q --tb=line
E   AttributeError: module 'tests.mutation_sweep' has no attribute 'DEFAULT_MODULE_PATTERNS'
2 failed, 3 passed in 0.08s
```

The census, before:

```
$ python -m pytest tests/test_mutation_census.py -q --tb=line
    from tests import mutation_census
E   ImportError: cannot import name 'mutation_census' from 'tests'
1 error in 0.06s
```

The self-review found that the census raised on a half-written record. Before the repair:

```
$ python -m pytest tests/test_mutation_census.py -q --tb=line
E   KeyError: 'candidates'
tests/mutation_census.py:112: KeyError: 'candidates'
4 failed, 25 passed in 0.07s
```

After every change:

```
$ python -m pytest tests/test_mutation_census.py tests/test_mutation_sweep_commit.py \
    tests/test_mutation_sweep_module_list.py -q
38 passed in 0.12s
```

### The sweep names every case that pytest collects

| Test file | `collect_cases` | `pytest --collect-only -q` | Equal |
|---|---|---|---|
| `tests/test_documentation_samples.py` | 22 | 22 | yes |
| `tests/test_ja4h_cookie_list.py` | 10 | 10 | yes |
| `tests/test_foxio_rust_parity.py` | 202 | 202 | yes |

The third file includes the parametrized identifier
`test_every_shell_sample_runs[README.md:74]`, which is the wrapped shape an earlier guard
missed.

### The five gates

```
$ python -m pytest tests/ -m "not spec_validation" -q
2543 passed, 3 skipped, 1809 deselected, 8 xfailed, 16 warnings, 98 subtests passed in 87.66s

$ python -m pytest tests/ -m spec_validation -q
1532 passed, 143 skipped, 2554 deselected, 134 xfailed, 1 warning in 9.12s

$ python -c "import json;print(len(json.load(open('tests/foxio_deviations.json'))))"
134

$ ruff check ja4plus/ tests/
All checks passed!

$ ruff format --check ja4plus/ tests/
180 files already formatted

$ mypy --strict ja4plus/

$ git status --porcelain ja4plus/
```

The register invariant holds: 1532 passed, 143 skipped, 134 xfailed against 134 keys.

### Coverage does not fall

```
$ python -m pytest tests/ -q --cov=ja4plus --cov-report=term
TOTAL                                 4232    273    94%
```

The base reports the same 4232 statements, the same 273 missed and the same 94 percent. The
38 new cases read no module of `ja4plus`.

## The candidates the unmutated suite state already excuses

The issue asks for the count of candidates the report reports as skipped or as xfailed.
**No report exists on this branch, so that count is not stated here**; it belongs to the
reports #412, #413 and #414 write. The whole suite reports 146 skipped and 142 xfailed
today, so 288 collected cases cannot fail, and each one that reaches a candidate list
reaches it for a reason the suite already states.

## What #412, #413 and #414 read

Each issue runs one sweep of the form below and writes one report under
`docs/mutation_reports/`.

```bash
python tests/mutation_sweep.py --max-per-module 0 \
  --module "ja4plus/utils/tls_utils.py" --tests tests/test_tls_utils.py \
  --report docs/mutation_reports/412-utils.json
```

Each issue then writes one settlement record under `docs/mutation_settlements/`. The record
names the sweep it settles, so no file is shared and the three run at the same time.
`.claude/rules/conformance.md` holds the schema under `How to read the census`.


Crank-Git:
## Self-review

A reviewer read the diff for correctness, for the writing standard and for the repository
conventions. It found two defects and both are fixed in `3d60ebb`.

1. `census` rebuilt the candidate set inside a comprehension, so it built one set for each
   claim. A sweep of the whole package names more than a thousand candidates, so the cost
   grows with the square of that count.
2. `candidates` carried a `type: ignore` that suppressed nothing.

The reviewer checked three further questions and each one holds.

- **No other caller of `markdown_report` breaks.** `main` builds the one report dict that
  reaches it, and no code path renders a report from disk.
- **No new case is a vacuous comparison.** Each one asserts a value or a position that the
  reverted code does not produce, and the failure output before each change is in the pull
  request body.
- **`mypy --strict` reads `ja4plus/` alone**, by the decision `pyproject.toml:139` records,
  so the new files under `tests/` are outside that gate.

`FR-pre-release-validation-18` asks a reader to prove the commit is an ancestor of the head
of the branch. **No code here checks the ancestry**, because no report exists to check.
That requirement belongs to the issue that runs the sweep.

Crank-Git:
## Second self-review, after the project manager settled the route

A reviewer read the three commits that carry the partition. It found one blocking defect
and five smaller ones, and `e457b1b` fixes all six.

**Blocking. The census raised instead of reporting a fault.** `read_claims` indexed
`settlement["candidate"]` while `settlement_faults` read the same field with a default, and
`census` calls `read_claims` first. A settlement record that names no candidate therefore
crashed the whole census with `KeyError: 'candidate'`. A report that holds no `candidates`
key or no `modules` key crashed it the same way.

**That defect matters here more than the traceback suggests.** #412, #413 and #414 write
these two directories at the same time, so the census meets a half-written record by
design. A traceback from a dictionary lookup names the key alone, and a reader needs the
name of the broken file. `report_faults` now names each key a report does not hold, and
`read_claims` passes over a settlement that names no candidate. Four cases in
`TestARecordThatBreaksItsOwnSchema` hold the repair, and each one reproduced the crash
first.

**A report that holds no `candidates` key counts zero and states one fault.** A count that
falls to zero in silence is the shape #206 exists for, so the count alone is not enough.

**Two cases measured git and never reached the sweep.**
`test_the_double_star_pathspec_drops_the_processor_and_the_command_line_program` now
measures the module list the sweep reads. `test_the_default_patterns_reach_the_depth_of_every_tracked_module`
replaces a bare depth reading, and it fails when a module sits below depth 2.
`TestThePathspecThatDropsSevenModules` states in its docstring that the pair measures git
and records the reason, so no reader takes it for a check of the sweep.

**Four sentences ran past the 25-word limit of `.claude/rules/ste.md`**, in
`FR-pre-release-validation-16a`, in two paragraphs of `.claude/rules/conformance.md` and in
the module docstring of `tests/mutation_census.py`. Each one is split, and the four-way
conjunction that listed what the census names is now a vertical list.

**One stale reading is marked stale.** `docs/specs/features/11-pre-release-validation.md`
still cited the 976 candidates of the report of 2026-08-07 as a current count. It now
states that the count comes from one sweep against a commit `dev` no longer holds.

The reviewer confirmed four things that hold. `unclaimed` is correct when two sweeps name
the same case. `unknown` excludes every legitimate claim. `settlement_faults` reports
nothing twice. The settlement schema in `.claude/rules/conformance.md` matches field for
field what `tests/mutation_census.py` reads.

Issue #433, opened 2026-08-09
Give a module-body mutation a reader in the cover rule


Crank-Git:
## What the measurement found

`.claude/rules/conformance.md`, section `How to scope one sweep: the minimal cover`,
states step 2 of the cover procedure:

> Subtract the lines the import runs. Every test file that imports `ja4plus` runs every
> module-level constant, so a cover built on import reach names every file.

**The step is correct for its purpose and it defeats the purpose of the cover for a module
whose mutation lines sit in the module body.** A cover exists so that every mutation keeps
a reader. A line the import runs has every test file as a reader, so the tie is broken by
the second rule, greatest lines for each second, and the cover then holds the cheapest
file rather than the file that reads the value.

**#414 measured it on `ja4plus/__init__.py`.** The module holds 32 mutations, 31 of them on
the entries of `__all__`, which the interpreter builds at import.

- The cover the rule produces is `tests/test_parity.py`, at 1.01 seconds.
- The sweep reports **1 killed and 31 survived**.
- `tests/test_public_interface.py::test_the_package_declares_its_public_interface` compares
  `ja4plus.__all__` against a hand-written list, so it kills every one of those 31.

**The survivors are over-reported, which the recorded limits already permit.** The defect
is that the rule produces that outcome by construction for this shape of module, rather
than by an accident of cost.

## The size of the reach

Over the seven modules of `git ls-files 'ja4plus/*.py'`, **95 of 279 surviving mutations
sit on a line the import runs**.

| Module | Survived | On a line the import runs |
|---|---|---|
| `ja4plus/__init__.py` | 31 | 28 |
| `ja4plus/cli.py` | 100 | 8 |
| `ja4plus/ja4db.py` | 32 | 2 |
| `ja4plus/output.py` | 52 | 22 |
| `ja4plus/processor.py` | 16 | 2 |
| `ja4plus/types.py` | 11 | 9 |
| `ja4plus/watch.py` | 37 | 24 |

`docs/mutation_reports/414-interface.json` holds the runs and
`docs/mutation_settlements/414-interface.json` records the limit.

## What a repair looks like

The cover rule needs a reader for a line the import runs, and coverage cannot supply one,
because the line carries the empty context. Two candidate rules, and the user decides:

1. **Read the value rather than the line.** For a module-level name, take the test files
   that name that value, read from the test source. A cover then holds
   `tests/test_public_interface.py` for `ja4plus/__init__.py`.
[...]

Issue #429, opened 2026-08-09
Fail a change that records no round


Crank-Git:
## What was measured

**A branch that changes code and adds no round entry passes the whole gate.** #412 proved
it by accident: commit `46aa502` carried eleven sweeps, two repairs and a new test file,
**no `CHANGELOG.md` entry and no `Changelog` row**, and the whole unit gate was green.

**The project manager caught it by reading the diff at the merge gate.** That is not a
check, and the next one will pass unread.

## Why the existing cases do not catch it

`tests/test_changelog_round_agreement.py` holds the two documents **to each other**:

- `test_every_numbered_entry_names_a_round_the_specification_records`
- `test_an_unassigned_entry_matches_an_unassigned_row`
- `test_a_numbered_entry_holds_no_row_that_waits_for_its_number`

**Each compares one file against the other. When both are absent, they agree, so every
case passes.** The invariant they hold is agreement, and the invariant that is missing is
existence.

## What to build

**A check that a branch which changes a tracked file also records a round.**

The hard part is the reference point, and it must be stated as a condition and not as a
habit. Some candidate readings, and the issue is free to find a better one:

- The commits of the branch name an issue number, and the round entry names the same
  number.
- The change set holds a file outside `CHANGELOG.md` and `docs/specs/spec.md`, and the two
  documents each gained one unassigned entry.

**Prove it fails.** Build a change set that touches a tracked file and adds no entry, and
show the case red. **Then prove it passes** on a change set that adds one. **A check proved
in one direction is a check that may pass always** — this repository has recorded sixteen
instances of a comparison that is never made reading as a comparison that passes.

**State plainly what it cannot test.** A check can test that an entry exists. **It cannot
test that the entry is true**, and the issue says so rather than implying otherwise.

## Acceptance criteria

- [ ] A case fails on a change set that edits a tracked file outside the two documents and
      adds no unassigned round entry.
- [ ] The same case passes on a change set that adds one.
- [ ] The pull request holds both readings, and names the reference point the case uses.
- [ ] The case does not fail on a change set that edits only `CHANGELOG.md` or
      `docs/specs/spec.md`.
- [ ] The pull request states what the case cannot test.
- [ ] `pytest tests/ -m spec_validation` reports 1532 passed, 143 skipped and 134 xfailed
      against 134 register keys. No file under `ja4plus/` changes.

## Links

- Found by #412 on 2026-08-09.
[...]

Issue #419, opened 2026-08-09
Repair the network requirement that every install case must violate


Crank-Git:
## The defect

**Two requirements of `docs/specs/features/11-pre-release-validation.md` contradict each
other. #408 found it while it built the first of them, and it changed no requirement
text, correctly.**

> `FR-pre-release-validation-1` — One case installs the built wheel into a clean
> environment, produces a fingerprint through the public interface and through the
> command-line program, and compares the output against the output of the source tree.

> `FR-pre-release-validation-26` — No case of this feature set opens a network
> connection.

**A case cannot satisfy both.** `pip install <the wheel>` resolves `scapy` and
`cryptography` from an index, so installing into an environment that holds no
development extra reaches the network by construction. The same holds for
`FR-pre-release-validation-1` of the source distribution in #409.

## Why the rule exists, and what it was protecting

The rule came from #414, where it is correct and narrow: **`ja4plus/ja4db.py` refuses a
remote lookup until the caller opts in**, and a case that asserts no request happened
must not itself send one. `FR-pre-release-validation-26` states that rule with the scope
widened to the whole feature set, and the widening is the defect.

## What to build

Replace `FR-pre-release-validation-26` with a requirement that states the rule as a
condition a check can test, over the scope it was written for.

- **The lookup cases open no network connection.** They use the recorded transport and
  reach no live address. That is the rule #414 needs.
- **The install cases reach an index, and that is the interface under test.** State it
  as permitted and name why: an install that resolves no dependency does not measure the
  shipped dependency list, which is the thing `FR-pre-release-validation-1` exists to
  check.

Read `docs/specs/features/11-pre-release-validation.md` on the integration branch and
check whether any other requirement of the set carries the same widened scope.

## Acceptance criteria

- [ ] `docs/specs/features/11-pre-release-validation.md` holds no requirement that a
      case satisfying `FR-pre-release-validation-1` must violate.
- [ ] The replacement names the scope it binds, and a check can test it. A case reads
      the requirement text and fails a requirement that states a rule over "this feature
      set" without naming which cases it binds.
- [ ] The lookup rule still binds #414: the requirement text names the recorded
      transport and the live address.
- [ ] `pytest tests/test_specification_changelog.py tests/test_changelog_round_agreement.py`
      passes.
- [ ] `pytest tests/ -m "not spec_validation"` passes, and `pytest tests/ -m spec_validation`
      reports 1532 passed, 143 skipped and 134 xfailed against 134 register keys.
[...]

Issue #206, opened 2026-08-08
Settle the 976 candidates the mutation sweep names


Crank-Git:
## Context

#172 built the sweep and shipped `docs/mutation_sweep.md`. **The sweep names 976 candidate
cases. Two were settled.** This issue holds the rest, so the number is not lost.

That is not a defect of #172. Its own body states the rule: *"Do not rewrite a test from
the report alone; a case may be correct and the mutation wrong."* A candidate needs a
reader, and 976 readings do not fit in one issue.

## What the report already states, and what it does not

The report carries two limits, and both bound this issue:

1. **A case the baseline reports as skipped or as xfailed cannot fail**, so it reaches the
   candidate list for a reason the suite already states. **248 of the 976 are that.**
2. **A sample of 12 mutations per module answers for that sample alone.** A case that no
   sampled mutation reaches is unmeasured, not proven vacuous.

**So 976 is an upper bound on the work, not a defect count.** The measured defect count so
far is two, both in `tests/test_ja4l.py`, both repaired and both proved.

## The modules the sample ranks worst

| Module | Survived of run | Reading |
|---|---|---|
| `ja4plus/collector.py` | 12 of 12 | **#48 removes this module as deprecated.** Settle #48 first, and this row disappears |
| `ja4plus/fingerprinters/__init__.py` | 9 of 9 | Exports and names. A survived string mutation here may be correct |
| `ja4plus/cli.py` | 10 of 12 | Epic 5 and Epic 6 rebuild the command-line program |
| `ja4plus/ja4db.py` | 10 of 12 | Epic 7 rebuilds the lookup client |

**Three of the four worst modules are rebuilt by an epic that has not started.** Adjudicate
those after the epic lands, not before, or the reading is thrown away.

## What to build

Take the modules that no epic rebuilds, one issue at a time, and settle their candidates
properly:

1. **Sweep the module whole against the file that holds the case**, which is the escape
   from limit 2: `--max-per-module 0 --tests tests/<file>.py`.
2. **Read each candidate.** Repair the vacuous ones, and prove each repair against the
   mutation that exposed it.
3. **Record the ones that are correct and the mutation wrong**, with the reason, so the
   next reader does not re-derive it. #172 did this for the three candidates of
   `tests/test_tcp_stream.py`.

## Acceptance criteria

- [ ] Every module that no open epic rebuilds is swept whole against the test files that
      name it.
- [ ] Every candidate in those modules is settled: repaired and proved, or recorded as
[...]

PR #434, opened 2026-08-09
Settle the sweep candidates of the interface modules

Crank-Git:
Settles every sweep candidate of the seven Python modules of the top directory of the
package, and
settles `FR-pre-release-validation-23`, which #419 routed here.

Refs #414. Part of #406. Targets `epic/406-pre-release-validation`, as a draft, and every
commit carries `[skip ci]`.

## What ran

Seven single-module sweeps, `--max-per-module 0`, each against the minimal cover of its
own module, at base commit `6fd61f8432453d9cf14eb150132448644e29b74d`.

| Module | Mutations | Killed | Survived | Timeout | Unusable | Candidates | Seconds |
|---|---|---|---|---|---|---|---|
| `ja4plus/__init__.py` | 32 | 1 | 31 | 0 | 0 | 5 | 35.9 |
| `ja4plus/cli.py` | 226 | 126 | 100 | 0 | 0 | 110 | 2391.7 |
| `ja4plus/ja4db.py` | 68 | 36 | 32 | 0 | 0 | 53 | 217.9 |
| `ja4plus/output.py` | 68 | 16 | 52 | 0 | 0 | 5 | 123.0 |
| `ja4plus/processor.py` | 54 | 38 | 16 | 0 | 0 | 70 | 446.5 |
| `ja4plus/types.py` | 11 | 0 | 11 | 0 | 0 | 22 | 107.1 |
| `ja4plus/watch.py` | 97 | 56 | 37 | 4 | 0 | 42 | 867.4 |
| **Total** | **556** | **340** | **279** | **4** | **0** | **307 rows, 236 distinct** | **4188.7** |

`docs/mutation_reports/414-interface.json` holds the seven runs, and
`docs/mutation_settlements/414-interface.json` holds 236 rows: **231 `correct` with a
reason and 5 `repaired` with the case.**

**Do not compare 236 against the 976 of #206 or the 205 of #412.** The three count
different sets, and `.claude/rules/conformance.md` states why.

## The cost was measured before the run started

Amendment 2 asks for the product, stated before the run, and Amendment 3 rules the scope to
the minimal cover. Both are in [the comment on #414](https://github.com/Crank-Git/ja4plus/issues/414#issuecomment-5233129947).
The product is **1.27 hours** against the four-hour ceiling, so the stop condition did not
fire. The seconds are wall clock for the whole `pytest` command, so they hold the
interpreter start and the collection that `--durations` omits.

### 1. A `raises` sample block asserted the type of the error and never the line

`tests/test_documentation_samples.py` ran a `raises` block and compared
`type(caught.value).__name__` against the marker. **The type names no line**, so a block
whose first line raises reads exactly like a block whose documented last line raises.

`docs/api_reference.md:284` holds two lines. The first reads `result["fingerprint"]` and
succeeds; the second reads `result["method"]` and raises `KeyError`. The mutation
`not in` -> `in` at `ja4plus/types.py:74` moves the raise to the **first** line, keeps the
type `KeyError`, and the case stayed green. That is why `ja4plus/types.py` reported
**0 killed of 11**.

```
== the mutation is live
   inspect.getsource -> '        if key in _FIELD_NAMES:'
   pytest exit 1
   FAILED tests/test_documentation_samples.py::test_every_python_sample_of_the_page_runs[docs/api_reference.md]
   1 failed, 1 warning in 1.26s

== the module is reverted
   inspect.getsource -> '        if key not in _FIELD_NAMES:'
   pytest exit 0
   1 passed, 3 warnings in 1.36s
```

The harness now reads the line the error came from and requires it to be the last line of
the block that carries code. **Both `raises` blocks of the documentation raise on their
last line**, measured before the change, so the rule fits every block that exists.

### 2. The constructor case read the type of the error and never the message

`tests/test_db_offline.py::TestTheConstructorRefusesAnAmbiguousCall::test_a_value_that_is_no_bool_raises_type_error`
read `pytest.raises(TypeError)` alone, so the mutation of the message at
`ja4plus/ja4db.py:294` changed nothing it compared.

**The first repair was itself unable to fail, and the sweep method caught it.**
`pytest.raises(match=...)` runs `re.search`, so the unanchored pattern
`allow_remote takes True or False` matches the mutated message
`allow_remote takes True or False_mutated`:

```
== the mutation is live
   inspect.getsource -> '            raise TypeError("allow_remote takes True or False_mutated")'
   pytest exit 0
   5 passed, 1 warning in 0.55s
```

The pattern now carries both anchors:

```
== the mutation is live
   inspect.getsource -> '            raise TypeError("allow_remote takes True or False_mutated")'
   pytest exit 1
   4 failed, 1 passed, 1 warning in 0.60s

== the module is reverted
   inspect.getsource -> '            raise TypeError("allow_remote takes True or False")'
   pytest exit 0
   5 passed, 1 warning in 0.51s
```

Four cases fail, one for each parameter, and those are the four repaired candidates.

## The four candidates a mutation reaches and no repair fits

Each reason is measured live rather than argued. The record holds the transcript of each.

| Candidate | Mutation | Reading |
|---|---|---|
| `tests/fuzz/…::test_a_foxio_malformed_capture_produces_no_exception[ssh2-malformed.pcap]` | `0` -> `1` at `processor.py:131` | The case asserts that no exception reaches the caller. The first value of a packet count changes no exception. `tests/test_processor_stats.py` reads the count and the cover drops it. |
| `tests/test_documentation_samples.py::…[README.md]` | `>` -> `>=` at `processor.py:362` | The case asserts that the samples run, and no sample of that page compares a shard key. **The second `>` of that line is equivalent**: the swap it guards runs only where `src_ip` equals `dst_ip`, and swapping two equal ports changes no key. |
| `tests/test_watch.py::…::test_the_key_of_a_packet_that_carries_no_port_is_none` | `is` -> `is not` at `watch.py:127` | A second check masks the guard: `packet_endpoints` reports no address, so `connection_key` returns `None` by the later path. The sweep records the mutation as **killed**, so another case measures the guard. |
| `tests/test_watch.py::…::test_the_handler_raises_no_system_exit` | `True` -> `False` at `watch.py:557` | The case asserts that the handler raises no `SystemExit`, and the flag value raises nothing. The sweep records the mutation as **timeout** and not as survived, because a monitor whose stop flag never sets does not end. |

## The census assertion is not vacuous, in both readings

#412 reported that a criterion reading "0 unsettled candidates" passes on an empty
directory. Both readings, measured on this branch:

```
reading one: the census names 236 candidates of sweep 414-interface
             unclaimed 0, claimed twice 0, unknown 0, faults 0

reading two: one row leaves docs/mutation_settlements/414-interface.json
             dropped tests/fuzz/test_malformed_captures.py::test_a_foxio_malformed_capture_produces_no_exception[CVE-2018-6794.pcap]
             the census now names 1 unclaimed: ['tests/fuzz/test_malformed_captures.py::test_a_foxio_malformed_capture_produces_no_exception[CVE-2018-6794.pcap]']
```

## The scope addition: `FR-pre-release-validation-23` is split, and both halves are stated

**I did both of the two the issue offers, because the requirement held two statements of
different kinds and only one of them is checkable.**

- **`FR-pre-release-validation-23` now states the artefact**, which a check tests: a
  `repaired` verdict names a case the suite collects. New file
  `tests/test_settlement_procedure.py` reads it, along with the two verdict shapes of
  `FR-pre-release-validation-22`.
- **`FR-pre-release-validation-23a` carries the procedure** — a worker proves each mutation
  live in both directions with `inspect.getsource` — and it joins the
  `Two statements here are not checkable` list, now three, with its reason: no case
  observes a worker.

**The check reads a necessary condition and not a sufficient one.** A record that names a
collected case proves the case exists. It does not prove the worker ran the mutation. The
requirement, the list entry and the module docstring each say so, rather than leaving a
reader to infer it.

The failing direction is proved. One `repaired` verdict was pointed at an absent case:

```
== the defect is live
FAILED tests/test_settlement_procedure.py::test_the_suite_collects_every_case_a_repaired_verdict_names
1 failed, 4 passed in 0.99s
== the record is reverted
5 passed in 0.95s
```

`pytest tests/test_requirement_scope.py` reports `4 passed`, and the case #419 added is
unchanged.

## Three findings this pull request reports and does not repair

1. **The cover rule gives a module-body mutation no reader, by construction.** The rule
   drops the lines the import runs, which is right for a mutation inside a function and
   wrong for one in the module body. **95 of the 279 survivors sit on such a line.**
   `ja4plus/__init__.py` measures the cost: the cover the rule builds is
   `tests/test_parity.py` at **1 killed and 31 survived**, and the same 32 mutations
   against `tests/test_public_interface.py` read **25 killed and 7 survived**, so 24
   survivors are over-reported. Filed as **#433**, with the first claim there corrected
   after measurement.
2. **`tests/test_cli.py` runs all 27 reachable mutation lines of `ja4plus/output.py` on its
   own**, so the cover of that module holds it alone. `tests/test_output_schema.py` and
   `tests/test_cli_output_option.py` hold the larger part of the output cases, and **no
   case of either file is measured against `ja4plus/output.py` here.** That is unmeasured
   and not clean, and the record states it.
3. **Of 406 distinct mutation lines over the seven modules, 250 run under no test context
   at all.** They run at import, or the suite reaches them nowhere.
4. **The body of #414 asks a worker to confirm the module list against
   `git ls-files 'ja4plus/*.py'`, and that command names 31 files and not seven.** A git
   pathspec matches `*` across a directory separator. `Path.glob`, which
   `tests/mutation_sweep.py` uses, reads it as one path part and names the seven the issue
   enumerates, so the sweep measured the right files. The self-review found it, and the
   record, the report, the changelog entry and the specification row now name the seven
   modules rather than the command. `.claude/rules/conformance.md` carries the same
   reading for `ja4plus/**/*.py` and does not name the single-star form; that is a rules
   edit for its own issue.

## Two premises of the handoff brief, re-measured

- **"The unit gate reads 2590 passed, 4 skipped, nothing deselected."** The count agrees:
  the base reads 2590 and this branch reads 2595, which is the five cases of
  `tests/test_settlement_procedure.py`. **"Nothing deselected" is contradicted**:
  `-m "not spec_validation"` deselects **1825** cases by construction, and #410 recorded
  the same 1825 on Linux.
- **"One whole-suite run costs 72.75 seconds over 4325 collected cases."** This host reads
  **158.69 seconds over 4411 outcomes** with coverage on. No number in this pull request
  derives from the 72.75 figure; every sweep cost is measured on the scoped command.

## The sweep was stopped once, and it left no change behind

A signal stopped the driver during `ja4plus/processor.py`. The handler #412 added turned it
into `SystemExit`, the restore ran, `git status --porcelain -- ja4plus` reported nothing,
and the checkpoint resumed the run at mutation 37 of 54. **`git status --porcelain -- ja4plus`
reports nothing now.** One stray directory `~_mutated/` was created outside `ja4plus/` by a
mutation of a cache path; it is removed and it is not committed.

## Gates

| Gate | Result |
|---|---|
| `pytest tests/ -m "not spec_validation"` | `2595 passed, 4 skipped, 1825 deselected, 8 xfailed, 98 subtests passed in 127.53s` |
| `pytest tests/ -m spec_validation` | `1532 passed, 143 skipped, 2623 deselected, 134 xfailed in 9.20s`, against **134** keys of `tests/foxio_deviations.json` |
| `ruff check ja4plus/ tests/` | `All checks passed!` |
| `ruff format --check ja4plus/ tests/` | `188 files already formatted` |
| `mypy --strict ja4plus/` | `Success: no issues found in 31 source files` |
| Coverage, `--cov=ja4plus` | **94**, unchanged. No file under `ja4plus/` changes, so it cannot fall. |

The register invariant holds exactly: 1532 passed, 143 skipped, 134 xfailed, 134 keys.

## Documentation

`CHANGELOG.md` gains one entry ending `Round TBD.`, and the `Changelog` table of
`docs/specs/spec.md` gains one `| TBD | ... |` row. #429 records that no check catches
their absence, and #412 was held at the gate for exactly this.

**No file under `ja4plus/` changes. No file under `tests/foxio_vectors/` changes. No
fingerprint moves. No case of this pull request opens a network connection.**


Crank-Git:

Four changes read: `tests/test_settlement_procedure.py`, the `raises` branch of
`tests/test_documentation_samples.py`, the anchored pattern of `tests/test_db_offline.py`,
and the split of `FR-pre-release-validation-23`.

**Four readings confirmed, each against a run and not against the source alone.**

1. **The collection filter holds a parametrized identifier that carries a space.**
   `pytest --collect-only -q` writes one identifier for each physical line, so a filter
   that reads whole lines is safe where a filter that split on whitespace would not be.
   `tests/test_db_offline.py::TestARemoteLookupFailure::test_a_response_with_an_unexpected_shape_returns_none[list of one match]`
   returns as one unbroken line. **The earlier form of that filter read `" " not in line`
   and would have dropped it**; the committed form reads the leading character.
2. **The collection case is not vacuous.** The two records hold 7 `repaired` verdicts
   between them, so `named` and `collected` are both non-empty on this branch.
3. **The anchors behave as intended.** `pytest.raises(match=...)` runs `re.search` with no
   `re.MULTILINE`, and the message is one line with no trailing newline, so `^` and `$`
   bind the whole string.
4. **Neither new requirement trips `SET_SCOPE`.** The pattern reads
   `this feature set|\b(?:no|every|each) cases?\b`. `FR-pre-release-validation-23a` holds
   `each mutation` and not `each case`, and `FR-pre-release-validation-23` holds neither
   form and names `tests/test_settlement_procedure.py` regardless.
   `pytest tests/test_requirement_scope.py` reports `4 passed`.

**One finding, and it is repaired here rather than dismissed.**

`_last_code_line` counts physical lines and reads no statement. Where the last statement of
a `raises` block spans several lines, the interpreter reports the line the statement opens
and the reader reports the line it closes, so the new assertion would fail a block that is
correct. **No such block exists**: both `raises` blocks raise on a single-line statement,
measured before the change. **The failure is loud rather than silent**, so it misreports
nothing today. The docstring now carries the warning and names the two repairs a future
writer can take.

Crank-Git:
## Self-review, evidence lens

Ten readings of `docs/mutation_reports/414-interface.json` against
`docs/mutation_settlements/414-interface.json`, each computed and not read.

| Reading | Result |
|---|---|
| `sweep` equals the stem of the report | `414-interface` on both |
| Every report candidate is claimed exactly once, and no row claims a case the report omits | 236 and 236, 0 unclaimed, 0 claimed twice, 0 unknown |
| Row count equals the distinct candidate count | 236 = 236 |
| The report's `candidates` equals the union of the seven per-module lists | sum 307, union 236, equal to `candidates` |
| Every row carries a verdict with its own field | 5 `repaired` with a `case`, 231 `correct` with a `reason`, 0 other |
| Every `repaired` case collects | all 5 appear in a `--collect-only -q` run of the two files |
| The `modules` list of the report and of the record agree | the same seven files |
| The mutation counts sum, and each module balances | 556, and `killed + survived + timeout + unusable` equals the count for each of the seven |
| `commit` is the base commit | `6fd61f8432453d9cf14eb150132448644e29b74d`, and `git merge-base` against `origin/epic/406-pre-release-validation` returns it |
| No `correct` reason names a module whose sweep did not name that candidate | checked on all 231 rows, not sampled: 0 inconsistent pairs |

**One finding, and it corrects this pull request rather than the record.**

The body of #414 asks a worker to confirm the module list against
`git ls-files 'ja4plus/*.py'`. **That command names 31 files and not seven.** A git
pathspec matches `*` across a directory separator, so the pattern reaches
`ja4plus/fingerprinters/` and `ja4plus/utils/` as well:

```
$ git ls-files 'ja4plus/*.py' | wc -l
31
$ python -c "from pathlib import Path; print(len(list(Path('.').glob('ja4plus/*.py'))))"
7
```

**The module set is right and my description of it was wrong.**
`tests/mutation_sweep.py --module 'ja4plus/*.py'` selects the seven the issue enumerates,
because `module_paths` uses `Path.glob`, where `*` reads as one path part. The `--dry-run`
at the head of this issue reported those seven and no other, so the sweep measured the
right files throughout.

The record, the report, the changelog entry and the specification row now name the seven
modules instead of the command, and `scope.module_set` of the record states the finding.
**`.claude/rules/conformance.md` already carries the same reading for `ja4plus/**/*.py`,
which names 24**; the single-star form has the opposite fault and the rule does not name
it. I leave that rule as it stands, because it is a rules edit that belongs to its own
issue, and the instruction that failed here lives in the issue body rather than in the
rule.

**This is the ninth project-manager premise this epic has contradicted**, and the fourth
that could not be followed as written.

Issue #438, opened 2026-08-09
Read the change set of a pull request on the runner


Crank-Git:
## What was measured

**The GitHub runner reads no change set, so `tests/test_round_entry_existence.py` skips on
every job.** #429 built the case that fails a change set which records no round. It reads
the change set against the merge base of `HEAD` and the integration branch. The
`actions/checkout` step of `.github/workflows/test.yml` names no `fetch-depth`, so it makes
a clone of depth 1 and fetches the triggering ref alone. That clone carries no `origin/dev`
ref and no merge base, so the case reports a skip reason and runs no comparison.

The case therefore guards the local gate of a worker, whose worktree carries the ref. It
guards no pull request on the runner.

## Why this is a separate issue

#429 owns the case. This issue owns the checkout. A change to
`.github/workflows/test.yml` changes the clone of every job of the matrix, and the cost of
a full history against the cost of one extra fetch is a decision of its own.

## What to build

**A reading of the change set that the runner can make.** Some candidate readings, and the
issue is free to find a better one.

- `fetch-depth: 0` on the `actions/checkout` step of the jobs that run the unit suite.
- A second step that runs `git fetch --depth=1 origin dev`, which fetches one ref rather
  than the whole history.
- A reference point the runner already holds. A pull-request event carries
  `github.event.pull_request.base.sha`, and the case reads no environment variable today.

**Prove it in both directions on the runner.** Open a pull request that records no round
and show the case red. Then add the round and show it green. **A check proved in one
direction is a check that may pass always.**

## Acceptance criteria

- [ ] The case reports no skip on the `unit` job of `.github/workflows/test.yml`.
- [ ] A pull request that edits a tracked file and records no round fails the job.
- [ ] The same pull request passes the job after it records the round.
- [ ] The measured cost of the extra fetch reaches the pull-request body.
- [ ] No file under `ja4plus/` changes.

## Links

- Found by #429 on 2026-08-09.
- The case: `tests/test_round_entry_existence.py`.

Crank-Git:
Triaged `status:ready`. It stays out of batch #437.

The reason: an acceptance criterion of this issue is a pull request that records no round and fails the job. A batch pull request must reach green, so the red direction cannot be proved inside the gate that must pass. This issue runs as its own batch member with a deliberate red run of its own.

Crank-Git:
[...]

Issue #437, opened 2026-08-09
Batch: test hygiene — #429 #430 #371 #426


Crank-Git:
## The batch

Test hygiene. The user ruled this batch next on 2026-08-09. Each member weakens a gate
while it stands, so each one weakens every later gate.

- [x] #429 — a change that records no round passes the whole gate
- [x] #430 — two wall-clock races, one of them a control
- [x] #371 — the wall-clock race in the periodic statistics case
- [ ] #426 — four cases read ambient host state without a guard

## Why these four together

#430 and #371 are one defect class. #369 and #371 already recorded it twice, so the two
belong in one batch and one reading of the class.

#429 and #426 both hold the gate to itself: a check that passes when the thing it checks is
absent, and a case that fails on a correct package.

## The integration branch

`batch/437-test-hygiene`, forked from `dev` at `cb0ec62`.

## Rules that bind every member

- No file under `ja4plus/fingerprinters/` changes. No fingerprint moves.
- Do not repair a flaky case by widening a tolerance until nothing fails. A vacuous case
  is the defect this project keeps recording.
- Prove a guard and a check in both directions. A check proved in one direction is a check
  that may pass always.

Crank-Git:
#429 sub-merged as `f140a5c`. Round 156 recorded.

The gate read: 2612 passed against a base of 2595, so 17 new cases; the conformance suite unchanged at 1532 passed, 143 skipped, 134 xfailed against 134 register keys; coverage 94 percent, 4232 statements, 273 missed, identical to the base; the three files of the change set are `CHANGELOG.md`, `docs/specs/spec.md` and `tests/test_round_entry_existence.py`, so no file under `ja4plus/` changed.

**What #429 does not deliver, and it says so itself.** The new case reads the merge base of the branch, and `actions/checkout` makes a clone of depth 1, so the case skips on every runner job. It guards a worker's local gate and it guards no pull request. **#438 owns the checkout** and it is `status:ready` outside this batch, because proving it needs a deliberately red pull request.

The self-review of #371 filed #441, a vocabulary disagreement between the `## Terms` table and the prose of the package. It is parked `status:needs-feedback` for the user and nothing in this batch waits on it.

Crank-Git:
Batch #437 merged as `d123ab9`. Rounds 156 to 159. All four members closed.

[...]

PR #439, opened 2026-08-09
Fail a change set that records no round

Crank-Git:
Refs #429. Batch member of #437. Round 156.

## What changed

New file `tests/test_round_entry_existence.py` holds one invariant: **a change set that
edits a tracked file also records a Changelog round.** It adds 17 cases and it changes no
existing case. `CHANGELOG.md` and `docs/specs/spec.md` each gain the round 156 record of
this work. No file under `ja4plus/` changes.

## Why

`tests/test_changelog_round_agreement.py` holds the two records **against each other**.
Each of its three cases reads one file and compares it against the other, so **two absent
records agree and all three pass.** The invariant they hold is agreement. The invariant
that was missing is existence.

#412 proved the gap. Commit `46aa502` carried eleven sweeps, two repairs and a new test
file, it added no entry and no row, and the whole unit gate was green. The project manager
read the diff at the merge gate and caught it. That is not a check.

## The reference point, and the two readings

The issue offers two candidate readings. **Both are recorded here, and the second one is
the reading the case uses.**

**Reading 1 — the issue number.** The commits of the branch name an issue number, and the
round entry names the same number. **This reading is declined**, and the reason is
measurable rather than a preference. A change set reaches the case with every commit the
branch carries, and three commit shapes name a number that owns no round entry: a merge
commit names a pull-request number, as `Merge pull request #435 from ...` on the base of
this branch; an entry cites another issue in passing, which is the defect
`ISSUE_CITATION` of `tests/test_changelog_round_agreement.py` exists to bound; and a
repair folds its detail into the shipment row of an epic and writes no row of its own,
which #258 settled. Each shape would fail a legitimate change set.

**Reading 2 — the count of round entries. The case uses this one.**

- **The reference point is the merge base of `HEAD` and the integration branch.**
  `reference_commit` reads it with `git merge-base`, against `origin/dev` and then `dev`.
  The merge base is a condition on the commit graph. It reads no local configuration, so
  every checkout of one branch reaches the same commit, and no habit of a worker changes
  it. `CLAUDE.md` names `dev` as the integration branch.
- **The change set joins two readings.** `git diff --name-only <merge base>` reports every
  tracked file the branch changed, including a change no commit holds yet.
  `git ls-files --others --exclude-standard` reports every new file that no ignore rule
  covers. #412 shipped exactly one such file, so a reading of tracked files alone would
  reach it only after the commit.
- **A change set of `CHANGELOG.md` and `docs/specs/spec.md` alone demands no entry**,
  because the round assignment of the project manager edits those two files alone.
- **The case counts an unassigned entry and a numbered entry alike.** The project manager
  assigns the number at the batch gate on one batch, and hands the number to the worker on
  another, as this brief does with round 156. A case that demanded `TBD` would fail on the
  second. `tests/test_changelog_round_agreement.py` already holds the number itself against
  the specification row, so nothing is lost.
- **Where the change set cannot be read, the case skips and the reason names the state.**
  A shallow clone carries neither ref and no merge base. The `actions/checkout` step of
  `.github/workflows/test.yml` names no `fetch-depth` and makes such a clone, so **the case
  skips on the GitHub runner and runs on the local gate of a worker.** New issue #438 owns
  the ref the runner needs. It is filed and untriaged, and this pull request fixes no part
  of it.

## What the case cannot test

**A case here tests that an entry exists. It cannot test that the entry is true.** Prose
that names the wrong issue, that names the wrong round, or that states a change the branch
did not make passes every case in this file. A reader of the merge gate holds that
question.

Three more limits, stated rather than implied. The module docstring holds the same four.

1. The count reads the whole branch, so a change set that merges another branch inherits
   the entry of that branch. **An integration branch whose members each recorded a round
   therefore satisfies a member that recorded none.** This is the widest limit.
2. The count rises by one where an edit to the prose of an entry makes the parser match a
   round sentence it read before as prose.
3. The count reads no round number, so a wrong number reaches no case here.

## How this was tested

**The case is proved in both directions on this repository.** With the entry and the row
absent it failed, and the message carries the counts:

```
AssertionError: the change set holds these paths outside the two records: tests/test_round_entry_existence.py. CHANGELOG.md holds 68 round entries against 68 at the reference commit
```

With `CHANGELOG.md` recorded and the specification row removed, the second comparison
failed too, so neither comparison is the only live one:

```
AssertionError: the change set holds these paths outside the two records: tests/test_round_entry_existence.py. docs/specs/spec.md holds 155 Changelog rows against 155 at the reference commit
```

With both records present it passes. Each run restored the file it moved, and
`git status` reports the two records as the only modified files.

**One case reads the change set of #412 itself.** It reads commit `46aa502` against its
parent, where `CHANGELOG.md` holds 64 round entries on both sides, and it reports:

```
AssertionError: the change set holds these paths outside the two records: .claude/rules/conformance.md, docs/mutation_reports/412-utils.json, docs/mutation_settlements/412-utils.json and 3 more. CHANGELOG.md holds 64 round entries against 64 at the reference commit
```

So the new reading fails the change set that the whole unit gate passed. The case skips
where the clone holds no parent of that commit.

**Thirteen cases build a scratch repository with `git init` and prove the same pair over the
whole reading.** A committed change with no round fails and the same change with a round
passes. An uncommitted change behaves the same way. An entry that reaches one record alone
fails and names the record it missed. A change set of `CHANGELOG.md` alone, of
`docs/specs/spec.md` alone, and of the pair passes. A new
file that no commit holds fails, and a new file an ignore rule covers passes. A repository
that holds no integration branch produces a skip reason, and a directory that holds no
repository produces a skip reason and no failure. Two more cases hold the parser: it counts
a round sentence that wraps to the next line, which is the #302 defect, and it counts no
round that a quotation holds, which is the over-count of `grep -c 'Round TBD'`.

## Verification floor

| Gate | Result |
|---|---|
| `pytest tests/ -m "not spec_validation"` | 2612 passed, 4 skipped, 1825 deselected, 8 xfailed, 98 subtests passed |
| `pytest tests/ -m spec_validation` | 1532 passed, 143 skipped, 2640 deselected, 134 xfailed |
| Register invariant | `tests/foxio_deviations.json` holds 134 keys against 134 xfailed |
| `ruff format --check ja4plus/ tests/` | 189 files already formatted |
| Coverage | 94 percent, 4232 statements, 273 missed, unchanged against the base |
| Fingerprints | No file under `ja4plus/` changes, so no fingerprint moves |

The unit count rises from 2595 to 2612, which is the 17 new cases. The conformance run deselects every unit
case, so its deselected count reads 2640 and it carries the same 17. **Passed, skipped and xfailed hold the number the acceptance criterion
names on both suites**, and coverage, the register count and the statement count equal the
number the base reports.


Crank-Git:
## Self-review

Three lenses read the diff: can the check pass where it must fail, can it fail where it
must pass, and does the prose meet `.claude/rules/ste.md`. The diff holds one test module
and two prose records, so no security lens and no frontend lens applies.

### Findings the review fixed

1. **The exemption was proved on one reading of three.** The acceptance criterion names
   `CHANGELOG.md` **or** `docs/specs/spec.md`, and the first form of the case edited both.
   `test_a_change_set_of_the_records_alone_passes` now runs against each record alone and
   against the pair. Commit `57c589a`.
2. **Every red proof read a change set the case wrote.** A synthetic change set proves the
   comparison and not the reading of a real commit.
   `test_the_reading_fails_the_change_set_of_the_defect` now reads commit `46aa502` against
   its parent and fails it. That commit is the one #412 shipped, and the whole unit gate
   passed it. Commit `57c589a`.
3. **The failure message read `1 files`.** The message now names the paths and states no
   count of them.

### Findings the review rejected, with the reason

4. **`document_at` reads a failed `git show` as an empty document.** A reader asks whether
   that turns a defect into a pass. It does the opposite: an empty base document counts
   zero round entries, so every entry of the working tree reads as new and the comparison
   grows more lenient rather than falsely red. A false failure is the outcome this project
   cannot accept from a new gate, and no reading of an absent document produces one.
5. **The count reads the whole branch and a merge inherits the entry of the branch it
   merged.** This is real, and it is a limit rather than a defect. The pull-request body
   records it as the widest of the four limits. A reading that bounded it would need the
   pull-request base, which is local configuration on a worktree and absent on a runner,
   and the plan bars a reading that rests on a habit.
6. **A path with a space or a non-ASCII byte reaches the message through
   `core.quotePath`.** The path then reads as an escaped string. It changes no comparison,
   because the comparison reads the count of entries and the set difference against two
   ASCII paths, and no such path exists in this repository.

### What no lens can reach

The case tests that an entry exists. **It cannot test that the entry is true.** The
pull-request body states that limit and three more, and the module docstring holds the same
four.

PR #440, opened 2026-08-09
Remove the wall-clock race from the periodic statistics case

Crank-Git:
## What

`tests/test_watch_statistics.py::TheStatisticsGoToStandardError::test_the_periodic_line_reaches_standard_error`
ran `ja4plus watch` with `--stats-interval 0.05`, slept 0.3 seconds inside the capture, and
asserted that at least 2 statistics lines reached standard error. That count measures how
promptly the host schedules a thread. It is the second case of the fault #369 repaired, and
#371 records it.

The case now states the schedule. `report_statistics` gains a `wait` parameter and forwards
it to `StatisticsReporter`. `ja4plus/cli.py` passes nothing, so the default still waits on
the stop event and the shipped behaviour does not move.

## Why

#369 added the `wait` seam to `StatisticsReporter`, and that seam reaches a caller that
builds a reporter. This case builds none. It reaches the reporter through `cmd_watch`, so a
deterministic form needs a wait that travels from the command down to `report_statistics`.

Do not read the change as a wider margin. Criterion 1 of #371 bars any assertion whose
outcome depends on how promptly the host schedules a thread, and a wider margin still
depends on it. The sleep does not rise and the count does not fall; both are gone.

## How

`ScriptedReport` in the test file replaces the name `ja4plus.cli.report_statistics`, in the
same way `run_watch` already replaces `ja4plus.cli.read_interface`. It records the interval
and the stream `cmd_watch` passed, forwards both to the shipped call, and carries a
`ScriptedWait` to the reporter. `report_statistics` starts the thread and `stop` joins it,
so every line reaches the stream before `main` returns.

The case asserts five facts.

1. Four statistics lines reach standard error: three periodic and one exit summary.
2. No statistics line reaches standard output.
3. `cmd_watch` passes the interval `0.05` to `report_statistics`.
4. The reporter passes `0.05` to the wait four times, once for each of the three intervals
   and once for the call that reports the stop.
5. `cmd_watch` passes standard error to `report_statistics`.

Facts 3 and 4 carry FR-live-capture-9 end to end, and facts 1, 2 and 5 carry
FR-live-capture-10. The count is exact, where the earlier form asserted a lower bound, so a
reporter that writes an extra line now fails.

## Deviation from the plan, stated

The plan names `ja4plus/cli.py` as a file the seam touches. `cmd_watch` needs no change: the
test replaces the module-level name `ja4plus.cli.report_statistics` with a wrapper that
forwards to the shipped call, which is the seam this file already uses for `read_interface`.
A parameter on `cmd_watch` that only a test sets would need a module global, and it would
carry no reader. One line of `ja4plus/` changed.

## Deviation from the brief, stated

The brief assigns Changelog round 158. `tests/test_specification_changelog.py::test_the_changelog_row_count_equals_the_highest_round_number`
requires the row count of `docs/specs/spec.md` to equal the highest round number it records.
The table holds 155 rows and its highest round is 155, so a row numbered 158 fails that case
until rounds 156 and 157 land. `tests/test_changelog_round_agreement.py` states the standard
in its first sentence: "Every worker writes the literal `TBD` and chooses no round number,
because two members of one batch have claimed the same number before. The project manager
assigns each number at the batch gate."

This branch therefore writes `TBD` in both files. **Assign 158 at the batch gate.**

## How tested

Test-driven. The case failed against the unrepaired package with
`Error during capture: report_statistics() got an unexpected keyword argument 'wait'`,
and that commit precedes the seam.

**Six mutations prove the case discriminates.** Each one was proved live by reading
`inspect.getsource` of the loaded member back in both directions, with
`PYTHONDONTWRITEBYTECODE=1` and every `__pycache__` outside `.venv/` cleared.

| Mutation | Result |
|---|---|
| `report_statistics` passes `wait=None` and drops the injected wait | 1 failed |
| `cmd_watch` passes the constant `0.5` in place of `args.stats_interval` | 1 failed |
| `cmd_watch` passes `sys.stdout` in place of `sys.stderr` | 1 failed |
| `StatisticsReporter._run` writes two lines for one interval | 1 failed |
| The default wait never reports the stop | 1 failed of the 43 cases of the file |
| `_stats_interval` guards on `seconds <= 1` in place of `seconds <= 0` | 1 failed |

The fifth mutation binds the seam to the shipped path. It fails
`test_the_thread_ends_before_the_command_returns`, which is the case that measures the real
stop event through the command, so a `wait` of None still reaches `threading.Event.wait`.

**The sixth mutation holds a kill that this case alone carries.** The sweep of #414 recorded
the mutation `0` to `1` at `ja4plus/cli.py:871` as killed, at a killed count of 1, and it
named this case as the sample. That guard reads `seconds <= 0`, and `--stats-interval 0.05`
is the one interval below one that any case of the suite passes. The repair keeps that
argument, so it loses no coverage of the interval guard. The docstring of the case now warns
a later writer.

## Out of scope, and filed

The `## Terms` table of `docs/specs/spec.md` names `reporter` a word a writer does not use,
against `statistics thread`, and the prose of the package uses `the reporter` throughout.
A repair inside this diff would put both words in adjacent sentences of one paragraph, which
rule 7 of `.claude/rules/ste.md` bars. #441 records it, untriaged.

**The class runs 20 times in a row and passes 20 times**, at 7 cases for each run:

```
for i in $(seq 1 20); do .venv/bin/python -m pytest \
  "tests/test_watch_statistics.py::TheStatisticsGoToStandardError" -q | tail -1; done
```

Every run printed `7 passed, 1 warning`, between 0.70 s and 0.94 s.

**The five gates.**

| Gate | Result |
|---|---|
| `pytest tests/ -m "not spec_validation"` | 2595 passed, 4 skipped, 8 xfailed, 98 subtests passed |
| `ruff format --check ja4plus/ tests/` | 188 files already formatted |

**Coverage does not fall.** 94% before and after, on 4232 statements both times.
`ja4plus/watch.py` reads 99% before and after.

**No fingerprint moves.** No file under `ja4plus/fingerprinters/` changes and no file under
`ja4plus/utils/` changes. The conformance suite reports the same 1532 passed, 143 skipped and
134 xfailed as the base, and `tests/foxio_deviations.json` holds 134 keys against those 134
xfailed.

Refs #371


Crank-Git:
## Self-review

Two lenses read the diff: correctness, and the writing standard with documentation accuracy.

### Correctness — no finding

The lens traced every path a reviewer would suspect of a residual race, and each one is
structural rather than timed.

- `ScriptedWait` sleeps not at all. It returns False three times and True once, so the
  thread ends in microseconds on any host.
- The join of `stop` is a liveness bound. The thread blocks on nothing once the scripted
  wait returns True.
- The exit summary cannot arrive before the three periodic lines. `write_statistics` at
  `ja4plus/cli.py:650` runs after the `with report_statistics(...)` block at
  `ja4plus/cli.py:589` exits, and that exit joins the thread.
- `report.waits[0].timeouts` cannot read a partly filled list, for the same reason. The
  main thread reads it after `main` returns, and `main` returns after the join.
- No other writer reaches standard error while the thread runs. The `Starting live capture`
  line precedes the thread, and the fake capture writes nothing.

### The writing standard — six sentences repaired

Six sentences ran past the 25-word limit of `.claude/rules/ste.md`. Each one is now two
sentences, in `ja4plus/watch.py`, `docs/api_reference.md`, `tests/test_watch_statistics.py`,
`CHANGELOG.md` and the specification row.

### One finding declined, with the reason

The lens reported the literal `TBD` in `CHANGELOG.md` and in `docs/specs/spec.md` as a
blocker, and cited the entry of #302 as the precedent. **The citation reads the wrong
defect.** #302 repaired an assignment that covered one file and not the other, which leaves
an orphan. `tests/test_changelog_round_agreement.py` states the standard in its own
docstring:

> **Every worker writes the literal `TBD` and chooses no round number**, because two members
> of one batch have claimed the same number before. The project manager assigns each number
> at the batch gate.

and

> **A case here bars no `TBD`.** An integration branch carries one `TBD` for each member that
> has not reached the gate, and every one of them is correct.

`TBD` also holds the invariant of
`tests/test_specification_changelog.py::test_the_changelog_row_count_equals_the_highest_round_number`.
The base branch holds 155 rows and its highest round is 155, so a row numbered 158 fails that
case until rounds 156 and 157 land. Both files read `TBD` and neither is an orphan.

**Assign 158 at the batch gate.**

### One finding out of scope, and filed

The `## Terms` table of `docs/specs/spec.md` names `reporter` a word a writer does not use,
against `statistics thread`. The prose of the package uses `the reporter` throughout, in
`ja4plus/watch.py`, `docs/api_reference.md`, `tests/test_watch_statistics.py` and three
Changelog rows. A repair inside this diff would put both words in adjacent sentences of one
paragraph, which rule 7 bars. #441 records it, and the user decides which of three readings
settles it.

### One finding this review added

The sweep of #414 recorded the mutation `0` to `1` at `ja4plus/cli.py:871` as killed, at a
killed count of 1, and it named this case as the sample. That guard reads `seconds <= 0`, and
`--stats-interval 0.05` is the one interval below one that any case of the suite passes. The
repair keeps that argument, and the mutated guard fails the case as it did before. The
docstring of the case now warns a later writer, and the mutation joins the six the body
records.

Crank-Git:
Sub-merge gate. The conflict was the shape the handoff predicts: two members each appended one Changelog entry, so `docs/specs/spec.md` conflicted on the appended row. Both entries are kept in merge order, round 156 of #429 above the `TBD` row of #371. The merge is `6d6f63a`.

**The gates ran on the merged state, not on the state either branch measured.**

| Gate | Reading |
|---|---|
| Unit | 2612 passed, 4 skipped, 1825 deselected, 8 xfailed, 98 subtests. 2595 base plus the 17 cases #429 added |
| Conformance | 1532 passed, 143 skipped, 134 xfailed |
| Register | 134 keys of `tests/foxio_deviations.json` against those 134 xfailed |
| `ruff check` | All checks passed |
| `ruff format --check` | 189 files already formatted |
| Records | `docs/specs/spec.md` holds 156 numbered rows against a highest round of 156, and one `TBD` row |

**The `TBD` row takes its number at the batch gate.** The brief of this issue named round 158 and that was a project manager error; the worker measured the convention and reported it rather than writing the number.

Issue #430, opened 2026-08-09
Remove the load sensitivity from two timing cases


Crank-Git:
## What was measured

**Two cases failed once while the machine carried a mutation sweep, and both passed alone
and on every later run of the whole gate.**

1. `tests/test_tcp_stream.py::TestTCPStreamReassembler::test_the_cost_of_a_segment_does_not_grow_with_the_segment_count`
2. `tests/test_throughput.py::TestTheWorkControl::test_the_double_run_reports_a_longer_elapsed_time`

Neither reads a file the branch that observed them changed. The second landed with #425.

## Why this matters more than an occasional red run

**This project has repaired the same defect class twice: #369 and #371.** A case that
compares wall-clock durations passes on an idle machine and fails on a loaded one, so
**it reports the state of the host and not the state of the package**. A reader who meets
such a failure learns to disregard the gate, and the next real failure hides inside it.

**The second case is a control**, and that raises the cost. `TestTheWorkControl` exists to
prove that twice the packets take longer, which is what stops a throughput measurement
reading process start rather than traffic. **A flaky control is worse than a missing one,
because it is trusted.**

## What to build

Read each case and decide, per case, between two shapes.

- **Make the measurement insensitive to load.** Compare work performed rather than seconds
  elapsed, or take the minimum of several runs rather than one reading.
- **State the tolerance the case needs, and name what it then cannot detect.** A case that
  demands a duration ratio above one is a case that fails at a ratio of 0.999 for reasons
  that have nothing to do with the package.

**Do not repair either by widening a tolerance until nothing fails.** That converts a
flaky case into a vacuous one, and a vacuous case is the defect this project keeps
recording.

**Prove each repair in both directions.** The case still fails on the defect it exists to
catch — for case 2, a measurement that captures process start rather than traffic — and it
passes under a load that broke it before. **Reproduce the load deliberately rather than
waiting for it.**

## Acceptance criteria

- [ ] Each named case passes with the machine deliberately loaded, and the pull request
      states how the load was produced and what it measured.
- [ ] Each named case still fails on the defect it exists to catch, and the pull request
      holds that failing output.
- [ ] `TestTheWorkControl` still fails when the elapsed time is replaced by a constant.
- [ ] No tolerance is widened without the pull request naming what the case can no longer
      detect.
- [ ] `pytest tests/ -m spec_validation` reports 1532 passed, 143 skipped and 134 xfailed
      against 134 register keys. No file under `ja4plus/` changes.

## Links

[...]

Issue #426, opened 2026-08-09
Guard the four cases that read ambient host state


Crank-Git:
## What the sweep found

#424 repaired one case that reads ambient host state without a guard on that state. The
sweep that #424 ran found four more cases of the same shape. #424 repairs none of them,
because they sit outside its scope.

A case of this shape fails on a correct package. A reader who meets such a failure learns
to disregard the gate.

### 1. The filter cases depend on an ambient `libpcap`, and they catch the wrong class

`tests/test_watch_capture.py:155` — `the_filter_failure()`

```python
    from scapy.arch.common import compile_filter

    try:
        compile_filter(INVALID_FILTER, linktype=1)
    except Scapy_Exception as error:
        return error
```

`scapy` 2.7.0 raises `ImportError("libpcap is not available. Cannot compile filter !")`
from `compile_filter` where the host holds no `libpcap`. The call catches
`Scapy_Exception` alone, so the case ends with an error and not with a failure.

Three cases call it: `TheCommandReportsTheFilterError::test_the_message_names_the_filter_the_operator_stated`,
`TheCommandReportsTheFilterError::test_the_message_repeats_what_the_capture_layer_reported`
and `TheCaptureFailureEndsTheRunWithTheStatusOne::test_an_invalid_filter_ends_the_run_with_the_status_one`.
`TheCaptureFailuresNameTheClassesTheCaptureLayerRaises::test_the_three_real_failures_are_all_caught`
calls it too.

Severity: medium. A minimal Linux container holds no `libpcap`.

### 2. The interface list case requires the host to hold one interface

`tests/test_watch_capture.py:242` — `TheCommandListsTheInterfacesItFound::test_the_interface_list_of_this_host_holds_a_name`

```python
        names = available_interfaces()
        self.assertTrue(names, "the host reported no interface")
```

The case guards on nothing. A host whose capture layer reports no interface fails it.

Severity: low.

### 3. The absent interface helper requires the host to hold no `nosuchif0`

`tests/test_watch_capture.py:135` — `the_absent_interface_failure()`

The helper raises `AssertionError` where the host holds an interface named `nosuchif0`.
It guards on nothing. Four cases call it.

Severity: low.

### 4. The descriptor count case reads `/dev/fd` under no platform guard

`tests/test_watch_stop.py:433` — `TheMonitorHoldsNoFileDescriptorAcrossTheCalls::test_three_hundred_calls_hold_the_open_descriptor_count`

```python
[...]

Issue #424, opened 2026-08-09
Let the capture privilege case read a host that grants it


Crank-Git:
## What the measurement shows

`tests/test_watch_capture.py::TheCommandNamesTheCapturePrivilege::test_the_message_reads_the_failure_this_host_reports`
fails on a host that grants the capture privilege. The case asks `get_dev_bpf()` for the
error it raises, and it raises none where the host grants the privilege.

```
>       raise AssertionError("the host granted the capture privilege")
E       AssertionError: the host granted the capture privilege

tests/test_watch_capture.py:132: AssertionError
```

**The case reads the host and not the package.** It passes on a runner that denies the
privilege, and it fails on a developer host that grants it. A case that fails when nothing
is wrong teaches a reader to disregard it.

## The evidence

#415 measured it on 2026-08-09 on an Apple silicon laptop, `macOS-26.6.1-arm64-arm-64bit-Mach-O`,
Python 3.14.3. The case fails on commit `6862ac6` of `epic/406-pre-release-validation`,
which is the unchanged base, and it fails the same way on the branch of #415. #415 changes
no file under `ja4plus/` and no file of `tests/test_watch_capture.py`.

`git diff --stat 6862ac6..HEAD -- tests/test_watch_capture.py ja4plus/` prints nothing.

## What to decide

The issue states the finding and proposes no repair. Two shapes are open, and the reader
of this issue chooses one.

1. The case skips where the host grants the privilege, and it names the reason.
2. The case reads a stand-in that raises, so the message reaches the reader on every host.

Shape 2 states what the message says. Shape 1 states what the host reports. The two answer
different questions, so the choice is a decision and not a repair.

## Links

- Found by: #415
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)

Crank-Git:
## Plan, from the project manager. This is `priority:high` and it blocks local verification

**Base: `epic/406-pre-release-validation` at `6ff5fad`.**

### The cause, measured and not inferred

**The user granted this account read access to `/dev/bpf*` on 2026-08-09**, for #423. The
project manager asked for that grant.

```
sudo chown $(whoami) /dev/bpf*
python3 -c "os.close(os.open('/dev/bpf0', os.O_RDONLY))"   ->  READABLE
pytest tests/test_watch_capture.py::TheCommandNamesTheCapturePrivilege
[...]

PR #443, opened 2026-08-09
Guard the four cases that read ambient host state

Crank-Git:

It guards the four cases the sweep of #424 found, and it deletes none of them. Each case
states a real requirement, so each guard reads the ambient state and skips with a reason
that names it. **No file under `ja4plus/` changes and no fingerprint moves.**

Files: `tests/test_watch_capture.py`, `tests/test_watch_stop.py`, `CHANGELOG.md`,
`docs/specs/spec.md`.

## The premise of finding 1, re-measured rather than trusted

The issue states that `scapy` 2.7.0 raises `ImportError` from `compile_filter`. The
installed source holds it at `scapy/arch/common.py:86` and `:87`:

```python
    except OSError:
        raise ImportError(
            "libpcap is not available. Cannot compile filter !"
        )
```

`Scapy_Exception` is no subclass of `ImportError`, so the earlier `except Scapy_Exception`
never caught it. The red run below reproduces that as an **error** and not a failure,
exactly as the issue predicted.

## The four guards

| # | The state the case reads | The guard |
|---|---|---|
| 1 | Whether the host holds `libpcap` | `the_filter_failure` returns `None` on `ImportError`, and `the_filter_failure_or_skip` skips the four cases that read it |
| 2 | Whether the capture layer reports one interface | `test_the_interface_list_of_this_host_holds_a_name` skips on an empty list |
| 3 | Whether the host holds an interface named `nosuchif0` | `the_absent_interface_failure` returns `None`, and `the_absent_interface_failure_or_skip` skips the four cases that read it |
| 4 | Whether the host holds a `/dev/fd` directory | `the_open_descriptor_count` returns `None`, and the count case skips |

## Test-driven development: the cases came first

The four guard-proof classes ran against the unrepaired helpers. **Twelve cases failed, and
every failure sat in the skip direction**; the run directions already passed.

```
12 failed, 69 passed, 1 skipped, 1 warning, 8 subtests passed in 2.75s
```

The failures, quoted:

```
ImportError: libpcap is not available. Cannot compile filter !
  raised at .venv/.../scapy/arch/common.py:87, through the_filter_failure
AssertionError: the host holds an interface named nosuchif0
  raised at tests/test_watch_capture.py:211
AssertionError: [] is not true : the host reported no interface
  raised at tests/test_watch_capture.py:369
AssertionError: 0 != 1
  raised at tests/test_watch_stop.py:518, where the count case did not skip
```

After the guards landed: `73 passed, 1 skipped, 16 subtests passed`. The one skip is the
privilege case #424 guards, and it is unrelated.

## Every guard is proved in both directions

**Each guard-proof case passed and none of them skipped**, so both branches were reached on
this host. The whole run holds exactly one skip, which is #424's.

### How the skip branch was reached

| Guard | The route to the skip branch |
|---|---|
| 1, `libpcap` | A finder on `sys.meta_path` raises `OSError` from the import of `scapy.libs.winpcapy`. That is the failure the loader reports where the shared library is absent, so **`scapy` raises the `ImportError` itself** and no stand-in for `compile_filter` is involved. `test_the_probe_reads_the_import_error_scapy_raises` reads that class from the real call. |
| 2, the interface list | `patch("scapy.all.get_if_list", return_value=[])`, which is the read `available_interfaces` makes. |
| 3, `nosuchif0` | `patch("scapy.all.resolve_iface", ...)` returns an interface, which is the state a host that holds the name produces. |
| 4, `/dev/fd` | `DESCRIPTOR_DIRECTORY` points at `/dev/fd/ja4plus-no-such-directory`. `os.listdir` raises `FileNotFoundError` from the same call Windows fails on, and a case first asserts the path is absent. |

### How the run branch was reached

The host is this laptop, macOS, Python 3.14, `scapy` 2.7.0. It holds `libpcap`, it reports
interfaces, it holds no `nosuchif0`, and it holds `/dev/fd`. Each prover runs the guarded
cases unpatched and requires `testsRun == 1`, no error, no failure and **no skip**.

**The `libpcap`-absent host was not needed and `bigboy` was not used.** Controlling the
import reaches the branch on any host, and it reaches it through `scapy`'s own translation,
which a host without `libpcap` would also exercise. That is stronger evidence than a second
host, because it is reproducible on every host.

## A guard proved in one direction might skip always, and that hazard bit this branch

Each guard was mutated to skip on every host, and the mutation was confirmed live in the
running interpreter with `inspect.getsource` in both directions.

**The first form of the provers did not catch it.** Each run-direction prover read the
ambient state *through the guard it proves*, so a guard that always returned `None` made the
prover skip as well:

```
########## MUTATION filter: the guard skips always ##########
LIVE filter returns None x 2
1 failed, 68 passed, 6 skipped, 11 subtests passed
```

Only one subtest failed, and it failed by luck, through a case shared with another prover.
Every prover now reads the ambient state by an independent route — `compile_filter`,
`resolve_iface`, `os.path.isdir` — which is a route that passes the guard. The same
mutations now read:

| Mutation | What fails |
|---|---|
| `the_filter_failure` returns `None` always | 4 subtests of `test_each_case_runs_and_passes_where_the_host_holds_libpcap`, plus the shared case in the interface prover — `5 failed` |
| `the_absent_interface_failure` returns `None` always | 4 subtests of `test_each_case_runs_and_passes_where_the_host_holds_no_such_name`, plus 2 shared — `6 failed` |
| The interface-list guard skips always | `test_the_case_runs_and_passes_where_the_host_reports_one_interface` |
| `the_open_descriptor_count` returns `None` always | `test_the_case_runs_and_passes_where_the_host_holds_the_directory` and the leak case |

Each run restored the file, and `RESTORED: True` was read after each one.

## The second defect of case 4, and the tolerance

The count reads every descriptor of the process, so another part of the run that opens one
moves it, and the case compared for equality. The comparison is now
`assertLessEqual(max(counts) - before, DESCRIPTOR_TOLERANCE)` with the tolerance at 16.

**A tolerance that also covered the leak would measure nothing**, so the tolerance is
proved unable to hide the leak rather than asserted safe.
`test_the_case_fails_where_the_monitor_leaks_a_descriptor_for_each_call` opens one
descriptor inside each capture call, runs **the count case itself**, and requires it to
fail. The leak the case exists to detect opens two descriptors for each `sniff` call, which
is 600 over 300 calls against a tolerance of 16.

Two mutations prove the numbers are load-bearing, and each turns that case red:

| Mutation | Result |
|---|---|
| `DESCRIPTOR_TOLERANCE = 100000` | `1 failed` — `test_the_case_fails_where_the_monitor_leaks_a_descriptor_for_each_call` |
| The count case compares against a hardcoded `100000` | `1 failed` — the same case |

**One finding against my own first form.** The leak case derived its call count from the
tolerance, so the first tolerance mutation opened 200001 descriptors and had to be killed.
The count is now the fixed `DESCRIPTOR_CALLS`, and the case asserts
`DESCRIPTOR_CALLS > DESCRIPTOR_TOLERANCE`, so a raised tolerance fails the case cheaply
instead of exhausting the descriptor table.

## The rejection cases still reject

`the_filter_failure` keeps `raise AssertionError("libpcap accepted the expression ...")`.
A host that compiled `tcp port` reports no filter failure for any expression, and that is a
finding rather than a state of the host. **The repair widens no caught class until nothing
fails**: the case still fails when the filter is genuinely wrong, which the run direction
reads on every host that holds `libpcap`.

## The records read `TBD` and no number

The brief assigned round 159. **That premise is wrong and the project manager corrected it
mid-task; I re-measured the correction and it holds.**
`tests/test_specification_changelog.py:66-71` requires the row count of
`docs/specs/spec.md` to equal its highest round number, and the table holds 156 rows today,
so a row numbered 159 fails that gate until 157 and 158 land.
`tests/test_changelog_round_agreement.py:3-5` states the convention: every worker writes
the literal `TBD`. Both records now read `Round TBD` and they cite #426, which
`test_an_unassigned_entry_matches_an_unassigned_row` requires.

## The verification floor

The branch was rebased onto `d160d0f` after #371 and #430 landed, and every gate ran again on commit `503b07f`. The first comment below holds the self-review and the rebase record.

Every gate ran in this worktree, in a virtual environment inside the
worktree, with `PYTHONDONTWRITEBYTECODE=1` and every `__pycache__` cleared first.

| Gate | Result |
|---|---|
| `pytest tests/ -m "not spec_validation"` | **2625 passed, 4 skipped, 1825 deselected, 8 xfailed, 114 subtests** |
| `pytest tests/ -m spec_validation` | **1532 passed, 143 skipped, 134 xfailed** |
| `ruff format --check ja4plus/ tests/` | 189 files already formatted |

**The register invariant holds.** `tests/foxio_deviations.json` holds 134 keys and the
conformance gate reports 134 xfailed.

**Coverage does not fall.** Both measurements used `--cov=ja4plus` on this host.

| Commit | Passed | Skipped | Coverage |
|---|---|---|---|
| `d160d0f`, the unchanged base | 2612 | 4 | 4232 statements, 273 missed, 94% |
| `503b07f`, this branch | 2625 | 4 | 4232 statements, 273 missed, 94% |

The difference of 13 is exactly the 13 cases the four guard-proof classes add, and the
subtest count rises from 98 to 114. **The skip count does not move**, which is the reading
that shows no guard skips on this host.

**The new round-entry gate passes.** `tests/test_round_entry_existence.py` is part of the
green unit run above.

## Verified against

- `scapy` 2.7.0 as installed, `scapy/arch/common.py:79-88`, read on 2026-08-09 with
  `inspect.getsource`.
- The repair shape of #424, commit `9c487ce`.

Refs #426


Crank-Git:
## Self-review, and what it changed

Two Sonnet reviewers read the diff, one for correctness and one against
`.claude/rules/ste.md`. Commit `503b07f` answers both. **Three correctness findings were
real and all three are fixed. One was declined and the reason is below.**

### Correctness

1. **`the_open_descriptor_count` caught `OSError`, which is wider than the state it reads.**
   A `PermissionError` on a host that holds `/dev/fd` would have skipped the count case with
   the reason `this host holds no /dev/fd directory`, which names the wrong state. The catch
   now names `FileNotFoundError`. That is the class Windows raises, and it is the class
   `os.listdir("/dev/fd/ja4plus-no-such-directory")` raises, measured as `errno 2`, so the
   skip proof still reaches the branch.

2. **The shared case coupled the two capture guards, and the skip reason could name the
   wrong state.** `test_the_three_real_failures_are_all_caught` reads two ambient states,
   and it evaluates the absent interface before the filter. The `libpcap` prover asserts
   that the skip reason names `libpcap`, so **a host that held an interface named
   `nosuchif0` would have skipped on the other state and failed the prover.** Each prover
   now pins the state it does not prove: the `libpcap` prover pins `resolve_iface` to raise
   `ValueError`, and the interface prover pins `compile_filter` to raise the
   `Scapy_Exception` a host with `libpcap` reports. Both provers are now deterministic on
   any host, and the run-direction provers carry the same pin for the same reason.

3. **The tolerance stated no limit.** A leak of fewer than `DESCRIPTOR_TOLERANCE`
   descriptors over the whole run now passes the count case, where the earlier comparison
   for equality failed it. That is a real loss of strength and it is the price of repairing
   the second defect, because the equality read every descriptor of the process and failed
   on a run that opened one elsewhere. **The limit is now stated in three places** rather
   than implied: the constant, the leak case docstring, and both records.

**Declined: nothing.** The fourth item the reviewer raised was a request to confirm that
item 3 is intended scope. It is: #426 names that defect in its own words — "The count also
reads every descriptor of the process, so another part of the run that opens a descriptor
moves it."

### All four always-skip mutations still fail after the fixes

The pins could have made a prover vacuous, so the four mutations ran again against
`503b07f`. Each guard was made to skip on every host, and the mutation was confirmed live
with `inspect.getsource`.

| Mutation | Failures |
|---|---|
| `the_filter_failure` returns `None` always | 5 failed |
| `the_absent_interface_failure` returns `None` always | 6 failed |
| The interface-list guard skips always | 1 failed |
| `the_open_descriptor_count` returns `None` always | 2 failed |

Two more mutations hold the tolerance: `DESCRIPTOR_TOLERANCE = 100000` and a hardcoded
`100000` in the count case each turn
`test_the_case_fails_where_the_monitor_leaks_a_descriptor_for_each_call` red.

### Writing standard

Nine sentences came under the 25-word limit, across two docstrings, three comments, the
`CHANGELOG.md` entry and the `docs/specs/spec.md` row. One `-ing` noun went. One docstring
opened with an imperative and now opens with the result.

## The branch was rebased, and the base moved

#371 and #430 landed on `batch/437-test-hygiene` while this work ran, so the base is now
`d160d0f` and not `f140a5c`. **The branch is rebased and not merged**, because a merge commit
does not inherit `[skip ci]`. Both conflicts were in the two records, where three workers
each append an entry, and the resolution keeps all three. `gh run list` for this branch
returns `[]`, so no provider CI ran on either push.

**Every gate was measured again after the rebase.** `ja4plus/watch.py` moved in #371, and
these test files import it, so the earlier numbers could not be carried over.

| Gate | Result |
|---|---|
| `pytest tests/ -m "not spec_validation"` | **2625 passed, 4 skipped, 1825 deselected, 8 xfailed, 114 subtests** |
| `pytest tests/ -m spec_validation` | **1532 passed, 143 skipped, 134 xfailed** |
| `ruff format --check ja4plus/ tests/` | 189 files already formatted |

| Commit | Passed | Skipped | Subtests | Coverage |
|---|---|---|---|---|
| `d160d0f`, the unchanged new base | 2612 | 4 | 98 | 4232 statements, 273 missed, 94% |
| `503b07f`, this branch | 2625 | 4 | 114 | 4232 statements, 273 missed, 94% |

The register invariant holds: `tests/foxio_deviations.json` holds 134 keys against 134
xfailed. **The skip count does not move**, which is the reading that shows no guard skips on
this host.

Crank-Git:
Sub-merge gate. Base `d160d0f`, clean, no conflict.

| Check | Reading |
|---|---|
| Change set | `CHANGELOG.md`, `docs/specs/spec.md`, `tests/test_watch_capture.py`, `tests/test_watch_stop.py`. No file under `ja4plus/` |
| Records | one `TBD` row, the third of the batch |
| Provider runs | `gh run list` on this branch returns none, and both commits carry `[skip ci]` |
| Threads | none open |

**The hazard the brief named actually bit, and the worker reported it rather than working around it.** The first form of every run-direction prover read the ambient state through the guard it proves, so a mutation that made the guard skip always made the prover skip too, and the prover then measured nothing. Each prover now reads the state by an independent route — `compile_filter`, `resolve_iface`, `os.path.isdir`, a patched `get_all` — so no prover depends on a guard it does not prove. **That is the failure mode the acceptance criterion exists to catch, caught inside the issue that wrote the criterion.**

**Each guard was mutated to skip on every host and the suite went red each time**: filter 5 failed, interface 6 failed, interface list 1 failed, descriptor 2 failed. The skip count of the whole run does not move, so no guard skips on this host.

**One tolerance is introduced and it is bounded by a proof.** `DESCRIPTOR_TOLERANCE` reads 16 against the 600 descriptors a real leak opens over 300 calls, and `test_the_case_fails_where_the_monitor_leaks_a_descriptor_for_each_call` requires the count case to fail on that leak. Two mutations of the tolerance turn that case red. The limit it introduces, a leak below 16 across the whole run, is stated in the constant, the docstring and both records.

**No case was deleted**: `git diff -- tests/ | grep '^-.*def test_'` prints nothing.

Issue #441, opened 2026-08-09
Name the statistics thread by its controlled term


Crank-Git:
The `## Terms` table of `docs/specs/spec.md` holds this row:

```
| statistics thread | noun | The thread that writes a statistics line for each interval that passes. | reporter, timer, ticker |
```

The fourth column names the words a writer does not use. `reporter` is one of them. The
prose of the package uses `the reporter` throughout, so the controlled vocabulary and the
prose disagree.

The self-review of #371 found it. #371 declined the repair, because a repair inside one diff
would put `the reporter` and `the statistics thread` in adjacent sentences of one paragraph,
and rule 7 of `.claude/rules/ste.md` bars that.

## Where the word reads today

| File | What holds it |
|---|---|
| `ja4plus/watch.py` | The class docstring of `StatisticsReporter`, and the `wait` parameter of `StatisticsReporter` and of `report_statistics` |
| `docs/api_reference.md` | The two paragraphs on the `wait` seam |
| `tests/test_watch_statistics.py` | The class docstring of `TheReporterWritesOneLinePerInterval`, and several case docstrings |
| `docs/specs/spec.md` | The Changelog rows of #55, #369 and #371 |

`git grep -n 'the reporter'` reads the current set.

## What to decide

The identifiers `StatisticsReporter`, `report_statistics` and `TheReporterWritesOneLinePerInterval`
carry the word too. `.claude/rules/ste.md` exempts an identifier from the standard, so a
decision that renames the prose alone is available and a decision that renames the class is a
wider change.

Three readings are open, and the user decides.

1. The prose reads `the statistics thread`, and every identifier holds its name.
2. The prose and the identifiers both change.
3. The `## Terms` table admits `reporter`, and the row drops it from the rejected column.

## Notes

A Changelog row records a past round, so a row that a rename edits reads differently from
the round it records. Name what happens to those rows in the decision.

No file under `ja4plus/fingerprinters/` changes and no fingerprint moves under any of the
three readings.

Crank-Git:
Parked for the user. The three readings are a vocabulary decision, and the project manager does not rule on the controlled vocabulary.

The question: does the prose alone read `the statistics thread` (reading 1), do the prose and the identifiers both change (reading 2), or does the `## Terms` table admit `reporter` and drop it from the rejected column (reading 3)?

[...]

Issue #85, opened 2026-08-06
No Tests run fired on the merge to dev


Crank-Git:
## What happened

The Epic 0 batch merged into `dev` as `0254ffc` at 21:09 on 2026-08-06. The `Tests`
workflow did not run. No run was created and no check suite was attached to the merge
commit.

## What was verified

- `.github/workflows/test.yml` is on `dev` and parses. Top-level keys are `name`,
  `on`, `env`, `jobs`. Jobs are `lint`, `test`, `conformance`.
- The triggers read `push: branches: [master, dev]` and
  `pull_request: branches: [master, dev]`.
- `gh api repos/Crank-Git/ja4plus/actions/workflows/251432872` reports
  `state=active`, `path=.github/workflows/test.yml`.
- `gh api repos/Crank-Git/ja4plus/actions/permissions` reports `enabled=true`,
  `allowed_actions=all`.
- The head commit message on `dev` is `Merge pull request #81 from
  Crank-Git/epic/11-foundation`, which holds no `[skip ci]` directive.
- The workflow has run on `pull_request` before, most recently on
  `feat/foxio-spec-2026-05` in May, so the mechanism works on this repository.
- Dependabot runs were created on `dev` at the same minute, so Actions is processing
  events on this branch.

Pull request #81 also received no check. That part is explained: GitHub reads the
`on.pull_request.branches` filter from the workflow file on the **base** branch, and
`dev` still held `branches: [master]` until #81 merged. **The push to `dev` after the
merge is not explained.**

## Why it matters

The batch gate for Epic 0 was the project manager's own suite run, because no
continuous integration existed. #24 was supposed to end that. Until a run appears,
every batch gate stays a local run, which no reviewer can see and no branch
protection rule can require.

## What to do

1. Read the repository Actions tab and the workflow's own page. A run that GitHub
   declined to create sometimes states the reason there and nowhere in the API.
2. Confirm against the current GitHub Actions documentation how a push trigger is
   evaluated on a merge commit. Do not assume it.
3. If the cause is the merge commit, state it and add the trigger that covers it.
4. Add a `workflow_dispatch` trigger, so a run can be started by hand while the cause
   is unknown. That alone is not the fix.
5. Prove the fix: open a throwaway pull request into `dev`, show the run, close it.

## Acceptance criteria

- [ ] A `Tests` run appears for a pull request into `dev`, with the `lint`, `test`
      and `conformance` jobs.
[...]

PR #444, opened 2026-08-09
Batch #437: test hygiene — rounds 156 to 159

### LINKED ISSUE #371: Remove the wall-clock race from the periodic statistics case

### LINKED ISSUE #426: Guard the four cases that read ambient host state

### LINKED ISSUE #429: Fail a change that records no round

### LINKED ISSUE #430: Remove the load sensitivity from two timing cases

Crank-Git:
Batch #437, test hygiene. The user ruled this batch next on 2026-08-09. Each member held a
gate that could pass while the thing it checks was absent or wrong, so each one weakened
every later gate while it stood.

Closes #429
Closes #371
Closes #430
Closes #426

## The four members, in sub-merge order

| Round | Issue | What it repaired |
|---|---|---|
| 156 | #429 | A change set that edits a tracked file and records no round now fails a case. The three existing cases held the two records **to each other**, so they agreed when both were absent. |
| 157 | #371 | The periodic statistics case states the schedule through an injected wait that reaches `report_statistics` from `cmd_watch`. The 0.3 second sleep and the lower-bound assertion are gone. |
| 158 | #430 | Two timing cases read the work performed rather than the seconds elapsed. One of them, `TestTheWorkControl`, is a control, and a flaky control is worse than a missing one. |
| 159 | #426 | The four remaining cases that read ambient host state carry a guard, and every guard is proved in both directions. |

## What each member found that its own issue did not state

**#429 built a check that guards no pull request, and it said so.** `actions/checkout` clones
at depth 1, so the case reads no merge base on a runner and skips on every job. It guards a
worker's local gate alone. **#438 owns the checkout** and stays outside this batch, because
one of its criteria needs a deliberately red pull request and a batch gate must reach green.

**#430 found the reassembler case vacuous, not merely flaky.** It now fails at
`2499.5 not less than or equal to 4` against a duplicate check that scans the whole segment
list. A review proposal of a 1.5 ratio floor was **declined on measurement**: the two fastest
runs read 1.219 under a load average of 20 against 1.96 on a quiet host, so that floor fails
a sound package.

**#426 met the failure mode its own acceptance criterion names.** The first form of every
run-direction prover read the ambient state *through the guard it proves*, so a mutation that
made the guard skip always made the prover skip too, and the prover then measured nothing.
Each prover now reads the state by an independent route.

**#371's self-review filed #441**, a disagreement between the `## Terms` table, which names
`reporter` a word a writer does not use, and the prose of the package, which uses it
throughout. It is `status:needs-feedback` for the user and nothing here waits on it.

## The project manager error this batch recorded

**The brief of every member named a Changelog round number, and a worker writes `TBD`.**
`tests/test_changelog_round_agreement.py:3-5` states the convention and
`tests/test_specification_changelog.py:66-71` enforces it: the row count of
`docs/specs/spec.md` equals its highest round, so a row numbered 157 fails until 156 lands.
**#371's worker contradicted the premise rather than working around it**, and the correction
reached #430 and #426 while they ran. #429 wrote 156 and it was legal only because 156 was
the next number. The three numbers of this batch were assigned here, at the gate, in
sub-merge order.

## What changed under `ja4plus/`

**One line, and no fingerprint moves.** `report_statistics` of `ja4plus/watch.py` gains an
optional `wait` parameter and forwards it to `StatisticsReporter`. `ja4plus/cli.py` passes
nothing, so the shipped behaviour does not move. **No file under `ja4plus/fingerprinters/`
changed.**

## The sub-merge conflict

**One conflict, of the shape this project always produces**: #429 and #371 each appended one
Changelog row, so `docs/specs/spec.md` conflicted. Both entries were kept in merge order and
**the five gates were then run on the merged state**, not on the state either branch had
measured. #430 and #426 rebased onto the moved base and needed no resolution.


Crank-Git:
## The batch gate, and how the run was started

**No provider run was created by the pull-request event.** `gh run list` on this branch returned nothing 45 seconds after the pull request opened and nothing again a minute later, while the pull request of the previous batch had created two runs on its own event. **The head commit `2a589ca` carries no `[skip ci]`**, so the skip path is not the cause.

`.github/workflows/test.yml` records the reason it carries `workflow_dispatch`: "When GitHub throttles the webhook path, a push and a pull request create no run, and a manual run is the only way to get one. See issue #85." **This is that state.** The run was started by hand on `test.yml` and on `docs-build.yml`, both against `batch/437-test-hygiene`.

**`docs.yml` was not dispatched.** It publishes the site and it triggers on a push to `master`, so a manual run from an integration branch would publish from the wrong branch. `docs-build.yml` builds the site and publishes nothing; the two share the name "Documentation" in the interface and they are different files.

## The local gate on the merged state, with the rounds assigned

| Gate | Reading |
|---|---|
| Unit | 2625 passed, 4 skipped, 1825 deselected, 8 xfailed, 114 subtests. 2595 base, plus 17 from #429 and 13 from #426 |
| Conformance | 1532 passed, 143 skipped, 134 xfailed |
| Register | 134 keys of `tests/foxio_deviations.json` against those 134 xfailed |
| `ruff check` | All checks passed |
| `ruff format --check` | 189 files already formatted |
| Records | `docs/specs/spec.md` holds 159 rows numbered 1 through 159, contiguous, no gap, no duplicate, no live `TBD` |

**One `Round TBD` survives in `CHANGELOG.md` and it is correct.** It is the backticked string inside the entry of round 134, which records the eight orphans #302 found. **An aggregate that counts that string counts the string**, so the assignment read each entry rather than counting matches.

Issue #446, opened 2026-08-09
Decide the pin shape for the rest of the dev extra


Crank-Git:
Found by #378 on 2026-08-09 while it pinned the `ruff` version. **#378 pinned `ruff` alone and changed nothing else**, because its acceptance criteria name `ruff` and no other distribution. The Notes of #378 asked for this measurement and it is recorded below.

## The measurement

Every other entry of the `dev` extra of `pyproject.toml` still floats, and two of them have already drifted across a major line.

| Entry as declared | What `pip` installs today | Distance |
|---|---|---|
| `pytest>=7.0` | 9.1.1 | Two major lines above the floor |
| `mypy>=1.11` | 2.3.0 | One major line above the floor |
| `pytest-cov>=5.0` | 7.1.0 | Two major lines above the floor |
| `build>=1.0` | 1.5.0 | Same major line |

`ruff==0.16.2` is the one exact entry.

## Why this is worth an issue

**The defect #378 records is not repaired for `mypy`.** `mypy --strict ja4plus/` is a gate of every pull request, and a `mypy` release that adds a check turns it red on an unchanged tree. `pyproject.toml` already carries the evidence that a major `mypy` release moved under this project: the comment on `[tool.mypy]` states that `python_version` stays unset, because `mypy` 2.x rejects the value `3.9`.

**A `pytest` release can change which cases run.** A collection rule or a fixture rule that changes moves the suite result, and no commit records the move.

## The second defect: a tracked `requirements.txt` disagrees with `pyproject.toml`

`requirements.txt` is tracked and it holds a second dependency declaration:

```
# Core dependencies
scapy>=2.4.0
cryptography>=42.0.0

# Development dependencies (optional)
pytest>=7.0.0
pytest-cov>=3.0.0
```

**`pytest-cov>=3.0.0` contradicts the `pytest-cov>=5.0` of `pyproject.toml`.** The comment beside that entry states why 5.0 is the floor: 5.0 is the first release that requires `coverage` 7.5, which carries `coverage report --format=total`, and the continuous-integration floor reads that value. A contributor who installs from `requirements.txt` therefore gets an environment the coverage gate cannot read.

**No workflow reads `requirements.txt`.** Every job of `.github/workflows/test.yml` installs with `pip install -e ".[dev]"`. The file is a second record that nothing checks and that a reader can follow.

## The decision this needs

Three questions, and the second one is the one that carries a trade.

[...]

Issue #445, opened 2026-08-09
Batch: documentation repairs — #378 #380 #387 #393


Crank-Git:
## The batch

The documentation repairs, the group the user named after test hygiene on 2026-08-09. Each
member is a document that states something the repository contradicts, so a reader who
follows it reaches the wrong answer.

- [x] #378 — pin the ruff version so the lint gate cannot change without a commit
- [x] #380 — correct the stale register key count the rule document states
- [x] #387 — correct the three documents that state ten implemented methods
- [x] #393 — amend the documentation behaviour rule that names a doctest command collecting
      nothing

## Why these four, and not the other two

**The grouping is by file, because two workers on one file collide for no gain.**

- **#388 also edits `README.md`**, which #387 edits, and it also edits
  `.claude/rules/conformance.md`. It moves to the next batch.
- **#398 edits `docs/migration-0.6-to-1.0.md`** with `CHANGELOG.md`, and it is the last
  documentation repair. It moves to the next batch with #388 and the sweep-tool repairs.

**Every member edits `docs/specs/spec.md`, and that is not a collision to avoid.** Each
appends one Changelog row, the sub-merge conflict is the two appended rows, and the
resolution keeps both in merge order. The gates then run on the merged state.

## The integration branch

`batch/445-documentation-repairs`, forked from `dev` at `d123ab9`.

## Rules that bind every member

- **A worker writes the literal `Round TBD` in both records and chooses no number.** The
  project manager assigns each number at this gate. `tests/test_specification_changelog.py`
  requires the row count of `docs/specs/spec.md` to equal its highest round, so a number a
  worker picks fails the gate until the numbers below it land.
- No file under `ja4plus/fingerprinters/` changes and no fingerprint moves.
- **A document that states a count states a count the repository can produce.** Re-measure
  every number rather than copying it.

Crank-Git:
Batch #445 merged as `d94d8c1`. Rounds 160 to 163. All four members closed.

**Twelve provider jobs green on the integration branch**, started by hand because the pull-request event created no run for the second consecutive batch. **#459 owns that gap at `priority:high`.** The head SHA of the run was read against the pull-request head before the merge.

[...]

Issue #378, opened 2026-08-09
Pin the ruff version so the lint gate cannot change without a commit


Crank-Git:
Found by #297 on 2026-08-09 while it enabled the `ruff` rule F401. **#297 measured the drift and made no change to the pin**, because the pin was not in its scope.

## The measurement

`pyproject.toml` pins `ruff>=0.6`, so the version floats. **The counts #297 recorded against 0.14.5 and re-measured against 0.16.2 disagree on every line.**

| Measurement | `ruff` 0.14.5 | `ruff` 0.16.2 |
|---|---|---|
| F401 findings | 58 | 54 |
| Files holding one | 28 | 27 |
| `I001` findings | 82 | 76 |

## Why this is worth an issue

**The lint gate can turn red without a commit.** A `ruff` release that adds a rule to a selected group, or that widens an existing rule, changes the result of `ruff check ja4plus/ tests/` on an unchanged tree. Continuous integration installs the newest version that satisfies the pin, so **the run that passes today and the run that fails tomorrow read the same source.**

**A contributor and continuous integration can disagree for the same reason**, and neither is wrong.

**This is not theoretical for this project.** #297 was triaged with numbers measured under one version and worked under another, and every number differed.

## The decision this needs

Three shapes, and the trade is real rather than obvious.

1. **Pin an exact version**, for example `ruff==0.16.2`. The gate becomes reproducible and a version bump becomes a deliberate commit that shows its own diff. The cost is that the pin needs maintenance, and a project that forgets it drifts far behind.
2. **Pin a compatible range**, for example `ruff>=0.16,<0.17`. New rules within the minor line can still land.
3. **Leave the pin and accept the drift**, recording that a red gate on an unchanged tree is expected and is repaired by removing the new findings.

**Shape 1 is the reading this project's own rules point at**: a gate whose result depends on the day it runs is a comparison that measures something other than what it names.

## Acceptance criteria

- [ ] The `ruff` version the lint gate uses is stated in one place, and two runs of the gate on one tree cannot disagree because of a version difference.
- [ ] `ruff check ja4plus/ tests/` and `ruff format --check ja4plus/ tests/` pass at the chosen version.
- [ ] `pyproject.toml` records why the shape was chosen, in one sentence.
- [ ] No file under `ja4plus/` changes and no fingerprint moves.

## Notes

[...]

PR #448, opened 2026-08-09
Pin ruff to 0.16.2 so the lint gate cannot change without a commit

Crank-Git:
Batch member of #445. Draft, `[skip ci]` on every commit, base `batch/445-documentation-repairs`.

Refs #378.

## What changed

The `dev` extra of `pyproject.toml` declared `ruff>=0.6`, so `pip` resolved the newest release at the moment of each install. The entry now reads `ruff==0.16.2`.

| File | Change |
|---|---|
| `pyproject.toml` | `ruff>=0.6` becomes `ruff==0.16.2`, with the comment that states why |
| `tests/test_lint_gate_pin.py` | New. 6 cases hold the pin |
| `docs/specs/features/00-foundation.md` | `FR-foundation-8b`, `FR-foundation-8c`, and two behaviour rules that state the cost |
| `docs/specs/spec.md` | The `TBD` Changelog row, plus `pin` and `lint gate` in the `Terms` table |
| `CHANGELOG.md` | The `Round TBD` entry under `### Changed` |

**No file under `ja4plus/` changes, no vector changes, no register entry changes, and no fingerprint moves.** `git diff --stat <base> -- ja4plus/ tests/foxio_vectors/ tests/foxio_deviations.json` is empty.

## Why an exact pin

The project manager ruled shape 1 of the issue. A gate whose result depends on the day it runs measures something other than what it names.

**#297 measured the drift on this repository**, on an unchanged tree:

| Measurement | `ruff` 0.14.5 | `ruff` 0.16.2 |
|---|---|---|
| `F401` findings | 58 | 54 |
| Files holding one | 28 | 27 |
| `I001` findings | 82 | 76 |

## The pin moves what the gate installs today not at all

This is measured, not assumed. `pip index versions ruff` reports 0.16.2 as the newest release, so the floating specifier already resolved it. `ruff check ja4plus/ tests/` and `ruff format --check ja4plus/ tests/` each report no finding before the change and after it.

**The pinned release therefore reports no finding the floating one did not.** That is the condition the plan asked a worker to report rather than repair, and it did not arise.

## The `dev` extra is the one place a tool reads the version from

1. No file under `.github/workflows/` names `ruff` with a version specifier.
2. Every job of `.github/workflows/test.yml` installs with `pip install -e ".[dev]"`.
3. The tracked `requirements.txt` names no `ruff`.

`tests/test_lint_gate_pin.py` holds all three, so a later commit cannot open a second place quietly.

**Prose that quotes the pin reaches no case.** `CHANGELOG.md` and `docs/specs/spec.md` record `ruff==0.16.2` inside backticks, and no tool installs from a record, so the reader names the files a tool resolves a dependency from. The docstring states that limit.

## The tests came first

The unrepaired tree failed 4 of the 6 cases:

```
AssertionError: the dev extra states no exact version for ruff: 'ruff>=0.6'
```

The two cases that passed on the unrepaired tree are the two that hold the current single-place state in place, which is correct: nothing had opened a second place yet.

## Eleven mutations, each isolating to the case it targets

Every mutation restored from a snapshot of the implemented tree, under `PYTHONDONTWRITEBYTECODE=1` with every `__pycache__` cleared. The first harness restored from the index instead and contaminated six of its own readings; that is why the run was repeated.

|---|---|
| `ruff==0.16.1` | the installed-release case alone |
| the comment drops `#378` | the citation case alone |
| the comment drops the word `commit` | the commit rule alone |
| the pin moves above its own comment | both comment cases, on the empty string |
| a decoy `ruff-lsp` entry carries the comment | both comment cases |
| a workflow appends `"ruff>=0.6"` | the workflow case alone |
| a workflow appends `"ruff>0.6"` | the workflow case alone |
| `ruff==0.1.0` appended to `requirements.txt` | the dependency-file case alone |
| the extra names `ruff` twice | the pin case, at `the dev extra names ruff 2 times` |
| `ruff>=0.6` restored | the pin case and the installed-release case |
| a comment quotes `"ruff==0.14.5"` | none, by design — a comment is prose |

The decoy mutation and the last one are the two that matter, and both come from the self-review. The decoy proves the comment reader binds to the pin's own distribution name. The last one proves a comment's wording cannot move a case.

## The self-review found three defects, and all three are repaired

Full detail is in the review comment below. In short:

- **A substring test read the comment of the wrong entry.** A decoy `ruff-lsp` entry satisfied both comment cases while the pin carried no comment at all.
- **The specifier pattern missed a one-character operator.** `ruff>0.6` escaped it, which is the shape of the specifier this pin replaces.
- **The shared list parser counted a string inside a comment.** `_dependency_block` of `tests/test_documentation_site.py` collects every double-quoted substring of the block, and the `dev` extra carries comments inside its brackets, so a comment that quoted a version would have failed a case on the wording of a comment. This file now strips comment lines first and no longer imports that helper. `_dependency_block` itself is unchanged and stays correct for its own callers.

## Verification

Run on this host, in a virtual environment inside the worktree built with `pip install -e ".[dev]"`.

| Gate | Result |
|---|---|
| `pytest tests/ -m "not spec_validation"` | 2631 passed, 4 skipped, 8 xfailed, 114 subtests, 1825 deselected |
| Register invariant | `tests/foxio_deviations.json` holds 134 keys against those 134 xfailed |
| Coverage | 94%, 4232 statements, 273 missed |
| `ruff format --check ja4plus/ tests/` | 190 files already formatted |
| `mypy --strict ja4plus/` | no issues in 31 source files |
| `python -m build` | built `ja4plus-0.6.0.tar.gz` and the wheel |

The unit count rises from the 2625 of the base commit by exactly 6, which is the new file. **Coverage does not fall**: 94%, 4232 statements and 273 missed are the base readings unchanged, because this change adds no line under `ja4plus/`.

## Round

**`Round TBD` in `CHANGELOG.md` and `| TBD |` in `docs/specs/spec.md`.** The project manager assigns the number at the batch gate. `docs/specs/spec.md` still holds 159 numbered rows against a highest round of 159, because `CHANGELOG_ROW` of `tests/test_specification_changelog.py` matches a digit and not `TBD`.

## Two findings this PR reports and does not repair

The Notes of #378 asked whether `mypy`, `pytest` and the other development dependencies carry the same fault. They do, and **#446 carries both findings** because the acceptance criteria of #378 name `ruff` alone.

1. **The rest of the `dev` extra still floats, and two entries have drifted across a major line.** `pytest>=7.0` resolves 9.1.1, `mypy>=1.11` resolves 2.3.0, `pytest-cov>=5.0` resolves 7.1.0, `build>=1.0` resolves 1.5.0. `pyproject.toml` already carries evidence that a major `mypy` release moved under this project: the comment on `[tool.mypy]` states that `python_version` stays unset because `mypy` 2.x rejects the value `3.9`.
2. **A tracked `requirements.txt` disagrees with `pyproject.toml`.** It states `pytest-cov>=3.0.0` against the `>=5.0` of the extra, and the comment beside that entry states why 5.0 is the floor: 5.0 is the first release that requires `coverage` 7.5, which carries `coverage report --format=total`, and the continuous-integration floor reads that value. No workflow reads `requirements.txt`, so this is a second record that nothing checks.

#446 holds the `mypy` question open rather than pin by symmetry, because a pinned type checker that falls behind hides a defect a newer release would report.

## Practices

- **TDD** — the failing case is quoted above and it is the first commit, `c4a19ce`.
- **Coverage** — 94%, unchanged.
- **Conventional commits** — `test(lint): ...`, `build(deps): ...`, `test(lint): ...`.
- **Documentation** — two functional requirements, two behaviour rules, two `Terms` rows, and both records.
- **E2E** — none. This change ships no user-visible behaviour, and `practices` sets `e2e: user-facing`.
- **DDD** — not applicable. The change touches no domain model.

## Evidence for the external interface

Verified against `pip index versions ruff`, `pip index versions mypy` and `pip index versions pytest` run against PyPI on 2026-08-09, which is the machine-readable interface of the package index. 0.16.2 is the newest `ruff` release that command reports.


Crank-Git:
## Self-review of #378, by specialist lens

Two lenses read the diff: correctness with test quality, and the writing standard with record correctness. **Four findings were repaired in `9305ff1` and one was declined.** All three test defects are the shape this project records most often: a case that cannot fail.

### MUST-FIX 1 — a substring test read the comment of the wrong entry

`_lint_comment` matched the line that held `"ruff`, rather than comparing the distribution name. **A decoy entry defeated both comment cases.** With a `ruff-lsp` entry placed above the pin, carrying a one-line comment of its own, and the real pin stripped of its whole justification, `test_the_pin_cites_the_issue_that_chose_the_shape` and `test_the_pin_states_that_a_version_change_is_a_commit` both passed. The two cases exist to catch exactly that state.

`_distribution` now compares the whole name. The mutation is in the harness and both cases fail on it:

```
FAILED tests/test_lint_gate_pin.py::test_the_pin_cites_the_issue_that_chose_the_shape
FAILED tests/test_lint_gate_pin.py::test_the_pin_states_that_a_version_change_is_a_commit
2 failed, 4 passed
```

### MUST-FIX 2 — the specifier pattern missed a one-character operator

`ANY_SPECIFIER` read `[=<>!~]=` and demanded a trailing `=`. It therefore missed `ruff>0.6`, **which is the shape of the specifier this pin replaces**, and it missed `ruff[extra]==1.0`. A workflow that resurrected either one passed both "one place" cases.

The pattern now reads `ruff>=0.6`, `ruff>0.6`, `ruff===0.16.2`, `ruff[extra]==1.0` and `ruff == 0.6`, and it correctly declines `ruff-lsp>=1.0` and the bare word in prose. A workflow that appends `"ruff>0.6"` now fails the workflow case alone.

### MUST-FIX 3 — the shared list parser counted a string inside a comment

This is the hazard "an aggregate that counts a string counts the string", and the first form of the file walked into it.

`_dependency_block` of `tests/test_documentation_site.py` collects every double-quoted substring of the block. The `dev` extra carries comment lines **inside** its brackets, so a comment that quotes anything reads as a dependency entry. It already returns `not spec_validation` as an entry today, from the comment beside `build>=1.0`.

That was harmless only because the bogus entry does not name `ruff`. **A comment that quoted `"ruff==0.14.5"` would have failed a case on the wording of a comment**, which is a check a rewording defeats. `_dev_entries` now strips comment lines first, and the parser returns exactly the five real entries. A mutation that adds such a comment moves no case, at `6 passed`.

`_dependency_block` stays correct where it is used, and this PR does not change it: the `docs` list and the runtime list carry no comment inside the brackets, so no case of `tests/test_documentation_site.py` reads a comment. That is a latent defect of a shared helper with no live caller, so it is recorded here rather than repaired.

### MUST-FIX 4 — the records broke the writing standard, and one count was wrong

`.claude/rules/ste.md` caps a description sentence at 25 words. The first form of the `CHANGELOG.md` entry broke it in at least eight sentences, several by more than ten words, and the `docs/specs/spec.md` row inherited the same prose. Both are rewritten to short sentences, and the three-item and six-item enumerations are vertical lists in `CHANGELOG.md`.

**One sentence was also factually wrong**: it read "holds all four statements" and then enumerated six cases. It now states the count it holds. The `pyproject.toml` comment lost a 34-word sentence with four inline items in favour of a numbered list.

### REJECT-WITH-REASON — `EXACT_PIN` and a pre-release version

The review noted that `EXACT_PIN` rejected `ruff==0.16.2rc1` and `ruff==0.16.2+local`, and reported it at low severity because no pre-release pin is proposed. The pattern is widened anyway, because the cost was one character class and the failure mode was a misleading message: a pre-release pin **is** exact, and the case would have reported "states no exact version". The wider severity judgment is accepted; the finding is repaired rather than declined.

### Verification after the repairs

| Gate | Result |
|---|---|
| `pytest tests/ -m "not spec_validation"` | 2631 passed, 4 skipped, 8 xfailed, 114 subtests, 1825 deselected |
| Register invariant | 134 keys in `tests/foxio_deviations.json` against those 134 xfailed |
| Coverage | 94%, 4232 statements, 273 missed |
| `ruff format --check ja4plus/ tests/` | 190 files already formatted |
| `mypy --strict ja4plus/` | no issues in 31 source files |
| `docs/specs/spec.md` | 159 numbered rows, highest round 159, exactly one `TBD` row |

**Eleven mutations now prove the six cases**, up from eight. Each one restores from a snapshot of the implemented tree rather than from the index, under `PYTHONDONTWRITEBYTECODE=1` with every `__pycache__` cleared. The first harness restored from the index and contaminated six of its own readings, which is why the run was repeated.

`git diff --stat <base> -- ja4plus/ tests/foxio_vectors/ tests/foxio_deviations.json` is empty. No file under `ja4plus/` changes and no fingerprint moves.

Crank-Git:
Sub-merge gate. The conflict was the appended Changelog row of #380, which sub-merged first. Both `TBD` rows are kept in merge order.

**The gates ran on the merged state.**

| Gate | Reading |
|---|---|
| Unit | 2639 passed, 4 skipped, 8 xfailed, 114 subtests. 2625 base, plus 8 from #380 and 6 from #378 |
| Conformance | 1532 passed, 143 skipped, 134 xfailed against 134 register keys |
| `ruff check` | All checks passed, at the newly pinned 0.16.2 |
| `ruff format --check` | 190 files already formatted |

**The pin is held against the installed release, not only against the file.** `test_the_installed_lint_tool_holds_the_pinned_version` is the case that catches a drifted environment, and the eleven mutations include a decoy `ruff-lsp` entry carrying the justification comment — **a check that a decoy defeats is not a check**, and that direction was proved.

**Two findings were reported rather than repaired, and both are routed.**

- **#446**, the rest of the `dev` extra. It is `status:needs-feedback`, because the issue states that an exact `mypy` pin is not to be chosen by symmetry and the project manager agrees. **A floating formatter turns a green tree red, and a frozen type checker keeps a red tree green.** The first is noisy and safe.
- **#452**, the shared helper that reads every quoted substring of a dependency block, so a comment reads as an entry. **#378 stopped importing the helper rather than change a reader outside its scope, and that was right.**

Issue #393, opened 2026-08-09
Amend the documentation behaviour rule that names a doctest command collecting nothing


Crank-Git:
Found by #63 on 2026-08-09 while it made every code sample run in continuous integration. **#63 built the harness, met the requirements, and stopped at the behaviour rule rather than editing the specification inside a feature diff.** That was right, and this issue records the amendment.

## The divergence

**`docs/specs/features/08-documentation.md` names a mechanism that cannot work on this documentation.** Its Behaviour rules name:

```
pytest --doctest-glob="*.md" README.md docs/
```

**That command collects nothing and exits with the status 5.** Measured on `epic/19-documentation`:

```
grep -rc '^\s*>>>' README.md docs/*.md   ->  zero matches in every file
pytest --doctest-glob="*.md" README.md docs/   ->  "no tests ran in 0.05s"
```

**`--doctest-glob` collects an interactive session, written as `>>>` lines with expected output beneath. Every sample in this project is a fenced code block.** The two forms have nothing in common, so the named command is not a weak check — **it is no check at all, and a reader of the feature file would believe the samples are covered.**

## What the requirements actually ask, and whether they hold

**FR-documentation-4** — every code sample in the README runs in continuous integration.
**FR-documentation-5** — every code sample in `docs/` runs in continuous integration.

**Both hold today, through `tests/test_documentation_samples.py`.** It runs **45 samples**, and 14 more carry a marker naming why they cannot run, which a case enforces at four words or longer. `tests/test_examples.py` runs the five scripts of `examples/` under six command lines, which is FR-documentation-6.

**The requirements are met. The Behaviour rule names the wrong instrument.**

## The decision, already made

**The project manager ruled on 2026-08-09 that the criterion is amended and the samples are not rewritten.**

Rewriting 44 fenced samples into doctest sessions with exact expected output would make several unrunnable, would make the documentation worse to read, and would buy no coverage the harness does not already provide. **A token doctest added to make the command green was refused: it would prove nothing, and this project has recorded seventeen instances of a comparison that is never made reading as a comparison that passes.**

## What to change

Amend the Behaviour rule of `docs/specs/features/08-documentation.md` to name the instrument that exists:

- `pytest tests/test_documentation_samples.py` for FR-documentation-4 and FR-documentation-5,
- `pytest tests/test_examples.py` for FR-documentation-6.

[...]

Issue #387, opened 2026-08-09
Correct the three documents that state ten implemented methods


Crank-Git:
Found by #62 on 2026-08-09 while it corrected the README coverage claim. **#62 built to the correct number, changed none of the stale documents, and reported them.** Each sits outside its scope.

## The measurement

**Three documents state that this project implements ten of the twelve FoxIO methods. The number is eleven.**

| Document | What it states |
|---|---|
| #62, the issue body | ten of the twelve |
| `docs/specs/features/08-documentation.md:14-16` | ten of the twelve |
| `docs/specs/spec.html:126` | ten of the twelve |

**`docs/specs/spec.md:42` already reads that `ja4plus` implements eleven of them**, so the specification and the feature file disagree with each other today.

## The cause, which is worth stating precisely

**`JA4LFingerprinter` writes two methods.** It emits `JA4L-C=` at `ja4plus/fingerprinters/ja4l.py:482` and `JA4L-S=` at `ja4plus/fingerprinters/ja4l.py:446`.

**So ten fingerprinter classes carry eleven methods, and the ten is a count of classes read as a count of methods.** Twelve published, one declined, eleven implemented. JA4TScan is the decline.

**This is a counting error of the same family the register work keeps finding**: a number measured against one thing and then quoted about another.

## What to change

Correct the three documents to eleven, and **state the reason beside the number wherever a reader could recount it as ten** — a bare `11` invites the next reader to count the classes and correct it back.

**`docs/specs/spec.html` is generated. Read how it is built before you edit it.** If a generator produces it from `docs/specs/spec.md`, the repair is to regenerate rather than to hand-edit, and hand-editing would be undone by the next build.

**Prefer a case over a corrected number.** `tests/test_readme_contracts.py` landed with #62 and holds the README's count against the table's own rows rather than restating it. **The same shape reaches the feature file.** #345 and #258 set the precedent, and this project has now corrected a stale documented number seven times in one day.

## Acceptance criteria

- [ ] No document states that this project implements ten FoxIO methods.
- [ ] Every document that states a count states eleven, and the count agrees with `docs/specs/spec.md`.
- [ ] A case fails when a stated count and the implemented method set disagree, **or** the verdict states why a case is not worth its cost for a given document.
- [ ] The verdict lists every count of implemented methods found in `README.md`, `docs/`, `.claude/` and the root documents, and says which were changed and which were left and why.
- [ ] No file under `ja4plus/` changes and no fingerprint moves.

## Notes

[...]

Issue #380, opened 2026-08-09
Correct the stale register key count the rule document states


Crank-Git:
Found by #345 on 2026-08-09 while it replaced a stale count in `tests/foxio_deviations.py` with a case. **#345 measured this one, placed it out of its own scope, and filed nothing.** That was right, and this issue records the work.

## The measurement

**`.claude/rules/external-apis.md` states that the precedence exception reaches "6 rows of the 135 the register holds". The register holds 134.**

#272 removed `ssh2.pcapng/JA4L-S` from `tests/foxio_deviations.json` on 2026-08-09, because the repair to `ja4plus/fingerprinters/ja4l.py` made the comparison pass. **The reach of 6 is still correct. The key count alone is stale.**

## Why this is the same defect twice in one file pair

**#345 replaced exactly this shape of sentence** — a count in prose that nothing measures — **and the rule file holds another one.** `tests/foxio_deviations.py` now states its counts under `TestTheMarkerRuleCounts`, which reads them out of the register. **The rule file states its count in prose that no case reads.**

**A count that nothing measures is stale by the time someone reads it.** This project has now corrected that shape six times in one day, in five different files.

## What to change

**Prefer a case over a corrected number**, the same way #345 did.

**Read whether a case can reach this file at all.** `.claude/rules/external-apis.md` is a rule document rather than a module, so a case must read it as text. `tests/test_specification_changelog.py` reads `docs/specs/spec.md` as text and is the shape to follow, and #345's `TestTheMarkerRuleCounts` is the shape for reading the register.

**If a case is not worth its cost here, say so with a reason and correct the number.** That is an acceptable outcome, and it is a decision rather than a shortcut.

## What not to change

**`tests/test_foxio_deviations.py:660` comments "all 135 entries", and that is not this defect.** #345 read it as a record of what #341 measured at the time, and that reading is right. **A dated record of a past measurement is quoted, not rewritten.** Leave it.

**Read every other count of register entries in `docs/`, `.claude/` and the root documents before you decide.** A repair that fixes one sentence and leaves four is the defect wearing a smaller face. **Report what you find, even for the ones you leave.**

## Acceptance criteria

- [ ] No document states a register key count that disagrees with `tests/foxio_deviations.json`.
- [ ] Either a case fails when a stated count and the register disagree, or the verdict states why a case is not worth its cost here.
- [ ] The verdict lists every count of register entries found in `docs/`, `.claude/` and the root documents, and says which were changed and which were left and why.
[...]

PR #454, opened 2026-08-09
Name the instrument the documentation samples run under (#393)

Crank-Git:
Amends the documentation Behaviour rule for #393. Batch member of #445, targeting
`batch/445-documentation-repairs`. Draft, and every head commit carries `[skip ci]`.

## What diverged

`docs/specs/features/08-documentation.md` read:

```
- A code sample is tested by `pytest --doctest-glob`.
```

and its Acceptance criteria named `pytest --doctest-glob="*.md" README.md docs/`.

**That command collects nothing.** Measured on this branch's base commit `2c181fe`:

```
$ pytest --doctest-glob="*.md" README.md docs/
collected 0 items
============================ no tests ran in 0.08s =============================
$ echo $?
5
```

`--doctest-glob` collects an interactive session, written as `>>>` lines above the
expected output. Every code sample of this project is a fenced code block. The two forms
share nothing, so the rule was not a weak check — it was no check, and a reader of the
page believed the rule covered the samples.

## What the amendment claims

The requirements themselves hold, and they always did.

| Requirement | The command that runs it | Output |
|---|---|---|
| FR-documentation-4, FR-documentation-5 | `pytest tests/test_documentation_samples.py` | `22 tests collected`, and the harness runs 45 samples |
| FR-documentation-6 | `pytest tests/test_examples.py` | `9 passed in 4.50s`, over the five scripts of `examples/` |

The Behaviour rules now name those two commands, and they record **why** the fenced form
stays: `--doctest-glob` reads an interactive session, every sample here is a fenced block,
and a fenced block is what a reader copies. #393 asked for the reason in the rule, because
a later reader will otherwise propose the doctest form again.

**The defect stood in a second record.** The Documentation row of the Testing strategy
table of `docs/specs/spec.md` named the same command. It now names
`pytest tests/test_documentation_samples.py`. `CLAUDE.md` was read for the same fault and
its `## Commands` section holds no doctest command.

No sample is rewritten. A rewrite of 44 fenced samples into doctest sessions would make
several unrunnable, and it would buy no coverage the harness already provides. A token
doctest added to make the command green would prove nothing. The project manager ruled on
2026-08-09 that the criterion is amended.

## Lines touched, for the #387 sub-merge

#387 edits the same feature file. This change touches three places and nothing else:

- **Behaviour rules**, `docs/specs/features/08-documentation.md:113-120` — the old rule at
  the former line 113 becomes three rules.
- **Edge cases table**, line 158 — `The doctest run fails` becomes `The harness case fails`.
- **Acceptance criteria**, line 174 — the doctest command becomes
  `pytest tests/test_documentation_samples.py`.

`docs/specs/spec.md` takes one row change in the Testing strategy table and one new
Changelog row.

## The case, and how it fails

New file `tests/test_documentation_behaviour_rule.py`, 20 cases in four groups.

1. **Command.** Every backticked `pytest` command of the Behaviour rules runs under
   `--collect-only`, and each must collect at least one case. A separate case reads the
   Testing strategy row of `docs/specs/spec.md` the same way, so the second copy of the
   defect cannot return alone.
2. **Floor.** The Behaviour rules must name two such commands. An aggregate over an empty
   set passes, so a deleted rule would satisfy group 1 in silence.
3. **Discrimination.** The doctest command collects 0 cases and exits 5. The counter reads
   0 for a file that holds no case, and a positive count for one that does.
4. **Reader.** `stated_command` raises on an absent sentence, on a repeated sentence, and
   on the text the rules held before this round.
5. **The recorded reason takes two clauses**, and a case reads each one. A reader of the
   first clause alone would accept a sentence that reverses the claim.

**The count is parsed, never grepped.** A page holding the words `0 tests collected` would
match a substring search, so the reader matches the summary line and reads the exit status
beside it.

### Test-driven development

Six cases failed on the unamended records, before the amendment:

```
FAILED ...::test_the_behaviour_rules_name_a_sample_command_that_collects_cases
FAILED ...::test_the_behaviour_rules_name_a_script_command_that_collects_cases
FAILED ...::test_the_testing_strategy_table_names_a_command_that_collects_cases
FAILED ...::test_every_pytest_command_the_behaviour_rules_name_collects_cases
FAILED ...::test_the_behaviour_rules_name_at_least_one_pytest_command
FAILED ...::test_the_behaviour_rules_state_that_the_doctest_option_reads_a_session
6 failed, 12 passed in 1.41s
```

The floor case named the defect directly:
`AssertionError: the Behaviour rules name 1 pytest commands, and the floor is 2: ['pytest --doctest-glob']`

### Seven mutations, each written to disk, measured, then restored

`PYTHONDONTWRITEBYTECODE=1`, `__pycache__` cleared, and each file compared equal after its
restore.

| Mutation | Result |
|---|---|
| Delete the sample-command rule | 2 failed, the floor case among them |
| Name the doctest command in the rule sentence | 2 failed |
| Name `tests/test_no_such_harness.py` | 2 failed |
| Delete the recorded reason | 1 failed |
| Keep the first clause of the reason, drop the consequence | 1 failed, the new consequence case |
| Revert the `docs/specs/spec.md` row | 1 failed, alone |
| Rename the `## Behaviour rules` heading | 6 failed |
| Everything restored | 20 passed |

The removal direction and the wrong-text direction both fail, which is what #380 proved on
the precedence exception and the shape this batch follows.

## Verification, run locally because this member skips CI

- Unit: `2653 passed, 4 skipped, 8 xfailed, 114 subtests passed in 202.17s`. The base
  commit reports 2633 passed, and the 20 new cases are the difference.
- Conformance: `1532 passed, 143 skipped, 134 xfailed`, against the 134 keys of
  `tests/foxio_deviations.json`. The register invariant holds at 134 == 134.
- Coverage: **94%**, 273 missed of 4232 statements — identical to the base commit. Coverage
  does not fall.
- `ruff check ja4plus/ tests/` — `All checks passed!`
- `ruff format --check ja4plus/ tests/` — `190 files already formatted`
- `mypy --strict ja4plus/` — `Success: no issues found in 31 source files`
- **No file under `ja4plus/` changes and no fingerprint moves.**

## The sweep of every other feature file

#393 asked whether another feature file names a command that does not run. Every
command of all eleven `docs/specs/features/*.md` files was read and run where running it
was safe. The doctest command was the only one that collects nothing. Three findings are
left, and none is repaired here.

1. `08-documentation.md:122`, `:155`, `:171` name `mkdocs build --strict`. It is no
   `pytest` command, so the aggregate case passes over it and
   `tests/test_documentation_site.py` holds it. Round 140 measured it green from a clean
   environment. This round did not run it, because the worker environment holds no `docs`
   extra.
2. Two acceptance criteria state counts the repository no longer holds. `00-foundation.md:214`
   expects 75 files under `tests/foxio_vectors/` where `git ls-files` reports 90.
   `09-release.md:144` expects the string `0.6.0` in one file where `grep -rn` reports
   eight files. Both are the stale-count shape of #297, #345 and #387, in a column this
   issue does not own.
3. **One is a defect rather than a stale count, and #69 already owns it.**
   `09-release.md:156` requires the wheel to carry no file under `docs/`. `python -m build`
   on this base produces a wheel of 96 entries that holds 56 files under `docs/`. No
   duplicate issue is filed; the measurement is reported here and in the specification row.

## Premises re-measured

The brief and the issue are premises, and each was measured again.

- **45 runnable samples — confirmed.** `census()` reports
  `{'run': 43, 'raises': 2, 'skip': 95, 'output': 29}`.
- **14 marked samples — contradicted.** `disposition` reads **15** explicit `skip` markers
  today, and 17 markers with the two `raises` markers of `docs/api_reference.md`. This
  round names no page as the fifteenth, because the marker set moved across several rounds
  of this epic.
- **The base commit `2c181fe` — confirmed**, and `docs/specs/spec.md` held 159 numbered
  Changelog rows plus one `TBD` row belonging to #380.
- **The round convention — followed.** `CHANGELOG.md` reads `Round TBD` and the
  specification row reads `| TBD |`. No number is chosen.


Crank-Git:
## Self-review, two lenses

Reviewed by specialist lens in parallel: correctness, and the writing standard of
`.claude/rules/ste.md`. Findings, what I fixed, and what I rejected with the reason.

### Fixed in `52b39cc`

1. **The reason reader took one clause, so a reversal would have passed it.**
   `REFUSAL_REASON` matched `the `--doctest-glob` option collects an interactive session`
   alone. A sentence reading "The `--doctest-glob` option collects an interactive session,
   and this project uses it" would have satisfied the case while recording no refusal.
   `REFUSAL_CONSEQUENCE` now reads `that option collects nothing here` beside it, and two
   cases prove the pair. **A seventh mutation proves it discriminates**: keeping the first
   clause and dropping the consequence fails
   `test_the_behaviour_rules_state_that_the_doctest_option_collects_nothing_here` alone.

2. **A quoted timing is not stable evidence.** Both records read
   "It writes `no tests ran in 0.07s`". `no tests ran` is verbatim, and the duration is
   not: three runs of the same command on this host wrote `0.05s`, `0.07s` and `0.16s`.
   The records now quote the stable text alone. The `0.07s` came from the issue body's
   measurement rather than from a run of mine, which is the defect underneath it.

3. **`collect` inherited `PYTEST_ADDOPTS`.** A coverage option or an xdist option in the
   parent environment would reach the child run and change the summary line, so
   `COLLECTED_SUMMARY` would fail to parse for a reason unrelated to the rule under test.
   The child environment now drops that variable.

4. **A usage error exits 4 and writes its diagnostic to standard error.** The
   `AssertionError` of `collect` reported standard output alone, so such a failure named no
   reason. It now carries both streams.

5. **Prose.** Three passive sentences become active. Seven gerund subjects
   ("Deleting the sample rule fails...") become active sentences with a named subject. The
   opening sentence no longer rotates `instrument` against `command` for one referent, and
   the test class reads `TestTheCommandsTheRecordsName`.

6. **The case count reads 20** in both records, because items 1 and 5 added two cases.

### Rejected, with the reason

7. **"The command writes `no tests collected in 0.07s`, not `no tests ran`" — rejected.**
   The reviewer measured a different command. `--collect-only -q` writes
   `no tests collected`; the bare command the specification named writes `no tests ran`.
   Both measured here:

   ```
   $ pytest --doctest-glob="*.md" README.md docs/
   ============================ no tests ran in 0.16s =============================
   $ pytest --collect-only -q --doctest-glob="*.md" README.md docs/
   no tests collected in 0.13s
   ```

   The record quotes the command the specification named, so `no tests ran` is correct.
   The finding did surface a real defect next to it, which is item 2 above.

8. **"`instrument` is new vocabulary and needs no Terms entry" — rejected as stated.**
   `docs/specs/spec.md` already uses the word in three earlier Changelog rows, at lines
   748, 755 and 759, so it is established vocabulary rather than something this round
   introduced. The rotation inside one sentence was the real fault, and item 5 fixes that.

9. **The long-form sentence lengths of the two Changelog records — left as they read, and
   reported instead.** The review flags roughly a dozen sentences over the 25-word limit in
   the `CHANGELOG.md` entry and its `docs/specs/spec.md` row. **The finding is correct
   against the letter of the standard.** I split the sentences that split cleanly, and I
   left the rest, because **every one of the 159 numbered rows of that table reads the same
   way** — round 143 and round 156 carry sentences well past the limit, and the surrounding
   `### Fixed` entries do too. Rewriting my row to a different standard than its neighbours
   is a change to the house style of the whole record, which is wider than this issue and
   not a worker's call. **This is a question for the project manager**, and it is reported
   rather than decided here.

### Checked and clean

- No vacuous assertion. The one aggregate that iterates to `[]` is guarded by the separate
  floor case requiring two commands, and by the two sentence readers that raise on absence.
- `COLLECTED_SUMMARY` handles the singular and the plural form, and correctly does not
  match `no tests collected`; the exit-status short circuit carries that case.
- No recursion. A named command is collected and never executed under `--collect-only`.
- No residue, no network, no ordering dependence. `git status` is clean after two runs.
- No `--doctest-glob` reference survives as an instrument in either record, in `CLAUDE.md`,
  or in `README.md`. The only mentions are the deliberate warning and the historical record.

Crank-Git:
Sub-merge gate. Four members each appended one Changelog row, so `docs/specs/spec.md` conflicted and all four `TBD` rows are kept in merge order. **`docs/specs/features/08-documentation.md` merged without a conflict**, because #387 changed the method count near the head of the page and this branch changed the Behaviour rules, the Edge cases row and one Acceptance criterion. **The worker reported the exact lines it touched, which is why that merge could be trusted rather than inspected line by line.**

**The gates ran on the merged state.**

| Gate | Reading |
|---|---|
| Unit | 2810 passed, 4 skipped, 8 xfailed, 114 subtests. 2790 before this member, plus its 20 |
| Conformance | 1532 passed, 143 skipped, 134 xfailed against 134 register keys |
| The five gates | `ruff check` clean, 192 files formatted, `mypy --strict` no issue in 31 source files |

## The second record was not the file the brief named

**The brief told this worker to read `CLAUDE.md` for the same dead command. `CLAUDE.md` carries none.** The identical command stood in `docs/specs/spec.md:411`, the Documentation row of the Testing strategy table, so **the rule stood in two records and the project manager pointed at the wrong second one.** The worker swept for the fault instead of checking the named file and reporting a miss. A case now reads that row, and reverting it alone fails that case alone.

## A premise of the issue was stale and is re-measured

The issue reported 14 samples carrying a marker. **The measurement is 15 explicit `skip` markers, and 17 counting the two `raises` markers of `docs/api_reference.md`.** The 45-sample figure is confirmed. **The round names no page as the fifteenth**, because the marker set moved across several rounds and the attribution would be a guess. **Declining to attribute is the right call.**

## The floor case is the interesting one

`--collect-only` over every backticked `pytest` command would pass over an empty set, so a rule deleted entirely would satisfy the command cases in silence. **The floor case requires two such commands.** That is the "an aggregate over an empty set passes" hazard, closed by the worker rather than by the brief.

## Three findings reach no repair here, and all three are now filed

- **#455, `priority:high`** — the wheel carries 56 files under `docs/` where `09-release.md:156` requires none. **The worker named #69 as the owner and filed nothing. The project manager overrules that: #69 is held behind a user gate, and an issue that cannot be scheduled cannot own a repair.**
- **#456** — the two stale counts of `00-foundation.md:214` and `09-release.md:144`.
- **#457** — the sentence-length question for the Changelog record, `status:needs-feedback` for the user.

Issue #459, opened 2026-08-09
Fail a batch merge when the pull-request event created no run


Crank-Git:
## What was measured, twice

**A pull request from an integration branch into `dev` created no workflow run.** Measured on
2026-08-09 on two consecutive batch pull requests.

| Pull request | Head | Runs created by the pull-request event |
|---|---|---|
| #444, `batch/437-test-hygiene` into `dev` | `2a589ca` | **none**, re-checked at 45 seconds and again at 105 seconds |
| #458, `batch/445-documentation-repairs` into `dev` | `b593de5` | **none**, re-checked at 25 seconds and again at 80 seconds |

**Neither head commit carries `[skip ci]`**, so the documented skip path is not the cause. Each
assignment commit was written deliberately without the marker for this reason.

**Two controls hold.** A `workflow_dispatch` on the same ref starts every job and they pass.
**A push to `dev` after the merge created its runs normally**, both times. So the provider is
not refusing the repository or the workflow; it is the pull-request event on these branches
that produced nothing.

**The epic pull request of the previous session did create pull-request runs**, so this is a
change in behaviour rather than a standing condition.

## Why this is worth a high priority

**The batch pull request is the only gate that runs the full matrix on a member's code.** Every
member pull request carries `[skip ci]` by design, so a member's cases may never have run on
Linux until the batch gate.

**The failure mode is a batch merged in the belief that continuous integration ran.** A green
`gh pr checks` reading of "no checks reported" is not a failure; it reads like an absence. **A
project manager that treats an absent run as a passed run merges an ungated batch**, and
nothing on the pull request records that the gate never ran.

**The current mitigation is a person or a project manager remembering to look.** That is the
weakest kind of check, and this project has ruled repeatedly that a check nobody runs is no
check.

## What to build

**A condition that fails when the batch gate did not run**, rather than a habit of looking.
Some candidate readings, and the issue is free to find a better one.

- A required status check on `dev` through branch protection, so a pull request cannot merge
  while its run is absent. **Read whether this repository's plan carries required checks
  before you propose it.**
- A `merge_group` or `pull_request_target` trigger that the provider treats differently.
- A workflow that runs on a push to `epic/**` and `batch/**` and states plainly that it is the
  batch gate. **This trades the whole point of `[skip ci]` on member branches**, so it must
  exclude `issue/**` and it must be measured for cost.
- A pre-merge reading, in the loop rather than in the provider: the project manager refuses a
[...]

PR #458, opened 2026-08-09
Batch #445: documentation repairs — rounds 160 to 163

Crank-Git:
Batch #445, the documentation repairs. The user named this group after test hygiene on
2026-08-09. Each member was a document that stated something the repository contradicted, so
a reader who followed it reached the wrong answer.

Closes #380
Closes #378
Closes #387
Closes #393

## The four members, in sub-merge order

| Round | Issue | What it repaired |
|---|---|---|
| 160 | #380 | The precedence exception rule stated 135 register keys where the register holds 134. A case now measures both numbers of that sentence against the register |
| 161 | #378 | `ruff` floated on `>=0.6`, so the lint gate could turn red with no commit. The pin is `0.16.2` and a case reads the installed release |
| 162 | #387 | Six documents stated a count of ten implemented methods. The count is eleven, and 151 cases derive it from `ja4plus` rather than restate it |
| 163 | #393 | A Behaviour rule named `pytest --doctest-glob`, which collects nothing here. The rules now name the harness that runs, and 20 cases hold every named command to a real collection |

## Every member replaced a number with a measurement

**That is the shape this batch is really about.** A count in prose is stale by the time
someone reads it, and this repository had corrected that shape seven times in one day before
this batch. So no member simply corrected a number:

- #380's case reads the rule page as text and the register as data, and **it fails when the
  sentence is deleted** as well as when it is wrong.
- #387's cases derive the count from `__all__` and from the `JA4L-S=` emitter, so **removing
  that emitter fails all five documents that state eleven.**
- #393's cases run every backticked `pytest` command of the Behaviour rules under
  `--collect-only`, and **a floor case requires two of them**, because an aggregate over an
  empty set passes and a deleted rule would otherwise satisfy the check in silence.
- #378's cases read the installed `ruff` release, not only the pin, so **a drifted
  environment fails rather than passing quietly.**

## Four project manager premises were contradicted, and every worker reported rather than worked around

**This is the instruction earning its place, and it is recorded so the next session keeps it.**

1. **The round numbers.** Every brief named one. A worker writes `TBD` and the project
   manager assigns at the gate. `tests/test_changelog_round_agreement.py:3-5` states it.
2. **`docs/specs/spec.html` is not generated.** A brief told #387 to regenerate it. No
   generator exists, the page states a person maintains it, and round 70 records the same
   reading. The correction is a hand edit.
3. **The second record of the dead command is not `CLAUDE.md`.** A brief told #393 to check
[...]

Issue #388, opened 2026-08-09
Rule on the FoxIO licence list the README publishes


Crank-Git:
Found by #62 on 2026-08-09 while it stated the FoxIO attribution the README needed. **#62 measured the disagreement and changed nothing, correctly: a licence statement is the user's decision and not a worker's.**

## The measurement

**The README's FoxIO License 1.1 list names seven methods. FoxIO's own list names ten.**

| Source | Methods named |
|---|---|
| `README.md` | JA4S, JA4H, JA4T, JA4TS, JA4L, JA4X, JA4SSH |
| FoxIO `License FAQ.md:5`, at the pinned commit `27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8` | the same seven, **plus JA4LS, JA4D and JA4D6** |

**Three methods that FoxIO places under the licence are absent from the list this project publishes.**

## Why this is the user's decision and not the project manager's

**This is a statement about licence terms, and it is published.** A worker or a project manager who edits it is deciding, on the user's behalf, what this project asserts about someone else's licence. **`.claude/rules/conformance.md` governs what the software must output; it does not govern what the project asserts about a licence.**

**The project manager will not edit it and has not.** The finding is recorded here so it is not lost, and so a later reader does not mistake the shorter list for a considered position.

## What the user decides

1. **Does the README list follow FoxIO's list?** If the shorter list is an oversight, the repair is to name all ten and cite the FoxIO document that carries them.
2. **Or is the shorter list deliberate** — for example, naming only the methods this project implements, which would explain the absence of JA4TScan but not the absence of JA4D and JA4D6, both of which this project does implement.
3. **Is legal review wanted before any edit?** A licence statement is the one class of prose in this repository where being wrong is not merely untidy.

## What is not in question

**Nothing about the software changes.** No fingerprint moves, no output changes, and `__all__` is untouched. **This is a paragraph of prose in `README.md`.**

The pinned FoxIO checkout is the evidence and it is on disk in the session scratchpad as `foxio-ja4`, at `27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8`. **Re-read `License FAQ.md` before acting** — this issue records what it said on 2026-08-09.

Labelled `status:needs-feedback`. It gates nothing.

Crank-Git:
## The user ruled on 2026-08-09

**Option 1. The README list follows FoxIO's list. It names all ten methods.**

The short list was an oversight and not a considered position.

## What to build

[...]

PR #463, opened 2026-08-10
Fail a batch merge when the pull-request event created no run

Crank-Git:
## The cause, measured

**A skip keyword anywhere in a commit message creates no run for that commit.** GitHub reads
five keywords, matches the subject and the body, and reads no intent.

The head commit of pull request #444 and the head commit of pull request #458 each held this
line:

```
This commit carries no `[skip ci]`, so it is the head that starts the full run
on the batch pull request.
```

**The sentence that states the absence of the keyword is the keyword.** Neither pull request
created a run from the pull-request event.

**The correlation is perfect across four heads.**

| Head | Pull request | Holds a keyword | Runs from the pull-request event |
|---|---|---|---|
| `2a589ca` | #444, `batch/437-test-hygiene` | yes | none |
| `b593de5` | #458, `batch/445-documentation-repairs` | yes | none |
| `ec697c0e` | Epic 11 | no | Tests and Documentation, both green |
| `b21604ce` | Epic 8, #405 | no | Tests and Documentation, both green |

Verified against
https://docs.github.com/en/actions/how-tos/manage-workflow-runs/skip-workflow-runs
(retrieved 2026-08-09), which states that a keyword anywhere in the head commit message of a
pull request creates no run for a push event or a pull-request event.

## One premise of the brief re-measured false

The issue body reads **Neither head commit carries `[skip ci]`, so the documented skip path is
not the cause**. Both head commit messages hold the literal keyword. That premise is what
ruled the documented skip path out, and it is the one thing between the observation and the
cause.

The other four observations re-measured **true**. `gh run list` reports `workflow_dispatch`
runs alone on both batch branches, both green, and it reports the push runs on `dev` after
each merge.

## What this ships

**A condition that fails, not a habit of looking.** New file `tests/batch_gate.py`:

```bash
python -m tests.batch_gate --pr <number>
```

It exits 0 on one state alone: every required workflow holds a terminal successful run at the
head commit, and every other run of that commit concluded `success`. It exits 1 on an absent
run, on a run that has not finished, on any other conclusion, and on a failed read of the
provider. Where a required run is absent and the head commit message holds a keyword, the
failure names the keyword and this issue.

`.github/workflows/test.yml` is the one required workflow, because it accepts every pull
request into `dev` and it filters no path. `.github/workflows/docs-build.yml` filters four
paths, so an absent run of it proves nothing, and a red run of it still refuses the merge.

**The procedure reaches the loop.** New file `.claude/rules/batch-gate.md`. `CLAUDE.md` names
it under `## Branch model`. `.issue-flow.json` replaces the note that read "A batch pull
request into dev starts a run, so provider continuous integration is the batch gate" — that
sentence is the belief that turns an absence into a pass. The header comment of
`.github/workflows/test.yml` named a throttled webhook path that nobody measured, and it now
records the keyword and cites this issue, held by a case.

The run reader takes `GET /repos/{owner}/{repo}/actions/runs`, verified against
https://docs.github.com/en/rest/actions/workflow-runs (API version 2022-11-28, retrieved
2026-08-09) for the `head_sha` parameter and the `path`, `status`, `conclusion`, `head_sha`,
`event` and `created_at` fields.

## The strongest shape needs the user

**This repository holds no required status check.** A read of 2026-08-09:

| Read | Result |
|---|---|
| `gh api repos/Crank-Git/ja4plus/branches/dev/protection` | `404`, `Branch not protected` |
| `gh api repos/Crank-Git/ja4plus/rulesets` | `[]` |

The token carries the `repo` scope and both reads returned a determinate answer, so this is a
measurement and no stated limit. **A change to branch protection changes the repository
configuration, so the user makes it and no agent makes it.**
`.claude/rules/batch-gate.md` names the four steps and the eleven check names. A check name is not a job name: the `test` job runs a matrix, so a required check named `test` would match nothing.

## How this was tested

`tests/test_batch_gate.py` holds **48 cases**, and every one failed before the module existed.
They cover an absent run, a run of another commit, a queued run, a run in progress, four
terminal conclusions that are not `success`, a `skipped` conclusion, a re-run that passed
after a failure, a failure after a success, a red run of the unrequired workflow, an
abbreviated commit identifier, and the head commit message of both measured pull requests
against both controls. **The empty required set carries a case of its own, because an
aggregate over an empty set passes.**

**The condition was proved against the provider in both directions.**

- `python -m tests.batch_gate --pr 444`, `--pr 458` and `--pr 405` each exit 0 today, because
  the manual run of each ref holds a terminal successful run at that head. That is the
  accepted procedure and no defect.
- Against the head of this branch, which carries no run, it exits 1 and names the keyword.

**No provider run was started for this issue.** Every provider call was a read.

### The self-review found three defects and each is closed

See the review comment below. The sharpest one: `_newest` used `max`, which returns the first
of several equal keys, so **a green run listed before a red one at the same `created_at` made
the gate report a pass**. Measured on the unrepaired reader, one input, two orders:
`green,red: PASSED` and `red,green: REFUSED`. The verdict rested on the order of the response.
The tie now breaks toward a refusal.

### The local gate

| Gate | Result |
|---|---|
| `pytest tests/ -m "not spec_validation"` | 2862 passed, 4 skipped, 8 xfailed, 114 subtests |
| Coverage | 94%, 273 missed of 4232 — identical to the base commit `d94d8c1` |
| Register invariant | 134 keys against 134 xfailed |
| `ruff format --check ja4plus/ tests/` | 194 files already formatted |

The base commit `d94d8c1` reported 2810 passed. The rise to 2862 is 48 new cases, plus 2 cases
of `test_documented_method_count.py` — it parametrizes over Markdown pages and now covers
`.claude/rules/batch-gate.md` — plus 2 cases that arrived with #388 in the merge below.

### The merge with #388

`4c3691d` takes `origin/batch/460-release-gates-and-documentation` into this branch.
`CHANGELOG.md` and `docs/specs/spec.md` each held two new records at one place. **Both sides
read `TBD`, so the resolution keeps both and invents no number.** The entry of #388 sub-merged
first, so it opens the pair in both files. The merge commit carries the skip marker, because a
merge commit does not inherit it. Every gate above ran after the merge.

## Records

`CHANGELOG.md` and `docs/specs/spec.md` each gain a record that reads `TBD`. The project
manager assigns the number at the batch gate.

## Scope

Refs #459


Crank-Git:
## Self-review — three defects found and closed in `9dae880`

### 1. The gate's own verdict rested on the order of the response (real defect)

`tests/batch_gate.py`, `_newest`. One workflow holds **two runs at one commit** where a push
event and a pull-request event both reach it, and two runs can carry one `created_at`. `max`
returns the first of several equal keys.

Measured on the unrepaired reader, one input, two orders:

```
== tie-break, unhardened (the defect) ==
  green,red: PASSED
  red,green: REFUSED
```

**A green run listed before a red one made the gate report a pass.** That is the fault this
project records eighteen times — a comparison that never runs reads as a comparison that
passes — and it reached the gate built to stop it.

`_newest` now breaks the tie toward a refusal. Both orders now read `REFUSED`.
`test_two_runs_of_one_time_break_the_tie_toward_a_refusal` reads both orders, and
`test_an_older_red_run_moves_no_verdict` holds the rule to one creation time so it does not
become a rule about every run.

### 2. The branch-protection list named a check that does not exist (real defect)

`.claude/rules/batch-gate.md` listed the six `jobs:` keys of `.github/workflows/test.yml` as
the required check names. **The `test` job runs a matrix, so the provider publishes one check
per combination.** A read of commit `b593de5`:

```
lint
test (ubuntu-latest, 3.9)   test (ubuntu-latest, 3.10)  test (ubuntu-latest, 3.11)
test (ubuntu-latest, 3.12)  test (ubuntu-latest, 3.13)  test (macos-latest, 3.12)
fuzz  samples  installed-wheel  conformance  build
```

A required check named `test` matches nothing, so the guidance would have produced a
protection rule that guards less than it appears to. The rule now lists the eleven names, and
it bars `build`: that check belongs to `.github/workflows/docs-build.yml`, which filters four
paths, so a required `build` would block every batch that touches none of them.

`test_the_rule_names_every_job_of_the_required_workflow` reads the `jobs:` keys against the
rule file. Removing `installed-wheel` or `conformance` from the file names it in the failure.

### 3. Two cases tested a region they never found (real defect)

`test_the_required_workflow_reads_no_path_filter` and
`test_the_unrequired_workflow_reads_a_path_filter` sliced the file on the text
`pull_request:`. A slice that landed on a comment would have satisfied
`assert "paths:" not in trigger` **on a claim nobody read** — a case that cannot fail.

`_trigger_block` now takes the `on:` block by indent and requires exactly one line reading
`on:`. `_pull_request_trigger` raises where the region it took holds no `branches:` key.
Against `.github/workflows/publish.yml` it raises `accepts no pull-request event` rather than
returning a region. `test_the_workflow_comment_states_the_measured_cause` now splits on
`"\non:\n"` and asserts the split moved.

### 4. The upper form of a commit identifier read an absent run (weakness)

`COMMIT_IDENTIFIER` is lowercase-only and the comparison was case-sensitive, so a caller that
passed the upper form got "no run exists" for a commit that holds a green one. That sends a
reader after a cause that is not there. The comparison now folds the case, and a case reads
it. The direction was always safe — it refused rather than passed — so this is a weakness and
no defect.

### Findings I declined

**`assert reasons != []` in several cases.** Each of those cases also asserts on the content
where the content is the point (the absent-run case asserts the workflow path and the head
commit appear; the conclusion cases assert the conclusion appears). The remaining bare
assertions are on inputs that can fail for one reason only, and a second assertion there would
restate the implementation rather than the behaviour.

**No case starts a provider run.** Every provider call in this issue was a read.

Crank-Git:
Sub-merge gate. Base clean, no conflict, and no provider run exists on the branch. Change set is `.claude/rules/batch-gate.md`, `.github/workflows/test.yml`, `.issue-flow.json`, `CHANGELOG.md`, `CLAUDE.md`, `docs/specs/spec.md`, `tests/batch_gate.py`, `tests/test_batch_gate.py`. No file under `ja4plus/`.

## The cause is the project manager's own prose, and the project manager confirms it

**The premise of this issue was false, and it was the only thing standing between the observation and the cause.** The issue body reads *"Neither head commit carries a skip keyword, so the documented skip path is not the cause."* **Both heads carry it.**

Confirmed independently at the gate:

```
2a589ca:13  This commit carries no `[skip ci]`, so it is the head that starts the full run
b593de5:19  This commit carries no `[skip ci]`, so it is the head that starts the full run
```

**The sentence that denies the keyword contains the keyword.** The provider reads it anywhere in the head commit message and reads no intent. The two controls `ec697c0` and `b21604c` hold zero occurrences and each created its runs, so the correlation is four for four.

**The project manager wrote both of those sentences, believing it was making the run fire.** This is the project's own recorded fault turned on its author: **an aggregate that counts a string counts the string**, and prose about a marker is the marker. The throttling hypothesis was wrong, #85 was not the relative, and no provider incident occurred.

**The two batches did reach a full green gate**, by manual dispatch, with the head SHA read against the pull-request head. **So no batch merged ungated. The reasoning was wrong and the outcome was not**, and both halves are recorded.

## What the worker built, and the fault it found inside its own gate

**`python -m tests.batch_gate --pr <number>` exits 1 on an absent run**, measured live: `no run of .github/workflies/test.yml exists for the head commit ..., and an absent run is no passed run`, with the keyword diagnosis beside it. It exits 0 on #444, #458 and #405, whose manual runs are terminal and green.

**The self-review found the gate reporting a pass on a red run.** `_newest` used `max`, so a green run listed before a red one at the same `created_at` made the gate pass. **That is this project's signature fault appearing inside the instrument built to stop it**, and the repair carries cases that read both orders.

## The record the project manager was keeping is corrected here

**`.issue-flow.json` carried the sentence "A batch pull request into dev starts a run, so provider continuous integration is the batch gate."** That sentence turns an absence into a pass, and the worker replaced it. **It is the project manager's own file and the correction is accepted.**

## What still needs the user

**`dev` carries no branch protection and no rulesets**, read determinately rather than blocked by scope: `404 Branch not protected` and `[]`. **So the strongest shape — a required status check that refuses the merge at the provider — is not in place, and only the user can put it there.** `.claude/rules/batch-gate.md` names the four steps and the eleven check names. **A required check named `test` would match nothing, because that job runs a matrix.** The loop-level condition ships now and does not wait for it.

Issue #431, opened 2026-08-09
Skip a type annotation when the sweep builds a mutation


Crank-Git:
## What the sweep measures today

`tests/mutation_sweep.py` builds one mutation for each `BinOp` node whose operator sits in
`BINOP_SWAP`, and `ast.BitOr` is one of them. A type annotation written `str | None` holds
a `BitOr`, so the sweep changes it to `str & None`.

**Every module of `ja4plus/` carries `from __future__ import annotations`.** Python then
holds each annotation as a string and evaluates none of them, so the changed annotation
reaches no run-time value. **No case can fail for such a mutation.**

## The measurement

#413 swept `ja4plus/fingerprinters/` whole. **94 of the 675 surviving mutations sit inside
a type annotation, and all 94 survived.** That is 14 percent of the survivors of that
module group.

| Module | Annotation mutations |
|---|---|
| `ja4plus/fingerprinters/ja4x.py` | 18 |
| `ja4plus/fingerprinters/ja4l.py` | 13 |
| `ja4plus/fingerprinters/ja4s.py` | 13 |
| `ja4plus/fingerprinters/ja4.py` | 12 |
| `ja4plus/fingerprinters/ja4h.py` | 9 |
| `ja4plus/fingerprinters/ja4ts.py` | 8 |
| `ja4plus/fingerprinters/ja4ssh.py` | 6 |
| `ja4plus/fingerprinters/ja4t.py` | 5 |
| `ja4plus/fingerprinters/base.py` | 4 |
| `ja4plus/fingerprinters/ja4d.py` | 3 |
| `ja4plus/fingerprinters/ja4d6.py` | 3 |

`docs/mutation_reports/413-fingerprinters.json` holds every one of them.

## Why it matters

A survivor asks a reader for a settlement. A survivor that cannot fail spends that reading
for nothing, and it raises the survivor count that a later reader compares against.

`tests/mutation_sweep.py` already skips a docstring and a logger argument for the same
reason: `_docstring_nodes` and `_logging_argument_nodes` remove the nodes whose mutation
reaches no assertion. An annotation is the third such node set.

## Acceptance criteria

- [ ] `tests/mutation_sweep.py` holds a reader that returns the id of every node inside a
      type annotation, covering an argument annotation, a return annotation and an
      `AnnAssign` annotation.
- [ ] `generate_mutations` builds no mutation for a node that reader names.
- [ ] A case feeds a module that holds `def f(a: str | None) -> int | None:` and
[...]

PR #471, opened 2026-08-10
Skip a type annotation when the sweep builds a mutation

Crank-Git:
Builds #431.

## What changed

`tests/mutation_sweep.py` builds one mutation for each `BinOp` node whose operator sits in
`BINOP_SWAP`, and `ast.BitOr` is one of them. A type annotation written `str | None` holds
a `BitOr`, so the sweep changed it to `str & None`.

**Every module of `ja4plus/` carries `from __future__ import annotations`.** Python then
holds each annotation as a string and evaluates none of them, so a changed annotation
reaches no run-time value. No case can fail for such a mutation.

New reader `_annotation_nodes` returns the id of every node inside a type annotation. It
covers an argument annotation, a return annotation and an `AnnAssign` annotation, in the
shape `_docstring_nodes` and `_logging_argument_nodes` already set.

## Why the wiring stands apart from `skip`

`generate_mutations` reads the union `skip` in the constant branch alone. An annotation
holds a `BitOr` that the binop branch reads, so the annotation set is read for every node
kind, at the top of the walk. The change removes no mutation of a logger argument and no
mutation of a docstring.

## The count

`--dry-run --max-per-module 0` over `ja4plus/fingerprinters/*.py`:

| Read | Mutations |
|---|---|
| Before | 1569 |
| After | 1474 |

**The reader removes 95 mutations and not the 94 the issue states.**
`ja4plus/fingerprinters/base.py:43` holds `-> Literal[False]:`, and that constant reaches
no run-time value either. The per-module difference matches the table of #431 on every
other module. `TestTheConstantOfAnAnnotation` holds the extra one.

## How it was tested

New file `tests/test_mutation_sweep_annotation_skip.py` holds nine cases, and every one was
proved in both directions.

- The five reader cases failed with
  `AttributeError: module 'tests.mutation_sweep' has no attribute '_annotation_nodes'`
  before the reader existed.
- The three sweep cases failed against a control that reads an empty set in place of the
  reader: `assert 4 not in {4, 6, 7}` and `assert 4 == 1`.
- `test_the_sweep_builds_one_mutation_outside_an_annotation` proves that a `BitOr` outside
  every annotation still builds its mutation.
- `test_the_module_holds_four_bit_or_nodes` is the floor, so a reader that names nothing
  fails rather than passes an aggregate over an empty set.

Every run set `PYTHONDONTWRITEBYTECODE=1` and cleared `__pycache__` first.

## The five gates

| Gate | Result |
|---|---|
| `pytest tests/ -m "not spec_validation"` | 2874 passed, 4 skipped, 8 xfailed, 114 subtests passed |
| `ruff format --check ja4plus/ tests/` | 195 files already formatted |

Coverage reads 94 percent of `ja4plus`, measured with `--cov=ja4plus`. No file under
`ja4plus/` changes, so no fingerprint moves.

## Documentation

- `FR-pre-release-validation-34` and `FR-pre-release-validation-34a` state the rule.
- `CHANGELOG.md` holds the round entry, reading `Round TBD`.
- `docs/specs/spec.md` holds the matching `TBD` row of the changelog table.
- `docs/mutation_sweep.md` is generated, so this change edits none of it.

## The interface the reader depends on

The three field names come from the machine-readable interface of the `ast` module, read
on the interpreter of this worktree.

```
$ python -c "import ast; print(ast.arg._fields, ast.FunctionDef._fields, ast.AnnAssign._fields)"
('arg', 'annotation', 'type_comment') ('name', 'args', 'body', 'decorator_list', 'returns', 'type_comment', 'type_params') ('target', 'annotation', 'value', 'simple')
```

Verified against https://docs.python.org/3/library/ast.html (retrieved 2026-08-09) for
`ast.arg.annotation`, `ast.FunctionDef.returns`, `ast.AsyncFunctionDef.returns` and
`ast.AnnAssign.annotation`.


Crank-Git:
**Self-review.** Two lenses read the diff: correctness, and the writing standard of
`.claude/rules/ste.md`. Commit `9455bdc` holds the repairs.

## Accepted

1. **The premise was overstated.** The issue body and the first commit read
   `Every module of ja4plus/ carries from __future__ import annotations`. A read of the
   tree measures 28 of 31. `ja4plus/fingerprinters/__init__.py`, `ja4plus/utils/__init__.py`
   and `ja4plus/utils/loopback.py` carry none. Those three hold one annotation node between
   them, the `-> None` of `ja4plus/utils/loopback.py`, and no operator sits in it, so the
   reader removes no mutation there today. The prose now states the measured count in all
   four places. **The reader stays sound where a module evaluates its annotations**: a
   changed annotation raises at import, and the sweep records that mutation `unusable`
   rather than a survivor.
2. **Three sentences joined two facts.** The docstring of `_annotation_nodes`, the comment
   above the wiring and `FR-pre-release-validation-34a` each hold one fact per sentence now.
3. **One metaphor left the changelog entry.** `each one asks a reader for a settlement it
   cannot earn` states no mechanism. It now reads `A reader reads each one and finds no case
   that it can ever fail.`

## Checked and clean

- **Every annotation position is covered.** A probe module with an `async def`, a
  positional-only argument, a keyword-only argument, `*args`, `**kwargs`, a class-level
  `AnnAssign` and a nested function builds no `binop` mutation. `ast.arg` is one node type
  for every argument kind, so the reader needs no branch for each one.
- **The wiring over-skips nothing.** The reader reads `.annotation` and `.returns` alone,
  so the value of an `AnnAssign` and every argument default keep their mutations. A probe
  proves it: `k: bytes | None = b""` keeps the `bytes` mutation of `b""`.
- **The `id()` key is safe.** `generate_mutations` holds `tree` for the whole call, and
  every annotation node is reachable from it, so no node is collected and no id is reused.
  `_docstring_nodes` and `_logging_argument_nodes` already rest on that.
- **Python 3.9 accepts it.** `ast.AsyncFunctionDef`, `ast.AnnAssign`, `ast.arg.annotation`,
  `Optional` and `Set` all predate 3.9.

## Declined

- **A `Terms` row for the word `reader`.** `docs/specs/spec.md` already uses the word in the
  ordinary sense, and `CHANGELOG.md` already names a helper function a reader on entries
  that predate this issue. A row that fixes the word to one of the two meanings is a
  specification change, and this issue carries none.
- **The paragraph length of the changelog entry.** Rule 3 of `.claude/rules/ste.md` caps a
  topic paragraph at six sentences. Every entry of `CHANGELOG.md` runs longer, so this entry
  matches the file it lands in. A change of that convention belongs to its own issue.

PR #472, opened 2026-08-10
Give a module-body mutation a reader in the cover rule

Crank-Git:
## What this changes

Step 2 of the cover procedure in `.claude/rules/conformance.md` subtracts the lines the
import runs. Every mutation of a module body therefore lost its reader, and the cost rule
of step 3 then held the cheapest test file. New step 4 names a reader for that line class
by the value it holds, and never by the cost.

New file `tests/mutation_cover.py` applies the step. It reads the names a module body
binds and the identifier strings those statements build, and it names the test file whose
own source holds the most of those tokens.

```
$ python -m tests.mutation_cover ja4plus/__init__.py
28 tokens in the module body.
  26  tests/test_public_interface.py
  15  tests/test_edge_cases.py
  12  tests/test_comprehensive.py
  10  tests/test_no_retained_packets.py
   9  tests/test_bounded_state_table_adoption.py
```

## Why

#414 measured the outcome on `ja4plus/__init__.py`. The cover the cost rule builds is
`tests/test_parity.py`, and that sweep reads 1 killed and 31 survived. The defect follows
from the construction and not from an accident of cost, because the subtraction removes
the whole line class at once.

## How this is tested

The rule change is measured and not asserted. Both sweeps ran on commit `00a0c42`, over
all 32 mutations of `ja4plus/__init__.py`, with `--max-per-module 0`.

| Cover | Killed | Survived | Survivors on an entry of `__all__` | Seconds |
|---|---|---|---|---|
| `tests/test_parity.py`, from the cost rule | 1 | 31 | 25 | 43.2 |
| `tests/test_public_interface.py`, the reader alone | 25 | 7 | 0 | 39.9 |
| Both files, the repaired cover | 26 | 6 | 0 | 48.0 |

`tests/test_mutation_cover.py` holds 18 cases. Seven read the token set of one module body,
five read the ranked readers, and six read the rule text.

- The reader case fails when every line naming `tests/mutation_cover.py` leaves the rule
  text. The cost case fails when every line naming a second count leaves it. A reworded
  step passes both, so a rewording defeats neither prover.
- One case reads the floor of the ranked list, because an aggregate over an empty set
  passes. Two more read a module body that binds nothing and a module body that no test
  file reads.
- One case reads a module whose function body binds a name, and one reads a module whose
  class body binds a name. The token set drops both names.
- One case reads a tuple target and a starred target, and one reads an assignment inside an
  `if` block and inside a `try` block. The import runs both blocks, so both names count.

The first run of the four rule-text cases failed against the unrepaired rule file, with
`assert False` on `states_the_body_reader_rule` and on `states_the_reader_cost`.

## The limit the rule now states

The reader kills a mutation of the value it reads, and it kills no other mutation of the
module body. `__version__`, `__author__` and `__license__` of `ja4plus/__init__.py` survive
it, because `tests/test_public_interface.py` names all three and compares none against a
value.

## Local gate

- `pytest tests/ -m "not spec_validation"`: 2883 passed, 4 skipped, 8 xfailed, 114
  subtests passed.
- `pytest tests/ -m spec_validation`: 1532 passed, 143 skipped, 134 xfailed.
- `ruff check ja4plus/ tests/`: All checks passed!
- `ruff format --check ja4plus/ tests/`: 196 files already formatted.
- `mypy --strict ja4plus/`: Success: no issues found in 31 source files.
- Coverage: TOTAL 94%, 273 missed of 4232 statements. No file under `ja4plus/` changes, so
  the number holds.

## Scope

`tests/mutation_sweep.py` is unchanged, because #431 holds the node-selection path of that
file. `docs/mutation_settlements/414-interface.json` keeps the numbers #414 measured,
because that record states the sweeps that ran.

Part of #406. Found by #414. Refs #433.


Crank-Git:
## Self-review, two lenses

I reviewed this diff for correctness and against `.claude/rules/ste.md`. Six findings
reached a repair and three did not.

### Repaired

1. **`body_tokens` dropped the names of a tuple target and a starred target.** `A, B = "x",
   "y"` bound one `ast.Tuple` target, and the reader tested for `ast.Name` alone. No module
   of `ja4plus/` writes that form today, so no measured number moves. `_target_names` now
   reads a tuple, a list and a starred target, and
   `test_the_token_set_holds_the_names_of_a_tuple_target` reads it.
2. **`body_tokens` read no assignment inside an `if` block or a `try` block.** The import
   runs both bodies, so an assignment there is a module-body assignment. `_body_statements`
   now descends into `If`, `Try`, `With`, `For` and `While`, and never into a function body
   or a class body. Two cases read both directions.
3. **The cost prover read one line, and the rule file wraps at 90 columns.** A rewrap of my
   own paragraph broke it, which is the failure a reader would meet later.
   `states_the_reader_cost` now reads the paragraph, and
   `test_the_cost_prover_reads_a_sentence_that_wraps_across_two_lines` holds it.
4. **`reader` named two things.** `docs/specs/spec.md` already uses `reader` for the person
   who reads a page. The `## Terms` table now holds a `body reader` row for the test file,
   and the prose of this change uses that term.
5. **Three sentences passed the length limit of the writing standard**, one in
   `.claude/rules/conformance.md` and two in each record. Each one is now two sentences.
6. **`an accident of cost` is an idiom, and a case name read `beats`.** The prose now names
   the mechanism, which is the subtraction of step 2. The case reads
   `test_the_body_reader_names_more_tokens_than_the_cost_rule_file`.

### Declined, with the reason

7. **`body_readers` raises on a binary file, on a missing file and on unparsable source.**
   This tool reads the repository and it reads no packet, so the hostile-input rule of
   `CLAUDE.md` does not reach it. A raise names the file and a silent empty answer would
   name nothing.
8. **A rule text that names `tests/mutation_cover.py` inside a negative sentence passes the
   prover.** The two anchors are deliberate: a prover that reads more wording fails on a
   rewording, and this repository already carries that trade in its other prose provers.
   The limit is stated in the docstring of each prover.
9. **The measured numbers of `docs/mutation_settlements/414-interface.json` stay as #414
   wrote them.** That record states the sweeps that ran, and this change ran different
   sweeps.

The reviewer also reported the worktree dirty in the middle of the review. That was my own
uncommitted work at that moment, and it is committed now.

Issue #461, opened 2026-08-09
Eight sharded threads wrote a fingerprint one thread does not write


Crank-Git:
## What was measured

**`dev` is red. Eight sharded threads produced a fingerprint that one thread does not
produce, and the difference is one missing field.**

`tests/test_thread_safety.py::TestTheConcurrencyContract::test_eight_sharded_threads_read_one_timeline_the_way_one_thread_reads_it`
failed on `macos-latest` with Python 3.12, on the merge commit `d94d8c1` of batch #445, in
run `31342645430`.

```
At index 10 diff: '64240_2-1-1-4-1-3_1460_7' != '64240_2-1-1-4-1-3_1460_7_0'
```

The assertion is `values == expected`, where `values` is the eight-thread reading and
`expected` is the one-thread reading. **So the sharded run wrote
`64240_2-1-1-4-1-3_1460_7` where one thread wrote `64240_2-1-1-4-1-3_1460_7_0`.** The
sharded value holds one field fewer.

## Why this outranks a flaky case

**A fingerprint exists so that one tool's output can be compared against another tool's
output.** `CLAUDE.md` states it as the first rule of this project. A fingerprint that depends
on how many threads read the packets is a fingerprint that cannot be compared.

**Clause 2 of the concurrency contract is the claim under test**, and the claim is an
equality of output. This is not a timing assertion and not a wall-clock race, so **the
repairs of #430, #371 and #369 do not cover it.** The suite asserted a correctness property
and the property did not hold.

**The eleven other jobs of the same run passed**, including every `ubuntu-latest` row from
Python 3.9 to 3.13 and the `installed-wheel` job. **The same tree passed every job, macOS
included, on `b593de5` before the merge.** So the reading is not deterministic, and a
nondeterministic failure of an equality claim is the more serious shape, not the less: it
means an interleaving exists that changes the output.

## What this issue must not conclude without evidence

**Do not close this as a flaky case.** A case that fails once in N runs on an equality of
fingerprints is reporting either a real interleaving defect or a defect in the case's own
construction. **Both are findings. Neither is "it passed on the re-run".**

**Do not repair it by widening the comparison, sorting differently, or removing the field
from the assertion.** That converts a failing correctness claim into a vacuous one, which is
the defect this project has recorded more than eighteen times.

## Where to look first, and what not to assume

**The missing field is the last component of the value.** Read which method emits
[...]

PR #475, opened 2026-08-10
Read the age clock of the thread that touches the state table

Crank-Git:
## The cause

**`BoundedStateTable` held one clock for every thread, and its age pass read every
entry.** A caller who gives each thread whole connections puts each thread at its own
point of one timeline. The thread that stands furthest ahead therefore evicts a
connection that a slower thread still reads.

**`SynAckTracker.times` is the table that lost.** It runs one age pass on every packet
and it holds a maximum age of 120 seconds.

On the concatenated timeline of `TestTheConcurrencyContract`:

- Packet 578, at 19.2378 seconds, opens `('184.150.157.177', 80, '172.16.225.48', 57380)`
  and writes `64240_2-1-1-4-1-3_1460_7`.
- Packet 581, at 19.2437 seconds, retransmits that SYN-ACK. One thread writes
  `64240_2-1-1-4-1-3_1460_7_0` for it, where `_0` is part e, the delay list.
- Packet 1773, at 151.678 seconds, belongs to another connection and another shard. It
  announces its timestamp to the shared clock, the pass reads 132.44 seconds of age, and
  it evicts the connection.
- Packet 581 then finds no entry. `SynAckTracker.record` returns an empty delay list,
  `_part_e` returns an empty string, and the value writes four parts.

That is the reading the issue records:
`At index 10 diff: '64240_2-1-1-4-1-3_1460_7' != '64240_2-1-1-4-1-3_1460_7_0'`.

**The reproduction is three packets on two threads, and it holds no race.** It fails
every run rather than one run in four.

## The repair

`ja4plus/utils/state_table.py` scopes the clock and the pass to the thread.

- `on_packet` writes the timestamp into a `threading.local`, so each thread reads the
  timestamp of its own most recent packet. A thread that announces no packet reads the
  clock the table holds, which is the behaviour this project shipped.
- Each entry carries the identifier of the thread that read it last, in a third position
  `_SHARD`.
- `evict_aged` passes over every entry of another thread that a packet timestamp dated.
- One wall clock serves every thread, so those readings compare and no thread owns an
  entry the wall clock dated. `_read_shard` returns None for such an entry, and every age
  pass evicts it. `JA4DBClient` builds the one table that reads the wall clock.

**One thread owns every entry it stores, so a caller that runs one thread reads the pass
this project always ran.** The entry of a thread that ends stays until the entry count
bound removes it, and that bound is the one that holds the memory.

`FR-concurrency-safety-8`, `FR-concurrency-safety-8a`, `FR-concurrency-safety-8b` and
`FR-concurrency-safety-8c` state the rule.

## No assertion is widened

**No existing case changes.** The diff adds cases and it removes none.
`test_eight_sharded_threads_read_one_timeline_the_way_one_thread_reads_it` holds the same
`values == expected` it held before, over the same five captures.

## How this was tested

**The 50 consecutive runs on macOS**, which acceptance criterion 2 asks for:

```
for i in $(seq 1 50); do
  .venv/bin/python -m pytest tests/test_thread_safety.py::TestTheConcurrencyContract \
    -q -p no:randomly
done
passed=50 failed=0
```

**The five gates**, on Python 3.11 on macOS:

| Gate | Result |
|---|---|
| `pytest tests/ -m "not spec_validation"` | 2883 passed, 4 skipped, 8 xfailed, 114 subtests passed |
| `pytest tests/ -m spec_validation` | 1532 passed, 143 skipped, 134 xfailed, against 134 register keys |
| `ruff format --check ja4plus/ tests/` | 195 files already formatted |

**Coverage holds.** `--cov=ja4plus` reads 94 percent on the base and 94 percent here, at
273 missed lines on each. `ja4plus/utils/state_table.py` reads 99 percent on each.

**No fingerprint moves.** A replay of all 38 committed captures writes the type, the
value, the raw form and the raw form in original order of every result, 783 in all. The
SHA-256 of that output is
`04014a1d7a1863bc3244e1348ce61b5f39f9805a988310af657a6a931a2d8907` on the base commit and
the same here. No file under `ja4plus/fingerprinters/` changes.

## The mutation reversals

Each run cleared every `__pycache__` and set `PYTHONDONTWRITEBYTECODE=1`, and
`inspect.getsource` read the loaded source in both directions.

| Mutation | Reader, clean | Reader, mutated | The cases it kills |
|---|---|---|---|
| The shard filter of `evict_aged` removed | 1 | 0 | `test_a_packet_of_another_shard_holds_the_retransmission_field`, `test_the_age_pass_of_one_thread_holds_the_entry_of_another_thread` |
| The thread clock of `on_packet` removed | 1 | 0 | `test_a_write_reads_the_clock_of_the_thread_that_makes_it` |
| The owner write of `__contains__` removed | 1 | 0 | `test_a_read_on_another_thread_moves_the_entry_to_that_thread[the in operator]` |
| The owner write of `__getitem__` removed | 1 | 0 | `test_a_read_on_another_thread_moves_the_entry_to_that_thread[the read of one key]` |
| The wall-clock exemption of the pass removed | 1 | 0 | `test_the_age_pass_evicts_an_entry_that_the_wall_clock_dated` |

Every case passes clean and after the restore.

**Two control cases prove the failing direction.**
`test_the_age_pass_evicts_the_entry_of_the_thread_that_runs_it` and
`test_one_shard_that_reads_past_the_maximum_age_writes_no_retransmission_field` both fail
against a table that evicts nothing at all, so the new cases do not pass by removing the
age bound.

## Documentation

- `docs/specs/features/03-concurrency-safety.md` states the four requirements, three
  behaviour rules and the measurement.
- `docs/api_reference.md` states the thread scope of `on_packet` and `evict_aged`.
- `README.md` states it in the concurrency contract.
- `CHANGELOG.md` and `docs/specs/spec.md` hold the round entry, reading `Round TBD`.

Verified against: https://docs.python.org/3/library/threading.html#thread-local-data
(Python 3.11, retrieved 2026-08-10).

Refs #461.


Crank-Git:
## Self-review

Two reviewers read the diff: one for correctness and concurrency safety, one for test
quality and the writing standard. `a2ee389` addresses every finding I accepted.

### Accepted, and repaired

**1. The pass scoped by thread alone ages out nothing in `JA4DBClient._cache`.**
`ja4plus/ja4db.py:326-331` calls `self._cache.on_packet()` with no timestamp, under one
lock, from any thread. Its callers share every entry rather than owning whole
connections, so the entry owner is whichever thread read it last, and a different thread
runs the pass. The age bound then stops removing an idle entry, and the count bound alone
holds the memory.

The repair reads the distinction rather than adding a switch. One wall clock serves every
thread, so those readings compare and no thread owns such an entry. New reader
`_read_shard` returns None where the operation reads the wall clock, and the pass evicts
an entry that carries None. `FR-concurrency-safety-8c` states it, and
`test_the_age_pass_evicts_an_entry_that_the_wall_clock_dated` holds it. Removing the
exemption fails that case.

**2. An assertion inside a spawned thread reports no cause.** Python prints the
`AssertionError` of a thread to standard error and raises nothing in the thread that
joins, so `test_a_read_on_another_thread_moves_the_entry_to_that_thread` failed on a
later, less specific assertion. The case now collects its readings and asserts them on
the main thread.

**3. Six sentences passed the 25-word limit of `.claude/rules/ste.md`.**
`FR-concurrency-safety-8` also carried the passive voice and put its condition after its
action. Each one is split, and `FR-concurrency-safety-8` now reads
`When an entry of the calling thread waits past the maximum age, the age pass evicts that
entry.`

**4. The owner write of `__getitem__` was unmeasured.**
`test_a_read_on_another_thread_moves_the_entry_to_that_thread` is now parametrized over
the `in` operator and the read of one key, so a mutation of either write fails a case.

### Read and found sound

- No module outside `ja4plus/utils/state_table.py` indexes into the entry list, so the
  two-item to three-item change reaches no other call site.
- `__setitem__` is the only path that builds an entry, so no entry lacks the third slot.
- `threading.local` drops the slot of a thread that ends, so the clock store holds no
  bound of its own.
- `on_packet(None)` still moves the calling thread to the wall clock. That call sets the
  attribute, so `hasattr` reports True and `_read_clock` reads the wall clock.
- No caller passes an explicit `now` to `evict_aged` today.

### Rejected

**The long sentences of the `CHANGELOG.md` entry and the `docs/specs/spec.md` row.** Both
files are a historical ledger, and every adjacent row reads the same way. A row written to
a different standard than the rows beside it is harder to read, not easier. The writing
standard binds the normative pages, and those are repaired above.

Issue #483, opened 2026-08-10
Batch: the documentation and rule repairs


Crank-Git:
## Members

- [x] #477 — Correct the superseded fallback rule of spec.html and hold it against a shape
- [x] #473 — The document set of the count cases reaches into a harness worktree
- [x] #480 — The batch-gate rule states dev carries no required status check, and it now does
- [x] #482 — Two live integration branches cannot both assign round numbers

## The sequence

**#480 and #482 both edit `.claude/rules/batch-gate.md`.** #482 runs after #480 sub-merges.

## Round numbers stay `TBD` until the batch gate

**#482 records why.** A member writes `TBD` and keeps it through every sub-merge. The project
manager assigns every round number of this batch in one pass on the integration branch,
immediately before the batch pull request, when the row count of `dev` is known and fixed.

**Batch #476 is live at the same time.** Both batches therefore hold `TBD` rows, and neither
assigns a number until its own batch gate. The batch that merges second reads the row count of
`dev` after the first merge and numbers from there.

## The integration branch

`batch/483-documentation-and-rule-repairs`, off `dev`.

Issue #473, opened 2026-08-10
The document set of the count cases reaches into a harness worktree


Crank-Git:
## What was measured

**`tests/test_documented_method_count.py` collects 153 cases on `dev` and 543 cases in a
checkout that holds two live worker worktrees.** The difference is 390 cases, and no commit of
either measurement touches that file.

`tests/test_documented_method_count.py:340-343` builds `DOCUMENTS` from the file system at
import.

```python
found = sorted((REPO_ROOT / "docs").rglob("*.md"))
found += sorted((REPO_ROOT / ".claude").rglob("*.md"))
found += sorted((REPO_ROOT / "tests").rglob("*.md"))
found += sorted(REPO_ROOT.glob("*.md"))
```

**The harness places a worker worktree at `.claude/worktrees/agent-<id>`**, and that worktree is
a whole checkout of the repository. So the second line reaches every document of every live
worktree.

| Reading | Documents under `.claude/` |
|---|---|
| Whole tree, two worktrees live | 196 |
| Inside `.claude/worktrees/` | 189 |
| The project's own rule files | 7 |

**189 of the 196 documents the case reads are copies inside a worktree.** Three parametrized
cases read `DOCUMENTS` or a set derived from it, at lines 432, 444 and 455.

## Why this is a defect and not a curiosity

**The case measured a different document set than the one the change touched, and it passed.**
It passed because a worktree holds a valid copy, so the copies agree with the rule. That is the
shape this project has recorded more than eighteen times: a comparison whose result does not
depend on the thing under test.

**Three consequences follow.**

1. **The reading is not reproducible.** The same commit collects 153 cases or 543 cases
   according to how many workers happen to be running. A count stated in a Changelog entry is
   then a count of the host and not of the commit.
2. **A stale worktree turns the suite red for a reason unrelated to the change.** A worktree
   left by a killed session holds the documents of an older commit, and a case that reads a
   count from a document then reads two disagreeing answers.
3. **The runner and the local gate measure different sets.** The GitHub runner holds no
   worktree, so it reads 7 documents under `.claude/` where a local gate reads 196. **A local
   pass is therefore not evidence of a runner pass for these cases**, and the reverse holds too.

## What to build

- **Read tracked files, not the file system.** `git ls-files '*.md'` names exactly the documents
  the project owns and it names no worktree copy, no ignored file and no build output. That is
  the same correction `tests/mutation_sweep.py` already carries, which reads its module list
  rather than walking a directory.
[...]

Issue #436, opened 2026-08-09
State the module pathspec so its reading matches what it matches


Crank-Git:
## What was measured

**`FR-pre-release-validation-16` names a pathspec pair that reports the right count for the
wrong reason, and a reader who copies it will be wrong.**

```
git ls-files 'ja4plus/**/*.py'                     -> 24
git ls-files 'ja4plus/*.py'                        -> 31
git ls-files 'ja4plus/*.py' 'ja4plus/*/*.py'       -> 31
git ls-files ':(glob)ja4plus/*.py'                 ->  7
```

**In a default git pathspec, `*` crosses `/`.** Only `:(glob)` magic stops it at a
separator.

## The history, because this is the third pathspec error of one epic

1. **The original `'ja4plus/**/*.py'` listed 24 of 31 modules.** Git reads `**` as one or
   more directories, so it matched no file of the top directory and dropped `cli.py`,
   `processor.py`, `ja4db.py`, `output.py`, `types.py`, `watch.py` and `__init__.py` —
   556 of 3545 mutations. #411 found it.
2. **#411 replaced it with `'ja4plus/*.py' 'ja4plus/*/*.py'`, which reports 31.** That is
   the correct module count. **The reasoning is wrong: the first term alone already matches
   all 31, and the second term adds nothing.**
3. **#414 was told by its brief to confirm a seven-module list with `git ls-files
   'ja4plus/*.py'`, which names 31.** The worker measured it and reported the
   contradiction. Its module set was right throughout, because `tests/mutation_sweep.py`
   uses `Path.glob` and not a git pathspec.

**The count is right today and the rule is not, so the next writer who reasons from it
reproduces the defect.**

## What to build

- **State `FR-pre-release-validation-16` in a form whose reading matches its intent.**
  Either `git ls-files 'ja4plus/*.py'` alone, with a sentence stating that `*` crosses a
  separator so this names every module of the package, or `:(glob)` forms that mean what
  they appear to mean.
- **`tests/test_mutation_sweep_module_list.py` already compares the module list against
  the tracked files.** Extend it so the *pattern the requirement states* is the pattern the
  case reads, and **a requirement naming a pathspec that reports a different set than the
  sweep reads fails a case.**
- **Prove it fails.** Put a `**` form back and watch it go red, then revert.

## Acceptance criteria

- [ ] `FR-pre-release-validation-16` names a pathspec whose plain reading equals what it
[...]

PR #487, opened 2026-08-10
Read the document set of the count cases from the tracked file list

Crank-Git:
Closes nothing on its own. This pull request is a member of batch #483 and it builds #473.

## What was wrong

`tests/test_documented_method_count.py` built `DOCUMENTS` from a walk of `docs/`,
`.claude/`, `tests/` and the repository root. **The agent harness places a worker worktree
at `.claude/worktrees/agent-<id>`, and that worktree is a whole checkout**, so the walk of
`.claude/` read every document of every live worker. The reading counted the host and not
the commit.

**The cases passed on such a checkout, because a worktree holds a valid copy that agrees
with the rule.** That is the shape this project records more than eighteen times: a
comparison whose result does not depend on the thing under test.

## The measurement, inside a worker worktree, 2026-08-10

| Reading | Collected cases of the file |
|---|---|
| Present reader, no worktree page | 153 |
| Present reader, one page below `.claude/worktrees/` | 156 |
| Repaired reader, no worktree page | 166 |
| Repaired reader, one page below `.claude/worktrees/` | 166 |

Two of the three extra cases failed on that page against the present reader.

## What changed

`documents()` reads `git ls-files -z '*.md'`. That list names the tracked pages alone, and
it names no worktree copy, no ignored file and no build output. `tests/mutation_sweep.py`
already carried this correction for its module list.

**In a default git pathspec `*` crosses `/`**, so the one term reaches every depth and its
plain reading equals what it matches. `FR-pre-release-validation-16` and
`.claude/rules/conformance.md` state the rule that #436 repaired. A read of 2026-08-10
reports three counts: `git ls-files '*.md'` lists 59 files, `git ls-files '**/*.md'` lists
56, and `git ls-files ':(glob)*.md'` lists 3. The tracked list and the walk name the same
59 pages on a clean checkout, so the runner reads the set it read before.

New `FR-documentation-16` and its four parts state the rule.

## How it was tested

The repair is proven in both directions.

- `test_the_reader_names_no_markdown_page_of_a_worktree` writes one page below
  `.claude/worktrees/`, and it removes the directory it created and no other. Against the
  walk it fails with
  `AssertionError: .claude/worktrees/issue-473-qnt_uybr/docs/copy.md reaches a parametrized case`.
- Nine anchor cases name one page of each depth and one page of each root, so a document
  that should be read is missed when absent.
- `test_the_anchor_set_fails_a_reader_that_names_one_depth_of_one_directory` proves the
  anchor set fails a reader that drops a root.
- `test_the_reader_fails_where_the_pathspec_names_no_document` holds the floor, because an
  aggregate over an empty set passes.
- `test_the_pathspec_reaches_every_depth_of_the_corpus` reads the depth set `{0, 1, 2, 3}`.

The five gates, on macOS with Python 3.11 in a virtual environment of the worktree:

```
pytest tests/ -m "not spec_validation"   2918 passed, 4 skipped, 8 xfailed, 114 subtests passed
ruff format --check ja4plus/ tests/      197 files already formatted
```

Coverage reads 94%, 273 missed of 4253 statements, under `--cov=ja4plus`. No file under
`ja4plus/` changes and no fingerprint moves.

The whole suite collects 4739 of 4762 cases with a worktree page present and 4739 of 4762
with none.


Crank-Git:

A review of the diff ran six checks against live command output and it found no defect.

1. **The two readers name the same set on a clean checkout.** `walked_documents()` and
   `tracked_documents()` each return 59 paths, and the set difference is empty in both
   directions.
2. **The `-z` read handles an unusual path.** Against a tracked file
   `docs/tmp test dir/weird name é.md`, the reader names the file and reads its text. The
   `if name` filter drops the empty element the terminal NUL leaves.
3. **The floor fails at import and it hides nothing.** An empty pathspec raises
   `AssertionError` during collection with the message that names the count, and a failed
   read of git raises `subprocess.CalledProcessError` from `check=True`. A reader tells the
   two apart.
4. **The worktree case removes the directory it created and no other.**
   `tempfile.mkdtemp` names a unique directory, and `shutil.rmtree` in the `finally` block
   removes that path alone. The case globs no sibling.
5. **Neither new floor case is vacuous.** `git ls-files -z ':(glob)docs/*.md'` lists 9
   pages of one depth and misses `README.md`, `docs/specs/spec.md` and
   `.claude/rules/ste.md`, so the `missing` list is not empty.
   `git ls-files -z '*.no-such-suffix'` lists nothing, and the raised message reads
   `names 0 documents`.
6. **The git call anchors on the repository root.** `pytest test_documented_method_count.py`
   run from inside `tests/` reports 166 passed, because `cwd=str(REPO_ROOT)` holds.

The review also reproduced the three pathspec counts and the two collected counts this
pull request states.

Issue #453, opened 2026-08-09
Correct the state-table count of the concurrency-safety feature file


Crank-Git:
Found by the self-review of #387 on 2026-08-09 while it corrected the method count. **#387
changed the word `methods` to `fingerprinters` on the same lines and left the table count,
because that count is a different number.** It reports this rather than repairs it.

## The measurement

**Two documents state a different count of state tables, and the code agrees with one.**

| Document | What it states |
|---|---|
| `docs/api_reference.md:355` | `seventeen` state tables across the ten fingerprinters |
| `docs/api_reference.md:519-523` | 57400 remembered keys, 10.2 MiB, `Ten fingerprinters hold seventeen state tables` |
| `docs/specs/features/03-concurrency-safety.md:217-231` | sixteen state tables, 47400 keys, 8.5 MiB |

**The code holds seventeen.** Measured on the base commit `d123ab9`:

```
$ python -c "from ja4plus import Processor; \
  print(sum(len(s.tables) for s in Processor().stats().values()))"
17
```

`docs/api_reference.md` states the measured count and
`docs/specs/features/03-concurrency-safety.md` states a superseded one. Round 39 moved
thirteen tables, and `tests/test_processor_stats.py:29` records the count in a comment.

## What to change

Correct the feature file to the measured count, and **correct the two numbers that rest on
it**: the remembered-key count and the memory figure. A table count that a reader repairs
alone leaves two derived numbers wrong.

**Prefer a case over a corrected number.** `tests/test_processor_stats.py` already reads the
tables out of a live `Processor`, so a case can hold each stated count against
`Processor().stats()` rather than restate it. `tests/test_documented_method_count.py`, which
#387 built, holds the shape: read the number out of the package, then compare each document
against it.

Read `docs/api_reference.md:519-523` and the feature file together before you pick which
figures are derived, because the memory figure rests on the key count and the key count
rests on the table count.

## Acceptance criteria

- [ ] Every document that states a count of state tables states the count
      `Processor().stats()` reports.
- [ ] Every number derived from that count agrees with it, and the verdict names each
      derived number and how it was recomputed.
- [ ] A case fails when a stated table count and the live count disagree.
- [ ] The verdict lists every count of state tables found in `README.md`, `docs/` and
      `.claude/`, and says which were changed and which were left and why.
- [ ] No file under `ja4plus/` changes and no fingerprint moves.

## Notes

**Do not change the table count in the code to match a document.** The code is the
[...]

PR #489, opened 2026-08-10
Correct the state-table count of the concurrency-safety feature file

Crank-Git:
Closes nothing on its own — #453 closes with the batch pull request.

## What

`docs/specs/features/03-concurrency-safety.md` stated sixteen state tables, 47400
remembered keys and 8.5 MiB. `docs/api_reference.md` stated seventeen, 57400 and 10.2 MiB.
**The code holds seventeen**, so the feature file carried a superseded count and two
numbers that rest on it.

## Why the number moved, and where it hid

The state-bound table of the feature file lost one row. #215 added
`JA4TFingerprinter.connections` at 10000 entries and 600 seconds, and that table never
gained the row. 47400 plus 10000 is 57400, so one missing row explains all three numbers.

## The measurement, taken on the base commit `ec207d9`

```
$ python -c "from ja4plus import Processor; \
  print(sum(len(report.tables) for report in Processor().stats().values()))"
17
$ python -c "from ja4plus import Processor; \
  print(sum(t.max_entries for r in Processor().stats().values() for t in r.tables.values()))"
57400
```

57400 keys at the 187 bytes #41 measured cost 10.2 MiB.

## A case, not a corrected number

New file `tests/test_documented_state_table_count.py` reads the three numbers out of a
live `Processor` and compares every Markdown page of the corpus against them. It also
reads the state-bound table row by row, which is the check that catches the missing row
rather than the stale sentence.

**Four reversals prove the cases fail, and each one was restored.**

- The `JA4TFingerprinter.connections` row removed: `states the bound of 16 state tables, and one processor holds 17`.
- The superseded sentence restored: three failures, one for each number.
- The same mutation on `docs/api_reference.md`: two failures, so the reader reaches both pages.

Floors refuse a reader that matched nothing, because an aggregate over an empty set
passes. Eleven more cases prove the readers take each wording and take no count of another
thing.

## What was left, and why

`readable_text` cuts `CHANGELOG.md` and the `## Changelog` and `## Risks & open questions`
sections of `docs/specs/spec.md`, so round 85 keeps its record of fifteen tables, 46400
keys and 8.3 MiB. The #279 reading of `40200 entries across the seventeen state tables` is
a dated measurement that already agrees with the code, and it stands unchanged.

## How tested

On this host, in a virtual environment of the worktree:

- `pytest tests/ -m "not spec_validation"` — 3319 passed, 4 skipped, 8 xfailed, 114 subtests. The base commit reads 3114 passed.
- `pytest tests/ -m spec_validation` — 1532 passed, 143 skipped, 134 xfailed.
- `ruff format --check ja4plus/ tests/` — 199 files already formatted.
- `mypy --strict ja4plus/` — no issue in 31 source files.
- `pytest --cov=ja4plus` — 94%, 273 missed of 4253, which is the reading of the base commit.


Crank-Git:
## Self-review

Two Sonnet reviewers read `ec207d9..a0cdb15` by lens. `77ac953` answers every finding they raised, and it names the two they raised that need no change.

### Correctness lens

**Accepted — the `#279` sentence is unread.** `docs/specs/features/03-concurrency-safety.md` states `The run ends with 40200 entries across the seventeen state tables`, `readable_text` cuts no section that holds it, and `stated_table_counts` returns `[]` for it. That number is correct today and nothing measures it. **The reader stays as it is and the docstring now states the limit.** A sentence that records what one past run ended with is a dated measurement, and a reader that took it would demand a rewrite of the record on the day the code moves.

**Accepted — the bound reader anchored on a column position.** `bounds_table_rows` read `cells[1]`, so a column that a later writer inserts before the entry count would misread that column in silence. It now finds the column by the heading `Maximum entry count`, and it fails loud where that heading is absent. **Proven by its reversal, and restored.** A `Shard` column inserted before it fails the two bound cases with `states the bound of 0 state tables, and one processor holds 17`.

**Rejected — the `#215` attribution stands.** The reviewer read `git blame` on `ja4plus/fingerprinters/ja4t.py:51` and found the commit message names `#300`. `gh pr view 300` opens `Builds #215.`, so `#300` is the pull request and `#215` is the issue. This project cites the issue, and `docs/specs/features/03-concurrency-safety.md:533` already made the same attribution.

**No finding on the arithmetic, the floors, or the count history.** The reviewer reproduced all four reversals and read the same failure counts the record states.

### Writing-standard lens

**Accepted — five sentences passed the 25-word cap**, two of them chaining three instructions. Each one now reads as two or three sentences, across `CHANGELOG.md`, `docs/specs/spec.md` and the feature file. Three comma splices now read as separate sentences.

**Accepted — `remembered key` was not in the `## Terms` table.** The word is older than this pull request and this repair counts them in three places, so the table gains the row.

**Rejected — `and each one was restored` keeps its passive form.** It is quoted evidence of the shape the surrounding rounds already use, and the standard reproduces evidence verbatim.

**One sentence stands above the cap on purpose.** `Those three read ...` in the `docs/specs/spec.md` row carries three verbatim assertion messages and four carrier words. The standard reproduces an error message verbatim.

**No finding on the controlled vocabulary, the docstrings, the test names, or the comments.** The reviewer searched every added line for the count of fingerprinter classes applied to the word `method` and found none.

### The gate after the change

- `pytest tests/ -m "not spec_validation"` — 3319 passed, 4 skipped, 8 xfailed, 114 subtests.
- `pytest tests/ -m spec_validation` — 1532 passed, 143 skipped, 134 xfailed.
- `ruff check` and `ruff format --check` — clean over 199 files.
- `mypy --strict ja4plus/` — no issue in 31 source files.
- `pytest --cov=ja4plus` — 94%, 273 missed of 4253, which is the reading of the base commit.

Issue #482, opened 2026-08-10
Two live integration branches cannot both assign round numbers


Crank-Git:
## What was measured

**`tests/test_specification_changelog.py::test_the_changelog_row_count_equals_the_highest_round_number`
fails on the second of two integration branches that run at the same time.** Measured by the
project manager on 2026-08-10, on `batch/476-count-repairs`, at the sub-merge gate of #456.

```
AssertionError: the Changelog holds 169 rows and its highest round is 173
assert 169 == 173
```

**Neither member is at fault, and the merged state of each member is correct.** The invariant
is a condition on a sequence that two branches assign from at once.

| Branch | Rows it holds | Rounds it assigns |
|---|---|---|
| `batch/470-sweep-tools-and-shard-bug` | 171 | 168, 169, 170, 171 |
| `batch/476-count-repairs` | 169 | 172, 173 |

`batch/476` starts from `dev`, which holds 167 rows, so it holds 169 rows after its two members
and it records a highest round of 173. **The four rows of the other branch are the difference,
and they reach this branch only when that branch merges.**

## Why the rule is right and the procedure is wrong

**`row count == highest round` states that the record is contiguous**, which is the property a
reader needs: a citation of `Round N` reaches row N. #302 records what an orphan costs. **The
rule is not the defect.**

**The defect is that the project manager assigns a round number at the sub-merge gate**, which
is a per-batch event, from a sequence that is global to the repository. One live batch hides
this, because that batch is the only writer. Two live batches make the assignment a race that
the invariant then reports.

**The failure is benign in this instance and it will not always be.** It resolves itself the
moment `batch/470` merges into `dev` and `batch/476` takes `dev`: 171 rows plus 2 equals 173
rows against a highest round of 173. **But nothing enforced that ordering**, and a project
manager who assigned 168 and 169 on the second branch would have produced two rows numbered 168
and two numbered 169, which merges cleanly and which no case reads as a duplicate.

## The readings

1. **Assign the round number at the batch gate and never at the sub-merge gate.** A member
   writes `TBD` and it keeps `TBD` through every sub-merge. The project manager assigns the
   numbers in one pass on the integration branch, immediately before the batch pull request,
   when the row count of `dev` is known and fixed. **This is the smallest change and it removes
   the race rather than detecting it.**
2. **Serialize the batches.** One integration branch at a time. Correct, and it costs the
   concurrency the batch model exists to provide.
[...]

Issue #456, opened 2026-08-09
Correct the two stale counts the feature acceptance criteria state


Crank-Git:
## What was measured

**Two acceptance criteria of the feature files state counts the repository no longer holds.**
Measured by #393 on 2026-08-09 while it swept every command of `docs/specs/features/*.md`.

| Where | What it expects | What the repository holds |
|---|---|---|
| `docs/specs/features/00-foundation.md:214` | 75 files under `tests/foxio_vectors/` | **90**, by `git ls-files` |
| `docs/specs/features/09-release.md:144` | the string `0.6.0` in one file | **eight files**, by `grep -rn` |

## Why this is worth an issue

**A criterion that states a stale count is a criterion nobody can satisfy.** A reader who
runs it reads a failure and learns to disregard the page, which is the same failure mode the
guarded-case work of #424 and #426 repaired for the suite.

**The second one is the more interesting.** `09-release.md:144` expects one file to hold the
version string, and #67 exists to put the version in one place. **So the criterion may be
stating the intended end state rather than a stale measurement.** Read #67 before you decide
which of the two it is, and say which reading you took.

## What to change

**Prefer a case over a corrected number.** A count in prose that nothing measures is stale by
the time someone reads it, and this repository has now corrected that shape more than seven
times. `tests/test_requirement_scope.py` and `tests/test_specification_changelog.py` read
specification text as text and are the shape to follow.

**Re-measure both numbers.** They are premises of this issue.

**Decide, and record the decision, whether `09-release.md:144` is a stale count or a
statement of the end state #67 builds.** If it is the end state, this issue records that and
changes the sentence to say so, rather than raising the number to eight.

## Acceptance criteria

- [ ] No acceptance criterion of `docs/specs/features/*.md` states a count the repository
      contradicts, or the criterion states plainly that it describes an end state that is not
      yet built and names the issue that builds it.
- [ ] A case fails when a stated count and the measured count disagree, or the pull request
      states why a case is not worth its cost for a given criterion.
- [ ] The pull request records the measurement of both counts.
- [ ] No file under `ja4plus/` changes and no fingerprint moves.

## Links

- Measured by: #393 on 2026-08-09
- Related: #67, which puts the version in one place

Crank-Git:
[...]

PR #490, opened 2026-08-10
Assign every round number at the batch gate

Crank-Git:
Closes nothing on its own. The batch pull request of #483 closes #482.

## What

**One pass at the batch gate assigns every round number of a batch.**
`.claude/rules/batch-gate.md` gains the section `Assign every round number at the batch
gate`. A member writes the literal `TBD` in `CHANGELOG.md` and in the Changelog row of
`docs/specs/spec.md`, and it keeps `TBD` through every sub-merge. The file declines one
integration branch at a time and states the reason: that order costs the concurrency the
batch model exists to provide.

**`tests/test_specification_changelog.py` gains a reader that the row-count rule cannot
replace.** `_contiguity_failures` requires the rounds 1 to the row count, each on one row.

## Why

The round sequence is global to the repository and a sub-merge is an event of one batch,
so two live integration branches assign from one sequence at once. The project manager
measured the result on 2026-08-10 at the sub-merge gate of #456.

## The measurement the repair rests on

**`row count == highest round` passes on a table that carries 168 twice and 170 nowhere.**
Proved on the shipped page before the repair. A copy of `docs/specs/spec.md` whose round
170 reads 168 failed the new case and passed the older one.

```
FAILED test_every_changelog_round_number_names_one_row
FAILED test_the_changelog_assigns_every_round_from_one_to_the_row_count
E   AssertionError: assert ['these round numbers open more than one row: [168]',
E                          'these rounds from 1 to 171 open no row: [170]'] == []
2 failed, 7 passed
```

`test_the_changelog_row_count_equals_the_highest_round_number` is not in that failure
list. `test_the_row_count_rule_passes_on_a_table_that_repeats_a_round` keeps the
measurement as a case.

**One statement of the issue body is corrected by measurement.** The body states that the
present invariant passes on a duplicate and on a gap. **A gap alone raises the maximum
above the count**, so the row-count rule reads a pure gap and fails on it. A repeat inside
a table of N rows numbered 1 to N forces a gap, so the shape that hides from the older
rule carries both. The new reader reads each shape on its own, and
`TABLE_THAT_HOLDS_A_GAP` proves the gap direction.

## The cases

Written first. Against the module that held no reader they failed 6 of 9, and the three
that passed were the two older rules and the record of what the older rule reads as
correct.

| Case | Shape | Result |
|---|---|---|
| `test_the_changelog_assigns_every_round_from_one_to_the_row_count` | The shipped page | Green, and red on the 168 twice copy |
| `test_the_row_count_rule_passes_on_a_table_that_repeats_a_round` | 1, 2, 2, 4 | Records that the older rule passes |
| `test_the_reader_fails_on_a_table_that_repeats_a_round` | 1, 2, 2, 4 | Reports the repeat and the gap |
| `test_the_reader_fails_on_a_table_that_holds_a_gap` | 1, 2, 4 | Reports the gap |
| `test_the_reader_passes_on_a_table_that_assigns_every_round_once` | 1, 2, 3, 4 | Green |
| `test_the_reader_passes_on_a_table_whose_new_rows_wait_for_a_round` | 1 to 4 and two `TBD` | Green, so a live integration branch stays legal |
| `test_the_reader_fails_on_a_table_that_names_no_numbered_row` | Two `TBD` rows | Fails, because an aggregate over an empty list passes |

`MINIMUM_ROWS` holds the second floor at 124 for every case that reads the shipped page.

## How this was tested

```
pytest tests/ -m "not spec_validation"   3066 passed, 4 skipped, 8 xfailed, 114 subtests
ruff format --check ja4plus/ tests/      199 files already formatted
```

No file under `ja4plus/` changes, no fingerprint moves, and the coverage of `ja4plus`
holds.

Verified against https://docs.github.com/en/actions/how-tos/manage-workflow-runs/skip-workflow-runs
(retrieved 2026-08-09), which `.claude/rules/batch-gate.md` already cites for the head
commit rule this branch follows.


Crank-Git:
## Self-review

Two lenses read the diff: the writing standard `.claude/rules/ste.md`, and the correctness
of the new cases.

### The correctness lens found no defect

A brute-force sweep over every sequence of length 1 to 5 with values 1 to 6 reports that
`_contiguity_failures` returns `[]` for a permutation of 1 to N and for nothing else. It
therefore reports every repeat, every gap and every number outside the range, and it
reports no false failure on a legal table in any row order. A `TBD` row reaches no entry,
because `CHANGELOG_ROW` requires `\d+`.

The lens named `test_the_row_count_rule_passes_on_a_table_that_repeats_a_round` as a check
proved in one direction. **That is its purpose and the module docstring states it.** The
case records what the older rule reads as correct, so a later reader does not take that
rule for a complete one. It asserts nothing about the new reader, and five other cases
drive the new reader in both directions.

### The writing lens found twelve departures, and commit `8adbe57` repairs each one

- Nine sentences ran past the 25-word limit for a description.
- One sentence wrote a hypothetical in the conditional perfect rather than the present.
- One sentence read `The measurement is proved` in the passive voice.
- One sentence read `this project declines one integration branch at a time`, which states
  the opposite of the decision. It now reads `this project declines that order`.

### Two findings are declined, with the reason

**A vertical list inside a Changelog row of `docs/specs/spec.md` is impossible.** That row
is one cell of one table, and a cell holds no list. The sentence is split into four short
sentences instead, which meets the length rule.

**The same finding against `CHANGELOG.md` is declined for a second reason.** The items are
five fixed tables and no list of conditions, so rule 16 does not reach them. The sentence
is split there too.

Issue #457, opened 2026-08-09
Rule on the sentence-length rule for the Changelog record


Crank-Git:
## The question

**Does the `## Changelog` table of `docs/specs/spec.md` follow the sentence-length rule of
`.claude/rules/ste.md`, or is it exempt?**

Raised by #393 on 2026-08-09. Its self-review found about twelve sentences of its own
Changelog row over the 25-word limit. **The finding is correct against the letter of the
rule.**

## Why the worker did not repair it

**All 159 numbered rows read the same way.** Rounds 143 and 156 are well past the limit, and
so are the neighbouring `### Fixed` entries of `CHANGELOG.md`. #393 split the sentences that
split cleanly and left the rest, because **a row written to a different standard than its
159 neighbours is a house-style change for the whole record**, which is wider than any one
issue.

That reasoning is sound, and the decision belongs to the user.

## The two readings

1. **The record is exempt.** A Changelog row is a dense record of one round, written once and
   read by someone who wants the whole measurement in one place. `.claude/rules/ste.md` gains
   a stated exemption for the `## Changelog` table of `docs/specs/spec.md` and the entries of
   `CHANGELOG.md`, and a case holds every other document to the limit.
2. **The record is not exempt, and the whole table needs a pass.** 159 rows are rewritten to
   the limit, in one issue of its own, and a case then holds every new row.

**Reading 2 has a cost this issue must state: a Changelog row records a past round.**
Rewriting the prose of a row does not falsify what it records, but it does mean the text a
past reader saw is not the text a future reader sees. This repository has ruled repeatedly
that **a dated record of a past measurement is quoted, not rewritten**, and reading 2 sits
against that rule.

**The project manager recommends reading 1**, and states the reason: the exemption makes the
rule true, where the alternative makes 159 records disagree with what they recorded.

## What a decision unblocks

Nothing is blocked. Every worker writes a Changelog row today, and each one writes it the way
its 159 neighbours read. **A decision stops each worker's self-review from finding the same
correct-and-unactionable defect.** #393's is the second.

## Links

- Raised by: #393 on 2026-08-09
- The rule: `.claude/rules/ste.md`

Crank-Git:
**Ruling of the user, 2026-08-10. The record is exempt.**

`.claude/rules/ste.md` gains a stated exemption from the sentence-length rule for the
`## Changelog` table of `docs/specs/spec.md` and for the entries of `CHANGELOG.md`. **Every
[...]

PR #494, opened 2026-08-10
Exempt the two records from the sentence-length rule

Crank-Git:
Builds #457. The user ruled on 2026-08-10 that the record is exempt from the
sentence-length rule.

## What changes

`.claude/rules/ste.md` gains the section `## The one exemption`. It names two records and
it names no third.

1. The entries of `CHANGELOG.md`.
2. The `## Changelog` table of `docs/specs/spec.md`.

New file `tests/test_changelog_sentence_exemption.py` holds 25 cases. `exempt_records`
reads the list of that section alone. `evaluate` reads the exemption against the tracked
document list of git, and it reports three states: a record the ruling names nowhere, a
document the exemption reaches outside the two records, and a document set below the floor
of 40.

## Why

A Changelog row records one round, and this project quotes a dated record of a past
measurement rather than rewriting it. A rewrite of the rows falsifies nothing they record.
It does mean that the text a past reader saw is not the text a future reader sees.

## The blanket form fails

**The exemption names a path and never a class of files.** A blanket exemption for any file
whose name holds the word `changelog` reaches `docs/CHANGELOG.md`, which this ruling never
read. Two live mutations of the shipped rule prove the failing direction.

| Mutation | Cases killed |
|---|---|
| A third record, the entries of `docs/CHANGELOG.md` | 2, and one failure names `docs/CHANGELOG.md` |
| A blanket item that exempts any file whose name holds the word `changelog` | 4 |

Each mutation was restored, and the 25 cases are green after each restore.

## How this was tested

The cases came first. Against the unrepaired rule they failed 5 of 25.

```
FAILED tests/test_changelog_sentence_exemption.py::test_the_writing_standard_states_the_exemption
FAILED tests/test_changelog_sentence_exemption.py::test_the_exemption_names_the_two_records_and_no_other
FAILED tests/test_changelog_sentence_exemption.py::test_the_exemption_names_the_changelog_table_of_the_specification
FAILED tests/test_changelog_sentence_exemption.py::test_the_exemption_names_the_entries_of_the_changelog
FAILED tests/test_changelog_sentence_exemption.py::test_this_repository_exempts_the_two_records_alone
5 failed, 20 passed
```

The five gates ran in a virtual environment of this worktree. The verdict of the worker
records each result.

## What this change does not do

**A case here measures no sentence of a document outside the two records.** A sweep that
measured every sentence of every page is separate work, and this ruling orders none. The
module docstring states that limit.

## Records

`CHANGELOG.md` and the `## Changelog` table of `docs/specs/spec.md` each carry one new
entry, and both read `TBD` in place of the round number. The project manager assigns the
number at the batch gate.


Crank-Git:
## Self-review

A Sonnet child read the diff by three lenses: the readers of the new case file, the writing standard of the new rule section, and the two records.

**One finding was real and this branch repairs it.** `_record_of` kept the first path of a list item alone. An item that names the entries of `CHANGELOG.md` and `docs/CHANGELOG.md` therefore widened the exemption, and `evaluate` reported nothing. `exempt_records` now returns one record for each path of an item, and two cases hold that reading: `test_the_reader_reports_both_paths_of_one_item` and `test_a_second_path_of_one_item_that_claims_the_exemption_fails`. Commit `841a45a` carries the repair.

**One finding was a stale measurement and this branch corrects it.** The two records stated 190 and 786 sentences past the limit, and those counts were measured before this round appended its own entry and row. Both records now read 191 and 787, and each states that the count includes the entry and the row of this round.

The reviewer found no other defect. It confirmed the floor of 40 documents, the order that reads a path before a blanket word, the region readers, the 25-word boundary, the writing standard of the new section, and Python 3.9 compatibility.

Issue #452, opened 2026-08-09
Read the dependency entries rather than every quoted substring


Crank-Git:
## What was measured

**`_dependency_block` of `tests/test_documentation_site.py:364` collects every double-quoted
substring of a dependency block, and a comment inside the block reads as an entry.** It
already returns `not spec_validation` as an entry of the `dev` extra, which is a fragment of
a comment and not a distribution.

Found by #378 on 2026-08-09. **#378 stopped importing the helper rather than change it**,
because a change to a shared reader reaches callers outside its scope. That was right.

## Why it is harmless today, and why that is not a reason to leave it

The two callers that exist read the `docs` extra and the runtime list, and **neither of
those blocks carries a comment inside its brackets**. So the helper returns the right answer
for every caller it has.

**The next caller is the defect.** A reader of the `dev` extra gets a list that holds a
comment fragment, and a case built on that list either compares against the fragment or
silently accepts it. **A helper that is correct only because of what its callers happen to
read is a helper that fails on the caller that arrives next.**

## What to change

**Read the entries rather than the quoted substrings.** A `pyproject.toml` dependency block
is TOML, and the repository already installs a TOML reader through the standard library
`tomllib` on Python 3.11 and above; read what the package's own floor of Python 3.9 permits
before choosing the instrument.

**Prove it in both directions.** The helper returns no comment fragment for the `dev` extra,
and it still returns every real entry of the `docs` extra and the runtime list. **A repair
that returns an empty list passes a case that only checks for the absence of the fragment**,
so a case must assert the real entries too.

## Acceptance criteria

- [ ] `_dependency_block` returns no comment fragment for any extra of `pyproject.toml`,
      including the `dev` extra.
- [ ] It returns every real entry of the `docs` extra and of the runtime list, and a case
      asserts those entries by name.
- [ ] Every existing caller of the helper still passes.
- [ ] No file under `ja4plus/` changes and no fingerprint moves.

## Links

- Found by: #378
- Related: #446, which decides the pin shape of the `dev` extra and would add a caller.

Crank-Git:
[...]

PR #495, opened 2026-08-10
State one recorded pin shape for every dev entry

Crank-Git:
Builds the ruling the user gave on #446 on 2026-08-10. Three parts: pin `pytest`,
`pytest-cov` and `build` exactly; leave `mypy` floating; delete `requirements.txt`.

## The measurement of 2026-08-10

A clean virtual environment ran `pip install -e ".[dev]"` and reported these versions.

| Entry before | Resolved on 2026-08-10 | Entry after |
|---|---|---|
| `pytest>=7.0` | 9.1.1 | `pytest==8.4.2` |
| `pytest-cov>=5.0` | 7.1.0 | `pytest-cov==7.1.0` |
| `build>=1.0` | 1.5.0 | `build==1.4.4` |
| `mypy>=1.11` | 2.3.0 | `mypy>=1.11`, unchanged |
| `ruff==0.16.2` | 0.16.2 | `ruff==0.16.2`, unchanged |

**Two of the resolved versions would turn every Python 3.9 job red, and this change
declines both.** `.github/workflows/test.yml` runs Python 3.9 through 3.13 on Linux and
3.12 on macOS, and `FR-foundation-13` states that matrix.

| Distribution | Newest release | Its `requires_python` | Pinned |
|---|---|---|---|
| `pytest` | 9.1.1 | `>=3.10` | 8.4.2 |
| `pytest-cov` | 7.1.0 | `>=3.9` | 7.1.0 |
| `build` | 1.5.0 | `>=3.10` | 1.4.4 |

**How the check was made.** A script read the PyPI JSON interface for every release of
each distribution. For each release it read the `requires_python` marker of every wheel,
and it kept the release only when one wheel accepts each of 3.9, 3.10, 3.11, 3.12 and
3.13. It reported `pytest` 8.4.2, `pytest-cov` 7.1.0 and `build` 1.4.4 as the newest such
releases. The read used the distribution's own metadata and no recollection.

Verified against: https://pypi.org/pypi/pytest/9.1.1/json (retrieved 2026-08-10)
Verified against: https://pypi.org/pypi/build/1.5.0/json (retrieved 2026-08-10)
Verified against: https://pypi.org/pypi/pytest-cov/7.1.0/json (retrieved 2026-08-10)

## `mypy` floats

A pinned type checker falls behind, and a frozen release hides a defect that a later
release reports. A floating formatter turns a green tree red, which is noisy and safe. A
frozen type checker keeps a red tree green, which is quiet and unsafe. The comment beside
the entry states that reading and cites #446.

## Nothing read `requirements.txt`, and two documents named it

A read of every tracked file found no reader and two instructions.

- No file under `.github/workflows/` reads it. Every job installs `pip install -e ".[dev]"`.
- No `Makefile`, no `tox.ini`, no `Dockerfile` and no `.pre-commit-config.yaml` exists.
- `README.md` names it nowhere, and this repository holds no `CONTRIBUTING` file.
- The `ruff` comment of `pyproject.toml` named it. **Repaired in this change.**
- The note under `FR-foundation-8c` of `docs/specs/features/00-foundation.md` named it.
  **Repaired in this change.**
- `CHANGELOG.md` and `docs/specs/spec.md` name it inside a record of an earlier round. No
  reader installs from a record, so both stand as they read.

`test_no_instruction_names_the_deleted_dependency_record` now holds that state.

## The cases

`tests/test_lint_gate_pin.py` widens from 6 cases to 30, which the Notes of #446 asked
for. The file keeps its name, because `pyproject.toml`,
`docs/specs/features/00-foundation.md` and `CHANGELOG.md` each name that path.

**The tests came first.** The unrepaired tree failed 18 of the 30, among them
`AssertionError: the dev extra states no exact version for pytest: 'pytest>=7.0'`.

**Eight mutations prove the cases discriminate, and each one isolates to the case it
targets.** The list stands in `CHANGELOG.md`. Each mutation restored from a snapshot of
the implemented tree, and the restored form passed 30 of 30 every time.

**The reader reads dependency entries and never every quoted substring.** `_dev_entries`
strips the comment lines of the list first, so the version each new comment quotes moves
no case. #452 records that defect and this change does not reproduce it.

**Two floors fail a reader that collected nothing.** `MINIMUM_DEV_ENTRIES` holds the extra
at five entries and `MINIMUM_INSTRUCTION_FILES` holds the instruction set at ten files.

**One statement reaches no case, and the file states that.** No case reads whether a
pinned version installs on Python 3.9, because that read needs the package index and this
suite opens no network connection.

## Local gates

The worktree holds a virtual environment on Python 3.14.3, installed from
`pip install -e ".[dev]"` after the pin.

| Gate | Result |
|---|---|
| `pytest tests/ -m "not spec_validation" --cov=ja4plus` | 3506 passed, 4 skipped, 8 xfailed, 94% |
| `pytest tests/ -m installed_wheel` | 23 passed |
| `ruff format --check ja4plus/ tests/` | 201 files already formatted |

Coverage reads 94% before the change and 94% after it. No file under `ja4plus/` changes
and no fingerprint moves.

`FR-foundation-8d` through `FR-foundation-8g` carry the four new requirements, and the
`Terms` table gains `dependency record`. The round entry reads `TBD` in both records.

Refs #446


Crank-Git:
Self-review, four lenses over the diff.

**Stale prose.** One finding, repaired in c151a61. The docstring of `tests/test_lint_gate_pin.py` named `build>=1.0`, and the entry now reads `build==1.4.4`. The same paragraph now records that #446 widened the hazard it states: every entry carries a comment, and three of those comments quote a version.

**A second reader of `requirements.txt`.** No finding beyond the two this change already repairs. A read of every tracked file returns `CHANGELOG.md`, `docs/specs/spec.md` and the search constants of `tests/test_lint_gate_pin.py`. The first two are records, and no reader installs from a record.

**A case that cannot fail.** No finding. Every new case carries a floor or a fixed parameter list. `_dev_entries()` returns 5 entries against the floor of 5, and `_instruction_files()` returns 55 files against the floor of 10. Eight mutations fail the cases they target and no other case.

**A second place that states a version.** No finding. No file under `.github/workflows/`, `README.md`, `mkdocs.yml` or `CLAUDE.md` names `build`, `mypy`, `pytest`, `pytest-cov` or `ruff` with a version specifier. This repository holds no `setup.cfg`, no `tox.ini` and no `.pre-commit-config.yaml`.

**One reading a reviewer has to take from the environment.** `test_the_installed_release_holds_the_pinned_version` fails in an environment that installed the extra before this change. Run `pip install -e ".[dev]"` again and it passes. That is the drift the case exists to report.

PR #500, opened 2026-08-10
Read the dependency entries rather than every quoted substring

Crank-Git:
## What

`tests/dependency_entries.py` is the one reader of a dependency block of `pyproject.toml`.
`_dependency_block` of `tests/test_documentation_site.py` calls it, and `_dev_lines` and
`_dev_entries` of `tests/test_lint_gate_pin.py` call it too. No caller parses a block of its
own.

## Why

`_dependency_block` collected every double-quoted substring of a dependency block, so a
comment inside the block read as an entry. **A reader that is correct only because of what
its callers happen to read is the defect.**

The premise of #452 was re-measured against the current tree and it holds. The unrepaired
reader returns this list for the `dev` extra:

```
['pytest==8.4.2', 'pytest-cov==7.1.0', 'ruff==0.16.2', 'mypy>=1.11', 'not spec_validation', 'build==1.4.4']
```

`not spec_validation` is a fragment of the command `pytest tests/ -m "not spec_validation"`,
which stands inside the comment beside `build==1.4.4`. #446 widened the hazard, because
every entry of the extra now carries a comment above it, and it answered the hazard with a
second reader inside `tests/test_lint_gate_pin.py`. Two readers of one block can disagree, so
this change removes the second one.

**The module reads lines and it reads no TOML.** `tomllib` reaches the standard library at
Python 3.11, `pyproject.toml` states `requires-python = ">=3.9"`, and `FR-foundation-13`
runs the matrix from Python 3.9, so an import of `tomllib` fails two jobs of that matrix.
The `dev` extra names no `tomli`. A comment starts at a `#` that stands outside a string,
which is the rule TOML states, so one scan drops a comment line and a comment after an
entry.

**The reader refuses a block that holds no entry.** An aggregate over an empty set passes,
so the floor sits in the reader rather than in each caller.

## How this was tested

**The tests came first.** Against the unrepaired tree the new case failed:

```
tests/test_documentation_site.py::test_the_reader_returns_the_entries_of_the_dev_extra_and_no_comment_fragment
AssertionError: assert ['pytest==8.4...build==1.4.4'] == ['pytest==8.4...build==1.4.4']
  At index 4 diff: 'not spec_validation' != 'build==1.4.4'
  Left contains one more item: 'build==1.4.4'
```

Three cases of `tests/test_documentation_site.py` hold the `dev` extra, the runtime block and
the `docs` extra by name and in file order. **Each case names the entries it expects**,
because a repair that returns an empty list passes a case that reads the absence of the
comment fragment alone. New file `tests/test_dependency_entries.py` holds eleven cases
against the shapes `pyproject.toml` carries nowhere today: a comment after an entry, a `#`
inside an entry, an entry that a literal string holds, an escaped quote, a block of
comments alone, a renamed block and an unclosed block.

**Five mutations prove the cases discriminate, and each one isolates to the cases it
targets.**

| Mutation | Result |
|---|---|
| `_without_comment` returns the line unchanged | 4 failed, among them the `dev` extra case and `test_the_decision_covers_every_entry_of_the_dev_extra` |
| `_without_comment` cuts at the first `#` | 1 failed, the fragment-marker case, at `AssertionError: the 'dev = [' block holds no entry` |
| `dependency_entries` drops the floor | 1 failed, the empty-block case, at `Failed: DID NOT RAISE <class 'AssertionError'>` |
| `QUOTED` reads the double-quoted form alone | 1 failed, the literal-string case |
| `_without_comment` drops the open-string guard | 1 failed, the escape case |

Each mutation restored from a snapshot of the implemented tree, and the restored form passes
58 of 58 every time.

The five gates ran in a virtual environment of this worktree.

```
pytest tests/ -m "not spec_validation"   3845 passed, 4 skipped, 8 xfailed; coverage 94%
ruff format --check ja4plus/ tests/      206 files already formatted
```

**No file under `ja4plus/` changes and no fingerprint moves.** Coverage reads 94%, and the
package holds the code it held before.

## Records

`CHANGELOG.md` and the `## Changelog` table of `docs/specs/spec.md` each carry the entry of
this round, and each one reads `TBD` in place of the round number.
`docs/specs/features/00-foundation.md` names the module under `FR-foundation-8g`, and it
warns a later writer against a second reader. The `## Terms` table of `docs/specs/spec.md`
gains `dependency block`, because that table already rejects `dependency list` as a synonym
of `dependency record`.

Verified against https://toml.io/en/v1.0.0 (TOML 1.0.0, comment rule, retrieved 2026-08-10)
and https://docs.python.org/3/library/tomllib.html (`tomllib`, new in 3.11, retrieved
2026-08-10).

Refs #452


Crank-Git:
## Self-review

Two reviewers read the diff: one for correctness, one for the writing standard and the
records. Every finding below is repaired in `8babdc9`, except where the entry states a
decline.

### Correctness

1. **The reader read the double-quoted form alone.** TOML holds a literal string in single
   quotes, and an entry of that form would have disappeared from the returned list with no
   offender reported. That is the same shape as the defect this issue repairs: a reader
   that is correct only against the entries that exist today. `QUOTED` now closes a string
   on the quote that opened it, and `_without_comment` holds that quote too, so the
   apostrophe of an environment marker closes no basic string. **Repaired**, with two
   cases: `test_the_reader_returns_an_entry_that_a_literal_string_holds` and
   `test_the_reader_holds_an_entry_whose_marker_quotes_a_version`.

2. **An escaped quote inside a basic string produced half of an entry.** `"foo\"bar"`
   returned `foo\`. The reader reads no escape sequence, and a silent half of an entry is
   the worse result, so `_without_comment` now raises where a line ends inside a string.
   **Repaired**, with `test_the_reader_refuses_a_line_that_holds_an_escaped_quote`.

3. **No case passes in one direction alone.** Every case asserts list equality against
   named entries, so an empty result and a truncated result both fail. **No repair
   needed.**

4. **Every caller was found and covered.** `_dependency_block` and the `_dev_lines` and
   `_dev_entries` of `tests/test_lint_gate_pin.py` are the whole set, and the shared reader
   reproduces the earlier output for all three real blocks. **No repair needed.**

5. **The new floor breaks no caller.** All three blocks of `pyproject.toml` hold entries
   today. **No repair needed.**

### The writing standard and the records

6. **`dependency list` is a rejected word.** The `## Terms` table rejects it as a synonym
   of `dependency record`, which names a tracked file rather than a bracketed array. The
   prose of this change used it for the array. The table now holds `dependency block` for
   the array, and every new sentence takes that term. **Repaired.**

7. **Ten sentences ran past the 25-word limit** in files the exemption of
   `.claude/rules/ste.md` does not reach. The module docstring, the case file, the two
   caller files and the feature page each carry the split form now, and one sentence of
   three items became a vertical list. **Repaired.**

8. **One count was wrong.** The prose read that three comments of the `dev` extra quote a
   version. Four do: the comments beside `pytest`, `pytest-cov`, `ruff` and `build`. The
   `mypy` comment names the major line `2.x` and no release. **Repaired** in
   `tests/test_lint_gate_pin.py` and in both records.

9. **One claim was too wide.** The record read that `pyproject.toml` carries no comment
   after an entry, and `[tool.ruff.lint]` carries one. The record now scopes the claim to
   the dependency blocks the reader reads. **Repaired.**

### The mutation set after the repairs

| Mutation | Result |
|---|---|
| `_without_comment` returns the line unchanged | 4 failed, among them the `dev` extra case and `test_the_decision_covers_every_entry_of_the_dev_extra` |
| `_without_comment` cuts at the first `#` | 1 failed, the fragment-marker case, at `AssertionError: the 'dev = [' block holds no entry` |
| `dependency_entries` drops the floor | 1 failed, the empty-block case, at `Failed: DID NOT RAISE <class 'AssertionError'>` |
| `QUOTED` reads the double-quoted form alone | 1 failed, the literal-string case |
| `_without_comment` drops the open-string guard | 1 failed, the escape case |

Each mutation restored from a snapshot of the implemented tree, and the restored form
passes 58 of 58 every time.

Issue #423, opened 2026-08-09
Report the dropped-packet count from a macOS capture socket


Crank-Git:
## Split from #326 on 2026-08-09

**#326 held one criterion that names macOS and one Linux measurement that had never run.
The user ruled on 2026-08-09 that the Linux half runs on the granted Linux host, bounded
to the loopback interface. This issue holds the macOS half.**

> A monitor on macOS reports a whole number in the `dropped` field.

**Criterion 1 of #326 moved here rather than being waived.** It is not met by an injected
socket. **This project has recorded sixteen instances of a comparison that is never made
reading as a comparison that passes**, and marking this met against a fake socket would be
the seventeenth.

## What blocks it

**The capture device of the development host cannot be opened by this account.**

```
ls -l /dev/bpf0   ->   crw-------  (mode 600, owner root)
id -Gn            ->   the account holds no access_bpf
os.open('/dev/bpf0', os.O_RDONLY)  ->  [Errno 13] Permission denied
```

**The group `access_bpf` does not exist on this host.** It is created by the ChmodBPF
helper of Wireshark, which is not installed. `dscl . -read /Groups/access_bpf` answers
`eDSRecordNotFound`, and `/Library/LaunchDaemons/` holds no ChmodBPF entry. **An earlier
comment on #326 told a worker to add the account to that group, and the group was never
confirmed to exist.**

**The user ruled on 2026-08-09 that the grant is made.** Two routes are recorded on #326.
**Neither is applied by a worker or by the project manager**: it changes the host and it
widens what every process of this account may read.

## What unblocks it

```
python3 -c "import os; os.close(os.open('/dev/bpf0', os.O_RDONLY)); print('readable')"
```

**On that reading this issue becomes `status:ready`.** A worker then opens the capture
socket, passes it to `sniff` as `opened_socket`, passes its drop count as
`dropped_source`, and takes the measurement inside the worker rather than by paste.

## Acceptance criteria

- [ ] A monitor on macOS reports a whole number in the `dropped` field, measured from a
      real capture socket of this host.
- [ ] The measurement is taken from `_L2bpfSocket.get_stats`, which reads the `BIOCGSTATS`
      ioctl, and not from an injected object.
- [ ] The pull request states the macOS version and the `scapy` version that produced the
      measurement.

## Links

- Split from #326, which holds the build and the Linux measurement.
- Feature: [`docs/specs/features/06-live-capture.md`](docs/specs/features/06-live-capture.md)

[...]

PR #503, opened 2026-08-10
Report the drop count of the macOS capture socket

Crank-Git:
## What this changes

`ja4plus watch` reports the drop count of the capture socket it opened. The `dropped`
field of the statistics line read `null` on every host before this change: #55 built the
field, `MonitorStats` accepted a `dropped_source`, and `ja4plus/cli.py` passed none.

- `ja4plus/watch.py` gains `capture_drop_count` and `CaptureDropCount`.
- `read_interface` gains a `drop_count` argument. It attaches the socket it opened, and
  it reads the count once more before it closes the socket.
- `ja4plus/cli.py` builds the drop count and passes it to `Monitor` and to
  `read_interface`.

## Why

On macOS the capture socket reports a drop count and `sniff` hides the socket.
`_L2bpfSocket.get_stats` reads the `BIOCGSTATS` ioctl at
`scapy/arch/bpf/supersocket.py:297`. `AsyncSniffer._run` holds the socket it opens in a
local name at `scapy/sendrecv.py:1205`, so a caller reaches that socket through the
`opened_socket` argument alone. Both readings were taken again against the installed
`scapy` rather than trusted from the issue.

On Linux the capture socket of `scapy` 2.7.0 holds no `get_stats`, so the field still
reads `null` there. #326 owns the Linux half.

## The environment of the measurement

| Reading | Value |
|---|---|
| Platform | macOS 26.6.1, build 25G76, arm64 |
| `scapy` | 2.7.0 |
| Python | 3.14.3 |
| Date | 2026-08-10 |

## The capture grant, measured at the moment of the run

```
/dev/bpf0 READABLE
/dev/bpf1 READABLE
/dev/bpf2 READABLE
/dev/bpf3 READABLE
/dev/bpf4 ERRNO 13 EACCES Permission denied
/dev/bpf5 ERRNO 2 ENOENT No such file or directory
```

**A `bpf` device opens exclusively, so this run measured the grant rather than assumed
it.** The live cases opened `/dev/bpf0`.

## How this is tested

New file `tests/test_watch_drop_count.py` holds 18 cases. The cases came first, and the
file failed to collect against the base commit `cf77598` with
`ImportError: cannot import name 'CaptureDropCount' from 'ja4plus.watch'`.

**Two cases read a real capture socket of this host, and #423 bars an injected one.**
Both capture on the loopback interface and both generate every packet they read.

1. `test_a_monitor_reports_a_whole_number_in_the_dropped_field` runs the whole monitor
   through `read_interface` on `lo0`. It read 4 packets of one TCP connection this run
   opened, over an `L2bpfListenSocket`, and reported `dropped=0`.
2. `test_the_dropped_field_reads_a_count_above_zero` opens a real capture socket with a
   4096-byte kernel buffer, sends 50 UDP packets of 1400 bytes to the loopback address,
   and reads `received=55 dropped=46`.

**A count of 0 on a clean capture proves nothing**, because a field no code writes reads
the same way. Case 2 exists for that reason.

**Where no `/dev/bpf` node opens, both live cases skip and neither one passes.** A probe
that refused the device reported
`this host opened no capture socket on 'lo0': [Errno 13] Permission denied`.

## Mutations

Each mutation ran live and each one was restored.

| Mutation | Result |
|---|---|
| `capture_drop_count` reports no count | 10 cases fail |
| `read_interface` attaches no socket | 3 cases fail |
| `read_interface` reads no count before the close | 3 cases fail |
| The command passes no drop count | 2 cases fail |
| `CaptureDropCount` holds no last count | 4 cases fail |
| The drop case runs with the whole 65535-byte buffer | 1 case fails, `AssertionError: 0 not greater than 0` |
| The drop case sends one packet | 1 case fails, `AssertionError: 0 not greater than 0` |

## Gates

| Gate | Result |
|---|---|
| `pytest tests/ -m "not spec_validation"` | 3847 passed, 4 skipped, 8 xfailed |
| `ruff format --check ja4plus/ tests/` | 205 files already formatted |
| Coverage | 94%, misses hold at 273 while statements rise from 4253 to 4281 |

**No file under `ja4plus/fingerprinters/` changes and no fingerprint moves.** The
conformance suite reports the same three counts on the base commit `cf77598` and on this
branch.

## Records

`CHANGELOG.md` and the `## Changelog` table of `docs/specs/spec.md` each hold one entry
that reads `TBD`. The project manager assigns the round number at the batch gate.

(`scapy` 2.7.0, read from the installed package on 2026-08-10)

Refs #423


Crank-Git:
## The command-level reading, taken twice on one host

**The criterion of #423 asks a monitor on macOS to report a whole number in the `dropped`
field.** The two runs below are `ja4plus watch` itself, not a test double. Each one
captured on the loopback interface, under the filter `tcp port 18429`, and read three TCP
connections this run opened.

The base commit `cf77598`:

```
[ja4plus] packets=32 fingerprints=18 connections=3 evicted=0 dropped=null uptime=4s
```

This branch:

```
[ja4plus] packets=32 fingerprints=18 connections=3 evicted=0 dropped=0 uptime=4s
```

Both lines are the exit summary, which `write_statistics` writes after `read_interface`
closed the capture socket. The reading therefore also proves that `read_interface` reads
the count before the close.

The command line, for a reader who wants to repeat it:

```
ja4plus watch lo0 --bpf "tcp port 18429" --stats-interval 1
```

The run opened `/dev/bpf0`, on macOS 26.6.1 build 25G76, against `scapy` 2.7.0 and Python
3.14.3, on 2026-08-10.

Crank-Git:
## Self-review, by two lenses

Two reviews ran over the diff: one for correctness and safety, one against
`.claude/rules/ste.md`. Commit `841028c` carries the repairs.

### A race between the statistics thread and the close, and it is now closed

The correctness review found a defect this change introduced.

`report_statistics` stops the statistics thread after `read_interface` returns, and
`read_interface` closes the capture socket inside its own `finally`. The statistics
thread therefore reads the socket while the capture closes it. `SuperSocket.close` at
`scapy/arch/bpf/supersocket.py:324` calls `os.close`, then sets `closed`, then clears the
descriptor, and those are three statements. **A reader between the first two reads a
released file descriptor**, which the kernel may already have given to the next file this
process opened. The `closed` reading of `capture_drop_count` narrows that window and it
does not close it.

`CaptureDropCount` now holds one lock. `release` holds it over the last read and over the
drop of the socket, and `read_interface` calls `release` before `close`. No read is in
flight at the close.

### Two of the three new mutations passed against the first form of the cases

**That is the fault this project records sixteen times, and it appeared inside the guard
written to stop it.** Both cases were repaired until each mutation failed one.

| Mutation | First form | Repaired form |
|---|---|---|
| The release acquires no lock | 22 passed | 1 fails, `AssertionError: Lists differ: [4] != []` |
| The capture refreshes rather than releases | 22 passed | 1 fails, `AssertionError: 9 != 5` |
| The capture closes the socket before it releases it | 4 fail | 4 fail |

The first case made every read of the socket wait, so the release blocked inside the
ioctl rather than on the lock. Only the first read waits now. The second case read a
closed socket, which parts a release from a refresh nowhere, so it reopens the socket
under another count and models the reused file descriptor.

### Two more findings, and what this change does about each

- **An exception from `get_stats` inside the `finally` would leak the capture socket**,
  because it would abort the clause before `close` ran. The release now sits in its own
  `try`, and `close` runs in the `finally` of that.
- **A fixed sleep before the drop reading is a timing assumption.** The drop case now
  reads the count again until it rises or five seconds pass, so a loaded host fails no
  case for that reason. A burst that dropped nothing still fails the assertion.

### The writing standard

The prose review found four metaphors, two sentences past the 25-word limit, and a
rotation between `the count of dropped packets` and `the drop count`. The `## Terms`
table holds `drop count`, so every added line now reads that term. `holds` is the one
verb for the value a field carries.

### Gates after the repairs

| Gate | Result |
|---|---|
| `pytest tests/ -m "not spec_validation"` | 3851 passed, 4 skipped, 8 xfailed |
| `ruff format --check ja4plus/ tests/` | 205 files already formatted |
| Coverage | 94%, misses hold at 273 while statements rise from 4253 to 4292 |

Issue #67, opened 2026-08-06
Put the version in one place and gate it against the changelog


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/09-release.md`](docs/specs/features/09-release.md):

> FR-release-1 — The version number appears in one place in the repository.

> FR-release-2 — Continuous integration fails when the version in `pyproject.toml` does not match the version in `ja4plus/__init__.py`.

> FR-release-3 — Continuous integration fails when the version has no matching section in `CHANGELOG.md`.

## What to build

The version appears in `pyproject.toml` at line 7 and in `ja4plus/__init__.py` at line 71. Keep it in `pyproject.toml` and have `__init__.py` read it with `importlib.metadata.version`, so the two cannot disagree.

The changelog check looks for a `## [<version>]` section in `CHANGELOG.md` and fails when it is absent.

## Acceptance criteria

- [ ] `grep -rn "0\.6\.0" pyproject.toml ja4plus/` finds the version in one file only.
- [ ] A pull request that changes the version without changing the changelog fails continuous integration.

## Links

- Feature: [`docs/specs/features/09-release.md`](docs/specs/features/09-release.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #20

Crank-Git:
## Carried forward from #24

`CHANGELOG.md` has no `## [Unreleased]` section. Epic 0 has shipped #21, #22, #24,
#25 and #26 without a changelog entry, on purpose: a partial entry written per issue
would have to be rewritten here.

This issue owns opening the `## [Unreleased]` section and recording the Epic 0 work
in one pass, reconstructed from the commit history on `epic/11-foundation`. The
changelog gate this issue builds then applies from that point on.

Crank-Git:
## A measurement #456 made against the first acceptance criterion

#456 asked whether `docs/specs/features/09-release.md` states a stale count or the end state
this issue builds. **It is the end state, so nothing here changes and the criterion keeps its
number.** #456 records that reading and adds a case that holds the two apart.

**One finding belongs to this issue.** The command counts files and not declarations, and the
two differ. A read of `git ls-files pyproject.toml ja4plus/` on 2026-08-09 reports seven files
that hold the text `0.6.0`.

[...]

Issue #20, opened 2026-08-06
Epic 9: Release


Crank-Git:
## Goal

Publish version 1.0.0.

## Exit criteria

Version 1.0.0 is on PyPI. The wheel installs into a clean environment. The GitHub release carries the changelog entry.

## Spec

- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Feature: [`docs/specs/features/09-release.md`](docs/specs/features/09-release.md)

Depends on: #19

## Sub-issues

- [x] #67
- [x] #68
- [x] #69
- [x] #70

Depends on: #406. **Version 1.0.0 does not publish before Epic 11 closes.** The
user decided on 2026-08-09 that the evidence lands before the release.

Crank-Git:
**Unblocked. #19, Epic 8, closed on 2026-08-09 and merged into `dev` as `2702d10`.**

**The exit criterion of this epic is not the project manager's to satisfy, and the next one must not read it as though it were.** It reads *"Version 1.0.0 is on PyPI"*.

**The user has not authorized a publish, and #70 records that they are not publishing version 1.0.0 yet.** So this epic builds the release machinery and stops:

- **#67, #68 and #69 are buildable now.** They put the version in one place and gate it, verify the built wheel in a clean environment, and check the wheel contents.
- **#70 is `status:needs-feedback` and stays that way** until the user rules on the TestPyPI dry run and the release body.
- **The upload to PyPI is a separate step the user approves.** Nothing in this epic performs it, and `publish.yml` triggers on a published release only, so a merge to `dev` publishes nothing.

**Do not close this epic on its stated exit criterion.** When #67, #68 and #69 land, the machinery is ready and the epic waits on the user, not on work.

Crank-Git:
Every sub-issue of this epic is closed. #70 landed at round 191 in batch #510, and the release gates of #67, #68 and #69 landed in earlier batches. The release path now holds a manual dry run that reaches TestPyPI alone.

**One measurement of this epic stays with the user.** No dry run has run, because a run of the publish workflow publishes. The guardrail is proven structurally and the first dry run is the measurement.

PR #506, opened 2026-08-10
Declare the version in one file and gate it against the changelog

Crank-Git:
Closes nothing on its own. Part of #20, and it builds #67.

## What this changes

`pyproject.toml` declared `version = "0.6.0"` at line 7 and `ja4plus/__init__.py`
declared it again at line 94. Two records could state two versions.

`ja4plus/__init__.py` is now the one declaration. `pyproject.toml` names `version` in its
`dynamic` list and reads the value from `ja4plus.__version__`, so a build states the
version the package states.

## Why the shape differs from the feature page

`docs/specs/features/09-release.md` proposed that `__init__.py` read the version with
`importlib.metadata.version`. **That call reads the metadata of an installed
distribution, and it does not read the source tree.** Three readings of 2026-08-10 state
the reason.

1. A checkout that carries no install raises
   `PackageNotFoundError: No package metadata was found for ja4plus`. A reader therefore
   needs a literal fallback, and a fallback is the second declaration `FR-release-1` bars.
2. An editable install freezes its metadata at install time. With `pyproject.toml`
   changed to `0.7.0` and no reinstall, `importlib.metadata.version("ja4plus")` returned
   `0.6.0`.
3. That metadata resolves through `ja4plus.egg-info` of the working directory, so the
   answer follows the directory a command runs in. The system interpreter holds no
   `scapy`, so `import ja4plus` raises `ModuleNotFoundError` there, and
   `importlib.metadata.version("ja4plus")` still returned `0.6.0`.

**A gate that reads the version out of an install compares a value against itself**,
because continuous integration installs the project from `pyproject.toml`. `FR-release-2`
would then name a comparison that reports nothing. The two gates read tracked text
instead, and the two texts are independent inputs.

`FR-release-1` and `FR-release-2` pull against each other, and the `dynamic` shape
reconciles them. One declaration exists, so no second value can contradict it, and the
gate holds the structure that keeps it so: `pyproject.toml` must state either a static
version equal to the package version, or a `dynamic` read of `ja4plus.__version__`. It
passes on neither absence.

## How this is tested

`tests/version_gate.py` holds the readers and `tests/test_version_gate.py` holds fourteen
cases. **A comparison that is never made reads as a comparison that passes**, so each gate
carries control cases that feed it a contradicting pair and read the message it returns.
A reader that finds no version returns a message rather than agreement.

The four gates were also proven end to end against the real files.

- `__version__` at `0.7.0` with no changelog section:
  ``AssertionError: `CHANGELOG.md` holds no `## [0.7.0]` section, and it holds ['1.0.0', '0.6.0']``
- `pyproject.toml` given a second, contradicting declaration:
  ``AssertionError: `pyproject.toml` declares 0.9.9 and `ja4plus/__init__.py` declares 0.6.0``,
  `AssertionError: these tracked files declare a version, and one is right: ['ja4plus/__init__.py', 'pyproject.toml']`,
  and `AssertionError: these criteria state a count the repository contradicts: ['09-release.md states 1 and the repository holds 2']`

`tests/test_installed_wheel.py` reads the version from the package rather than from
`pyproject.toml`, and a new case reads the version the build resolved into the wheel
metadata. That case fails where `[tool.setuptools.dynamic]` reads the wrong attribute.

## The acceptance criterion #456 handed here

#456 measured seven files that hold the text `0.6.0` and it recorded that its command
counts files and not declarations. Five of the seven state in prose what version 0.6.0
did, a reader of `docs/migration-0.6-to-1.0.md` needs that prose, and a count of the text
can therefore never reach one. **The criterion now counts the declaration.**

A read of 2026-08-10 reports **one** declaration file and **six** files that hold the
text. The closing quote is part of the command, because `[tool.setuptools.dynamic]` writes
`version = {attr = ...}` and that line declares no version. This branch wrote the pattern
without the quote first, and the re-measurement caught it.

`MEASURED_CRITERIA` now holds no pending criterion, so its two pending cases aggregate
over an empty set. `PENDING_CONTROL` feeds the pending reader a criterion whose end state
this tree holds.

## Local checks

This is a batch member, so provider continuous integration never runs on it and the full
local suite replaces it.

- `pytest tests/ -m "not spec_validation"` — 3917 passed, 4 skipped, 8 xfailed
- `pytest tests/ -m spec_validation` — 1532 passed, 143 skipped, 134 xfailed
- `pytest tests/test_installed_wheel.py -m installed_wheel` — 24 passed
- `ruff format --check ja4plus/ tests/` — 209 files already formatted
- `mypy --strict ja4plus/` — Success, no issues in 31 source files
- Line coverage 94% against the floor of 70%

**A self-review found one silent pass and it is repaired on this branch.** `project_version` searched the whole file for a `version = "..."` line, so a `[tool]` table that carried the package version beside a `dynamic` list naming the wrong attribute read as agreement. Each reader searches one table now, and `test_the_gate_reads_the_version_of_the_project_table_alone` holds it against that input.

`python -m build` produced `ja4plus-0.6.0-py3-none-any.whl` and `ja4plus-0.6.0.tar.gz` in
an isolated build environment that holds neither `scapy` nor `cryptography`, which proves
`setuptools` resolves the attribute from the syntax tree and imports no module.

Verified against: https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata (retrieved 2026-08-10)
Verified against: https://docs.python.org/3/library/importlib.metadata.html (retrieved 2026-08-10)


Crank-Git:

One real finding, and it is the failure mode this project punishes: **a gate that reports nothing on a real disagreement.**

**REAL — `project_version` searched the whole file and not the `[project]` table.** `PROJECT_DECLARATION` matched `^version = "..."` anywhere in `pyproject.toml`. A `[tool]` table that carried a `version` key equal to the package version, beside a `dynamic` list that named the wrong attribute, therefore read as agreement.

Reproduced against the reader as it stood:

```
old reader static project version: 0.6.0
old reader verdict: AGREEMENT (silent pass)
```

on this input:

```toml
[project]
name = "ja4plus"
dynamic = ["version"]

[tool.setuptools.dynamic]
version = {attr = "ja4plus.cli.__version__"}

[tool.other]
version = "0.6.0"
```

The wrong attribute reached no report. `project_version` and `project_version_attribute` each search one table now, and `test_the_gate_reads_the_version_of_the_project_table_alone` holds the reader against that exact input.

**`declaration_files` still searches no single table, and that is deliberate.** It counts a `version` key of any table, so a second key fails the one-file case of `FR-release-1` rather than pass it. The two readers together close the hole: the count case reports the extra line, and the scoped reader no longer mistakes it for the project version.

### Findings rejected, with the reason

- **A single-quoted TOML string matches no pattern.** `version = {attr = 'ja4plus.__version__'}` would read as "declares no version". That is a loud failure and not a silent pass, so the direction is safe. The tracked file uses double quotes throughout.
- **A CRLF line ending matches no pattern.** `package_version` then returns `None` and the gate reports "`ja4plus/__init__.py` declares no `__version__`". Loud again. Every one of the three files is LF today, and the test matrix runs Linux and macOS.
- **`wheel_metadata_version` reads `^Version: ` with `re.search`.** Wheel `METADATA` is RFC 822 shaped, so every header precedes the blank line that opens the description. The first match is therefore always the header.
- **Python 3.9.** `tests/version_gate.py` annotates with `typing.List` and `typing.Optional` and uses no 3.10 syntax.

Crank-Git:
## Self-review, writing standard lens

`.claude/rules/ste.md` covers every comment, docstring, test name and page this branch adds. The exemption reaches the entries of `CHANGELOG.md` and the `## Changelog` rows of `docs/specs/spec.md`, and it reaches rule 1 alone.

**Four sentences ran past the 25-word limit, and each one is now two sentences.**

- `docs/specs/features/09-release.md` — 29 words, the `git ls-files` sentence.
- `tests/version_gate.py` module docstring — 26 words.
- `version_disagreement` docstring — 28 words.
- `declaration_files` docstring — 27 words.

**Rule 7 asks for one concept and one word, and this branch rotated `declares` with `states` for one fact.** The fact is that `ja4plus/__init__.py` holds the version value. `declares` is the word, in three places that had `states`.

- `CHANGELOG.md`: "so a build resolves the version the package declares".
- `ja4plus/__init__.py`: "so a release carries the version this line declares".
- `tests/version_gate.py`: "`ja4plus/__init__.py` declares it at the top level of the module".

One docstring wrote `derives none` where the branch says `reads` everywhere else, and it now reads `reads none`.

**Clean on the rest.** No passive voice, no `-ing` noun or heading, no noun cluster past three words, every docstring opens with the result, every test name reads as one behaviour. `the statistics thread` and the count of fingerprinters appear nowhere in this diff.

Issue #455, opened 2026-08-09
Exclude the documentation tree from the built wheel


Crank-Git:
## What was measured

**`docs/specs/features/09-release.md:156` requires the built wheel to carry no file under
`docs/`. The wheel carries 56 of them.**

Measured by #393 on 2026-08-09 with `zipfile` against a `python -m build` on the batch #445
base: **96 entries in the wheel, 56 of them under `docs/`.** `tests/` and `examples/` are
correctly absent, so the packaging configuration excludes some trees and not this one.

**This is a defect and not a stale count.** The acceptance criterion states the intent
correctly, and the artefact disagrees with it.

## Why it is not left to #69

#393 filed nothing and named #69 as the owner, which was a reasonable reading. **The project
manager rules otherwise: #69 is held behind a user gate.** #69 sets
`Development Status :: 5 - Production/Stable`, and the user ruled on 2026-08-09 that #69
stops and asks again. **A packaging defect cannot wait behind a classifier decision**, and an
issue that cannot be scheduled cannot own a repair.

So #69 keeps the classifier question, and this issue owns the wheel contents.

## What to build

**Read what the wheel holds before you change anything**, and record the entry list. Then
exclude `docs/` from the wheel by the same mechanism that already excludes `tests/` and
`examples/` — read `pyproject.toml` and find why those two are absent rather than adding a
second mechanism beside it.

**A case must fail on a wheel that carries a file under `docs/`.** `tests/` already holds
`installed_wheel` cases that build and read the artefact, and the `installed-wheel` job of
`.github/workflows/test.yml` is their only runner. **A case that reads the source tree
instead of the artefact does not test this**, because the defect is in what the build
includes.

**State what the wheel must carry.** The bundled `ja4plus/data/ja4plus-mapping.csv` is a
data file the package needs at run time, and a repair that excludes it breaks the lookup.
Prove the mapping file is still in the wheel after the change.

**The source distribution is a separate question and it is asked here.** State whether the
source distribution should carry `docs/`, and prove what it carries now. A source
distribution that carries documentation is normal; a wheel that carries it is not.

## Acceptance criteria

- [ ] The built wheel carries no file under `docs/`.
- [ ] The built wheel still carries `ja4plus/data/ja4plus-mapping.csv`, and a case proves it.
- [ ] A case fails when the wheel carries a file under `docs/`, and the pull request holds
      that failing output.
- [ ] The pull request records the entry count of the wheel before and after.
[...]

Issue #68, opened 2026-08-06
Verify the built wheel in a clean environment before publishing


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/09-release.md`](docs/specs/features/09-release.md):

> FR-release-4 — The publish workflow builds a source distribution and a wheel.

> FR-release-5 — The publish workflow installs the built wheel into a clean environment.

> FR-release-6 — The publish workflow runs the command-line program from that installation.

> FR-release-7 — The publish workflow runs the conformance suite against that installation.

> FR-release-8 — The publish workflow publishes only when every check passes.

> FR-release-9 — The publish workflow runs `twine check` on the built files.

## What to build

A publish to PyPI cannot be undone, so every check runs before the publish, not after.

The clean environment must not have the source tree on its path, or an import resolves to the working copy and the check proves nothing.

`.github/workflows/publish.yml` already triggers on a published release and uses trusted publishing with `id-token: write`, so no token is stored. Keep that.

Verified against https://docs.pypi.org/trusted-publishers/ (retrieved 2026-08-06).

## Acceptance criteria

- [ ] The publish workflow builds a source distribution and a wheel.
- [ ] `twine check dist/*` reports `PASSED` for both files.
- [ ] The workflow installs the wheel into a clean environment and runs `ja4plus --version`.
- [ ] The workflow runs the conformance suite against the installed wheel and it passes.
- [ ] The workflow publishes only when every check passes.

## Links

- Feature: [`docs/specs/features/09-release.md`](docs/specs/features/09-release.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #20
Depends on #67

Crank-Git:
## Inherited defect, found by #408 on 2026-08-09. Read this before you build

**`pip install <the wheel>` reports `Successfully installed` while installing only the
dependencies, whenever the package is already importable** — for example through a
stray `PYTHONPATH`. It never writes `ja4plus` into `site-packages`, and the message
gives no sign of it.

**So a `pip` success message does not establish that the package under test is
installed.** A case that reads the message and proceeds measures whatever was already
importable.

#408 repaired it by scrubbing `PYTHONPATH` and setting `PYTHONNOUSERSITE` for the build,
[...]

PR #507, opened 2026-08-10
Verify the built wheel in a clean environment before publishing

Crank-Git:
## What this does

`.github/workflows/publish.yml` held two steps before this branch: `python -m build` and
the upload. **Nothing stood between the build and PyPI, and a publish to PyPI cannot be
undone.** `FR-release-4` to `FR-release-9` state the checks that belong there.

New file `tests/release_verification.py` holds the whole check, and the workflow runs it in
one step in front of the upload step.

1. `python -m build` writes the source distribution and the wheel. `FR-release-4`.
2. `twine check` reads both files. `FR-release-9`.
3. `pip install <the wheel>` fills a clean environment. `FR-release-5`.
4. The console script of that environment writes the version. `FR-release-6`.
5. The conformance suite of the checkout runs against that installation. `FR-release-7`.

`FR-release-8` needs no step of its own. The publish step is the last step, no step carries
`continue-on-error` or `if: always()`, and GitHub stops a job at the first failed step.

## Why the suite runs outside the checkout

**The conformance suite lives under `tests/`, and #455 removed `tests/` from the wheel.** So
the suite comes from the checkout and the package comes from the environment.

**A run that starts in the repository root reads the source tree.** `python -m` puts the
working directory on `sys.path`, and `pytest` inserts the parent of the `tests` package there
as well. Both paths name the checkout, and the checkout holds `ja4plus/`.

`verification_root` copies `tests/` and `pyproject.toml` to a directory that holds no package
source, and the check reads `ja4plus.__file__` from that directory before it runs one case. A
measurement of 2026-08-10 reports both directions from one clean environment.

```
copied root:     <venv>/lib/python3.14/site-packages/ja4plus/__init__.py
repository root: <checkout>/ja4plus/__init__.py
```

## What this branch does not rewrite

**#408 repaired `pip install` reporting `Successfully installed` while it installs the
dependencies alone.** This branch imports `build_artifacts`, `create_clean_environment`,
`package_file_of` and `run_probe` of `tests/test_installed_wheel.py` rather than write the
check again. Those functions scrub `PYTHONPATH`, set `PYTHONNOUSERSITE`, and assert that
`pip` wrote the package into `site-packages`.

**This branch changes no version-resolution mechanism.** #67 owns that. The check reads
`declared_version` of `tests/test_installed_wheel.py` and changes nothing there.

## The floors

**An aggregate over an empty set passes**, so three floors stand in the check.

- `twine_check` refuses an empty file list.
- The `PASSED` count reads 2 against two files.
- `compare_collections` refuses a clean environment whose case list differs from the
  checkout. The clean environment collected 1809 cases against 1809 in the checkout, and the
  run reported `1532 passed, 143 skipped, 134 xfailed`.
- `passing_summary` refuses a conformance run that passed no case. **A status of zero is not
  a passing run**, because `pytest` reports a run of nothing but skips as a success. The
  self-review found this one; the review comment below records it.

## How the cases came first, and how they bite

New file `tests/test_publish_workflow.py` holds 28 cases, 10 of them under the
`installed_wheel` marker. **The whole file failed to collect against the base commit
`589c5ee`.**

```
E   ModuleNotFoundError: No module named 'tests.release_verification'
```

**Three mutations prove the cases bite, and each one was restored.**

1. The workflow of the base commit fails 2 cases.
   `AssertionError: publish.yml runs 'python -m tests.release_verification' in 0 steps`
2. A check that reads `ja4plus.__file__` from the repository root fails 1 case.
   `AssertionError: the conformance run imported <checkout>/ja4plus/__init__.py, which is not below <venv>/lib/python3.14/site-packages`
3. A conformance file list cut to its first entry refuses the release.
   `RuntimeError: the clean environment collected 202 cases against 1809 in the checkout`
4. A conformance run reduced to a collection refuses the release.
   `RuntimeError: the conformance run wrote no passed count: '1809 tests collected in 0.23s'`

**A workflow step that never runs cannot fail.** No case starts the publish workflow, because
a run of it would publish. The cases read the step order as text and run the check itself, and
one case holds the command the workflow states against the command the cases run.

The `installed-wheel` job of `.github/workflows/test.yml` now runs the marker over `tests/`
rather than over one file. **A command that names one file runs none of the cases this branch
adds**, and a case nobody runs is a check nobody has.

## The `twine` pin

The `dev` extra gains `twine==6.2.0`, `PINNED` of `tests/test_lint_gate_pin.py` gains the same
name, and `MINIMUM_DEV_ENTRIES` rises from 5 to 6. **7.0.0 is the newest release and this pin
declines it**, because it requires Python 3.10 and the matrix runs Python 3.9. That is the
reading #446 records for `pytest` and `build`.

## Local gates

`ci: skip`, so the full local suite replaces the provider run.

```
pytest tests/ -m "not spec_validation"   3929 passed, 4 skipped, 8 xfailed
pytest tests/ -m installed_wheel         33 passed
ruff format --check ja4plus/ tests/      209 files already formatted
coverage                                 94%, 4292 statements, 273 misses
```

Coverage holds at the base numbers. No file under `ja4plus/` changes and no fingerprint moves.

## Verified against

- https://docs.pypi.org/trusted-publishers/ (retrieved 2026-08-10)
- https://twine.readthedocs.io/en/stable/#twine-check (retrieved 2026-08-10)
- https://pypi.org/pypi/twine/json (retrieved 2026-08-10)

Part of #20. Refs #68.


Crank-Git:
## Self-review, by lens

Three reviewers read the diff: correctness, security and release safety, and the writing
standard. Every finding is below, with what it changed.

### The finding that mattered: the check could print `PASSED` over a suite that ran nothing

**`run_conformance` read the exit status of the conformance run alone.** `pytest` reports a
run whose every case skipped as a success. The cases of `tests/test_spec_validation.py` skip
where `have_vectors()` reports nothing, so a verification root that carried no vector tree
would have produced this.

```
release check: the conformance suite ran 1809 cases.
release check: PASSED. The release is ready to publish.
```

**The floor stood in the test file and not in the gate.** `test_the_conformance_run_reports_no_failure`
held the passed count, and that case runs in the `installed-wheel` job. It does not run in
`.github/workflows/publish.yml`. **That is the shape this project records more than eighteen
times: a comparison that is never made.**

`passing_summary` now reads the summary line, refuses a passed count of zero, and refuses a
summary that names a failure. `verify` keeps the line it read, and the case reads that line
rather than the raw output, so a check that stops reading the summary fails a case.

**Mutation, restored after the measurement.** A conformance run reduced to a collection
refuses the release.

```
E   RuntimeError: the conformance run wrote no passed count: '1809 tests collected in 0.23s'
```

Three cases hold the reader in both directions: a passing summary, an all-skipped summary,
and a summary that names a failure. A fourth holds `134 xfailed` as no failure, because a
registered deviation is not a failed case.

### Two readers of the workflow were wider than they read

**The step reader matched any line at six spaces.** A `strategy` block or any other list at
that indentation would have read as a step, and the assertion that the publish step is the
last step reads a step count. The reader now starts at the `steps:` key and stops at the
next key of the job.

**The trigger case named two events and read no third.** It asserted that `workflow_dispatch`
and `push:` are absent, so a `schedule:` or a `pull_request_target:` would have passed. One
case now reads the whole key set of the `on:` block and requires `['release']`.

### Writing standard

Six sentences ran past the 25-word limit and each one is now split, four of them into a
vertical list. Three figures of speech are gone: `load-bearing`, `trap`, and `green result`.
One aphorism, `a case nobody runs is a check nobody has`, now states the mechanism.

**One finding is declined.** The reviewer read `a publish to PyPI cannot be undone` as
passive voice under rule 11. That sentence is the existing wording of this repository:
`.claude/skills/release/SKILL.md:9` carried it before this branch, and
`docs/specs/features/09-release.md` states `A partial publish cannot be undone on PyPI`. Rule
7 asks for one concept and one word, so a rewrite here would part the new text from the three
records that already hold it. **A sweep of that phrase is separate work and this branch orders
none.**

**Two further findings are declined as third-party terms.** `trusted publishing` is PyPI's
own name for the feature, and `docs/specs/features/09-release.md` already uses it.

### Security and release safety

Clean. The reviewer confirmed the trigger, the `id-token: write` permission, the `pypi`
environment, and all three commit-pinned actions are byte-identical to the base. It also
took one measurement this branch needed: **`python -m build` uses build isolation by
default**, and no call site passes `--no-isolation`, so `pip install -e ".[dev]"` in the
publish job cannot reach the built artifacts. The workflow interpolates no `${{ github.event.* }}`
value into any `run:` step.

### Gates after the repair

```
pytest tests/ -m "not spec_validation"   3929 passed, 4 skipped, 8 xfailed
pytest tests/ -m installed_wheel         33 passed
ruff format --check ja4plus/ tests/      209 files already formatted
coverage                                 94%, 4292 statements, 273 misses
```

Issue #69, opened 2026-08-06
Check the wheel contents and mark the project stable


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/09-release.md`](docs/specs/features/09-release.md):

> FR-release-10 — The wheel contains the mapping file and the `py.typed` marker.

> FR-release-11 — The wheel does not contain the test suite, the examples, or the vectors.

> FR-release-12 — The project classifier reads `Development Status :: 5 - Production/Stable`.

## What to build

New file `tests/test_packaging.py`.

A wheel that omits the mapping file makes every lookup return nothing, and a wheel that omits `py.typed` makes the annotations invisible to a caller running `mypy`. Both fail silently at runtime, which is why they are checked at build time.

`pyproject.toml` currently declares `Development Status :: 3 - Alpha`. Version 1.0.0 states that the interface is stable, so a breaking change after it needs version 2.0.0.

## Acceptance criteria

- [ ] `unzip -l dist/*.whl` lists `ja4plus/data/ja4plus-mapping.csv` and `ja4plus/py.typed`.
- [ ] `unzip -l dist/*.whl` lists no file under `tests/`, `examples/` or `docs/`.
- [ ] `pyproject.toml` declares `Development Status :: 5 - Production/Stable`.

## Links

- Feature: [`docs/specs/features/09-release.md`](docs/specs/features/09-release.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #20
Depends on #68

Crank-Git:
## The user ruled on 2026-08-09: stop here and ask again

**The wheel-contents work is buildable. The stable classifier is not built on the project
manager's authority.**

When this issue is reached, **label it `status:needs-feedback`**, surface the evidence
Epic 11 produced plus whatever the intervening batches found, and let the user rule on
`Development Status :: 5 - Production/Stable`.

**The evidence as it stands**: 835 sweep candidates settled over 3545 mutations with 7
defects, none in a fingerprinter; the wheel and the source distribution installed into a
clean environment and fingerprinting byte-identically to the source tree; the five gates,
the fuzz job and the memory ceiling measured on a Linux host; a published throughput
measurement that gates nothing.

[...]

PR #508, opened 2026-08-10
Read the wheel against its two content requirements

Crank-Git:
## What this builds

`FR-release-10` and `FR-release-11` of `docs/specs/features/09-release.md`. New file
`tests/test_packaging.py` holds nine cases against the built wheel, under the
`installed_wheel` marker.

1. The wheel lists `ja4plus/data/ja4plus-mapping.csv` and `ja4plus/py.typed`.
2. The wheel lists no entry under `tests/`, `examples/` or `docs/`.
3. The wheel lists at least 30 entries. **An aggregate over an empty set passes**, so this
   floor stands in front of the exclusion rule.

## Why it reads rather than repairs

**#455 already repaired the wheel.** A read of 2026-08-10 reports 39 entries and 155900
bytes, and #455 reports the same entry count. The value of a case here is therefore the
direction it fails in, so four mutation cases prove both directions on a copy of the wheel:
a copy without the marker, a copy without the mapping file, a copy that gains one entry
under `tests/`, and an archive that lists no entry at all.
`test_the_mutation_leaves_the_built_wheel_unchanged` measures by digest that a mutation
edits no byte of the built wheel.

## The measurement that corrects a plausible reading

A read of 2026-08-10 cut `[tool.setuptools.package-data]` to `ja4plus = ["data/*.csv"]` and
cut the exclusion list to `["examples", "assets", "assets.*"]`. **The build still listed
`ja4plus/py.typed`.** `include-package-data` is true by default and the file finder of
`setuptools-scm` reports every tracked file, so a tracked file below the package ships
whatever the package-data list holds. That build listed 376 entries and 337 of them lay
under the three excluded trees, and the excluded-tree case failed with

```
AssertionError: the wheel carries 337 of its 376 entries under ['tests/', 'examples/', 'docs/'], and it must carry none
```

The measurement restored `pyproject.toml`, and `git status` reported the file unchanged.

Verified against https://setuptools.pypa.io/en/latest/userguide/datafiles.html and
https://setuptools.pypa.io/en/latest/userguide/package_discovery.html (retrieved
2026-08-10).

## The classifier is out of scope

**The user ruled on 2026-08-10 that the project holds `Development Status :: 3 - Alpha`
until version 1.0.0 is tagged.** The release commit of 1.0.0 writes that line. This change
edits no line of `pyproject.toml`, `FR-release-12` stays open, and the third acceptance
criterion of #69 is declined by that ruling.

## How it is tested

- `pytest tests/ -m "not spec_validation"`: 3949 passed, 4 skipped, 8 xfailed.
- `pytest tests/ -m spec_validation`: 1532 passed, 143 skipped, 134 xfailed.
- `pytest tests/ -m installed_wheel`: 43 passed.
- `ruff check ja4plus/ tests/`: all checks passed.
- `ruff format --check ja4plus/ tests/`: 212 files already formatted.
- `mypy --strict ja4plus/`: no issue in 31 source files.
- Coverage holds at 94% with 4292 statements and 273 misses.

This change repeats no reading of `tests/test_installed_wheel.py`, which holds the
documentation tree, the assets tree, the top-level name and the whole payload. The new file
imports `build_artifacts` and `wheel_entry_names` from it.

Part of #20. Read #69 for the ruling.


Crank-Git:
## Self-review, two lenses

Two Sonnet children read the diff, one for correctness and one for the writing standard.
`b7bf20d` carries the repairs.

### The correctness lens found one thing, and it is real

**The mapping file and the documentation tree reach a case in
`tests/test_installed_wheel.py` and a case here.** The module docstring claimed that this
file repeats no reading of that file, and that claim was wrong on two entries.
`FR-release-10` names the mapping file beside the `py.typed` marker, and `FR-release-11`
names the documentation tree beside `tests/` and `examples/`. A case that dropped half of
each sentence would read half a requirement, so both readings stay and all three records now
state the reason.

The lens confirmed four other readings and found no defect: no case passes vacuously,
`rewrite_wheel` opens the source read-only and a digest measures that it stays unchanged,
each acceptance criterion has a case that fails when the state is wrong, and the file imports
cleanly on Python 3.9 because it uses `typing` generics under `from __future__ import
annotations`.

### The writing lens moved eight sentences

- Four sentences past the 25-word limit become two sentences each.
- `test_the_wheel_lists_the_mapping_file_and_the_type_marker` becomes
  `test_the_wheel_lists_the_two_entries_the_release_requires`, and
  `test_the_reader_refuses_a_wheel_that_lost_the_type_marker` becomes
  `..._lost_the_py_typed_marker`. **`type marker` was a third name for one concept**, and the
  `## Terms` table holds no such term.
- The heading `How the wheel contents are read` becomes
  `How tests/test_packaging.py reads the wheel`, which is active.
- The four mutation steps become a vertical list.
- The fixture docstring opens with the result.
- `until version 1.0.0 is tagged` and `is declined rather than built` become active in all
  three records.

### Two findings declined

**The quoted ruling stays verbatim.** The lens asked to rewrite
`it is written by the commit that makes the promise true` into the active voice. That
sentence is the user's own text, and the writing standard reproduces a quote verbatim.

**The bold markers inside `#` comments stay.** `pyproject.toml` and
`tests/test_installed_wheel.py` already carry that form, so the file matches the repository.

### The gates, run again after the repairs

- `pytest tests/ -m "not spec_validation"`: 3950 passed, 4 skipped, 8 xfailed.
- `pytest tests/ -m spec_validation`: 1532 passed, 143 skipped, 134 xfailed.
- `pytest tests/ -m installed_wheel`: 43 passed.
- `ruff check`, `ruff format --check`: clean. `mypy --strict ja4plus/`: no issue.
- Coverage holds at 94% with 4292 statements and 273 misses.

Issue #510, opened 2026-08-10
Batch: the release dry run, the Linux drop count and two gates


Crank-Git:
## Members

- [x] #502 — Widen the writing-standard exemption to rule 3
- [x] #326 — Report the dropped-packet count of the capture layer (the Linux half)
- [x] #438 — Read the change set of a pull request on the runner
- [x] #70 — Add the TestPyPI dry run and the release body

## Each member carries a ruling of 2026-08-10

**#502** widens the exemption to rule 3. **#326** runs on the granted Linux host with the
standing guardrails. **#438** may open a deliberately red pull request and must then close it.
**#70** is built rather than split, and the project manager adds one guardrail that is not
optional: **the manual trigger reaches TestPyPI alone and cannot reach PyPI.**

## What the provider read on 2026-08-10, for #70

The `testpypi` environment exists, holds no protection rule and has **zero deployments**, and no
workflow reaches it. **The TestPyPI half is unreadable by any agent**, so the first dry run is
the measurement: an absent publisher fails at the upload step and publishes nothing.

## The integration branch

`batch/510-dry-run-and-gates`, off `dev` at `1c29898`.

Crank-Git:
Four members, rounds 190 to 193, merged as https://github.com/Crank-Git/ja4plus/pull/521 at `7949d8f`. The batch gate read a terminal successful run of every required workflow at the head commit `ca41065`.

Issue #502, opened 2026-08-10
Rule on rule 3 of the writing standard against the Changelog record


Crank-Git:
## The question

**Does the exemption of `.claude/rules/ste.md` cover rule 3, the six-sentence paragraph limit, as
well as rule 1, the sentence-length limit?**

Raised by the self-review of #484 on 2026-08-10. **The reviewer is right and the worker was right
to decline the repair**, so the question reaches the user.

## What #457 settled and what it did not

**#457 exempted rule 1 for two records**, the entries of `CHANGELOG.md` and the `## Changelog`
table of `docs/specs/spec.md`. The user ruled it on 2026-08-10 and it shipped at round 180.

**The file states plainly that the exemption reaches rule 1 alone.**

> Every other rule of this standard reaches both records, among them one word one meaning,
> active voice, and the verbatim list above.

**Rule 3 is one of those other rules.** It reads: `One topic per paragraph, six sentences at
most.`

## The measurement

**Every row of the Changelog table is one paragraph, and a row runs well past six sentences.**
The row #484 wrote runs to roughly 24. **All 183 rows hold that same shape**, and so do the
matching entries of `CHANGELOG.md`.

**So the standard states a rule that the record has never held.** That is the shape #457 existed
to remove: a rule that a reader measures and finds false teaches the reader to disregard the
file.

## Why the worker declined rather than repaired

**A row written to rule 3 parts from the 183 rows around it.** Restructuring one row makes the
record inconsistent, and restructuring the record is a house-style change for the whole project.
#457 settled that a change of that size belongs to the user. **The decline is recorded on pull
request #501 with that reason, and the project manager agrees with it.**

## The readings

1. **Widen the exemption to rule 3 for the same two records.** One row records one round, which
   is one topic, and the sentence count follows from how much that round measured. This keeps
   the record as it is and makes the standard true. **The project manager recommends this
   reading**, for the same reason the user gave on #457: the exemption makes the rule true where
   the alternative makes 183 records disagree with what they recorded.
2. **Hold rule 3 and rewrite the record.** 183 rows are restructured into paragraphs of six
   sentences. This carries the cost #457 already weighed and declined for rule 1, and a dated
   record of a past measurement is quoted rather than rewritten.
3. **Exempt the record from the paragraph rules generally**, naming which rules of the `Sentences`
   group apply and which do not, rather than listing them one at a time as each is found. **This
   is the widest reading and it needs the most care**, because a blanket exemption reaches rules
[...]

Issue #484, opened 2026-08-10
Correct the eight places under ja4plus that call ten fingerprinters ten methods


Crank-Git:
Found by #450 while it widened the class-count gate to the Python files under `tests/`.

## The measurement

**The Python files under `ja4plus/` state the count of fingerprinter classes as a count of
methods at eight places.** A read of the branch of #450, with `python_prose` and
`class_counts_of_methods` of `tests/test_documented_method_count.py`, reports them.

| File | Line | What it states |
|---|---|---|
| `ja4plus/__init__.py` | 113 | `One fingerprinter class for each of the ten methods.` |
| `ja4plus/cli.py` | 160 | `The processor runs the ten methods in one fixed order` |
| `ja4plus/fingerprinters/base.py` | 76 | `ten methods.` |
| `ja4plus/processor.py` | 124 | `so ten threads work at once on ten methods.` |
| `ja4plus/processor.py` | 264 | `The loop reaches ten methods, and one method` |
| `ja4plus/processor.py` | 283 | `The report holds one entry for each of the ten methods.` |
| `ja4plus/processor.py` | 295 | `across the ten methods stops the packet source first.` |
| `ja4plus/processor.py` | 332 | `from all ten methods or it leaks state` |

**The count is eleven methods and ten fingerprinter classes.** `JA4LFingerprinter` writes
both `JA4L-C=` and `JA4L-S=`, so a count of classes read as a count of methods reads one
short. #387 records the measurement.

## Why #450 left them

**#450 owns four lines of `ja4plus/watch.py` and no other line under `ja4plus/`.** The
triage comment of #450 rules on exactly those four, and it enumerates them. These eight sit
outside that ruling.

`tests/test_thread_safety.py:373` mirrors the sentence of `ja4plus/processor.py:124`. #450
corrected the test docstring and left the comment, so the two read differently today.

## The repair

1. Correct the eight comments and docstrings. Each one counts fingerprinters where it
   counts methods now.
2. Add `ja4plus/` to `python_sources` in `tests/test_documented_method_count.py`. That
   function holds the Python corpus of the class-count case, and it reads `tests/` alone
   today.

**`ProcessorStats.method` keeps its name.** It is a published output field, and #450
records the decision that a rename is the user's to make.

## Acceptance criteria

- [ ] No comment and no docstring under `ja4plus/` states the count of fingerprinter
      classes as a count of methods.
- [ ] `python_sources` reads the Python files under `ja4plus/` as well as those under
[...]

PR #514, opened 2026-08-10
Widen the writing-standard exemption to rule 3

Crank-Git:
Builds #502. The user ruled on 2026-08-10 that the exemption of `.claude/rules/ste.md`
widens to rule 3 for the same two records that #457 exempted from rule 1.

## What changed

`.claude/rules/ste.md` renames `## The one exemption` to `## The exemption`. The section
states `The exemption covers rule 1 and rule 3.` beside
`Every other rule reaches both records: rules 2, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15
and 16.` The checklist gains one item for the paragraph limit. The dated read reports 178
of the 190 rows of the specification table past six sentences, and 113 of the 136 entries
of `CHANGELOG.md` past six.

`tests/test_changelog_sentence_exemption.py` rises from 27 cases to 51. `standard_rules`
reads the 16 numbered rules of the standard, `exempt_rules` and
`rules_that_reach_the_records` read the two placement sentences, and `rule_failures`
reports five states, among them a rule the section places nowhere. `uncovered_failures`
measures the exempt region under every rule the exemption drops. `paragraphs` and
`record_units` read one entry and one table row as one paragraph, and a list item is a
paragraph of its own.

`CHANGELOG.md` gains one entry and the `## Changelog` table of `docs/specs/spec.md` gains
one row. Both read `TBD` for the round number, which the project manager assigns at the
batch gate.

## Why

Rule 3 reads `One topic per paragraph, six sentences at most.` The self-review of #484
found its own Changelog row past that limit, and that finding is correct against the
letter of the rule. One row records one round, which is one topic, so the sentence count
of a row follows from how much that round measured. A rule that a reader measures and
finds false teaches the reader to disregard the file.

## How this was tested

The cases came first. The 51 cases failed 17 against the writing standard of the base
commit, among them:

```
FAILED tests/test_changelog_sentence_exemption.py::test_the_exemption_covers_rule_one_and_rule_three
  the ruling exempts the rules [1, 3], and the exemption covers []
```

Five mutations run inside cases and prove the failing direction: an exemption that drops
rule 3 reports every long entry and every long row, an exemption that drops rule 1 reports
every long sentence, an exemption that names rule 2 on both sides fails the
placement, and an exemption that places rule 2 on neither side fails it too. `mutated_rule` refuses a mutation that matches no
span, so a rewording that defeats the reader fails rather than reports a clean rule.

The five gates, run in this worktree on macOS under Python 3.14.3:

| Gate | Result |
|---|---|
| `pytest tests/ -m "not spec_validation"` | 3974 passed, 4 skipped, 8 xfailed |
| `ruff format --check ja4plus/ tests/` | 212 files already formatted |

Coverage holds at 94% with 4292 statements and 273 misses, which are the counts the base
commit reports. No file under `ja4plus/` changes and no fingerprint moves. The conformance
counts match the base commit.

This pull request targets `batch/510-dry-run-and-gates` and it is a draft, because #502 is
a member of batch #510. Every head commit carries the skip keyword by design.


Crank-Git:
## Self-review

Three lenses read this diff: the correctness of the reader, the truth of the two records
against the cases they cite, and the reach of the change outside the two records. The
first and the third found nothing to repair. The second found one defect and `c9e377b`
repairs it.

**The records named the wrong rule for two mutations.** Each record stated that an
exemption which places rule 5 nowhere fails a case. Both placement cases move rule 2:
`test_an_exemption_that_places_a_rule_nowhere_fails` reads
`places these rules of the standard nowhere: [2]`, and
`test_an_exemption_that_covers_a_rule_and_states_that_it_reaches_fails` reads
`the exemption both covers these rules and states that they reach: [2]`. A record states a
past measurement, so a record that cites a case must name what that case measures. The
specification row now quotes the failure text of each one.

**Two readings that hold, and the reason each one holds.**

`uncovered_failures` returns an empty list against the shipped standard, because the
exemption covers both rules the reader measures. **A comparison that is never made reads
as a comparison that passes**, and three cases close that hole rather than leave it. An
exemption that drops rule 3 reports 113 entries and 178 rows, an exemption that drops
rule 1 reports every long sentence, and each floor stands at 50.

`_placed_rules` reads every whole number of a sentence that holds its marker. The
checklist of the standard holds `The exemption above covers rule 1 for two records.`, and
the marker reads `the exemption covers`, so no checklist item reaches the placement. A
rewording that defeats a marker fails the placement rather than reports a clean rule, which
`mutated_rule` and the third live mutation of the prior worker both measure.

**No prose outside the two records loses rule 3.**
`test_the_writing_standard_itself_holds_the_paragraph_limit` measures `.claude/rules/ste.md`
against the limit, and `test_a_document_outside_the_two_records_holds_the_paragraph_limit`
reports a paragraph of seven sentences. No file under `ja4plus/` changes and no fingerprint
moves.

Issue #513, opened 2026-08-10
State a description that a stripped docstring does not remove


Crank-Git:
Two modules under `tests/` read `__doc__` for the description of their argument parser. **`python -OO` sets `__doc__` to None on every module**, so each one raises `AttributeError` before it reads one argument.

## The two sites

```
tests/mutation_cover.py:211:    parser = argparse.ArgumentParser(description=__doc__.splitlines()[0])
tests/mutation_sweep.py:520:    parser = argparse.ArgumentParser(description=__doc__.splitlines()[0])
```

`tests/memory_ceiling_run.py:109` passes `__doc__` itself, which argparse accepts as None, so that line raises nothing.

## The measurement

#70 repaired the same pattern at `tests/release_body.py:139` and measured the failure first.

```
  File "tests/release_body.py", line 139, in main
    parser = argparse.ArgumentParser(description=__doc__.splitlines()[0])
AttributeError: 'NoneType' object has no attribute 'splitlines'
```

`pyright` reports the same site as `reportOptionalMemberAccess`, so a reader finds it without the run.

## Acceptance criteria

- [ ] `python -OO -m tests.mutation_cover --help` exits 0.
- [ ] `python -OO -m tests.mutation_sweep --help` exits 0.
- [ ] `pyright` reports no `reportOptionalMemberAccess` against either file.

## Links

- The repair to follow: `DESCRIPTION` of `tests/release_body.py`, and `test_the_module_runs_where_the_interpreter_strips_every_docstring` of `tests/test_release_body.py`.
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)

Found while building #70. It is out of the scope of that issue and no part of its change set.

Crank-Git:
A third site landed at round 197, after this issue was filed.

```
tests/skip_gate.py:333:    parser = argparse.ArgumentParser(description=__doc__.splitlines()[0])
```

**A read of `dev` at `31cca24` names three sites of the defect and one site that raises nothing.**

| Site | Form | Under `python -OO` |
|---|---|---|
| `tests/mutation_cover.py:211` | `__doc__.splitlines()[0]` | raises |
| `tests/mutation_sweep.py:520` | `__doc__.splitlines()[0]` | raises |
| `tests/skip_gate.py:333` | `__doc__.splitlines()[0]` | raises |
| `tests/memory_ceiling_run.py:109` | `__doc__` | argparse accepts `None` |

[...]

Issue #512, opened 2026-08-10
Read the version from ja4plus/__init__.py in the release skill


Crank-Git:
Steps 1 and 2 of [`.claude/skills/release/SKILL.md`](.claude/skills/release/SKILL.md) name the wrong version declaration. #67 reversed the direction they state, and the two steps still hold the old one.

## What the file states today

Step 1 reads:

```
1. Set the version. It lives in `pyproject.toml` only. `ja4plus/__init__.py` reads it
   with `importlib.metadata.version`.
```

Step 2 then reads `pyproject.toml` with `tomllib` to confirm the version against `CHANGELOG.md`.

## What the repository holds

#67 made `ja4plus/__init__.py` the one version declaration. `tests/version_gate.py` is the reader, `declaration_files` takes the file list from `git ls-files`, and `docs/specs/features/09-release.md` records the decision under `### The version count, measured on 2026-08-09 and settled on 2026-08-10`. Round 187 of the `## Changelog` table of `docs/specs/spec.md` records the change.

**A maintainer who follows step 1 edits a file that declares no version.** `pyproject.toml` reads the version from the package today, so the release ships the old number.

## Acceptance criteria

- [ ] Step 1 of `.claude/skills/release/SKILL.md` names `ja4plus/__init__.py` as the one version declaration.
- [ ] Step 2 reads the version through `tests/version_gate.py` and not through `tomllib`.
- [ ] A case fails where a step of the skill names `pyproject.toml` as the version declaration.

## Links

- Feature: [`docs/specs/features/09-release.md`](docs/specs/features/09-release.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)

Found while building #70. It is out of the scope of that issue and no part of its change set.

Issue #70, opened 2026-08-06
Add the TestPyPI dry run and the release body


Crank-Git:
## Context

This issue satisfies the following requirements, quoted from [`docs/specs/features/09-release.md`](docs/specs/features/09-release.md):

> FR-release-13 — The release publishes to TestPyPI first when the maintainer asks for a dry run.

> FR-release-14 — The GitHub release body holds the changelog section for that version.

## What to build

**This issue needs a decision before it can be built.** TestPyPI trusted publishing is a separate one-time configuration on the TestPyPI site, and only the repository owner can create it. I could not verify whether it exists for this project.

Two ways forward:

1. The owner configures the TestPyPI trusted publisher, and this issue implements the dry-run input on the publish workflow.
2. The dry run is dropped from version 1.0.0, `FR-release-13` moves to a later release, and this issue covers `FR-release-14` only.

The release body is extracted from the `## [<version>]` section of `CHANGELOG.md`.

Verified against https://test.pypi.org/ and https://docs.pypi.org/trusted-publishers/ (retrieved 2026-08-06).

## Acceptance criteria

- [ ] A dry run publishes to TestPyPI and not to PyPI.
- [ ] The GitHub release body holds the `1.0.0` changelog section.
- [ ] A dry run with TestPyPI unconfigured fails and names the missing configuration, and does not fall back to PyPI.

## Links

- Feature: [`docs/specs/features/09-release.md`](docs/specs/features/09-release.md)
- Design: [`docs/specs/spec.md`](docs/specs/spec.md)
- Writing standard: [`.claude/rules/ste.md`](.claude/rules/ste.md)
- Conformance rules: [`.claude/rules/conformance.md`](.claude/rules/conformance.md)

Part of #20

Crank-Git:
## The user ruled on 2026-08-09

**Way forward 1. The owner configures the TestPyPI trusted publisher, and this issue
implements the dry run in full.** `FR-release-13` stays in version 1.0.0.

## What was measured on 2026-08-09

**The project `ja4plus` does not exist on TestPyPI.** `https://test.pypi.org/simple/ja4plus/`
answers 404. It exists on PyPI: `https://pypi.org/simple/ja4plus/` answers 200.

A first measurement read `https://test.pypi.org/project/ja4plus/` and saw 200 on both
indexes. **That reading was wrong.** The 200 is the page shell, which the site serves
before it knows whether the project exists. **The simple index is the check that
answers the question.**

[...]

PR #516, opened 2026-08-10
Reach TestPyPI on a manual event and build the release body

Crank-Git:
Builds #70. `FR-release-13` and `FR-release-14`.

## What this changes

**The publish workflow holds two events and two jobs, and each job accepts one event.** A manual trigger on a publish workflow creates a path to PyPI, so the project manager ruled that the dry run must be structurally incapable of reaching the real index.

| Path | Event | Job | Environment | Repository URL |
|---|---|---|---|---|
| Release | `release: published` | `publish` | `pypi` | none, so the action uploads to PyPI |
| Dry run | `workflow_dispatch` | `dry-run` | `testpypi` | `https://test.pypi.org/legacy/` |

**The manual trigger declares no input.** Each environment name and the repository URL are literals of the file, so no value a caller chooses reaches either one. The dry run runs the same verification as the release.

**`FR-release-14` is restated, on a ruling of the user of 2026-08-10.** The whole `## [1.0.0]` section is 242778 characters and the provider accepts 125000, so the first form cannot hold. The body now holds the summary and the two breaking-change tables, and it links `CHANGELOG.md` at the tag for the rest. `CHANGELOG.md` keeps every row it holds, no round entry moves, and nothing is rewritten. The body of version 1.0.0 measures 4727 characters.

**The reader truncates nothing.** A named part above the limit fails the reader, the step fails, and the release publishes nothing.

## Why

A dry run that a caller could point at PyPI is the hazard the separation removes. A release body the provider refuses stops the release at the provider rather than at a named step.

## How this was tested

**The value of a case here is the direction it fails in**, so thirteen mutations prove both directions. Each was written to disk, measured, and restored, and `.github/workflows/publish.yml` compared equal by digest `36941ecb201aed94a19d8c63ee920144beafbda00e205e6c55059f7f5d0916d1` after the restore.

| Mutation | The case that failed |
|---|---|
| The `dry-run` job takes the environment `pypi`. | `test_each_job_of_the_publish_workflow_names_its_environment_as_a_literal` |
| The `dry-run` job reads its repository URL from an input. | `test_the_manual_path_of_the_publish_workflow_reaches_no_real_index` |
| The `publish` job takes the TestPyPI repository URL. | `test_the_release_path_of_the_publish_workflow_reaches_no_test_index` |
| Both jobs lose their `if:` condition. | `test_each_job_of_the_publish_workflow_accepts_one_event` |
| The manual trigger declares an input. | `test_the_manual_trigger_of_the_publish_workflow_declares_no_input` |
| The `publish` job loses the release-body step. | `test_the_publish_job_writes_the_release_body_before_it_publishes` |
| `release_body` returns the whole section. | `test_the_release_body_of_this_repository_stands_below_the_provider_limit` |
| `named_part` keeps the entry list. | `test_the_named_part_holds_no_entry_of_the_entry_list` |
| `changelog_link` names the default branch. | `test_the_link_names_the_changelog_at_the_tag_of_the_release` |
| `body_fault` reads no limit. | `test_the_release_body_reader_refuses_a_named_part_above_the_limit` and two more |
| `main` reads the module docstring again. | `test_the_module_runs_where_the_interpreter_strips_every_docstring` |
| `named_part` matches the heading as a substring. | `test_the_named_part_reader_passes_over_the_heading_named_inside_a_paragraph` and the case below |
| `named_part` matches a whole line and reads the code fence not at all. | `test_the_named_part_reader_passes_over_the_heading_quoted_in_a_code_block` |

**A workflow step that never runs cannot fail**, and no case here starts the publish workflow, because a run of it would publish. `python -m tests.release_verification --dist dist` therefore ran on this host against a real build.

```
release check: built ja4plus-0.6.0-py3-none-any.whl and ja4plus-0.6.0.tar.gz.
release check: twine check reported PASSED 2 times.
release check: the console script wrote 'ja4plus 0.6.0'.
release check: 1532 passed, 143 skipped, 134 xfailed, 1 warning in 11.23s
release check: PASSED. The release is ready to publish.
```

The whole gate ran on this host.

| Gate | Result |
|---|---|
| `pytest tests/ -m "not spec_validation"` | 3990 passed, 4 skipped, 8 xfailed |
| `pytest tests/ -m installed_wheel` | 43 passed |
| `ruff format --check ja4plus/ tests/` | 214 files already formatted |
| Coverage | 94%, 273 misses of 4292 statements |

Coverage holds at the reading round 143 records, and no file under `ja4plus/` changes.

## The self-review found one defect, and this branch repairs it

`named_part` ended the named part at any text that read like the breaking-change heading, so a heading quoted in a sentence or a code block dropped both tables. **Such a reader reports no fault**, because it returns a part that reads as complete and is not. `breaking_heading_end` now takes a whole line and it tracks the code fence. The two comments below record the whole review.

## Two Pyright diagnostics

**One reproduces and this branch repairs it.** `main` read the module docstring for the description of its argument parser. `python -OO` sets `__doc__` to None on every module, so the command raised `AttributeError: 'NoneType' object has no attribute 'splitlines'` before it read one argument. The case came first and it failed on that message. `DESCRIPTION` is now a literal of the module.

**The other does not reproduce.** `tests/release_body.py:38` imports `tests.version_gate`, which `tests/release_verification.py:47` imports in the same form. `pyright --outputjson tests/release_body.py` reports zero diagnostics from the repository root and from `tests/`.

## No dry run has run

The `testpypi` environment exists at the provider, holds no protection rule and carries zero deployments. No agent reads the TestPyPI publisher page. An absent publisher fails the index at `invalid-publisher` and the upload sends no artifact, and the literal repository URL keeps the job away from PyPI whatever the publisher state is. **The first dry run is the measurement, and this branch starts none**, because a run of this workflow publishes.

## Out of scope, filed rather than fixed

- #512 — step 1 of `.claude/skills/release/SKILL.md` still names `pyproject.toml` as the version declaration that #67 removed.
- #513 — `tests/mutation_cover.py:211` and `tests/mutation_sweep.py:520` hold the same stripped-docstring defect.

## Verified against

- https://docs.pypi.org/trusted-publishers/ (retrieved 2026-08-10)
- https://docs.pypi.org/trusted-publishers/troubleshooting/ (retrieved 2026-08-10), for `invalid-publisher`
- https://github.com/pypa/gh-action-pypi-publish (retrieved 2026-08-10), for `repository-url`
- https://docs.github.com/en/rest/releases/releases (retrieved 2026-08-10), for the `body` field
- https://github.com/cli/cli/issues/7705 (retrieved 2026-08-10), for the 422 answer
- https://cli.github.com/manual/gh_release_edit (retrieved 2026-08-10), for `--notes-file`

Refs #70.


Crank-Git:
## Self-review, the security lens

A reviewer read the diff against one question: can any value chosen at run time move the `dry-run` job from TestPyPI to PyPI? It found no such path, and it reports the six readings below.

1. **The job routes on `github.event_name` alone.** `if: github.event_name == 'workflow_dispatch'` compares a context value, and `workflow_dispatch:` declares no input, so the file holds no `inputs.*` expression a caller could set. No `workflow_call`, no matrix, and no reusable workflow stands between the event and the job.
2. **Both environment names and the repository URL are YAML literals.** `environment: testpypi`, `environment: pypi` and `repository-url: https://test.pypi.org/legacy/` hold no `${{ }}`. The `publish` job states no `repository-url` at all, which is how the action reaches PyPI.
3. **Each job holds the permission it needs and no other.** The `dry-run` job holds `id-token: write` alone, so it writes no repository content and it edits no release. The `publish` job adds `contents: write`, which is the permission `gh release edit` needs. The provider offers no finer grain for that endpoint.
4. **Every action pins a full commit SHA.** `actions/checkout`, `actions/setup-python` and `pypa/gh-action-pypi-publish` each carry 40 characters and a version comment. No tag and no branch reference stands in the file.
5. **`workflow_dispatch` needs write access at the provider**, so the caller of a dry run already holds write access. Such a caller reaches TestPyPI and the `testpypi` environment, and reaches PyPI nowhere.
6. **`RELEASE_TAG` carries no injection.** `github.event.release.tag_name` reaches the step through `env:` and never through `${{ }}` inside the script text, and the step reads it as `"${RELEASE_TAG}"`. The shell expands the value once and it parses the result no second time, so a tag that holds `$(...)` or `;` reaches `gh` as one inert argument.

`tests/release_body.py` reads `CHANGELOG.md` and `ja4plus/__init__.py` from a root it resolves from its own path, and it writes the one file `--output` names. It runs no subprocess and it builds no path from an untrusted value.

**One caveat stands outside this file, and this branch records it rather than closes it.** The protection rules of the `pypi` and `testpypi` environments, and the trusted-publisher configuration on TestPyPI, are the second half of the claim. `### The TestPyPI publisher, and what proves it` of `docs/specs/features/09-release.md` states what a read of the provider reports and what no agent can read.

No change follows from this lens.

Crank-Git:
## Self-review, the correctness lens

The reviewer read the diff and raised one defect that the cases did not hold, plus two notes. `a6a75ba` carries the repair.

### The defect, accepted and repaired

**`named_part` ended the named part at any text that read like the breaking-change heading.** The first form called `str.find`, so the heading text inside a sentence, a table cell or a code block matched. The reader then returned the summary alone and it dropped both tables.

**Such a reader reports no fault.** It returns a part that reads as complete and is not, so `body_fault` sees nothing wrong and the release carries a body that omits every breaking change. That is the failure this project records many times.

**A line anchor alone does not close it.** The reviewer measured the intermediate form and found that a line inside a fenced code block reading exactly `### The breaking changes` still satisfies `^...$` under `re.MULTILINE`. That reading is correct and the repair follows it.

`breaking_heading_end` now takes a whole line and it tracks the code fence. Two mutations measure the two readings, and each one was written to disk, measured and restored.

| Reader | The cases that failed |
|---|---|
| A substring match. | `test_the_named_part_reader_passes_over_the_heading_named_inside_a_paragraph`, `test_the_named_part_reader_passes_over_the_heading_quoted_in_a_code_block` |
| A whole-line match that reads the code fence not at all. | `test_the_named_part_reader_passes_over_the_heading_quoted_in_a_code_block` |

### The two notes

**The mutation count of `CHANGELOG.md` disagreed with the list below it.** The entry read "five of the reader" against a list of more. It reads seven now, and the list names each one.

**The `dry-run` job declares no `contents: read`.** That is correct as it stands and this branch changes nothing. A job that declares a `permissions:` block sets every unlisted scope to `none`, and `actions/checkout` reads a public repository with no token. The job writes no repository content by design, which is the reason the block names `id-token: write` alone.

### What the reviewer found clean

- `len(body) > BODY_LIMIT` accepts a body of exactly 125000 characters.
- `changelog_section` escapes the version, so `1.0` does not match `1.0.0`.
- A breaking-change heading that is the last heading of a section reads to the end of the section.
- No case asserts a value it computes the same way the reader computes it.
- The step order of the workflow, the `contents: write` scope, `${RUNNER_TEMP}` and `RELEASE_TAG` are each correct.

The whole gate ran again after the repair: 3990 passed in the unit suite, 1532 passed in the conformance suite, coverage 94% with 273 misses of 4292 statements, `ruff` and `mypy --strict` clean, and `pyright` reports zero diagnostics against both files.

PR #518, opened 2026-08-10
Read the change set of a pull request on the runner

Crank-Git:
## The finding that outranks the rest of this change set

**The round-entry check passed by construction, and it never refused the pull request it
exists to refuse.**

`.github/workflows/test.yml` read `pull_request: branches: [master, dev]`. That filter
matches the **base** branch of a pull request, so the only pull requests that reached the
`test` job were the batch pull request and the promotion. A batch pull request reads its
change set against the tip of `dev`, and every member of a batch has already recorded a
round, so the check found an entry whatever one member did. **The member pull request is
the change set the check exists to refuse, and the filter kept it away from the job.**

The user ruled on 2026-08-10 that the filter widens. It now reads:

```yaml
  pull_request:
    branches: [master, dev, "batch/**", "epic/**"]
```

**The widened filter costs no run that the batch model saved.** A member commit ends with a
skip keyword by design, and #459 measured that such a commit creates no run for any event.
A member pull request therefore still creates no run, and a keyword-free head is the one
head that now starts one.

`.claude/rules/batch-gate.md` gains the section `## Which pull request creates a run`. It
states the two conditions that each stop a run, and it states this shape so that a reader
holds the next check against it. Its opening paragraph rested on the old filter and now
names both conditions.

## Two readings of the runner

A green run proves nothing here on its own, because the case skipped before this change and
a skip is not a pass. The user approved one deliberately red pull request for this proof,
and no other issue of this project may open one. **#520 targeted
`batch/510-dry-run-and-gates`, it targeted neither `dev` nor `master`, each of its two heads
carried no skip keyword, it merged into nothing, and it is closed with its branch deleted.**

| Direction | Head | Run | Conclusion | `test_the_change_set_of_this_branch_records_a_round` |
|---|---|---|---|---|
| Red — three tracked files, no round | `8702322` | https://github.com/Crank-Git/ja4plus/actions/runs/31402557575 | `failure` | `FAILED`, not `SKIPPED` |
| Green — the same change set plus the entry and the row | `ecd1677` | https://github.com/Crank-Git/ja4plus/actions/runs/31403217183 | `success` | `PASSED` on all six jobs of the matrix |

The red assertion, from the run log:

```
E   AssertionError: the change set holds these paths outside the two records: .claude/rules/batch-gate.md, .github/workflows/test.yml, tests/test_round_entry_existence.py. CHANGELOG.md holds 105 round entries against 105 at the reference commit
```

## Which branch the filter was read from — measured, not assumed

**GitHub read the widened filter from the merge commit of the proof pull request, and not
from the base branch.** `batch/510-dry-run-and-gates` still carries the old filter, and the
run started anyway. The run itself is that measurement, so the filter change took effect
inside the pull request that carried it and it did not have to land on the base branch
first.

The documentation agrees, by contrast with `pull_request_target`: "This event runs in the
context of the default branch of the base repository, rather than in the context of the
merge commit, as the `pull_request` event does." Verified against
https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows
(retrieved 2026-08-10).

**#519 measured the older filter before the ruling.** It carried a keyword-free head into
the same integration branch and the provider held no run for it after 150 seconds. It is
closed and its branch deleted.

## What was wrong with the checkout

`tests/test_round_entry_existence.py` also skipped on every job. The `actions/checkout` step
names no `fetch-depth`, so it makes a clone of depth 1. That clone carries no `origin/dev`
ref and no history behind the checked-out commit, so `git merge-base` fails and the case
reports a skip reason.

## The reading this pull request chose

**The `test` job fetches the base commit of the pull request at depth 1 and names it.** The
step writes the commit into the `ROUND_ENTRY_REFERENCE` environment variable.
`environment_reference` reads that variable, `named_commit` resolves it against the
repository, and `reference_commit` prefers it over the merge base.

**`git diff` and `git show` read a commit that no history connects to `HEAD`, and
`git merge-base` does not.** A diff compares two trees, so one extra commit is the whole
requirement. This is the measurement that decided the reading, taken on 2026-08-10 against
a clone of depth 1 of this repository:

```
=== fetch that one commit, depth 1, by sha ===
From https://github.com/Crank-Git/ja4plus
 * branch            864b70d222ef800f113e9a355c38514ae595d537 -> FETCH_HEAD
real 0.68
=== can git read the tree of that commit? ===
commit
# Changelog
=== diff against HEAD (no shared history needed) ===
.claude/rules/ste.md
.github/workflows/test.yml
CHANGELOG.md
=== merge-base against HEAD in the shallow clone ===
merge-base exit: 1
```

A name the repository cannot read reads as no name, so a stale value of the variable leaves
the local gate of a worker exactly as it is.

**The fetch writes the ref `refs/ja4plus/round-entry-base`, because a commit that no ref
reaches is unreachable.** A bare `git fetch origin <sha>` writes `FETCH_HEAD` and no ref. A
read of 2026-08-10 on a clone of depth 1 reports that the commit survives
`git gc --prune=now` under the ref form and that `git diff` and `git show` still read it.

## The measured cost, and the readings this pull request declined

A read of 2026-08-10 against `https://github.com/Crank-Git/ja4plus`, on the `dev` branch.

| Reading | Size of `.git` | Cost over the clone of depth 1 |
|---|---|---|
| The clone of depth 1 that the workflow makes today | 14844 KB | — |
| The clone of depth 1 and the extra fetch of one commit | 15452 KB | 608 KB, 0.68 s |
| `fetch-depth: 0`, the whole history of every branch | 18728 KB | 3884 KB |

**`fetch-depth: 0` costs 3884 KB on each of the six jobs of the matrix, and that cost rises
with every commit the project makes.** The cost of one commit does not rise. #438 declined
it on that measurement.

**A second step that runs `git fetch --depth=1 origin dev` was measured and it does not
work.** Two shallow histories share no commit, so `git merge-base` fails between them and
the case would skip exactly as before. The `merge-base exit: 1` line above is that
measurement.

**The base commit is also a stricter reference than the merge base with `dev`.** It reads
the change set of one pull request rather than the change set of a whole integration
branch, which is the widest limit row 156 of the specification Changelog records.

## How this was tested

The cases came first. `test_the_test_job_fetches_the_base_commit_of_a_pull_request` failed
against the workflow of the base commit, and
`test_the_batch_gate_rule_records_the_widened_filter` failed against the rule file of the
base commit:

```
E       AssertionError: assert 'BASE_SHA: ${{ github.event.pull_request.base.sha }}' in 'name: Tests\n\n# `dev` is the default branch and every pull request targets it. ...'
E       assert 'batch/**' in "# The batch gate\n\n**An absent run is not a passed run.** The batch pull request is the only run of a\nmember's case..."
```

Twelve new cases hold the reading. Two build an orphan branch, which reproduces the runner
state where `git merge-base` reports nothing and the diff still answers. Two read
`.github/workflows/test.yml` for the fetch step, and two hold the widened filter and the
new section of the rule file.

The five local gates, run on 2026-08-10 in a virtual environment outside the worktree. The
branch carries `batch/510-dry-run-and-gates` at `737d3f5`, because #70, #326 and #517 landed
after it started and the tail of the `## Changelog` table conflicted. The merge keeps all
rows in arrival order.

| Gate | Result |
|---|---|
| `pytest tests/ -m "not spec_validation"` | 4039 passed, 7 skipped, 8 xfailed |
| `ruff format --check ja4plus/ tests/` | 214 files already formatted |
| `mypy --strict ja4plus/` | no issues in 31 source files |

**Coverage holds at 94%** with 4316 statements and 273 misses. The 273 misses are the count
the base commit reports, and the statement count rises with #326, which arrived through the
merge. The conformance suite reports the three counts of the base commit. **No file under
`ja4plus/` changes in this change set and no fingerprint moves.**

## Evidence

Verified against https://github.com/actions/checkout (README, retrieved 2026-08-10), which
states of `fetch-depth`: "Number of commits to fetch. 0 indicates all history for all
branches and tags."

Verified against https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28
(retrieved 2026-08-10) for `pull_request.base.sha`, and against
`gh api repos/Crank-Git/ja4plus/pulls/508`, which returned
`"base_sha": "eec6b3a60991bd7d9717e0c73ace246e8da1952e"` for base ref `epic/20-release`.

Verified against
https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows
(retrieved 2026-08-10) for the base-branch `branches` filter and for the merge-commit
context of `pull_request`.

Verified against
https://docs.github.com/en/actions/how-tos/manage-workflow-runs/skip-workflow-runs
(retrieved 2026-08-09) through `.claude/rules/batch-gate.md`.

Part of #510. Builds #438.


Crank-Git:
## Self-review: security and CI-policy lens

A Sonnet reviewer read the workflow diff against `.claude/rules/batch-gate.md`. Five findings, and this comment records what each one produced.

1. **Script injection: clean.** The step reads `env: BASE_SHA: ${{ ... }}` and uses `"$BASE_SHA"` inside `run:`, which is the safe pattern. `github.event.pull_request.base.sha` is a commit the provider computes on the base repository, so a fork author cannot set it to free text. **Accepted.** The step comment now states the reason, so a later reader does not move the expression into the shell command.

2. **A format assertion on the SHA: declined.** The value never carries `=` or a newline, so the write to `GITHUB_ENV` is safe. An assertion against a state that cannot occur is code this project does not add.

3. **No token or credential reaches the step.** Confirmed, and the step needs no permission the job does not already hold.

4. **`continue-on-error: true` on the fetch step: declined, and this is the one finding worth arguing.** The reviewer is right that a failed fetch would red a required check for an infrastructure reason. It is also true that `continue-on-error` leaves `ROUND_ENTRY_REFERENCE` empty, so the case skips and the job goes green. **`.claude/rules/batch-gate.md` states that an absent result is not a passed result**, and a green job over a silent skip is the exact failure #438 exists to remove. A red step names the state. The step comment records the decline.

5. **Reachability, found while reading finding 4.** A bare `git fetch origin <sha>` writes `FETCH_HEAD` and creates no ref, so the commit is unreachable. The fetch now writes `refs/ja4plus/round-entry-base`. A read of 2026-08-10 on a clone of depth 1 reports that the commit survives `git gc --prune=now` under the ref form, and that `git diff` and `git show` still read it.

6. **No violation of the batch gate and no interaction with the skip keyword.** The eleven required check contexts do not move, no job name changes, and the step reads no commit message.

Crank-Git:
## Self-review: correctness lens

A Sonnet reviewer read the diff and ran the file and the linter rather than reading alone. Eight findings.

1. **No regression on a local worktree.** `environment_reference()` returns an empty string where the variable is unset, `named_commit` short-circuits, and `reference_commit` falls through to the `git merge-base` loop it always ran. Confirmed by the local run.

2. **`named_commit` rested on its one caller to drop the space around a name: accepted and repaired.** `environment_reference` stripped the value, and a caller that reached `reference_commit` or `evaluate` directly with ` abc ` would have failed `rev-parse --verify` and fallen back to the merge base without saying so. `named_commit` now strips the name itself, and `test_a_named_commit_reads_through_the_space_around_the_name` holds it. That is the tenth new case.

3. **The bare fetch wrote no ref: repaired before this comment.** The fetch now writes `refs/ja4plus/round-entry-base`.

4. **Step order is correct.** Git is on the runner image before `actions/setup-python` runs, and a write to `GITHUB_ENV` reaches every later step.

5. **A push event and a manual run are unaffected.** The `if` condition skips the step, the variable stays unset, and the case skips as the module docstring states.

6. **The orphan-branch helper is not flaky.** `git checkout --orphan` keeps the index and the working tree, so the commit differs from the pre-orphan base by the write the case makes, which is the runner state exactly.

7. **`ruff check tests/` reports nothing.** `mypy --strict` runs on `ja4plus/` alone by project convention, so the `Mapping[str, str]` default is unenforced and it is not wrong.

8. **The declined `continue-on-error` widens the blast radius of a rare failure, and that is the intended trade.** A base commit that a rewrite of the integration branch makes unreachable would red the whole matrix rather than skip one case. The step comment states the reason, and `.claude/rules/batch-gate.md` states the doctrine behind it.

No blocker and no major finding. The two repairs are in the branch.

PR #519, opened 2026-08-10
PROOF, NEVER MERGE: show the round-entry case red on the runner (#438)

Crank-Git:
## Warning: never merge this pull request

**This is a proof and not a change set.** It carries the smallest change that makes
`tests/test_round_entry_existence.py` fail on the runner, and it lands in no branch. #438
closes it and deletes its branch as soon as the two runs are read. It targets
`batch/510-dry-run-and-gates`, and it targets neither `dev` nor `master`.

The user approved this one pull request for this one purpose. No other issue of this
project may open a deliberately red pull request.

## What it proves

#438 makes the `test` job fetch the base commit of the pull request and read the change set
against it. **A green run would prove nothing**, because the case skipped before #438 and a
skip is not a pass. The proof needs two readings of the runner.

1. This head edits two tracked files and records no round, so the job fails.
2. The next commit records the round, so the same job passes.

The head commit carries no keyword that stops a run.

Part of #438.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -72,6 +72,31 @@ jobs:
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
 
+      - name: Fetch the base commit of the pull request
+        # `tests/test_round_entry_existence.py` reads the change set of the branch against
+        # a reference commit. The step above makes a clone of depth 1, which holds no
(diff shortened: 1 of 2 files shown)


Crank-Git:
Closed with no run. The base-branch filter of `.github/workflows/test.yml` reads `branches: [master, dev]`, so a pull request into an integration branch creates no run. #438 records the measurement and names the decision it needs.

PR #520, opened 2026-08-10
PROOF, NEVER MERGE: show the round-entry case red on the runner (#438)

Crank-Git:
## Warning: never merge this pull request

**This is a proof and not a change set.** It carries the smallest change that makes
`tests/test_round_entry_existence.py` fail on the runner, and it lands in no branch. #438
closes it and deletes its branch as soon as the two runs are read. It targets
`batch/510-dry-run-and-gates`, and it targets neither `dev` nor `master`.

The user approved this one pull request for this one purpose. No other issue of this
project may open a deliberately red pull request.

## What it proves

#438 makes the `test` job fetch the base commit of the pull request and read the change set
against it. **A green run would prove nothing**, because the case skipped before #438 and a
skip is not a pass. The proof needs two readings of the runner.

1. This head edits two tracked files and records no round, so the job fails.
2. The next commit records the round, so the same job passes.

The head commit carries no keyword that stops a run.

Part of #438.


Crank-Git:
Both runs are read and this pull request lands in no branch.

- **Red**, head `8702322`, run https://github.com/Crank-Git/ja4plus/actions/runs/31402557575, conclusion `failure`. `tests/test_round_entry_existence.py::test_the_change_set_of_this_branch_records_a_round` reported `FAILED` and not `SKIPPED`.
- **Green**, head `ecd1677`, run https://github.com/Crank-Git/ja4plus/actions/runs/31403217183, conclusion `success`. The same case reported `PASSED` on all six jobs of the matrix.

**The measurement the ruling asked for: GitHub read the widened base-branch filter from the merge commit of this pull request, and not from the base branch.** `batch/510-dry-run-and-gates` carries the old filter, and the run started anyway. #438 records it.

Issue #524, opened 2026-08-10
Fail a suite that holds a case which skips in every environment


Crank-Git:
#438 measured a case that ran on the runner and refused nothing, because its precondition was unsatisfiable there. **`tests/test_round_entry_existence.py` reported a skip on every job of the matrix and it had done so since it was written.** A skip is not a pass, and a case that runs nowhere is not a case.

**This issue builds the detector for that shape and takes the census of the shape as it stands today.**

## The surface

A read of 2026-08-10 counts **40 skip sites across `tests/`**, in `skipTest(`, `pytest.skip(`, `skipIf(` and `skipUnless(` form. Some are correct: a macOS case skips on Linux and a Linux case skips on macOS, and each runs somewhere in the matrix. **The defect is a case that skips in every environment, so no job ever runs it.**

## What to build

**A gate that reads the skip report of every job and refuses a case that skips in all of them.**

`pytest -rs` writes one line per skip with the case identifier and the reason. `.github/workflows/test.yml` runs six matrix jobs: ubuntu with Python 3.9, 3.10, 3.11, 3.12 and 3.13, and macOS with 3.12. **The union of the six reports names every case the suite ran nowhere.**

The reading is free of a new dependency. A job writes its skip report as an artifact, and a job that depends on all six reads them and fails where the intersection is not empty.

**An allowlist is permitted and every entry of it carries a reason.** A case that needs a capture grant no runner holds is a legitimate entry. **An entry with no reason is the defect this issue exists to remove**, so a case fails where its allowlist entry names none.

## The census

**Classify all 40 sites.** Each is one of these, and the count of each reaches the pull-request body.

- a) The case runs in at least one job of the matrix. No action.
- b) The case runs in no job and the reason is a real environment limit, such as a capture grant. It reaches the allowlist with that reason.
- c) **The case runs in no job and no limit explains it.** That is the shape of #438 and each one is a finding. Name it, and open an issue for each rather than repairing it here.

## What this issue does not cover

**A case that runs and asserts over an empty set** is a different shape, and floors are the answer to it. **A check that no workflow job invokes** is a third shape; `publish.yml` is the recorded example, which #68 measured. Both are out of scope and each takes its own issue.

## Acceptance criteria

- A job of `.github/workflows/test.yml` reads the skip reports of all six matrix jobs and fails where a case skips in every one.
- The gate is proven in both directions on the runner: a deliberately universal skip fails the job, and the suite as it stands passes it. Report both run URLs.
[...]

PR #526, opened 2026-08-10
PROOF ONLY, do not merge: the skip gate names a case that runs nowhere

Crank-Git:
**Warning: this pull request merges into nothing, and it closes as soon as the run is read.**

The user approved a deliberately red pull request for #524, under the rules #438 followed.

- It targets `batch/523-conformance-and-claims`. It targets neither `dev` nor `master`.
- Its head commit carries no keyword that stops a run, because the point is that a run starts.
- It merges into nothing. It closes and its branch is deleted as soon as the run is read.

## What this run proves

`tests/test_universal_skip_proof.py` holds one case that skips in every environment. The
`skip-gate` job of `.github/workflows/test.yml` reads the six `test-results-*` artifacts of
the `test` job, and it must name that case and fail.

A gate proved in one direction is a gate that may pass always. #438 records that sentence.
A second head restores the tree and shows the same job green, which is the other direction.

Refs #524.


Crank-Git:
Both directions are read. The red run is https://github.com/Crank-Git/ja4plus/actions/runs/31408763259 and the green run is https://github.com/Crank-Git/ja4plus/actions/runs/31412758369. This pull request merges into nothing, so it closes and its branch goes.

Issue #515, opened 2026-08-10
JA4TS reaches no FoxIO reference value


Crank-Git:
A conformance audit of 2026-08-10 read every method against the FoxIO material. **JA4TS is the one method that reaches no FoxIO-produced expected output in this repository.**

## What was measured

`git grep -i ja4ts tests/foxio_vectors/` matches one file, `tests/foxio_vectors/NOTICE`. **No vector holds a JA4TS value.**

`tests/test_foxio_rust_parity.py:138` states the cause: the local snapshot holds a `ja4ts` field, so JA4TS reaches no reference value there.

## What stands in place of a vector today

1. **Transcribed prose of a file FoxIO deleted.** Rules R12 and R13 of `docs/specs/foxio/JA4T.md` come from `technical_details/JA4T.md`, which FoxIO removed in commit `b6f3ff4` on 2024-02-22.
2. **A capture this project builds itself.** `tests/build_ja4ts_rst.py` writes it and `tests/test_ja4ts_reset.py` reads it back. **A self-generated fixture proves the implementation against itself and not against the reference.**
3. **A Zeek corroboration that no gate runs.** `tests/compare_zeek_baselines.py:39` reads its baselines from an external FoxIO checkout, no `.log` baseline is committed, and `tests/test_compare_zeek_baselines.py` drives the comparator on constructed input alone.

## Why this is the gap that matters

Every other method is held against a value FoxIO produced: JA4, JA4S, JA4H, JA4X and JA4L against the FoxIO Python expected-output files, JA4T against the FoxIO Rust snapshots, JA4D and JA4D6 against the Wireshark dissector. **A JA4TS value this library emits is unverified against any FoxIO implementation.** It is not a known disagreement. It is an unmeasured one, and the two are not the same.

## What to build

**A JA4TS value that FoxIO produced, committed as a vector.** The issue is free to find the source. Candidates, in the order the audit ranks them.

- A Zeek baseline `.log` committed to the repository, produced by the FoxIO Zeek scripts at the pinned commit, so `tests/compare_zeek_baselines.py` runs as a gate rather than as a tool.
- A Wireshark reading of a committed capture, which is the source JA4D and JA4D6 already use.
- A FoxIO Rust snapshot that carries the field, if one exists at the pinned commit.

[...]

PR #531, opened 2026-08-10
Compare JA4TS against two FoxIO implementations

Crank-Git:
Builds #515.

## What this changes

JA4TS reached no FoxIO-produced expected value in this repository. It stood on transcribed
prose of a file FoxIO deleted, plus a capture this project writes itself, and a
self-generated capture proves the implementation against itself.

**The finding that outranks the rest of this work is that this project recorded a wrong
measurement, and that record is what hid the reference.** The table `The search for a
reference value` in `docs/specs/foxio/JA4T.md` read `No ja4t value and no ja4ts value` for
`wireshark/test/testdata/`. That directory holds 118 `ja4.ja4t` values and 58 `ja4.ja4ts`
values, in 24 of its 37 files. The earlier search read the key `ja4t`, and the FoxIO
Wireshark dissector writes the key `ja4.ja4t`.

## What it adds

- `tests/foxio_vectors/wireshark_expected/` rises from 2 files to 26.
  `tests/test_foxio_wireshark_ja4ts.py` compares all 58 values. **52 match byte for byte
  and 6 reach the register.**
- `tests/foxio_vectors/zeek_expected/` is new and holds all seven Zeek baselines, which
  `docs/specs/foxio/zeek.md` asked a later issue to commit.
  `tests/test_foxio_zeek_ja4ts.py` compares 9 of their 10 JA4TS values.
- `tests/compare_zeek_baselines.py` reads the committed baselines when it gets no path, so
  a gate runs it. It ran by hand until now.
- `docs/specs/spec.md` gains `## The search for a JA4TS reference value`, which records
  every source this work read and what each one returned.

## The six register entries

All six are the RST decline that R13 of `docs/specs/foxio/JA4T.md` already records, and
#246 owns each one. The dissector writes a second JA4TS value on a RST packet of a
connection with no delay, that value repeats the value the SYN-ACK produced, and
`.claude/rules/conformance.md` declines a value that describes no packet of its own.

`TestTheDifferencesAreTheRecordedResetDecline` measures three facts of each row, so the
decline stays proven rather than asserted: each frame carries the RST flag alone, each
value equals the value of the first SYN-ACK of that stream, and `ja4plus` writes one value
for the stream.

## The one barred Zeek value

`zeek/ja4t/main.zeek:66-68` returns an empty option record when the link layer is not
Ethernet, and the link type of `ipv6.pcapng` is `DLT_NULL`. The Zeek baseline writes
`65535_00_00_00` where `ja4plus` writes `65535_2-1-1-4-1-3_1346_10`. **The FoxIO Wireshark
dissector writes the `ja4plus` value for that same connection**, so a second FoxIO
implementation corroborates the bar and it rests on no reading of this project.
`TestTheBarredBaselineRestsOnAProvenZeekDefect` measures four facts of that row.

## No fingerprint moves

No file under `ja4plus/` changes.

## How this was tested

Test-driven. The ten comparison cases of `tests/test_foxio_zeek_ja4ts.py` failed against
the base index with `assert () == ('65535_2-1-1-4-1-3_1346_10',)`, because
`tests/conformance_index.py` ran no JA4TS fingerprinter.

The five gates, on macOS 26.6.1 against Python 3.14.3:

- `pytest tests/ -m "not spec_validation"` — 4041 passed, 7 skipped, 8 xfailed.
- `pytest tests/ -m spec_validation` — 1635 passed, 143 skipped, **140 xfailed**.
- `ruff format --check ja4plus/ tests/` — 216 files already formatted.
- `mypy --strict ja4plus/` — no issues in 31 source files.

**The register invariant holds on both sides.** `tests/foxio_deviations.json` held 134
keys against 134 xfailed, and it holds 140 against 140.

Coverage holds at 94%, with 4316 statements and 273 misses, which are the counts round 192
records.

## Evidence

Verified against: https://github.com/FoxIO-LLC/ja4/tree/main/wireshark/test/testdata (retrieved 2026-08-10, commit `27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8`)
Verified against: https://github.com/FoxIO-LLC/ja4/tree/main/zeek/tests/Traces (retrieved 2026-08-10, commit `27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8`)


Crank-Git:
## The mutation that proves the comparison bites

Acceptance criterion 2 asks for a mutation that is restored. This is it.

`ja4plus/fingerprinters/ja4ts.py`, in `generate_ja4ts`:

```python
-        return prefix + _part_e(packet, tracker, prefix)
+        return prefix + "x" + _part_e(packet, tracker, prefix)
```

The mutation moves every JA4TS value and no value of another method. It kills **62 cases**:

```
62 failed, 36 passed, 6 xfailed
```

The 62 are the 52 Wireshark comparisons, the 9 Zeek comparisons, and
`TestTheBarredBaselineRestsOnAProvenZeekDefect::test_ja4plus_reads_the_options_the_synack_carries`,
which reads the `ja4plus` value of the barred row.

**The six registered cases stay `xfailed` under the mutation, and that is correct.** A
mutation moves no declined value onto the dissector value, so the strict marker holds.

The mutation is restored. After the restore:

```
98 passed, 6 xfailed
```

`git status` reports no change under `ja4plus/`, and the branch holds no edit to any file
of that directory.

Crank-Git:
## Self-review

A specialist review read `tests/test_foxio_wireshark_ja4ts.py`, `tests/test_foxio_zeek_ja4ts.py`, `tests/test_compare_zeek_baselines.py`, `tests/compare_zeek_baselines.py`, `tests/conformance_index.py` and the six new register entries, against `.claude/rules/ste.md` and the section `Ask whether a case can fail` of `.claude/rules/conformance.md`.

**It found no vacuous assertion, no comparison that is never made, and no self-referential count.** It confirmed the reading with its own mutation, of `tcp.window` in `ja4plus/utils/tcp_options.py`, which killed 65 cases across the three test files. That mutation was restored, and `git status` reports the worktree clean.

### The one real finding, and it is fixed

**No value of either FoxIO source carries a delay list of more than one delay, so part e stays measured by a constructed case alone.** A mutation of the delay-list separator in `ja4plus/fingerprinters/ja4ts.py`, from `"-".join(...)` to `",".join(...)`, killed no case of either new module.

**The gap belongs to the FoxIO material and not to the two modules.** No capture of the vector set holds a connection the server answered three times, which `What part e moved` in `docs/specs/foxio/JA4T.md` already measures: the largest reading is two SYN-ACKs. `tests/test_ja4ts_part_e.py` keeps the constructed cases that measure the separator, the rounding rule and both bounds.

`docs/specs/foxio/JA4T.md` now records the reading as a warning, so the next reader does not take these 67 comparisons for whole coverage of part e. `857b29e` carries it.

### One criterion reported the change, as it was built to

`git ls-files tests/foxio_vectors | wc -l` reports 121 files where the criterion of `docs/specs/features/00-foundation.md` states 90. That criterion reads: "so a change to the vector set fails a case rather than a review." It did exactly that, and the count now states 121 with the reason.

### Two nits, both declined

- `test_the_comparison_holds_the_match_count_the_reading_records` is arithmetically implied by the two cases beside it, under the current writer. It re-derives the count through its own filter on the verdict cell, so it holds a row the writer labels neither `yes` nor `NO`. Kept as the reader of that one state.
- `test_the_suite_adopts_nine_of_the_ten_values` reads a partition the module builds. It pins which baseline is barred, so `BARRED_BASELINES` cannot grow or shrink in silence. Kept.

Issue #523, opened 2026-08-10
Batch: the branch-protection claim, the JA4TS reference gap and the ALPN ruling


Crank-Git:
## Members

- [x] #515 — JA4TS reaches no FoxIO reference value
- [x] #511 — Hold the branch-protection claims of the batch-gate rule against a case
- [x] #522 — Record the ALPN ruling in the divergence register
- [x] #524 — Fail a suite that holds a case which skips in every environment

## Where these came from

**Three findings of the session of 2026-08-10, and none of them came from the backlog.**

**#515 is the one conformance gap the audit of this date found.** Every other method is held against a value a FoxIO implementation produced. JA4TS is held against transcribed prose of a file FoxIO deleted and against a capture this project builds itself. **A self-generated fixture proves the implementation against itself.**

**#511 came from a review of the #480 change set.** `app_id` 15368 is stated in `.claude/rules/batch-gate.md`, in `CHANGELOG.md` and in the specification, and `tests/test_batch_gate_protection_rule.py` reads `app_id` in a comment alone. **That is a claim with no floor, which is the shape #480 existed to remove.**

**#522 records a ruling of the user**, and a ruling is a specification change that takes an issue of its own.

## The integration branch

`batch/523-conformance-and-claims`, off `dev` at `7949d8f`.

Crank-Git:
Four members, rounds 194 to 197, merged as https://github.com/Crank-Git/ja4plus/pull/534. The batch gate read a terminal successful run of every required workflow at the head commit `8a8956e`.

Issue #522, opened 2026-08-10
Record the ALPN ruling in the divergence register


Crank-Git:
A ruling of the user, 2026-08-10, needs a record in the specification. **A ruling is a spec change and it takes an issue of its own**, so it rides no feature change set.

## The ruling

**The form of `ja4plus` stands.** A first ALPN value that is non-alphanumeric reads `99`, and a one-byte value reads `hh`.

## The measurement behind it

The conformance audit of 2026-08-10 named this the one condition where `ja4plus` matches no FoxIO implementation. The two references disagree with each other, and each reads its own tooling rather than the packet.

| Implementation | Value | What it reads |
|---|---|---|
| `ja4plus` | `99`, `hh` | the packet bytes |
| FoxIO Python | U+FFFD | a replacement character that no packet byte holds |
| FoxIO Rust | `h9` | the escape text of `tshark` |

**Matching either reference copies an artefact of that reference tool into the fingerprint.** The user declines that.

The sixteen register entries of #162 stand. `tests/foxio_deviations.json` keeps them and the register invariant holds.

## What to build

- The divergence register row of `docs/specs/spec.md` states the ruling and its date, and states that no FoxIO reading is available to adopt here.
- A case reads that row, so a later edit that drops the ruling fails rather than passing quietly.
- `Round TBD.` reaches `CHANGELOG.md` and the `## Changelog` table of `docs/specs/spec.md`.

## Acceptance criteria

- `docs/specs/spec.md` records the ruling, the date, and the three values of the table above.
- A case fails when the row loses the ruling. Prove it with a mutation that is restored.
- The sixteen `#162` entries of `tests/foxio_deviations.json` are unchanged, and the key count still equals the xfailed count.
- No file under `ja4plus/` changes and no fingerprint moves.

PR #532, opened 2026-08-10
Record the ALPN ruling in the divergence register

Crank-Git:
Records the ALPN ruling of the user in the divergence register. Part of #523. Builds #522.

## What this changes

**The user ruled on 2026-08-10 that the form of `ja4plus` stands.** A first ALPN value
that is not alphanumeric writes `99`, and a first ALPN value of one byte writes `hh`.

The divergence register of `docs/specs/spec.md` gains one row. The row states the ruling,
the date, and the value of each of the three implementations.

| Implementation | Value | What it reads |
|---|---|---|
| `ja4plus` | `99`, `hh` | the packet bytes |
| FoxIO Python | U+FFFD | a replacement character that no packet byte holds |
| FoxIO Rust | `h9` | the escape text of `tshark` |

## Why

The conformance audit of 2026-08-10 named this the one condition where `ja4plus` matches
no FoxIO implementation. **The two references disagree with each other**, and each one
reads its own tooling rather than the packet. A match with either reference copies an
artifact of that reference tool into the fingerprint, and the user declines that. **No
FoxIO reading is available to adopt here**, so the row states the ruling rather than a
value to adopt.

**A ruling is a spec change and it takes an issue of its own.** This change set moves no
fingerprint and it repairs no defect.

## How this is tested

New file `tests/test_alpn_ruling_register.py` holds eight cases. **Seven of the eight
failed before the row landed**, and the eighth reads the deviation register, which this
change set does not move.

```
FAILED tests/test_alpn_ruling_register.py::test_the_register_holds_a_row_for_the_alpn_ruling
FAILED tests/test_alpn_ruling_register.py::test_the_register_states_the_ruling
FAILED tests/test_alpn_ruling_register.py::test_the_register_states_the_date_of_the_ruling
FAILED tests/test_alpn_ruling_register.py::test_the_register_states_the_value_of_each_implementation
FAILED tests/test_alpn_ruling_register.py::test_the_register_states_that_no_foxio_reading_reaches_this_condition
FAILED tests/test_alpn_ruling_register.py::test_the_register_names_this_the_one_condition_that_matches_no_reference
FAILED tests/test_alpn_ruling_register.py::test_the_register_cites_the_issue_that_records_the_ruling
7 failed, 1 passed in 0.09s
```

**Four mutations prove that the cases bite.** The run restored each mutation, and
`git diff --stat` then reported one insertion.

1. A row that drops the ruling sentence.

```
AssertionError: the ruling row holds no 'the form of `ja4plus` stands'
```

2. A row that gives FoxIO Rust the value `hh`.

```
AssertionError: the sentence that names FoxIO Rust states no value `h9`
assert '`h9`' in 'FoxIO Rust writes `hh`, which is the escape text of `tshark`'
```

3. A row that renames the item.

```
AssertionError: the divergence register holds no row named 'The JA4 ALPN form that matches no FoxIO implementation'
```

4. A row that moves the date out of the ruling sentence into the sentence after it. **The
self-review added this reading**, because the ruling sentence closed its bold mark after the
period and the reader therefore read it together with the sentence after it. The row now
closes the bold mark before the period.

```
AssertionError: the sentence that states the ruling names no date 2026-08-10
assert '2026-08-10' in '| 1 | **The user ruled that the form of `ja4plus` stands**'
```

**A floor refuses a reader that found no row**, because an aggregate over an empty set
passes. `MINIMUM_REGISTER_ROWS` reads 30, which is the row count of the base commit.

## The invariants this change set holds

**The sixteen entries of `tests/foxio_deviations.json` that name #162 stand.**
`tests/foxio_deviations.json` is byte-identical to the base commit, and one case reads the
count of 16.

| Reading | Before | After |
|---|---|---|
| Entries that name #162 | 16 | 16 |
| Deviation register keys | 134 | 134 |
| Conformance `xfailed` | 134 | 134 |

**No file under `ja4plus/` changes and no fingerprint moves.** The change set holds three
files: `CHANGELOG.md`, `docs/specs/spec.md` and `tests/test_alpn_ruling_register.py`.

## The five gates

Run on macOS 26.6.1 against Python 3.14 in a virtual environment outside the worktree.

| Gate | Result |
|---|---|
| `pytest tests/ -m "not spec_validation"` | 4076 passed, 6 skipped, 8 xfailed |
| `ruff format --check ja4plus/ tests/` | 215 files already formatted |

Coverage holds at 94%, with 4316 statements and 273 misses. **All six skips read the platform
or the host**: two name the Linux cache convention, three name the `AF_PACKET` sockets of
Linux, and one names the capture privilege of this account. `mkdocs build --strict` also
passes, which is the check `.github/workflows/docs-build.yml` runs on the batch pull request.

## The round record

`CHANGELOG.md` and the `## Changelog` table of `docs/specs/spec.md` each carry `TBD`. The
project manager assigns the number at the batch gate.

Verified against: https://github.com/FoxIO-LLC/ja4/tree/main/python/test/testdata (pinned
commit `27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8`). This change set makes no call against
an external interface and it reads no new one.

Refs #522


Crank-Git:
## Self-review

Two Sonnet lenses read the change set: one against `.claude/rules/ste.md`, one against the
failure this project calls "a comparison that is never made reads as a comparison that
passes". Six findings. Four are fixed, and two are declined with the reason.

### Fixed

**1. The date case was weaker than its docstring.** `_sentence_of` splits the row on `". "`,
and the ruling sentence closed its bold mark after the period, as `stands.**`. The literal
`". "` never appeared there, so the reader held the date against the ruling sentence and the
sentence after it, fused into one chunk. **A later edit could then move `2026-08-10` into the
conformance-audit sentence and still pass.** The row now closes the bold mark before the
period, as `stands**.`, and the reader holds the date against the ruling alone. A fourth
mutation proves it.

```
AssertionError: the sentence that states the ruling names no date 2026-08-10
assert '2026-08-10' in '| 1 | **The user ruled that the form of `ja4plus` stands**'
```

**2. `MINIMUM_REGISTER_ROWS` read 29, and the register held 31 rows before this round held
30.** The constant is now 30, which is the row count of the base commit, and the comment
states that. The floor never was the reader that catches the loss of the new row —
`test_the_register_holds_a_row_for_the_alpn_ruling` is — and the floor refuses a reader that
found no row.

**3. Rule 7, one concept one word.** The register row wrote `ruling` six times and then
`decision` once, for one concept. The closing sentence now reads `The ruling is reversible.`

**4. Rule 11, active voice.** Both round records wrote "Each mutation was restored". Neither
record is exempt from rule 11, because the exemption of `.claude/rules/ste.md` covers rule 1
and rule 3 alone. They now read "The run restored each mutation".

### Declined

**5. "Add `ruling` to the `## Terms` table."** Declined for this issue, and filed instead.
The word is already load-bearing across the tracked documents: `docs/specs/foxio/JA4D.md`
carries the headings `### JA4D — the rulings` and `### JA4D6 — the rulings of #271`,
`docs/specs/features/09-release.md:80` and `:209` use it, and eleven entries of
`CHANGELOG.md` use it. **The gap predates this change set and it reaches the whole
repository**, so a one-row addition inside a spec-record issue would settle a vocabulary
question wider than #522. The `Terms` table defines `reading` as "One recorded conclusion
about what a source states", which is a different concept from a determination the user
makes, so `reading` does not cover it either.

**6. "`test_the_ruling_moves_no_entry_of_the_deviation_register` verifies nothing about this
change."** Declined, and the reading is correct as far as it goes. That case passed before
the row landed, because this change set does not touch `tests/foxio_deviations.json`. **It is
a standing guard and not a proof of this diff**, and the third acceptance criterion of #522
asks for exactly that: the sixteen `#162` entries stay as they are. The case catches a later
edit that moves them.

### Read and not raised

`_sentence_of` asserts that exactly one chunk holds each phrase. That assertion never
branches into failure against the current row. It is over-strict rather than silently
permissive, so a future edit that added a second sentence holding `FoxIO Rust` would fail the
case loudly.

### The measurement moved

The first record draft read "rises from 4066 passed to 4074 passed, with 7 skipped". A later
run reported 4076 passed and 6 skipped, and the collected count rose by one, so the earlier
sentence stated a number this branch does not hold. **A record states a measurement, so both
records now state the reading of `-rs`**: 4076 passed, 6 skipped, 8 xfailed, and all six
skips read the platform or the host.

PR #536, opened 2026-08-10
State a description that a stripped docstring does not remove

Crank-Git:
Builds #513.

## What

**`python -OO` sets `__doc__` to None on every module.** A parser that reads
`__doc__.splitlines()[0]` for its description raises `AttributeError` before it reads one
argument. A parser that passes `__doc__` itself raises nothing, and it reports no
description at all.

**The deliverable is the case, not the four repairs.** #70 measured this defect at round
191 and repaired one site. `tests/skip_gate.py:333` carried the same line into the
repository at round 197, two rounds later. A repair of the known sites stops nothing.

## Why

New file `tests/test_parser_description.py` reads every module rather than the known
sites, so the fifth site fails here on the round that writes it.

- The static case parses every tracked Python file with `ast` and fails a parser whose
  description reads `__doc__`. It reads an `add_parser` call beside an `ArgumentParser`
  call, because a subcommand parser accepts the same keyword.
- The run case runs every tool module of `tests/` twice, once plain and once under `-OO`,
  and requires the same `--help` from both runs. **That comparison covers both failure
  modes at once**, because the first raises and the second drops the description.
- **The reader takes its file list from `git ls-files` and it walks the checkout
  nowhere.** #473 measured a reader that walked the repository root and picked up a worker
  worktree under `.claude/`, so its corpus grew with the number of live workers.

## The measurement, before the repair

The case came first and it failed on the four sites a read of `dev` at `31cca24` named.

```
AssertionError: these parser descriptions read the module docstring: ['tests/memory_ceiling_run.py:109', 'tests/mutation_cover.py:211', 'tests/mutation_sweep.py:520', 'tests/skip_gate.py:333']
```

```
  File "tests/skip_gate.py", line 333, in main
    parser = argparse.ArgumentParser(description=__doc__.splitlines()[0])
AttributeError: 'NoneType' object has no attribute 'splitlines'
```

`tests/mutation_cover.py:211` and `tests/mutation_sweep.py:520` raised the same message.
`tests/memory_ceiling_run.py:109` exited 0 and reported no description.

## Both directions

- `test_the_reader_finds_a_parser_that_splits_the_module_docstring` and
  `test_the_reader_finds_a_parser_that_passes_the_module_docstring` prove the reader
  reports each form.
- `test_the_reader_passes_a_parser_that_states_its_own_description` and
  `test_the_reader_passes_a_call_that_is_no_parser` prove it reports neither.
- `test_the_help_comparison_fails_a_module_that_reads_the_module_docstring` writes a
  module that reads `__doc__` and measures its two outputs apart, so the comparison bites.
- **The floor.** `test_the_tracked_corpus_holds_an_argument_parser` and
  `test_the_tracked_corpus_holds_a_tool_module` fail a reader that finds nothing, so a
  pathspec that lists nothing fails rather than reports a clean corpus.

## The repair

Four sites now name a `DESCRIPTION` constant, which is the repair `tests/release_body.py`
already carries: `tests/memory_ceiling_run.py`, `tests/mutation_cover.py`,
`tests/mutation_sweep.py` and `tests/skip_gate.py`. The issue body names the first three
of these and its newest comment names the fourth.

## How tested

The five gates, on macOS with Python 3.11.15, in a virtual environment outside the
worktree.

| Gate | Result |
|---|---|
| `pytest tests/ -m "not spec_validation"` | 4129 passed, 7 skipped, 8 xfailed |
| `ruff format --check ja4plus/ tests/` | 220 files already formatted |
| `mypy --strict ja4plus/` | Success, 31 source files |

`pyright` reports no `reportOptionalMemberAccess` against the four files. It reports one
`reportAttributeAccessIssue` at `tests/mutation_cover.py:105`, which stands before this
change and which this change does not touch.

`FR-pre-release-validation-39`, `FR-pre-release-validation-39a` and
`FR-pre-release-validation-39b` state the three requirements. `CHANGELOG.md` and the
Changelog table of `docs/specs/spec.md` each carry the round as `TBD`.

Verified against https://docs.python.org/3/using/cmdline.html#cmdoption-OO (Python 3.11,
retrieved 2026-08-10), which states that `-OO` discards docstrings.


Crank-Git:
## Self-review

Two lenses ran over the diff: correctness and the writing standard. Three findings reach a repair and three do not.

### Repaired

**1. The reader read no positional description.** `argparse.ArgumentParser` takes `description` as its third positional argument, and the reader read `node.keywords` alone. `ArgumentParser(prog, usage, __doc__)` therefore passed. No call site of the repository writes that form, so the gap was latent. **The point of this issue is the site the reader has not met yet**, so a latent gap in the reader is the defect this issue exists to remove. The reader now reads that position too, and `test_the_reader_finds_a_parser_that_takes_the_module_docstring_in_the_third_position` holds it.

**2. `tool_modules` matched the text `__name__ == "__main__"`.** A module that writes `__name__=='__main__'`, or that writes the comparison the other way round, reached no run case. Every guard of `tests/` writes the form the reader expected, so the gap was latent here too. The reader now reads the guard from the syntax tree, and `test_the_tool_reader_reads_a_main_guard_whatever_its_quotes_and_order` holds both other forms.

**3. Rule 1 of `.claude/rules/ste.md`, at `tests/test_parser_description.py:14`.** One sentence of the module docstring ran to 27 words, and the limit for a description sentence is 25. The sentence is now two.

### Declined

**4. The reader reports `description=__doc__ or "Read the run."`, which raises nothing.** That form is safe under `-OO`, so the report is stricter than the hazard. **The title of this issue asks a module to state a description**, and a fallback beside `__doc__` states one twice. A reader that accepts the fallback needs control-flow reasoning to tell it from the hazard, and no call site of the repository writes either form. The reader keeps the stricter reading and a comment states it.

**5. A module-level alias defeats the reader.** `DOC = __doc__` and then `description=DOC` reaches no report, because the reader reads the name and never the data flow. The run case covers that module, because it compares the two `--help` outputs and the alias moves one of them.

**6. `pyright` reports `reportAttributeAccessIssue` at `tests/mutation_cover.py:105`.** That report stands before this change, it is another rule than the one the issue names, and this change does not touch the line.

Crank-Git:
## The gates, after the review repairs

macOS, Python 3.11.15, in a virtual environment outside the worktree.

| Gate | Result |
|---|---|
| `pytest tests/ -m "not spec_validation"` | 4135 passed, 7 skipped, 8 xfailed |
| `ruff format --check ja4plus/ tests/` | 220 files already formatted |
| `mypy --strict ja4plus/` | Success, 31 source files |

`tests/test_parser_description.py` reports 22 passed, and it covers eight tool modules of `tests/` in the run case.

**Line coverage reads 94% against the floor of 70%.** `COVERAGE_FLOOR` of `.github/workflows/test.yml` states that floor, and the measurement runs `--cov=ja4plus`. No file under `ja4plus/` changes, so this change moves no coverage of the library.

`pyright` reports no `reportOptionalMemberAccess` against the four repaired files or the new case file.

Every thread of this pull request is answered.

PR #537, opened 2026-08-10
Run the change-set defect case on the runner

Crank-Git:
Repairs #528. **A skip is not a pass, and this case had never run on any job.**

## What was wrong

`tests/test_round_entry_existence.py::test_the_reading_fails_the_change_set_of_the_defect`
reads commit `46aa502` against its parent, which is the change set #412 shipped: eleven
sweeps, two repairs, a new test file, no `CHANGELOG.md` entry and a green gate. It is the
one case that proves the reading finds a real change set which records no round.

The `actions/checkout` step of `.github/workflows/test.yml` makes a clone of depth 1, and
that clone holds neither the commit nor its parent. The case therefore reported
`this clone holds no parent of commit 46aa502` on all six jobs. The census of #524
measured it in run https://github.com/Crank-Git/ja4plus/actions/runs/31408763259 and
classed it c, which names a site that no limit of the runner explains.

## What this change does

The `test` job fetches the two recorded commits, each at depth 2, into
`refs/ja4plus/recorded-defect` and `refs/ja4plus/recorded-control`. A fetch of depth 2
carries the commit and its parent, which is the pair a diff needs.

**A green reading of one commit proves nothing on its own**, so the case set holds two
directions.

| Case | Commit | The reading |
|---|---|---|
| `test_the_reading_fails_the_change_set_of_the_defect` | `46aa502`, from #412 | one failure |
| `test_the_reading_passes_the_change_set_that_records_a_round` | `f140a5c`, from #429 | no failure |

`tests/universal_skips.json` loses the entry that named this case, so the allowlist holds
one open finding and #529 names it.

## Three measurements

**The cost, taken on 2026-08-10 after `git gc --prune=now` on each read.** The clone of
depth 1 holds 14884 KB, the defect fetch raises it to 14912 KB, and the control fetch
raises it to 14916 KB. The two commands cost 32 KB together, against the 3884 KB of
`fetch-depth: 0` that round 193 records.

**The fetch names the whole identifier, because git resolves no abbreviation at the
remote.** `git fetch --depth=2 origin "+46aa502:refs/ja4plus/abbrev"` on a clone of depth 1
answered `fatal: couldn't find remote ref 46aa502`.

**A shallow clone reads both commits after the fetch.** `git diff --name-only` and
`git show` each answered on a clone of depth 1, and the two refs survived
`git gc --prune=now`.

## How this was tested

**The cases came first.** `test_the_test_job_fetches_the_two_recorded_change_sets` failed
against the workflow of the base commit:

[...]

Issue #528, opened 2026-08-10
Run the change-set defect case on the runner


Crank-Git:
## What was measured

**`tests/test_round_entry_existence.py::test_the_reading_fails_the_change_set_of_the_defect`
skips on all six jobs of the matrix.** The census of #524 measured it on the runner, in run
https://github.com/Crank-Git/ja4plus/actions/runs/31408763259. The case reports one of two
reasons.

```
tests/test_round_entry_existence.py:521: pytest.skip(f"this clone holds no parent of commit {DEFECT_COMMIT}")
tests/test_round_entry_existence.py:525: pytest.skip(f"git reports no change set of commit {DEFECT_COMMIT}")
```

**No environment limit explains it.** The `actions/checkout` step of
`.github/workflows/test.yml` makes a clone of depth 1, and that clone holds neither
`DEFECT_COMMIT` nor its parent. A runner can fetch those commits. #438 measured the cost of
one extra commit at 608 KB against a clone of depth 1, so the reading is affordable.

**This is a class c site of the #524 census.** The case runs in no job and the reason is a
choice of the checkout, not a limit of the runner.

## Why this matters

`DEFECT_COMMIT` is the commit #412 measured: eleven sweeps, two repairs, a new test file, no
`CHANGELOG.md` entry and a green gate. The case holds the reading against that recorded
defect, so it is the one case that proves the reading finds a real change set which records
no round. **It has never run on the runner.**

## What to build

**A reading of `DEFECT_COMMIT` that the runner can make.** Some candidate readings, and this
issue is free to find a better one.

- A step that fetches `DEFECT_COMMIT` and its parent at depth 1, in the shape the `test` job
  already fetches the base commit of the pull request.
- A recorded change set the case reads from a tracked file, so the case needs no commit at
  all.

**Prove it on the runner in both directions**, under the rules #438 followed.

## The allowlist entry this issue removes

`tests/universal_skips.json` holds this case today, and its reason names this issue. The
entry goes out with the repair, and `python -m tests.skip_gate` then passes without it.

## Acceptance criteria

- [ ] The case reports no skip on any job of `.github/workflows/test.yml`.
- [ ] `tests/universal_skips.json` holds no entry for this case.
- [ ] No file under `ja4plus/` changes and no fingerprint moves.

## Links

- Found by the census of #524 on 2026-08-10.
- The checkout reading: #438.
- The recorded defect: #412.

Crank-Git:
## The runner proof

[...]

PR #538, opened 2026-08-10
Prove the change-set defect case on the runner

Crank-Git:
**Warning: this pull request merges into nothing. Close it and delete its branch.**

The user approved one deliberately red pull request for the proof of #528, under the
standing rules. It targets `batch/535-skip-gate-repairs`, it targets neither `dev` nor
`master`, and each of its heads carries no skip keyword.

**A green run alone proves nothing here.** `test_the_reading_fails_the_change_set_of_the_defect`
skipped on all six jobs of the matrix before #528, and a skip reports a green job. That is
the error #438 recorded, so this pull request reads the two cases in both directions on the
runner.

## The red head

Each case raises rather than asserts, so the run reports the reading the runner made of
each recorded commit. A skip reason fails the first assertion instead, which names a fetch
that did not answer.

## The green head

The second head carries the two cases as #537 ships them. All twelve jobs then run,
`skip-gate` reads the six reports, and it names neither case.


Crank-Git:
The proof is read. The red head 4da756b ran https://github.com/Crank-Git/ja4plus/actions/runs/31416675503, where each of the two cases reported FAILED on all six jobs of the matrix. The green head 837f88e ran https://github.com/Crank-Git/ja4plus/actions/runs/31417379359, where all twelve jobs concluded success and the skip-gate job named neither case. This pull request merges into nothing.

Issue #535, opened 2026-08-10
Batch: the three skip-gate repairs and the stripped docstring


Crank-Git:
## Members

- [x] #528 — Run the change-set defect case on the runner
- [x] #529 — Run the documentation slug case in a job that holds the docs extra
- [x] #530 — Reach the skip gate to the jobs outside the test matrix
- [x] #513 — State a description that a stripped docstring does not remove

## Where these came from

**Three of the four are the output of round 197.** The skip gate of #524 named two cases that run
on no job of the matrix, and it named the jobs its own reader does not reach. **A detector that
reports a finding and leaves it unrepaired is half a gate**, so the repairs land here.

**#513 is the shape that outlives its repair.** #70 measured `__doc__.splitlines()` against
`python -OO` at round 191 and repaired one site. **A new file carried the same line into the
repository two rounds later**, at `tests/skip_gate.py:333`. A case that reads every module with a
parser is what stops the fourth site.

## The integration branch

`batch/535-skip-gate-repairs`, off `dev` at `31cca24`.

Crank-Git:
Four members, rounds 198 to 201, merged as https://github.com/Crank-Git/ja4plus/pull/542. The batch gate read a terminal successful run of every required workflow at the head commit `07cf258`.

Issue #529, opened 2026-08-10
Run the documentation slug case in a job that holds the docs extra


Crank-Git:
## What was measured

**`tests/test_documentation_site.py::test_the_slug_of_a_case_matches_the_slug_of_the_build`
skips on all six jobs of the matrix.** The census of #524 measured it on the runner, in run
https://github.com/Crank-Git/ja4plus/actions/runs/31408763259. The case reports one reason.

```
SKIPPED [1] tests/test_documentation_site.py:222: the `docs` extra installs pymdownx, and `dev` does not
```

**No environment limit explains it.** Every job of `.github/workflows/test.yml` installs
`pip install -e ".[dev]"`, and the case needs the `docs` extra. A runner installs either
extra, so the skip records a choice of the install step.

**This is a class c site of the #524 census.** The case runs in no job and no limit explains
it.

## Why this matters

The case holds the slug a documentation anchor produces against the slug the site build
produces. **A slug that moves breaks every link into that anchor, and nothing measures it
today.** The `build` job of `.github/workflows/docs-build.yml` runs `mkdocs`, and it runs no
case of `tests/`.

## What to build

**A job that runs this case with the extra it needs.** Some candidate readings, and this
issue is free to find a better one.

- The `build` job of `.github/workflows/docs-build.yml` runs
  `pytest tests/test_documentation_site.py`, because that job already installs the docs
  dependencies.
- One job of `.github/workflows/test.yml` installs `".[dev,docs]"`.

**Prove it on the runner in both directions.** A slug this project changes must turn the job
red.

## The allowlist entry this issue removes

`tests/universal_skips.json` holds this case today, and its reason names this issue. The
entry goes out with the repair.

## Acceptance criteria

- [ ] One job of the repository runs the case, and the case reports no skip there.
- [ ] `tests/universal_skips.json` holds no entry for this case.
- [ ] No file under `ja4plus/` changes and no fingerprint moves.

## Links

- Found by the census of #524 on 2026-08-10.

PR #539, opened 2026-08-10
Run the documentation slug case on one job of the matrix

Crank-Git:
## What this changes

**One job of the matrix installs the `docs` extra, so the documentation slug case runs.**
`tests/test_documentation_site.py::test_the_slug_of_a_case_matches_the_slug_of_the_build`
holds `_slug` against `pymdownx.slugs.slugify` on every published heading. The census of
#524 read it as a class c site: it skipped on all six jobs and no limit of the runner
explained the skip.

The `test` job of `.github/workflows/test.yml` gains one step,
`Install the documentation extra on one job of the matrix`. It runs
`pip install -e ".[docs]"` under
`if: matrix.os == 'ubuntu-latest' && matrix.python-version == '3.13'`.
`tests/universal_skips.json` loses its entry for the case.

## The reading this pull request chose

**The `docs` extra reaches the Python 3.13 job on `ubuntu-latest`, and no other job.**

| Reading | Cost, measured on 2026-08-10 | Decision |
|---|---|---|
| The extra on one job | 28 packages, 125476 KB, one job | Chosen |
| The extra on all six jobs | 627380 KB more, and the Python 3.9 job cannot install it | Declined |
| The case in `.github/workflows/docs-build.yml` | No new install, and the skip gate never reads it | Declined |

**The all-six reading is impossible, not merely costly.** `griffe` 2.1.0 requires Python
3.10 and the matrix runs Python 3.9, so the Python 3.9 job can never install this extra.

**The `docs-build.yml` reading leaves the gate naming the case.** The `skip-gate` job
downloads the `test-results-*` artifacts of its own run, so a case that runs in another
workflow reaches no report the gate reads. That workflow also filters four paths, so a
batch that touches no documentation path creates no run of it. **The case would then run
sometimes and not always, which is a weaker gate than one job that always runs it.**

**The step names one job rather than a matrix variable.** A matrix variable reaches the job
name, and the branch protection rule of `dev` holds each of eleven required check names. The
six check names of the matrix therefore stay as they are.

## Cost of the mixed environment

A read of 2026-08-10 installed `.[dev]` and then `.[docs]` into one environment. `pip`
reported no conflict, and the unit suite reported 4136 passed and 6 skipped in that
environment. `.github/workflows/docs-build.yml` stays the one job that measures the
committed pins in an empty environment.

## How this is tested

Two new cases of `tests/test_documentation_site.py` hold the repair, and both failed before
it:

- `test_one_job_of_the_test_matrix_installs_the_documentation_extra` reads the step and its
  condition.
- `test_the_skip_allowlist_holds_no_entry_for_the_slug_case` reads the allowlist.

The first case reads the `test` job alone, because every job of the file indents a step the
same way. A script moved the step into the `lint` job and the case failed, which measures
that reading.

Five gates on the branch: `pytest tests/ -m "not spec_validation"` 4138 passed, 6 skipped,
8 xfailed; `pytest tests/ -m spec_validation` 1635 passed, 143 skipped, 140 xfailed;
`ruff check` clean; `ruff format --check` 220 files already formatted; `mypy --strict`
no issues in 31 source files.

## The proof on the runner

**The head commit of a member carries the skip keyword, so this pull request creates no
run.** `.claude/rules/batch-gate.md` states that rule, and it states `workflow_dispatch` as
the way to start a run where the webhook path creates none. Two manual runs of the branch
head hold the reading.

| Run | Head | Result |
|---|---|---|
| https://github.com/Crank-Git/ja4plus/actions/runs/31418704343 | `e20e69e` | The case runs on `test (ubuntu-latest, 3.13)` |
| https://github.com/Crank-Git/ja4plus/actions/runs/31419880962 | `6f36fa0` | The same, after the self-review repair |
| https://github.com/Crank-Git/ja4plus/actions/runs/31421263768 | `bb3e018` | The same, after this branch took the base again |

**The report of `test (ubuntu-latest, 3.13)` records the case as run**, in 0.048 s on the
first two runs and 0.050 s on the third. The five other reports carry the skip, which is
correct, because those jobs install the `dev` extra alone. **The census of `skip-gate` on the
third run reads 10 cases that ran on no job of the matrix, and this case is none of them.**
Eleven jobs of that run concluded `success`.

**This branch took `batch/535-skip-gate-repairs` again after #537 landed.**
`tests/universal_skips.json` lost one entry on each side, so the merge keeps neither, and the
allowlist holds nine entries that each name a limit of the runner. The Changelog table of
`docs/specs/spec.md` keeps both rows, and each one carries `TBD`. All five gates ran again on
the merge result: 4154 passed, 6 skipped, 8 xfailed on the unit suite.

**`skip-gate` concludes `failure` on both manual runs, and one case explains it.** A manual
run carries no pull request, so the step `Fetch the base commit of the pull request` runs
nowhere and
`tests.test_round_entry_existence::test_the_change_set_of_this_branch_records_a_round` skips
on all six jobs. **That case reads a pull-request event**, so the batch pull request runs it,
and no manual run of any branch can pass this job.

`FR-pre-release-validation-38a` states the requirement. `CHANGELOG.md` and the `## Changelog`
table of `docs/specs/spec.md` each carry `TBD` for the round.

Verified against
https://docs.github.com/en/actions/reference/workflows-and-actions/contexts (retrieved
2026-08-10): the `matrix` context reaches `jobs.<job_id>.steps[*].if`.

Member of #535. Reads #529.


Crank-Git:
## Self-review

Four lenses read the diff: correctness of the workflow step, the two new cases, the writing
standard, and `.claude/rules/batch-gate.md` beside the requirement text. Three findings
stand, and each one is repaired in the second commit.

**1. The step reader read the whole workflow file.** `tests/test_documentation_site.py`
split every `- name:` line of `.github/workflows/test.yml`, and each of the seven jobs
indents a step the same way. A step in the `lint` job would therefore have passed the case,
and `matrix.python-version` resolves to nothing in a job that runs no matrix. `_job_block`
now cuts the `test` job out first, and the case reads that block alone. The repair is
measured: a script moved the step into the `lint` job and
`test_one_job_of_the_test_matrix_installs_the_documentation_extra` failed.

**2. Two sentences passed 25 words.** Rule 1 of `.claude/rules/ste.md` holds the limit. The
docstring of `test_the_slug_of_a_case_matches_the_slug_of_the_build` carried a 29-word
sentence, and the comment of the new workflow step carried two of 26. Each one is now two
sentences.

**3. The `matrix.os` clause of the condition is redundant today.** `python-version` reads
`3.13` on the `ubuntu-latest` job alone, so `matrix.os` decides nothing. **The clause
stays.** The matrix carries an `include` list, and an entry that adds a second operating
system at Python 3.13 would otherwise install the extra on two jobs.

The other readings found nothing. `FR-pre-release-validation-38a` follows the letter-suffix
form the same file already carries at `FR-16a` and `FR-39a`. Both round records hold the
literal `TBD`. The head commit ends with the keyword once and names it nowhere else.

Issue #546, opened 2026-08-10
Record the twelfth required context of dev


Crank-Git:
The provider now requires twelve status checks on `dev`, and `.claude/rules/batch-gate.md` lists eleven. Two cases of `tests/test_batch_gate_protection_rule.py` fail against that drift.

## The measurement

A read of 2026-08-10 on `batch/544-release-blockers` at `61e04f1`, with an administrator token, reports the state. The two failures stand on the base commit and no change of #541 produces them.

```
FAILED tests/test_batch_gate_protection_rule.py::test_the_provider_requires_the_contexts_the_rule_file_lists
FAILED tests/test_batch_gate_protection_rule.py::test_the_provider_carries_the_stated_application_on_every_context
```

```
AssertionError: the provider requires ['conformance', 'fuzz', 'installed-wheel', 'lint', 'samples', 'skip-gate', 'test (macos-latest, 3.12)', 'test (ubuntu-latest, 3.10)', 'test (ubuntu-latest, 3.11)', 'test (ubuntu-latest, 3.12)', 'test (ubuntu-latest, 3.13)', 'test (ubuntu-latest, 3.9)'] and .claude/rules/batch-gate.md lists ['conformance', 'fuzz', 'installed-wheel', 'lint', 'samples', 'test (macos-latest, 3.12)', 'test (ubuntu-latest, 3.10)', 'test (ubuntu-latest, 3.11)', 'test (ubuntu-latest, 3.12)', 'test (ubuntu-latest, 3.13)', 'test (ubuntu-latest, 3.9)']
```

## The cause

`.claude/rules/batch-gate.md` states the warning that `skip-gate` is a twelfth check and that the required list holds eleven names. #524 wrote that sentence, and it names the user as the one who adds the check. **The user has now added it**, so the record is stale and the case reports the drift it exists to report.

## Why the runner does not report it

The provider grants the branch-protection read to a repository administrator alone. A runner holds no such token, so both cases skip there and `tests/universal_skips.json` allows the skip. **The failure reaches a local gate alone**, which is where a worker of the release batch meets it.

## What to build

- `.claude/rules/batch-gate.md` lists twelve required contexts, and it names `skip-gate` among them.
- The warning that names the eleven-name list reads the new state, and it records the read date.
- The two cases pass against `gh api repos/Crank-Git/ja4plus/branches/dev/protection` with an administrator token.

## Acceptance criteria

[...]

Issue #544, opened 2026-08-10
Batch: the release blockers and the 1.0.0 bump


Crank-Git:
## Members

- [x] #546 — Record the twelfth required context of dev
- [x] #512 — Read the version from `ja4plus/__init__.py` in the release skill
- [x] #541 — The skip gate refuses every manual run
- [x] #533 — Settle the ruling and decision vocabulary in the Terms table
- [x] #545 — Read the README and the documentation against version 1.0.0
- [x] #543 — Bump the version to 1.0.0 and move the development status

## This is the last batch before version 1.0.0

**The user rules on 2026-08-10 that the release is version 1.0.0 and that it happens today.**
Every epic of this project is closed, and `docs/specs/spec.md:545` names the one condition:
version 1.0.0 does not publish before Epic 11 closes. **It closed.**

## Two members are release blockers rather than repairs

**#512.** The release skill states that the version lives in `pyproject.toml`. **#67 reversed
that and the version lives at `ja4plus/__init__.py:101`.** An operator who follows the skill as
it stands bumps a file that declares no version.

**#541.** A manual run of `Tests` concludes `failure` on every branch, and
`.claude/rules/batch-gate.md` names the manual run as the recovery path. **A release that needs
that path finds it red.**

## Order

**#543 lands last.** The version bump follows the repair of the skill that describes it, and #545 reads the documentation against the version the bump declares.

**This batch holds six members against a batch size of four.** The user rules that the release happens today, and every member is a condition of that release.

## The integration branch

`batch/544-release-blockers`, off `dev` at `61e04f1`.

Crank-Git:
Six members, rounds 202 to 207, merged as https://github.com/Crank-Git/ja4plus/pull/564. The batch gate read a terminal successful run of every required workflow at the head commit `174d0f6`. **This batch declares version 1.0.0.**

PR #547, opened 2026-08-10
Read the version from ja4plus/__init__.py in the release skill

Crank-Git:
Refs #512. Member of batch #544.

## What this round repairs

**Step 1 of `.claude/skills/release/SKILL.md` named `pyproject.toml` as the one version
declaration, and it said that `ja4plus/__init__.py` reads the value with
`importlib.metadata`.** #67 reversed both directions two rounds earlier. The declaration is
the plain string assignment at `ja4plus/__init__.py:101`, and `pyproject.toml` names
`version` in its `dynamic` list and reads it from `ja4plus.__version__`. A maintainer who
followed step 1 edited a file that declares no version, so the release shipped the old
number.

**Step 2 held the same defect in a second form.** It read the version out of
`pyproject.toml` with `tomllib`, which the `[project]` table no longer holds. `tomllib`
also reaches no Python 3.9 job of the matrix.

## Every step of the file, and what this round did to it

| Step | Reading | Change |
|---|---|---|
| Before you start | Names `pytest tests/`, `ruff check`, `ruff format --check`, `mypy ja4plus/` and the `CHANGELOG.md` section. `[tool.mypy]` of `pyproject.toml` sets `strict = true` and `files = ["ja4plus"]`, so `mypy ja4plus/` runs the gate the project names. | None. |
| 1 | Named `pyproject.toml` as the version declaration. | Rewritten. |
| 2 | Read `pyproject.toml` with `tomllib`. | Rewritten. |
| 3 | Runs `python -m tests.release_verification --dist dist`. That module reads the built artifacts and no version declaration. | None. |
| 4 | Runs `python -m pytest tests/ -m installed_wheel -q`. `pyproject.toml` declares that marker and `tests/test_installed_wheel.py` reads `package_version` of `tests/version_gate.py`. | None. |
| 5 | Runs `python -m tests.release_body --output <path>`. That module already reads the `## [<version>]` section for the version `ja4plus/__init__.py` declares. | None. |
| 6 | Runs `gh run list` and `pip download`. Neither reads a version declaration. | None. |
| How to run the dry run | Runs `gh workflow run publish.yml`. | None. |
| Rules | States three rules about a partial publish, a bump and version 1.0.0. | None. |

**The table of `### The version count, measured on 2026-08-09 and settled on 2026-08-10` in
`docs/specs/features/09-release.md` reads `The version declaration at line 94` for
`ja4plus/__init__.py`.** That table records a read of 2026-08-09 and this round rewrites
none of it.

## What step 1 carries now

1. **The version lives in `ja4plus/__init__.py` at line 101, and a bump edits that line
   alone.**
2. **`setuptools` reads the value from the syntax tree of the module**, so the assignment
   stays a plain string. #67 measured that a computed value makes a build import the
   package and every dependency it loads, and that failure appears at release time.
3. **A bump needs a matching `## [<version>]` section of `CHANGELOG.md`.**
   `tests/release_body.py` reads that section, so an absent section fails step 5.

Step 2 now reads `package_version`, `version_disagreement` and `changelog_disagreement` of
`tests/version_gate.py`. That is the reader `tests/test_version_gate.py` holds cases
against, so the manual check and continuous integration read one condition.

## The case set

**A skill file carried no case until this round, and that is why the defect survived two
rounds.** `tests/test_release_skill.py` holds 15 cases, and every one of them reads the
repository rather than a second copy of the skill text.

- The source cases open the file the skill names, read the line the skill names, and
  require `PACKAGE_DECLARATION` of `tests/version_gate.py` to match it. One case requires
  `declaration_files` to report that same file and no other.
- The command case extracts the fenced block of step 2, removes the indent of the numbered
  step, runs it in a subprocess at the repository root, and requires the version that
  `ja4plus/__init__.py` declares in the output.
- The rule cases read the plain string rule and the changelog rule.
- The discrimination cases read the superseded text of step 1, a step that names
  `pyproject.toml`, a step that names the wrong line, and a step that names a file the
  repository holds nowhere.

**Every pattern of the file joins its words with `\s+`**, because the skill wraps at 90
columns and a literal space matches no sentence that wraps between two of its words.

## How this was tested

**Eight of the 15 cases failed against the unrepaired file**, which is the failing case the
test-driven practice asks for. The first of them reports this line.

```text
assert 'the release skill names no version source and no line' is None
```

**A mutation that returns step 1 to `pyproject.toml` fails three cases, and the mutation is
restored.** The first of them reports this line.

```text
assert 'line 20 of `pyproject.toml` reads \'dynamic = ["version"]\' and declares no `__version__`' is None
```

The five gates run green on this branch, on Python 3.11 in a virtual environment outside
the worktree.

| Gate | Result |
|---|---|
| `pytest tests/ -m "not spec_validation"` | 4175 passed, 7 skipped, 8 xfailed, 2 failed |
| `ruff format --check ja4plus/ tests/` | 221 files already formatted |

**The two failures belong to `tests/test_batch_gate_protection_rule.py` and #546 holds
them.** The provider now requires twelve contexts and `.claude/rules/batch-gate.md` lists
eleven, so the two cases report `At index 5 diff: 'skip-gate' != 'test (macos-latest,
3.12)'`. This change set touches neither file, and the failure stands on the base commit.

**No file under `ja4plus/` changes and no fingerprint moves**, so coverage falls nowhere.

Verified against
https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata
(retrieved 2026-08-10).


Crank-Git:
## Self-review

A correctness lens read the change set. It found no defect that changes behaviour. It
raised two readings, and this comment records both.

**1. `CHANGELOG_RULE` held a literal space where every other pattern of the file joins its
words with `\s+`.** The space stands inside the code span `` `## [<version>]` ``, and an
author who wrapped that span would break the rendering, so the pattern was not wrong
today. The file states the rule in a comment on itself, so an exception to it is a reading
a later editor has to make again. The pattern now reads
`` matching\s+`##\s+\[<version>\]`\s+section\s+of\s+`CHANGELOG\.md` ``. 106 cases of the
six affected files pass after the change.

**2. `tests/release_verification.py` copies `tests` and `pyproject.toml` outside the
checkout, and it copies no `.claude/` directory.** A run inside that copy would raise
`FileNotFoundError` on `SKILL.read_text()` rather than skip. **That run collects no case of
this file today.** `run_conformance` names its targets, and every one of them carries the
`spec_validation` marker. This round adds no guard, because a hard failure names the cause
where a skip would leave the reader with a green report. **A skip that reached every job
would also fail the skip gate**, and `.claude/rules/batch-gate.md` states that rule. A
change that widens the copied selection to `pytest tests/` reads this note first.

**No finding on the other three lenses.** No case passes by construction, each pattern
matches once in the file and nothing in the superseded text, and the heredoc of step 2 runs
from the repository root and uses no syntax later than Python 3.9.

Issue #541, opened 2026-08-10
The skip gate refuses every manual run


Crank-Git:
Round 197 added the `skip-gate` job. **A manual run of `Tests` now concludes `failure` on every branch, and the failure names no defect of that branch.**

## The measurement

Run https://github.com/Crank-Git/ja4plus/actions/runs/31421263768 reads `event=workflow_dispatch` and `conclusion=failure`, and its `skip-gate` job reads `failure`. **Every other job of that run concluded `success`.**

## The cause

A `workflow_dispatch` run carries no pull request. The step that fetches the base commit runs under a pull-request condition, so it runs nowhere, and `tests.test_round_entry_existence::test_the_change_set_of_this_branch_records_a_round` skips on all six matrix jobs. **The skip gate reads that as a case the suite ran nowhere, which is exactly what it exists to refuse.**

**The gate is right about what it measures and wrong about what it concludes.** A case that reads a pull-request event runs on a pull-request event, and a manual run is a different event rather than a suite that ran the case nowhere.

## Why this matters beyond a red badge

**`.claude/rules/batch-gate.md` names the manual run as the recovery path.** The section `Start a run by hand where the gate refuses` states that `workflow_dispatch` starts a run through the provider interface where a push event and a pull-request event created none. **That procedure now produces a red run every time**, so a reader cannot tell a real failure from this one.

A push event is unaffected, and that is measured rather than assumed: the push run of `dev` at `31cca24` holds the `skip-gate` job and concluded `success`.

## What to build

**The gate reads the event.** A case that a given event does not select is no finding of that event. The reading is free to take one of these, or a better one.

- The `skip-gate` job runs on a pull-request event and a push event, and it runs nowhere on a manual event.
- The gate holds a set of cases that read a pull-request event, and it accepts their skip where the run carries no pull request.
- The case reads a reference the manual run also holds, so it stops skipping there.

**The third reading is the strongest and it costs the most**, because it makes the case run on every event rather than excusing it on one.

## Acceptance criteria

- A manual run of `Tests` on a branch that holds no defect concludes `success`.
- The gate still refuses a case that skips on every job of a pull-request run. Prove it with the same deliberately universal skip round 197 used, and report both run URLs.
- `.claude/rules/batch-gate.md` states the reading beside the recovery procedure it documents.
- No file under `ja4plus/` changes and no fingerprint moves.

PR #549, opened 2026-08-10
Prove the skip gate refuses a universal skip (#541)

Crank-Git:
This pull request lands in no branch. #541 opens it to read one run and then closes it with its branch deleted.

**A green manual run alone proves nothing.** A repair that turned the skip gate off would produce one too. This branch carries the deliberately universal skip that round 197 used, `tests.test_universal_skip_proof::test_the_case_that_skips_in_every_environment`, so the run must refuse it.

The head commit carries no skip keyword, and the base is `batch/544-release-blockers` and neither `dev` nor `master`.

Expected result: every other job concludes `success` and the `skip-gate` job concludes `failure`, naming that case.


Crank-Git:
The run is read and recorded. Run https://github.com/Crank-Git/ja4plus/actions/runs/31427370413 concluded `failure` on `skip-gate` and `success` on every other job. The census names `tests.test_universal_skip_proof::test_the_case_that_skips_in_every_environment` with no allowlist entry, held by six reports. This pull request lands in no branch.

Issue #480, opened 2026-08-10
The batch-gate rule states dev carries no required status check, and it now does


Crank-Git:
## What is stale

**#468 closed on 2026-08-10 and `dev` now carries a required status check.**
`.claude/rules/batch-gate.md` states the opposite, in a section headed
`The strongest shape needs the user`.

> **A required status check refuses the merge inside the provider, and this repository holds
> none.** A read of 2026-08-09 reports the state.

| Read | 2026-08-09 | 2026-08-10 |
|---|---|---|
| `gh api repos/Crank-Git/ja4plus/branches/dev/protection` | `404`, `Branch not protected` | `200`, eleven contexts |
| `gh api repos/Crank-Git/ja4plus/rulesets` | `[]` | `[]` |

The eleven contexts are the eleven the file already lists, each carrying `app_id` 15368. No
context reads `build` and none reads the bare `test`.

**The file is a rule a reader follows, and not a dated record.** A reader who follows it today
concludes that nothing at the provider refuses an ungated merge, and that conclusion is now
false.

## What to change

- **State that the rule is live**, with the date of the read and the eleven names, and keep the
  two warnings that make the list correct: a check name is not a job name, and `build` stays
  out because `docs-build.yml` filters four paths.
- **Keep the 2026-08-09 reading as the dated record it is.** A past measurement is quoted, not
  rewritten. Mark it as superseded rather than deleting it, in the shape this repository
  already uses.
- **Record the two limits the same call returned**, because each one bounds what the rule
  proves.
  - **`enforce_admins` reads `false`.** The file says a required check binds everyone. That is
    true for a contributor and not true for a repository administrator.
  - **`strict` reads `false`**, so a branch need not be up to date with `dev` before it merges.
    **State that this suits the batch model**: `strict: true` would demand every integration
    branch take `dev` again after another batch lands, and the run that proved the batch would
    then not be a run of the head.
- **Keep `python -m tests.batch_gate --pr <number>` in the procedure.** It reads the same
  condition the provider now holds, and it reads it before the merge rather than at it.
- **Check `CLAUDE.md`**, which names `.claude/rules/batch-gate.md`, for any sentence that rests
  on the absent-protection reading.

## Acceptance criteria

- [ ] `.claude/rules/batch-gate.md` states that `dev` carries the required status check, with
      the date of the read.
- [ ] The 2026-08-09 reading is present and marked superseded, not deleted.
[...]

PR #560, opened 2026-08-10
Record the twelfth required context of dev

Crank-Git:
## What

`.claude/rules/batch-gate.md` stated eleven required contexts on `dev`, and the provider
holds twelve. This pull request records the twelfth, `skip-gate`. It changes the record
and it changes no case assertion.

Refs #546.

## Why

The user added `skip-gate` to the branch protection rule of `dev` on 2026-08-10. Two cases
of `tests/test_batch_gate_protection_rule.py` measured the drift and failed on `dev` at
`61e04f1`:

```
FAILED tests/test_batch_gate_protection_rule.py::test_the_provider_requires_the_contexts_the_rule_file_lists
FAILED tests/test_batch_gate_protection_rule.py::test_the_provider_carries_the_stated_application_on_every_context
```

**A red case here turns every batch pull request red.** The two cases are correct and they
report a real change, so the repair is the record.

## The measurement

`gh api repos/Crank-Git/ja4plus/branches/dev/protection` on 2026-08-10 returns `200` with
twelve required contexts.

```
lint
test (ubuntu-latest, 3.9)
test (ubuntu-latest, 3.10)
test (ubuntu-latest, 3.11)
test (ubuntu-latest, 3.12)
test (ubuntu-latest, 3.13)
test (macos-latest, 3.12)
fuzz
samples
installed-wheel
conformance
skip-gate
```

Every entry of `required_status_checks.checks[]` carries `app_id` 15368, including
`skip-gate`. `enforce_admins` reads `false` and `strict` reads `false`, which the file
already states. `gh api repos/Crank-Git/ja4plus/rulesets` returns `[]`.

Verified against
https://docs.github.com/en/rest/branches/branch-protection (retrieved 2026-08-10).

## What the record now carries

1. **The twelve names**, and the count wherever the file states one.
2. **A statement of what `skip-gate` refuses.** The job downloads the report of every job
   that runs cases, and it fails a case that every one of those reports records as skipped.
   Round 201 built the job.
3. **A dated record of the change.** The protection section holds two reads of one date,
   so each record names the issue that took it: #480 read eleven contexts and #546 read
   twelve. The eleven-context reading stays as a quoted record under a paragraph that marks
   it superseded, which is the form the 2026-08-09 reading already holds.

## The records this change leaves as they stand

`CHANGELOG.md` Round 197 and the Round 174 row of `docs/specs/spec.md` each state the
eleven. Both are dated records of a past reading, so this round supersedes them rather
than rewrites them. `.github/workflows/test.yml` and `docs/specs/features/09-release.md`
each stated the eleven in live prose, and both now state the twelve.

## How this is tested

New case `test_the_rule_file_states_what_the_twelfth_required_context_refuses` requires a
sentence that names `skip-gate` beside what it refuses, and it requires the round
reference. It failed before the repair.

```
3 failed, 117 passed   (before the repair)
120 passed             (after it)
```

Local gates, run in a virtual environment outside the worktree:

- `pytest tests/ -m "not spec_validation"` — 4162 passed, 7 skipped, 8 xfailed
- `pytest tests/ -m spec_validation` — 1635 passed, 143 skipped, 140 xfailed
- `ruff format --check ja4plus/ tests/` — 220 files already formatted
- `mypy --strict ja4plus/` — no issues in 31 source files


Crank-Git:
## Self-review

A specialist review read the diff under three lenses: the writing standard, the internal
consistency of the record, and the strength of the cases. Commit `8f18fa9` holds every
repair.

### Repaired

**The new case was satisfiable by prose that named the check and stated no condition.**
`sentences` splits on a full stop and a space, and the check-name list holds neither, so
that list joined the sentence beside it and supplied the name from a fenced block. Prose
such as `` `skip-gate` exists, and this file refuses to explain more.`` would have passed.
Two repairs close it.

1. `prose_lines` cuts every fenced block before the case reads a sentence, and
   `test_the_prose_reader_cuts_a_check_name_of_the_fenced_list` proves the cut.
2. The case requires the word `skipped`, which names the input that turns the check red.

**Two mutations prove the case bites.** Prose that drops `skipped` fails the second
assertion at `tests/test_batch_gate_protection_rule.py:834`. Prose that names no check
fails the first at `:830`.

**Three compound sentences of `.claude/rules/batch-gate.md` each carried two facts.** Each
one is now two sentences: the reader list at line 292, the name-list warning at line 364
and the `skip-gate` statement at line 382.

### Declined

**A line-width finding of seven lines at 91 and 92 columns.** The base file already holds
26 non-table lines past 90 columns and `CHANGELOG.md` holds 593, so 91 and 92 columns are
the shape of these two files. The finding names no rule of `.claude/rules/ste.md`. The
table rows it reports are longer still, and every existing table row of the same section
is longer than 90 columns.

**A finding that the two comparison tables repeat the eleven-context reading.** The
suggested repair rewrites the right-hand column of the first table from the #480 reading
to the #546 reading. That table is the record of the 2026-08-09 read against the
2026-08-10 read of #480, so the repair would rewrite a dated record of a past measurement.
`.claude/rules/ste.md` and this file both hold that such a record is quoted rather than
rewritten. The second table records the new supersession beside it, which is the form the
2026-08-09 record already holds.

### Read and found clean

No live sentence of the repository states eleven required contexts. Every remaining
`eleven` is attributed to #480, to #524 or to the 2026-08-09 read. The fenced list holds
the twelve names the provider returns. No assertion of the two failing cases was weakened:
the floor assertions and the value-equality assertions stand as they were, and only the
recorded reading and the message strings moved.

Issue #568, opened 2026-08-10
Batch: the last three repairs and version 1.0.1


Crank-Git:
## Members

- [x] #566 — Write an absolute link in the release body
- [x] #562 — State the interface table and the migration sentence a reader can use
- [x] #548 — Reach the ruling vocabulary case to the Python prose corpus
- [x] #567 — Bump the version to 1.0.1

## This batch empties the board

**Version 1.0.0 published on 2026-08-10 and these are every issue that remained open.** The user asks for a clear board and for version 1.0.1 to carry the repairs.

## The hazard of this batch

**#548 edits files under `ja4plus/`, and it is the first change to the package since the release.** Nine instances of the barred noun sit in comments and docstrings of `ja4plus/fingerprinters/ja4ssh.py`, `ja4plus/utils/http_utils.py` and `ja4plus/fingerprinters/ja4l.py`.

**A prose change under `ja4plus/` is proven and not asserted.** #484 records the technique: parse each changed file at the base and at the head, strip every docstring, and compare the syntax trees. **Statement-identical trees prove the change is prose.**

## Order

**#567 lands last.** The version bump follows the repairs it carries.

## The integration branch

`batch/568-final-repairs`, off `dev` at `d0d7c81`.

Crank-Git:
Four members, rounds 208 to 211, merged as https://github.com/Crank-Git/ja4plus/pull/573. The batch gate read a terminal successful run of every required workflow at the head commit `6b37a0e`. **This batch declares version 1.0.1 and it empties the issue board.**

Issue #548, opened 2026-08-10
Reach the ruling vocabulary case to the Python prose corpus


Crank-Git:
#533 settled the vocabulary on `ruling` and built `tests/test_ruling_vocabulary.py`. That case reads the Markdown corpus and `docs/specs/spec.html`, and it reads no Python file.

**The acceptance criteria of #533 state that no file under `ja4plus/` changes.** A read of 2026-08-10 finds the noun `decision` in nine comments and docstrings under `ja4plus/`.

| Record | Count |
|---|---|
| `ja4plus/fingerprinters/ja4ssh.py` | 5 |
| `ja4plus/utils/http_utils.py` | 3 |
| `ja4plus/fingerprinters/ja4l.py` | 1 |

A reader of the Python corpus would therefore fail a case that #533 bars the repair of. #533 states the hole in the module docstring of `tests/test_ruling_vocabulary.py` rather than leave it for a later reader to find.

`tests/` holds about forty more uses of the noun, and no rule of #533 bars a change there.

## What to build

- Every comment and every docstring under `ja4plus/` and `tests/` names a determination of the user `ruling`.
- `tests/test_ruling_vocabulary.py` reads the Python corpus beside the Markdown corpus, with `python_prose` of `tests/test_documented_method_count.py`.
- The module docstring drops the paragraph that records this hole.

## Acceptance criteria

- A case fails where a comment or a docstring of `ja4plus/` or `tests/` uses the barred word. Prove it with a mutation that is restored.
- No fingerprint moves.

**Warning: a change under `ja4plus/` touches comments and docstrings alone.** No line of code moves.

Crank-Git:
**The hole reaches prose outside the Python corpus as well, and #533 measured it at the sub-merge gate.** `tests/test_ruling_vocabulary.py` reads the tracked Markdown pages and `docs/specs/spec.html`. It reads no other file type, so prose inside a configuration file or a JSON fixture reaches no case.

A read of 2026-08-10 on the merge result of `batch/544-release-blockers` names three records.

| Record | Count | Shape |
|---|---|---|
| `tests/universal_skips.json` | 4 | The `reason` field of four entries reads `records the decision that the case skips there and passes never`. |
| `pyproject.toml:62` | 1 | A comment reads `#446 records the decision`. |
| `tests/foxio_deviations.json` | several | The `cause` field of an entry reads `and the decision is reversible`. |

[...]

Issue #533, opened 2026-08-10
Settle the ruling and decision vocabulary in the Terms table


Crank-Git:
`.claude/rules/ste.md` states the rule: "The project's controlled vocabulary is the
`## Terms` table in `docs/specs/spec.md`. Read it before you write a domain word. ... When
you need a word the table does not hold, add it to the table."

**Two words carry a recurring concept of this project, and the table holds neither.**

- `ruling`, one determination the user makes where no source settles the question.
- `decision`, which the divergence register uses for the same concept.

The table holds 91 terms today, and `reading` is the nearest one. It reads "One recorded
conclusion about what a source states, with the evidence that supports it." **That is a
conclusion about a source, and a ruling is a determination of a person**, so `reading` covers
neither word.

## Where the two words already stand

| Record | Reading |
|---|---|
| `docs/specs/foxio/JA4D.md:444` | The heading `### JA4D — the rulings` |
| `docs/specs/foxio/JA4D.md:559` | The heading `### JA4D6 — the rulings` |
| `docs/specs/foxio/JA4D.md:618` | The heading `### JA4D6 — the rulings of #271` |
| `docs/specs/features/09-release.md:80` | "records the measurement and the ruling" |
| `docs/specs/features/09-release.md:209` | "words of the ruling" |
| `CHANGELOG.md` | Eleven entries use `ruling` as a noun |
| `docs/specs/spec.md` § Divergence register | About twenty rows close with "The decision is reversible." |

**The two words therefore rotate against each other across the specification**, which rule 7
of the writing standard bars: "One concept, one word. Never rotate synonyms for variety."

## What to build

- The `## Terms` table names one of the two words, with one meaning, and names the other in
  its `Do not use` column.
- Every live record follows the choice.
- A case reads the table against the tracked document set of git, so a later rotation fails
  rather than passing quietly.

**Warning: read the document set with `git ls-files`, and never with an `rglob` over the
repository root.** #473 records the cause: a reader that walks the checkout picks up a worker
worktree under `.claude/`, and the corpus then grows with the number of live workers.

## Acceptance criteria

- The `## Terms` table holds one row for the chosen word, and that row names the other word
  under `Do not use`.
- A case fails where a tracked document uses the barred word. Prove it with a mutation that
  is restored.
- No file under `ja4plus/` changes and no fingerprint moves.

#522 found this gap in its self-review and declined it as wider than one spec record.

Crank-Git:
[...]

PR #571, opened 2026-08-10
Reach the ruling vocabulary case to the Python prose corpus

Crank-Git:
Refs #548. Batch member of #568, into `batch/568-final-repairs`.

## What

Round 205 built `tests/test_ruling_vocabulary.py` over the Markdown corpus, and its module
docstring stated the hole rather than hid it: the case read no Python file, because #533
barred a change under `ja4plus/`. This member extends that reader — it writes no second one
— and it sweeps the prose the reader then reports.

## Why a Python file reads as prose and not as text

`python_prose` of `tests/test_documented_method_count.py` parses the file and returns the
comments and the docstrings alone. `without_quoted` then cuts every code span, every fenced
block and every blockquote line, exactly as it cuts one of a Markdown page.

**A string literal that is no docstring reaches no case, and the reader leaves it exactly as
it reads.** Such a literal is data and not prose: a case compares it, an assertion states it,
a schema names it, or a fixture holds the sentence the case measures. The verbatim list of
`.claude/rules/ste.md` covers each of those. `tests/test_foxio_deviations.py` is the standing
control, because lines 39, 578 and 580 each hold `records the decision` inside a literal and
that file passes the new case.

**An identifier reaches no case either**, because `python_prose` reads no code.
`unmarked_decisions`, `DECISION_CITATION`, `DECISION_ISSUE` and the field `decided_by` of
`ja4plus/fingerprinters/ja4ssh.py` all stay.

**The corpus comes from `git ls-files` and it walks no directory**, because #473 measured
that a walk reads the sources of every live worker worktree. The pathspec is `*.py` and it
names 227 tracked sources: 31 under `ja4plus/`, 191 under `tests/` and 5 under `examples/`,
which hold none of the barred word. The package carries its own floor of 25 beside the corpus
floor of 120, because 191 suite files alone pass the wider floor.

**The pattern stays `\bdecisions?\b` and it matches no verb form.** Rule 6 states one word,
one meaning, one part of speech, so `decides` and `decided` stay. The `decided` schema key of
`tests/foxio_deviations.json` does not move, and the rule of `CLAUDE.md` that the
specification decides intent and schema keeps its verb.

## The sweep

25 sources failed the new case. The sweep took the barred noun out of 84 places across them:
9 under `ja4plus/` and 75 across 22 files under `tests/`. The nine under the package are 5 in
`ja4plus/fingerprinters/ja4ssh.py`, 3 in `ja4plus/utils/http_utils.py` and 1 in
`ja4plus/fingerprinters/ja4l.py`.

**A self-review found that 15 of the 84 named something other than a determination of the
user, and each of the 15 names the mechanism instead.** Round 205 met the same class once, at
`docs/specs/spec.html:92`, and reverted it there. The other 69 take the controlled term.
Three sites of `ja4plus/fingerprinters/ja4ssh.py` and one of
`tests/test_ja4ssh_server_side.py` name the endpoint rules, because `_decide_endpoints` picks
the client and the server from the port, the handshake and the lower port. The other 11 stand
in `tests/test_documentation_fallback_rule.py`, where the word named precedence handed to a
source, and each now reads `authority` or `authority clause`. The review comment on this pull
request holds the whole reading.

## The failing case, stated first

Two mutations prove the case bites, and both are restored.

```
AssertionError: ja4plus/fingerprinters/ja4l.py names 'decision' at ['ne as a reference
value. #225 holds the decision'], and the `## Terms` row of 'ruling' rejects that word
```

```
AssertionError: tests/test_ja4x_raw.py names 'decision' at ['the same value, and #267
holds the decision'], and the `## Terms` row of 'ruling' rejects that word
```

Two floors fail a reader that collected nothing, because an aggregate over an empty set
passes. The package floor reports `the corpus holds 0 sources under ja4plus/, below the floor
of 25, and the suite alone passes the floor above`.

## The proof this change needs beyond the five gates

**A change under `ja4plus/` is proven and not asserted.** This member repeats the technique of
#484. Each of the three changed package files parses at the base commit `d0d7c81` and at the
head, every docstring comes out of both trees, and `ast.dump` reports the two trees identical.

| File | Result |
|---|---|
| `ja4plus/fingerprinters/ja4l.py` | identical |
| `ja4plus/fingerprinters/ja4ssh.py` | identical |
| `ja4plus/utils/http_utils.py` | identical |

**No fingerprint moves, and this member measures that.** The conformance suite reports the
same three counts before and after.

| Read | Before | After |
|---|---|---|
| `pytest tests/ -m spec_validation` | 1635 passed, 143 skipped, 140 xfailed | 1635 passed, 143 skipped, 140 xfailed |

**A docstring carries no statement, so the coverage count does not move.**

| Read | Before | After |
|---|---|---|
| `--cov=ja4plus` statements | 4316 | 4316 |
| Missed | 273 | 273 |
| Percent | 94 | 94 |

The unit suite runs 4355 cases before and 4595 after. The 240 new cases are the 227
parametrized sources, 4 anchors and 9 readings of the reader itself.

## The five gates

| Gate | Result |
|---|---|
| `pytest tests/ -m "not spec_validation"` | 4595 passed, 7 skipped, 8 xfailed |
| `ruff format --check ja4plus/ tests/` | 223 files already formatted |

## Records

`CHANGELOG.md` and the `## Changelog` table of `docs/specs/spec.md` each carry `TBD` in place
of the round number. The project manager assigns the number at the batch gate.

Verified against https://spec.commonmark.org/0.31.2/#fenced-code-blocks (CommonMark 0.31.2,
retrieved 2026-08-10), which round 205 already cites for `FENCE_DELIMITER`.


Crank-Git:
## Self-review

A specialist review of the whole diff ran on the correctness of the reader, the sense of
every swapped word, readability and the writing standard. It found three real classes and
this push repairs all three. One further finding is declined below with the reason.

### Finding 1, blocking. 15 of the 84 swapped places named no determination of the user.

Round 205 met this class once, at `docs/specs/spec.html:92`, and reverted it there rather
than swapped it. The review found 15 more.

**Three sites under `ja4plus/fingerprinters/ja4ssh.py` name the endpoint rules and no
person.** `_decide_endpoints` picks the client and the server from the port, from the
handshake and from the lower port, and its docstring opens `Three rules decide`. A program
issues no ruling.

| Site | Now reads |
|---|---|
| `ja4plus/fingerprinters/ja4ssh.py:160` | `move the endpoint pair` |
| `ja4plus/fingerprinters/ja4ssh.py:208` | `never from the endpoint rules again` |
| `ja4plus/fingerprinters/ja4ssh.py:433` | `the rule that named the server` |
| `tests/test_ja4ssh_server_side.py:147` | `names the rule that named the server` |

`ja4plus/fingerprinters/ja4ssh.py:388` and `:623` name #214 and #262, which are rulings of
the user, and both keep the controlled term.

**The other 11 stand in `tests/test_documentation_fallback_rule.py`, where the word named
precedence handed to a source.** That file reads the vector fallback, which gives the
expected-output files the authority over an ambiguous image, and an expected-output file
issues no ruling. Each of the 11 now reads `authority` or `authority clause`, which is the
word that file already binds in `AUTHORITY`, `AUTHORITY_SOURCE` and `AUTHORITY_FIRST`.

**The barred word reaches no case there either**, so each repair states the mechanism rather
than restores it. 69 of the 84 places take the controlled term.

### Finding 2, blocking. `needs no verb of ruling` reads as nonsense.

`tests/test_documentation_fallback_rule.py:200` now reads `needs no authority verb after the
expected-output files`. It falls out with finding 1.

### Finding 3, worth fixing. Four sentences of the new module prose passed 25 words.

`.claude/rules/ste.md` exempts two records from rule 1, and this module is neither. Each of
the four is now two sentences, and the list of four literal shapes becomes a vertical list
under rule 16.

### Finding 4, note. The record stated a line count where a place count is stable.

The record said `repaired 84 lines`. The repairs of finding 1 rewrapped several comments, so
a line count moves where the count of swept places does not. Both records now read `took the
barred noun out of 84 places`.

### Declined: the reader has no undisclosed blind region.

The review read `python_prose` against six shapes: an inline comment on a line of code, a
module docstring, a nested function docstring, a file that fails to parse, a fenced block
inside a docstring, and a bare string below a class attribute. Five reach the reader. The
sixth is a real limit, because `ast.get_docstring` reads the first statement of a node
alone, and the module docstring now states it. **A read of 2026-08-10 counted zero such
strings over the 227 sources**, so the limit costs this reader nothing today. A file that
fails to parse raises `SyntaxError`, which fails the case loudly rather than skips it.

### The gates after the repairs

| Gate | Result |
|---|---|
| `pytest tests/ -m "not spec_validation"` | 4595 passed, 7 skipped, 8 xfailed |
| `ruff format --check ja4plus/ tests/` | 223 files already formatted |

The syntax-tree proof holds after the repairs. All three changed package files report
`identical` against `d0d7c81` under a docstring strip. Coverage reports 4316 statements, 273
missed and 94 percent, which is the base reading.

Issue #589, opened 2026-08-11
Cut a code span before the floating-date reader matches


Crank-Git:
`floating_sentences` of `tests/test_batch_gate_protection_rule.py` reads a code span as prose, so an identifier that holds a barred word fails a case that no prose sentence broke.

## The measurement

#586 added this sentence to `.claude/rules/batch-gate.md`.

```
The command costs 40 KB, after `git gc --prune=now` on each read.
```

The case reported the sentence as a floating date.

```
AssertionError: .claude/rules/batch-gate.md dates these sentences against the day a reader reads them, and this file records the cost of a reading quoted as a standing fact: ['The command costs 40 KB, after `git gc --prune=now` on each read.']
```

**The sentence names no reading day.** `FLOATING_DATE` holds `\bnow\b`, and `--prune=now` carries that word inside a code span, because `=` closes a word boundary. The comment beside the pattern already states the neighbouring case: "`\bnow\b` matches no `nowhere`, because the boundary needs the word to end."

## Why this is a defect and not a finding

**`.claude/rules/ste.md` states that code, configuration, commands, identifiers and paths are reproduced verbatim**, under `## What is not rewritten`. A reader that measures prose therefore reads no code span. `without_quoted` of `tests/test_documented_method_count.py` already cuts a code span, a fenced block and a blockquote line, and #548 records that reader.

**#586 worked around it** and rewrote the sentence as `A read of 2026-08-10 puts the cost at 40 KB`, which is better prose anyway. The reader still reads the next `--prune=now`, `HEAD~1`, or any other identifier that holds a barred word, and the next writer meets the same false report.

## What to build

`floating_sentences` cuts every code span before it matches `FLOATING_DATE`. `readable_text` is the place, or a call of the existing `without_quoted` reader.

## Acceptance criteria

- A sentence of `.claude/rules/batch-gate.md` that holds `` `git gc --prune=now` `` and names no reading day passes `test_no_live_sentence_of_the_rule_file_dates_itself_against_the_reader`.
- A sentence that holds the bare word `now` outside a code span still fails that case.
- A case holds each of the two readings above, and the first one fails before the repair.

Issue #587, opened 2026-08-11
Batch: the shallow-fetch race


Crank-Git:
## Members

- [x] #586 — Fetch the two recorded change sets in one command
- [x] #589 — Cut a code span before the floating-date reader matches

## Why this batch exists after version 1.1.0

**The push of `master` at `2129964` failed**, and the cause is a race between two shallow fetches of one step rather than a defect of the release. Version 1.1.0 is published and correct; the runner is red on a step that carries no case.

**A failure of that step skips every case of the job**, so the repair matters more than its size.

## The integration branch

`batch/587-shallow-fetch-race`, off `dev` at `90b93e2`.

Crank-Git:
Two members, rounds 216 and 217, merged as https://github.com/Crank-Git/ja4plus/pull/591. The batch gate read a terminal successful run of every required workflow at the head commit `026bb59`.

Issue #586, opened 2026-08-11
Fetch the two recorded change sets in one command


Crank-Git:
The `test` job fails intermittently, and the failure is a race between two shallow fetches.

## The measurement

Run https://github.com/Crank-Git/ja4plus/actions/runs/31452971377, a push of `master` at `2129964`, job `test (ubuntu-latest, 3.13)`.

```
* [new ref]  46aa502ca47f3c29f3c5ece15e4e78500e2f59c5 -> refs/ja4plus/recorded-defect
fatal: shallow file has changed since we read it
##[error]Process completed with exit code 128.
```

**The first fetch succeeded and the second one failed.** The step runs two commands, and each one rewrites `.git/shallow` of a clone of depth 1.

```
git fetch --depth=2 origin "+$DEFECT_SHA:refs/ja4plus/recorded-defect"
git fetch --depth=2 origin "+$CONTROL_SHA:refs/ja4plus/recorded-control"
```

**A failure of this step skips every case of the job**, and `skip-gate` then reports nothing because it never runs. One earlier run of pull request #585 failed at the same step, and a re-run passed, which is the shape of a race rather than of a broken reference.

## What to build

**One fetch that carries both refspecs**, so one command reads and writes the shallow file once.

```
git fetch --depth=2 origin "+$DEFECT_SHA:refs/ja4plus/recorded-defect" "+$CONTROL_SHA:refs/ja4plus/recorded-control"
```

**Read whether the step above it races the same way.** `Resolve the reference commit of a run that carries no pull request` and `Fetch the base commit of the pull request` each fetch too, and a third shallow write stands between the checkout and this step. **State whether those need the same repair or why they do not.**

**Prove it on the runner.** A single green run proves little against a race, so **report several runs of the same head**, and say how many you read.

## One reading recorded beside this issue

**Two tags now pin the commits this step fetches**, `record/412-defect` and `record/429-control`, because `46aa502` was reachable from no branch after the branch sweep of 2026-08-10. **That reachability was never proven to be the cause of any failure**, and the project manager stated it as one before reading the error. The tags stand on their own: a commit that a workflow fetches is a commit some reference must hold.

**Name the tags in the workflow rather than the bare identifiers**, so a reader of that step sees what the step depends on, and a later branch sweep finds the reference.

## Acceptance criteria

- The step runs one `git fetch` command, and it names both refspecs.
- The step names the two tags beside the identifiers, so the dependency is readable.
[...]

PR #590, opened 2026-08-11
Cut a code span before the floating-date reader matches

Crank-Git:
Member of batch #587. Refers to #589.

## What

`floating_sentences` of `tests/test_batch_gate_protection_rule.py` cuts every code span before it matches `FLOATING_DATE`.

## Why

The reader read a code span as prose, so an identifier that holds a barred word failed a case that no prose sentence broke. #586 wrote a sentence that states the cost of a fetch and names the command `git gc --prune=now`, and the case reported it.

```
AssertionError: .claude/rules/batch-gate.md dates these sentences against the day a reader reads them, and this file records the cost of a reading quoted as a standing fact: ['The command costs 40 KB, after `git gc --prune=now` on each read.']
```

The sentence names no reading day. `FLOATING_DATE` holds `\bnow\b`, and `--prune=now` carries that word inside a code span, because `=` closes a word boundary. `.claude/rules/ste.md` reproduces code, configuration, commands, identifiers and paths verbatim, under `## What is not rewritten`, so a reader that measures prose reads no code span.

## How

This round writes no second reader. `floating_sentences` calls `without_quoted` of `tests/test_ruling_vocabulary.py`, which #548 records and round 210 repaired. The call reaches the match alone, so the reader reports the sentence the file holds and never the cut form.

## How it is tested

Sixteen cases came first and eight of them failed against the committed reader.

- Six parameter sets of `test_the_floating_date_reader_reads_no_barred_word_of_a_code_span`, which name `git gc --prune=now`, `git log --since=now`, `--date=today`, `docs/nowadays.md`, `pytest -k recently` and `CURRENTLY=1`.
- `test_the_floating_date_reader_reads_no_barred_word_of_a_fenced_block`.
- `test_the_live_floating_date_case_passes_on_a_code_span_of_the_rule_file`, which writes the #586 sentence into the rule file and requires a pass.

Three cases drive the live case in both directions. Two of them write a sentence that carries a barred word in prose and require a failure, and `PROSE_MUTATIONS` holds the bare `now` beside `currently`. Each one restores the file in a `finally` block and then reads the restored text. Three further cases prove that the cut removes the span alone, where a barred word stands in the prose beside a code span.

`test_the_floating_date_reader_reads_the_sentences_of_the_rule_file` states the floor, because a reader that finds no sentence reports a clean file rather than a measurement. The rule file holds 203 sentences against a floor of 100, and the cut keeps `required status check` in the prose.

Three cases record a limit rather than a repair. A fenced block reaches the sentence reader, and the cut removes its content because the join pairs the opening fence with the closing one, which is an accident of the join and not a design. A code span that a full stop and a space split reaches two sentences, each half then carries one backtick, and the reader still reports it. An unclosed code span moves every pair after it, so the next span reports. No sentence and no fenced block of the rule file meets any of the three readings today.

## The sweep of the other readers

A sweep read all 49 readers of `tests/` that match a pattern over prose. **No second reader carries a reachable instance of this defect.** Six readers already cut a span: `without_quoted` and `readable_prose` of `tests/test_ruling_vocabulary.py`, `without_code` of `tests/test_statistics_thread_term.py`, `sentences` of `tests/test_changelog_sentence_exemption.py`, `readable_prose` of `tests/test_documentation_version_claims.py`, and `unfenced_lines` of `tests/release_body.py`.

Three readers carry the same shape and no collision reaches them today. **This round records them as findings and repairs none of them.**

| Reader | Pattern | Why no collision reaches it today |
|---|---|---|
| `COVERAGE_CLAIMS` of `tests/test_readme_contracts.py` | a coverage phrase over the raw `README.md` | the pattern needs a multi-word phrase |
| `BLANKET_WORD` of `tests/test_changelog_sentence_exemption.py` | `\b(?:any|every|each)\b` | it reads two records and not general prose |
| `SET_SCOPE` of `tests/test_requirement_scope.py` | `no`, `every` or `each` beside `case` | the pattern needs the neighbouring word |

## The five gates

| Gate | Result |
|---|---|
| `pytest tests/ -m "not spec_validation"` | 4707 passed, 7 skipped, 8 xfailed |
| `ruff format --check ja4plus/ tests/` | 226 files already formatted |

`pytest --collect-only` reads 6667 cases at the base commit `4a08fea` and 6683 here. `--cov=ja4plus` reports 4316 statements, 273 missed and 94 percent, which does not move, because this round adds no statement to the package. No file under `ja4plus/` changes and no fingerprint moves.

Verified against https://docs.github.com/en/actions/how-tos/manage-workflow-runs/skip-workflow-runs (retrieved 2026-08-09), which `.claude/rules/batch-gate.md` records.


Crank-Git:
## Self-review

A specialist lens read the diff for correctness, for the two live cases that write a tracked file, for a vacuous pass, and against `.claude/rules/ste.md`. Four findings, and this comment records each one with what the branch does about it.

### 1. The docstring claimed more than the reader holds. Repaired in `8f45b25`.

The first form read **"The reader cuts every code span before it matches"**. The cut reads one sentence, and it pairs a backtick of that sentence alone, so three readings stand outside the claim. Each one reports a code span that a reader must read verbatim.

- **A code span that a full stop and a space split reaches two sentences.** Each half then carries one backtick and `CODE_SPAN` pairs neither. `test_the_floating_date_reader_reports_a_code_span_that_a_full_stop_splits` holds it.
- **An unclosed code span moves every pair after it, so the next span reports.** The review measured this one, and `test_the_floating_date_reader_reports_a_code_span_after_an_unclosed_one` now holds it.
- **A fenced block reaches the sentence reader**, because `readable_text` cuts a blockquote, the `## Changelog` section and a superseded paragraph alone. The cut removes the content of that block anyway, because the join pairs the opening fence with the closing one. **Read that result as an accident of the join.** `test_the_floating_date_reader_reads_no_barred_word_of_a_fenced_block` holds the measurement.

**No sentence and no fenced block of `.claude/rules/batch-gate.md` meets any of the three readings today.** The docstring, the module docstring under `## What these cases cannot test`, and both round records now state all three.

**A wider repair is declined here.** A cut that ran over the raw text before `sentences` splits it would pair a fence line by line, and it would then report the cut form of a sentence rather than the text of the file. #589 asks for the code-span cut, and a reader of a failure reads the file better with the sentence the file holds. A later round repairs the three readings where the rule file meets one.

### 2. The two live cases restore the rule file. Confirmed, no change.

The review ran both cases and read `git status --porcelain .claude/rules/batch-gate.md` clean afterwards, on the `pytest.raises` path as well. **This repository configures no parallel runner**, so the serial warning the module docstring carries states a condition and reports no live race. The warning stays, because a later round that adds a parallel runner needs it.

### 3. No vacuous pass. Confirmed, no change.

Every parameter tuple holds entries, and `test_the_floating_date_reader_reads_the_sentences_of_the_rule_file` states the floor: 203 sentences against a floor of 100, and the cut keeps `required status check` in the prose.

### 4. One passive sentence. Repaired in `8f45b25`.

The comment above `PROSE_MUTATION` read "so the reader is proven in the direction that must still fail". It now reads "so that case proves the reader in the direction that must fail".

### The counts the records carry

The review reported that the round records would go stale as the branch grew, and they take the final reading: **fifteen cases, eight of which failed against the committed reader**, `pytest --collect-only` 6667 at the base commit `4a08fea` and 6682 here, and the unit suite 4706 passed, 7 skipped and 8 xfailed.

Issue #598, opened 2026-08-12
Expose a public flush for the open window of one connection


Crank-Git:
`Crank-Git/ja4plus-go` issue #216 found this gap, and the maintainer ruled on it on
2026-08-12. **The port carries the same gap.** `.claude/rules/parity.md` states that a ruling
lands in both repositories or in neither, so this issue holds the port half.

## The finding

`cleanup_connection` removes a connection from the state table, and it emits nothing. A
caller that evicts a connection therefore loses the window that connection holds open.

| Location | What it holds | Does it emit? |
|---|---|---|
| `ja4plus/fingerprinters/ja4ssh.py:530` | `cleanup_connection` | No. It pops the connection, it pops the handshake entry, and it returns `None`. |
| `ja4plus/fingerprinters/ja4ssh.py:408` | `_close_window` | Yes, for one connection. The leading underscore makes it private, so no caller reaches it. |
| `ja4plus/fingerprinters/ja4ssh.py:382` | `close_open_windows` | Yes, for every connection at once. It evicts nothing. |

**The port holds the per-connection emission already.** `_close_window` does the whole job,
and no public name reaches it.

## Why it matters

A long-running monitor is the caller that evicts. It calls `cleanup_connection` when a
connection ends, which is the moment the reference publishes the final window.
`rust/ja4/src/ssh.rs:45-55` and `zeek/ja4ssh/main.zeek:160-164` both emit at teardown.

`close_open_windows` reaches every connection at once, which is the wrong instrument for one
connection that just ended.

## The ruling the Go side already made

The maintainer ruled candidate 2 on `Crank-Git/ja4plus-go` issue #216, on 2026-08-12.

- **A second method emits the open window of one connection, and it then removes the
  connection.**
- `cleanup_connection` keeps its signature and its behaviour. It emits nothing, so a caller
  that only reclaims memory receives no fingerprint it did not ask for.
- The method names the connection by the same key `cleanup_connection` accepts.

## The name

Neither side shipped a name for this behaviour, so rule 2 of the parity rules leaves the
choice to whichever side names it first. The Go side names it in
`Crank-Git/ja4plus-go` pull request #263.

- Go: `CloseConnectionWindow`
- Python: **`close_connection_window`**

The Go pull request states the four names it rejected and the reason for each one.

## What this issue asks for

1. Add `close_connection_window(src_ip, src_port, dst_ip, dst_port, proto)`. It calls
[...]

PR #634, opened 2026-08-15
Expose a public flush for the open window of one connection

Crank-Git:
## What

`JA4SSHFingerprinter` publishes `close_connection_window(src_ip, src_port, dst_ip, dst_port, proto)`. It emits the window one connection holds open, and it then removes that connection. Refs #598.

## Why

`cleanup_connection` removes a connection and it emits nothing, so a caller that evicts a connection lost the window that connection held open. `_close_window` does the whole job for one connection, and its leading underscore kept every caller away from it. `close_open_windows` reaches every connection at once, which is the wrong instrument for one connection that just ended.

A long-running monitor is the caller that evicts, and the reference publishes the final window at that moment. `rust/ja4/src/ssh.rs:45-55` and `zeek/ja4ssh/main.zeek:160-164` both emit at teardown.

## The interface

The maintainer ruled the method on `Crank-Git/ja4plus-go` issue #216, on 2026-08-12. Parity rule 2 gives the interface to the side that names it first, and `Crank-Git/ja4plus-go` pull request #263 shipped this signature.

```go
func (f *JA4SSHFingerprinter) CloseConnectionWindow(srcIP string, srcPort uint16, dstIP string, dstPort uint16, proto string) []FingerprintResult
```

This branch adopts that parameter order under the Python name and the Python types: `close_connection_window(src_ip: str, src_port: int, dst_ip: str, dst_port: int, proto: str) -> list[dict[str, Any]]`. The order is also the order `cleanup_connection` already accepts on this side, so one 5-tuple names the connection at both call sites.

Verified against: /Users/christiancrank/Development/Personal/ja4plus-go/ja4ssh.go:841 (read-only, retrieved 2026-08-15).

## The behaviour

- The method takes the lock for the whole call, as every other public method of the fingerprinter does.
- It removes the handshake entry of the endpoint pair first, because a connection the state table does not hold still carries an entry that a SYN wrote.
- It names the connection by the same key `cleanup_connection` accepts, so a caller names the two endpoints in either order.
- It returns an empty list for a connection the state table does not hold.
- It returns an empty list for a window that holds no SSH packet, because #97 declines a fingerprint of an empty window. It removes the connection in that case too, so a second call returns an empty list.
- `cleanup_connection` keeps its signature and its behaviour, so a caller that only reclaims memory receives no fingerprint it did not ask for.

## How this was tested

New file `tests/test_ja4ssh_close_connection_window.py` holds nine cases. The cases came first and all nine failed:

```
E       AttributeError: 'JA4SSHFingerprinter' object has no attribute 'close_connection_window'
tests/test_ja4ssh_close_connection_window.py:68: AttributeError
```

No FoxIO vector separates the two answers, because the conformance suite reads each capture to its end and calls no eviction. The file is the record of the ruling, and it names `Crank-Git/ja4plus-go` issue #216.

The five gates, all green.

```
pytest tests/ -m "not spec_validation"   4726 passed, 8 skipped, 1961 deselected, 8 xfailed, 114 subtests passed
pytest tests/ -m spec_validation         1635 passed, 143 skipped, 4783 deselected, 140 xfailed
ruff format --check ja4plus/ tests/      228 files already formatted
```

No fingerprint moves. The conformance suite reports 1635 passed, 143 skipped and 140 xfailed before this branch and the same three counts after it.

`CHANGELOG.md` and the `## Changelog` table of `docs/specs/spec.md` each record `TBD`. `docs/api_reference.md` gains the section `When to call close_connection_window`.


Crank-Git:
Self-review, correctness lens. No defect found, and four readings stand behind that verdict.

1. `BaseFingerprinter.__init__` at `ja4plus/fingerprinters/base.py:77` builds `threading.RLock()`, so the lock this method holds re-enters. `_close_window` takes no lock of its own, and its three callers each hold the lock already.
2. The forward key and the reverse key match the two forms `process_packet` writes at `ja4plus/fingerprinters/ja4ssh.py:159` and `ja4plus/fingerprinters/ja4ssh.py:163`, and `cleanup_connection` reads the same pair.
3. The lock covers the emission and the read of `self.fingerprints[-1]`, so no other thread appends between them. `close_open_windows` reads the same entry the same way.
4. A reviewer reintroduced a defect in memory, which left the connection in the state table, and four of the nine cases failed on it. The cases are not vacuous.

Coverage reads 4329 statements, 273 missed and 94 percent. The count of missed statements does not move.

Crank-Git:
Self-review, writing-standard lens. It found three readings and this branch repairs all three, in commit `2d7d156`.

1. `the wrong instrument for one connection that just ended` is a metaphor, which rule 9 of `.claude/rules/ste.md` bars. Four places carried it: the docstring, `docs/api_reference.md`, the entry of `CHANGELOG.md` and the Changelog row of `docs/specs/spec.md`. The exemption of that rule file covers rule 1 and rule 3 alone, so the two records carry rule 9 as well. Each place now reads `it serves no single connection that just ended`.
2. One sentence of `docs/api_reference.md` held 26 words against the limit of 25 for a description sentence. Two sentences now carry it.
3. `lost` and `kept` read as past tense inside a present-tense clause chain. The leading underscore still keeps every caller away, and `cleanup_connection` still loses the window, so both verbs now read as present tense.

The reviewer raised three further readings and this branch takes none of them.

- Second person in `docs/api_reference.md` is a nit rather than a rule, and this branch takes it anyway, because the section above it names `the caller`.
- `behaviour` is the spelling of the whole repository, and `.claude/rules/ste.md` itself carries it. A change of that spelling is a repository-wide ruling and no part of this issue.
- `handshake` is ordinary English here and no controlled term, so the `## Terms` table needs no entry for it.

The five gates run again after the repair: 4726 passed on the unit suite, 1635 passed with 143 skipped and 140 xfailed on the conformance suite, `ruff check` clean, `ruff format --check` clean over 228 files, and `mypy --strict` clean over 31 sources.

Issue #619, opened 2026-08-14
JA4X hashes an empty object identifier list under the empty-list ruling of 2026-08-14


Crank-Git:
**The maintainer ruled the empty-list question on 2026-08-14, and the ruling settles JA4X R12 as
well as JA4H rule 5b.** `Crank-Git/ja4plus#612` carries the JA4H half. **This issue carries the
JA4X half**, and the Go repository carries it at `Crank-Git/ja4plus-go#582`.

## What was ruled

**An empty list hashes, and it does not write the `000000000000` sentinel.**

**The deciding fact is a rank 1 image rule.** R18 of the JA4H transcription states
`Truncated SHA256 hash of Headers, in the order they appear`, and it names no sentinel. **R27
confines the zero sentinel to part c and part d.**

## What this port holds today

**Read at tag `v1.1.0` on 2026-08-14.** `ja4plus/fingerprinters/ja4x.py:70-78`:

```python
# Hash these elements - SHA256 truncated to 12 hex chars
# Use '000000000000' sentinel for empty values (consistent with other fingerprinters)
issuer_hash = (
    hashlib.sha256(issuer_str.encode()).hexdigest()[:12] if issuer_str else "000000000000"
)
subject_hash = (
    hashlib.sha256(subject_str.encode()).hexdigest()[:12] if subject_str else "000000000000"
)
ext_hash = hashlib.sha256(ext_str.encode()).hexdigest()[:12] if ext_str else "000000000000"
```

**All three parts of `generate_ja4x` short-circuit an empty list to the sentinel.** So this port
holds the behaviour that the ruling changes, and **an empty list produces a different value in
each repository once the Go half lands.**

## The Go half, so both repositories move together

`Crank-Git/ja4plus-go#582` moves all three call sites to a helper that hashes the empty string.
The Go tree already holds that helper, and `Crank-Git/ja4plus-go#527` built it for JA4H.

**The Go issue also records this reading**, so no one measures it twice.

## What no measurement of this reaches

**Zero vectors of the shared FoxIO corpus exercise an empty object identifier list.** The Go half
measured it directly rather than assuming:

- **The corpus holds 120 JA4X values across both vector sets, and 0 of them hold `000000000000`
  in any part.**
- **The Go conformance report carries 78 JA4X rows, and none names a value that holds the
  sentinel.**

**So this change is behaviour-only, and a constructed test is the only thing that can guard it.**
The Go project's rule states that where no vector reaches the value, the ruling carries a test
[...]

Issue #614, opened 2026-08-14
Match either line ending in the HTTP header block terminator


Crank-Git:
The maintainer ruled `Crank-Git/ja4plus-go#298` answer 1 on 2026-08-14. **This is the Python half.** A ruling lands in both repositories or in neither.

## What was ruled

**The header block terminator matches one line ending followed by another**, and it does not match two literal `\r\n` sequences alone.

## What the port changes

**`ja4plus/utils/http_utils.py:43` at tag `v1.1.0` holds the two literals.** It matches `(\r\n|\n)(\r\n|\n)` instead.

## What it repairs

1. **A header block that ends `\n\r\n` reaches no value today.** The rule reads it.
2. **A header block that ends `\r\n\n` matched one byte early**, and the value trim saved it by coincidence. The rule removes the coincidence.

## What it moves

**No value of the shared FoxIO vector set.** No capture of the corpus reaches the input. **The rule reaches every mixed-line-ending header block outside it.**

**No FoxIO source addresses this input**, which is a stop condition in both repositories. The maintainer ruled it rather than leaving the two parse gaps open.

## Reversal path

**`Crank-Git/ja4plus-go#298`**, and this issue.

Crank-Git:
Plan. Batch #631, integration branch `batch/631-header-terminator`.

**Approach.** `ja4plus/utils/http_utils.py:43` holds `HEADER_BLOCK_TERMINATORS = (b"\r\n\r\n", b"\n\n")`. Neither literal matches `\n\r\n`, and `\r\n\n` matches one byte early because `\n\n` is a substring of it. Replace the two literals with the ruled form, one line ending followed by another line ending, and route `header_block_end` through it so the end offset it returns is the offset past the whole terminator.

**Files.** `ja4plus/utils/http_utils.py`. Tests under `tests/`.

**Out of scope.** The second copy at `ja4plus/fingerprinters/ja4l.py:402-403`. #630 adopts the reader this issue builds, and it runs after this one sub-merges.

**The four cases the tests must state.** `\r\n\r\n`, `\n\n`, `\n\r\n`, and `\r\n\n`. The fourth case must assert the end offset, because the present code returns an offset one byte early on it rather than failing outright.

No fingerprint may move. Run both pytest markers, `ruff check`, `ruff format --check` and `mypy --strict`.

Crank-Git:
status repair: removed status:in-review, kept status:batched

Crank-Git:
[...]

Issue #612, opened 2026-08-14
Read a JA4H request path that holds a space, and rule the empty header list


Crank-Git:
**The Go library at `Crank-Git/ja4plus-go` read cause 5 of its JA4H deviation cluster, and
it found two rules that the port also carries.** `Crank-Git/ja4plus-go#527` holds the Go
half, and `Crank-Git/ja4plus-go#548` holds the reading.

**Rule 1 is a defect, and rule 2 is a question for the maintainer.** Neither repository has
built either one.

## Rule 1 — the request line pattern reads no path that holds a space

`ja4plus/utils/http_utils.py:31-34` at tag `v1.1.0` holds:

```
REQUEST_LINE_PATTERN = (
    r"^([" + METHOD_TOKEN_CHARACTERS + r"]+)"
    r"\s+(\S+)\s+(HTTP/(?:\d+\.\d+|[23]))(?=[ \t\r\n]|$)"
)
```

**The path group `(\S+)` reads no space.** Frame 4 of `gre-erspan-vxlan.pcap` holds
`GET /Hello Arkime HTTP/1.0\r\n\r\n`, so the match reads `/Hello`, then needs `HTTP/1.0`
and finds `Arkime`. **The parse fails, and the request reaches no JA4H value.**

**The FoxIO per-packet vector holds a value for that frame**:
`ge10nn000000_e3b0c44298fc_000000000000_000000000000`.

**The Go library holds the same `(\S+)` path group**, so both implementations carry rule 1.

## Rule 2 — part b of an empty header list, and the FoxIO implementations split

`ja4plus/fingerprinters/ja4h.py:480` at tag `v1.1.0` holds:

```
hashlib.sha256(headers_str.encode()).hexdigest()[:12] if headers_str else "000000000000"
```

**So the port writes the zero sentinel for an empty header list.** The Go library writes the
same sentinel.

**The four FoxIO implementations split two against two**, read on 2026-08-14 at the pin
`27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8`.

| Implementation | Part b of an empty header list | Evidence |
|---|---|---|
| Python | `e3b0c44298fc` | `python/ja4h.py:72`, `python/common.py:127` |
| Wireshark | `e3b0c44298fc` | `wireshark/source/packet-ja4.c:628`, `:641` |
| Rust | `000000000000` | `rust/ja4/src/http.rs:202`, `rust/ja4/src/lib.rs:184` |
[...]

Issue #228, opened 2026-08-08
Decide the JA4X empty-list form and the JA4X raw form


Crank-Git:
#202 transcribed `technical_details/JA4X.png` into `docs/specs/foxio/JA4X.md`. The
transcription found two questions the image does not settle. Each one needs the user,
because each changes what this project publishes. **#202 changed no fingerprinter.**

## Decision 1 — which empty form does part c write?

R8 of `docs/specs/foxio/JA4X.md` holds the reading. **The image draws both forms.**

- Its two `Sliver, Havoc C2` rows write `000000000000` in part a.
- Its `Qakbot` row writes `2bab15409345_af684594efb4_e3b0c44298fc`, and `e3b0c44298fc` is
  the truncated SHA-256 of the empty string.

The two forms are the two FoxIO implementations.

| Source at commit `27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8` | Empty form |
|---|---|
| `rust/ja4x/src/lib.rs`, `hash12` | `000000000000` |
| `wireshark/source/packet-ja4.c:590` | `000000000000` |
| `README.md` line 146, the same Qakbot certificate | `000000000000` |
| `python/ja4x.py`, no test on the empty list | `e3b0c44298fc` |
| `technical_details/JA4X.png`, the Qakbot row | `e3b0c44298fc` |

`ja4plus/fingerprinters/ja4x.py:65` to `ja4x.py:71` write `000000000000`, so this project
follows the FoxIO Rust implementation, the Wireshark dissector and the `README.md`.

**No vector in this repository reaches the case.** No expected-output file under
`tests/foxio_vectors/` holds `e3b0c44298fc`, and none holds a `JA4X` value with a zero
sentinel. **No fingerprint moves either way**, so this is a schema question rather than a
defect.

Reproduce the arithmetic:

```bash
python3 -c "import hashlib; print(hashlib.sha256(b\"\").hexdigest()[:12])"
```

## Decision 2 — does this project publish a JA4X raw form?

R11 of `docs/specs/foxio/JA4X.md` holds the reading.

- `rust/ja4x/src/lib.rs` writes `ja4x_r` as the three unhashed lists, joined with `_`.
- `wireshark/source/packet-ja4.c:1726` registers the field `ja4.ja4x_r`.
- The FoxIO Python implementation writes no raw form for JA4X.
- The image states no raw form.

`ja4plus` publishes no `JA4X_r` value, and no expected-output file under
[...]

PR #636, opened 2026-08-15
Hash an empty object identifier list in JA4X

Crank-Git:
## What this change does

**JA4X hashes an empty object identifier list.** `ja4plus/fingerprinters/ja4x.py:72` to
`:78` substituted the zero sentinel `000000000000` for each of the three parts whose
joined list was empty. Each part now reads `hashlib.sha256(part.encode()).hexdigest()[:12]`
with no guard, so an empty part reads `e3b0c44298fc`.

Refs #619.

## Why

**The maintainer ruled the empty-list question on 2026-08-14.** The deciding fact is a
rank 1 image rule. The JA4H image states `Truncated SHA256 hash of Headers, in the order
they appear` for the hashed part and it names no sentinel there, which R12 of
`docs/specs/foxio/JA4H.md` transcribes. R17 of that page confines the zero sentinel to
part c and part d of JA4H. No rule of the JA4X image gives a sentinel to any part.

**The ruling reverses the ruling of 2026-08-08 that #228 recorded under round 77.** R8 of
`docs/specs/foxio/JA4X.md` now quotes the superseded wording rather than rewrite it, and
it keeps the record of the contradiction the FoxIO image carries.

**`Crank-Git/ja4plus-go#582` is the other half, and the two ports agree.** The Go tree
writes `parser.TruncatedHashNoSentinel` into each of the three parts, at `ja4x.go:490` to
`:492`, and `internal/parser/hash.go` states that the function returns `e3b0c44298fc` and
never the zero sentinel. Both ports therefore produce `e3b0c44298fc` for an empty part.

**This change rules nothing on JA4H.** R19 of `docs/specs/foxio/JA4H.md` holds that half,
and #612 carries it.

## How it is tested

**Nine cases came first and six of them failed against the committed guard.** One failure
reads `AssertionError: assert '000000000000' == 'e3b0c44298fc'`. All nine pass against the
hashed form.

**The mutation reversal runs in both directions.** Against the guarded code the file
reports `6 failed, 3 passed`. Against the hashed code it reports `9 passed`.

**No committed capture reaches the empty list, so the change moves no fingerprint.** A
replay of the 38 captures of `tests/foxio_vectors/` produces 60 JA4X values before the
change and 60 after. The two records are byte identical, at SHA-256
`07ee974800c623f79026695f37f3c9be380343e3cf394d0cf6266c52d427292f`.

The five gates, all green:

```
pytest tests/ -m "not spec_validation"  4726 passed, 8 skipped, 1961 deselected, 8 xfailed
pytest tests/ -m spec_validation        1635 passed, 143 skipped, 4785 deselected, 140 xfailed
ruff format --check ja4plus/ tests/     228 files already formatted
mypy --strict ja4plus/                  Success: no issues found in 31 source files
```

`tests/foxio_deviations.json` holds 140 keys against 140 xfailed cases.

## Two readings beside the change

**The change removes one relation defect.** `generate_ja4x_raw` writes an empty part for
an empty list, and the earlier form wrote a sentinel that hashes nothing into the same
position. The raw form is now the exact preimage of the fingerprint on every part, and
`test_the_hashed_form_is_the_hash_of_each_part_of_the_raw_form` holds that relation.

**Two readers model a hash and they part on this rule.** `hash12` of
`tests/test_ja4x_raw.py` models `ja4plus`, so it loses its guard. `hash12` of
`tests/test_foxio_rust_parity.py` models the FoxIO Rust snapshot, and the Rust
implementation writes the sentinel, so it keeps it. No local snapshot holds an empty part,
so the second reader meets no case.

Verified against: https://github.com/FoxIO-LLC/ja4/tree/main/technical_details (retrieved
2026-08-15, through `docs/specs/foxio/JA4X.md` and `docs/specs/foxio/JA4H.md`).

## Two states this branch took again

**The integration branch moved under this pull request, and the two round records
conflicted.** #614 sub-merged, so `gh pr view 636 --json mergeable` read `CONFLICTING` and
the provider builds no merge commit for that state. The branch takes
`batch/631-header-terminator` again, and the resolution keeps the entry of #614 and the
entry of #619 in both records. The five gates run again on the merged tree, and the replay
reports the same SHA-256.

**Coverage holds.** The unit suite reports 94 percent on the base commit `4497e9f` and 94
percent after the change, over 4316 statements and 273 missed in each run.


Crank-Git:
**Self-review, correctness lens.** No code path of `ja4plus/` writes the zero sentinel for a JA4X part after this change, and `ja4plus/fingerprinters/ja4x.py:75` to `:77` hold the three hashes. Each of the nine cases of `tests/test_ja4x_empty_ext.py` asserts the new value and the absence of the old one, so none passes by construction against the guarded form. The two `hash12` readers part on purpose: `tests/test_ja4x_raw.py:62` models `ja4plus` and loses its guard, and `tests/test_foxio_rust_parity.py:539` models the FoxIO Rust snapshot and keeps it.

**One finding stands outside this change, and #639 holds it.** Three JA4X rows of `ja4plus/data/ja4plus-mapping.csv` carry the sentinel, at lines 26, 27 and 35, so a lookup matches no `ja4plus` value for a certificate that carries an empty list. The mapping file is FoxIO data at a pinned commit, and this pull request rewrites no row of it and changes no lookup path.

Issue #631, opened 2026-08-15
Batch: the header block terminator and three small parity gaps


Crank-Git:
This batch holds four members. Every one is verified against the code before it was scheduled.

- [x] #614 — match either line ending in the header block terminator, under the maintainer ruling of 2026-08-14. `ja4plus/utils/http_utils.py:43`.
- [x] #630 — the second copy of the same check, at `ja4plus/fingerprinters/ja4l.py:402-403`. **Sequenced after #614**, because it adopts the reader #614 builds.
- [x] #598 — a public flush for the open window of one connection. `ja4plus/fingerprinters/ja4ssh.py:530`.
- [x] #619 — JA4X hashes an empty object identifier list, under the ruling of 2026-08-14. `ja4plus/fingerprinters/ja4x.py:71-78`.

Continuous integration runs once, on the batch pull request into `dev`. Every member pull request is a draft that carries the skip keyword.

- [x] #612 — the request line path group, and the empty header list. `ja4plus/utils/http_utils.py:32-34` and `ja4plus/fingerprinters/ja4h.py:479-481`. **Sequenced after #630.** It joined this batch because it writes `http_utils.py`, which #614 already changed on this branch, and `dev` does not hold that change yet.

Crank-Git:
finding: batch cross-check, `batch/631-header-terminator`

**What was compared.** The four plan comments of #614, #630, #598 and #619, against each other, on file, function, interface, migration, configuration key and route.

**The pairs found.**

1. **#614 and #630 own the same rule in two files.** #614 builds the reader in `ja4plus/utils/http_utils.py` and #630 deletes the copy in `ja4plus/fingerprinters/ja4l.py` and calls that reader. #630 therefore assumes a shape #614 creates. **Resolved by sequence.** #630 launches after #614 sub-merges, and it forks the integration branch at that point. No plan was edited.

**The clean negatives.**

- **No two members write the same file.** `http_utils.py`, `ja4l.py`, `ja4ssh.py` and `ja4x.py` are four files, one to each member.
- **No two members own the same new name.** #598 adds `close_connection_window`. No other member adds a public name.
- **No member changes a configuration key, a route or a migration.** This project holds none of those.
- **#619 is the one member that can move a fingerprint**, and no other member reads the JA4X path. Its byte-identical replay proof therefore rests on its own change alone.
- **#598 and #619 assume nothing that another member changes.** Neither reads the HTTP header block.

[...]

Issue #630, opened 2026-08-15
Read a mixed line ending in the JA4L request completeness gate


Crank-Git:
`Crank-Git/ja4plus-go#685` read the JA4L request completeness gate on 2026-08-15 UTC. **This is the Python half.** A ruling lands in both repositories or in neither.

## The port carries a second copy of the terminator reading

**`ja4plus/fingerprinters/ja4l.py:385-402` at tag `v1.1.0` holds `_holds_a_complete_http_request`, and it reads two fixed byte groups.**

```python
    if not is_http_request(payload):
        return False
    return b"\r\n\r\n" in payload or b"\n\n" in payload
```

**That pair is a second copy of `HEADER_BLOCK_TERMINATORS` at `ja4plus/utils/http_utils.py:43`**, which #614 replaces with one line ending followed by another. **#614 does not reach this line**, because the line lives in the JA4L fingerprinter and never in `http_utils`.

## Why this is a defect

**The pair declines the terminator `\n\r\n`.** So a request whose header block ends `\n\r\n` reaches a JA4H value after #614, and `_holds_a_complete_http_request` still reports `False` for it. The packet then moves the JA4L client measurement point, and a request with any other terminator does not.

## The reading, and its deciding source is not #614

**#614 answers where a JA4H value ends its header block. This gate answers a different question: does the reference route the packet to a cache that holds no measurement point.**

- **`python/common.py:78`** at the pinned FoxIO commit `27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8` routes a packet whose `hl` is `http` to `http_cache`.
- **`python/ja4.py:398`** sets `hl` from the layer name that tshark reports.
- **`python/ja4.py:572-573`** writes point C and calls `calculate_ja4_latency`, and **`python/ja4.py:151`** reads `get_cache(x)`. So an `http` packet reads a cache that holds no `A` and no `B`, and it completes no JA4L-C value.

**So the Wireshark HTTP dissector decides, and it reads `\n\r\n` as a complete header block.**

- **`epan/tvbuff.c:4203`** of Wireshark v4.6.0 compiles the line-end pattern with `ws_mempbrk_compile(&pbrk_crlf, "\r\n")`, so a bare line feed ends a line.
- **`epan/req_resp_hdrs.c:143`** breaks out of the header loop on the first line of length 0, with the comment `We found the end of the headers.`

**The two readings reach one repair, and they reach it for two reasons.**

## What the port changes

[...]

PR #640, opened 2026-08-15
Read the header block terminator of http_utils in the JA4L request gate

Crank-Git:
Refs #630. Batch #631, into `batch/631-header-terminator`.

## What it changes

**`ja4plus/fingerprinters/ja4l.py:402-403` held a second copy of the header block terminator rule.**

```python
    return b"\r\n\r\n" in payload or b"\n\n" in payload
```

That line called no reader of `ja4plus/utils/http_utils.py`, so #614 moved the rule in one file and left the copy in the other. `_holds_a_complete_http_request` now calls `header_block_end`, and one reader answers the question for every caller.

## Why it is a defect

**The pair declined the terminator `\n\r\n`.** The reference routes a packet that holds a whole HTTP request to a cache that holds no measurement point, so such a packet completes no client value. A request whose header block ends `\n\r\n` therefore moved the JA4L client measurement point, where a request with any other terminator moved it nowhere.

**The deciding source is the Wireshark HTTP dissector and not #614**, because this gate answers which cache the reference gives the packet. `epan/tvbuff.c:4203` of Wireshark v4.6.0 compiles the line-end pattern with `ws_mempbrk_compile(&pbrk_crlf, "\r\n")`, so a bare line feed ends a line. `epan/req_resp_hdrs.c:143` leaves the header loop on the first line of length 0.

## How it is tested

**No FoxIO capture holds a mixed line ending, so the separating packet is built.** Three packets carry a SYN at 0.000 seconds, a SYN-ACK at 0.001 seconds and a request at 0.051 seconds. That request holds the head `GET / HTTP/1.1\r\nHost: example.com`, the terminator `\n\r\n` and the body `BODY`. It read `JA4L-C=25000_128` on the base commit `727a10f` and it reads no client value here.

New file `tests/test_ja4l_header_block_terminator.py` holds seven cases. **Five failed on the base commit and two passed.** The two that passed are the controls: a request that holds no whole header block moves the client point, and a payload that starts no HTTP request moves it too.

```
FAILED tests/test_ja4l_header_block_terminator.py::test_the_gate_reads_a_request_that_ends_its_header_block_with_a_mixed_line_ending
FAILED tests/test_ja4l_header_block_terminator.py::test_a_mixed_line_ending_request_completes_no_client_value
FAILED tests/test_ja4l_header_block_terminator.py::test_the_gate_reads_every_terminator_two_line_endings_build
FAILED tests/test_ja4l_header_block_terminator.py::test_no_terminator_two_line_endings_build_completes_a_client_value
FAILED tests/test_ja4l_header_block_terminator.py::test_the_fingerprinter_holds_no_second_copy_of_the_terminator_rule
==================== 5 failed, 2 passed, 1 warning in 0.32s ====================
```

## The five gates

| Gate | Result |
|---|---|
| `pytest tests/ -m "not spec_validation"` | 4734 passed, 8 skipped, 8 xfailed |
| `ruff format --check ja4plus/ tests/` | 229 files already formatted |

**No fingerprint moves.** The conformance suite reports the same three counts as the base commit, and `tests/foxio_deviations.json` holds 140 keys. `pytest --collect-only` reads 6668 cases against 6659 on the base commit.

Verified against: https://gitlab.com/wireshark/wireshark (`epan/tvbuff.c` and `epan/req_resp_hdrs.c`, cited from #630, retrieved 2026-08-15)

`Crank-Git/ja4plus-go#685` holds the Go half.


Crank-Git:
Self-review, correctness lens. **No defect found.**

**The two readers agree outside the mixed terminator.** The old pair and `header_block_end` were each unanchored, so a terminator that stands inside the body reports the same under both. `is_http_request` gates both, so an empty payload, a payload that starts no request and a truncated request behave as they did.

**`header_block_end` never returns 0 here**, because `HEADER_BLOCK_TERMINATOR_PATTERN` matches two bytes at least, so `match.end()` is 2 or more. The `is not None` test is correct on its own reading, and it stays correct whatever the pattern becomes.

**No case is vacuous.** A run against the old body restored by hand failed the four behaviour cases and the source case, and it passed the two controls. That is the 5 and 2 the pull-request body states.

**The gate raises on no input.** `re.search` reads bytes and needs no decode, `is_http_request` decodes with `errors="ignore"`, and the pattern nests no quantifier.

**The control value is right.** `JA4L-C` is half the time from the SYN-ACK to the request, so `(0.051 - 0.001) / 2` is 25000 microseconds, and the client packets carry TTL 128.

Crank-Git:
Self-review, conventions and writing lens. **One finding repaired, two declined with a reason.**

**Repaired: three sentences passed the 25-word limit of `.claude/rules/ste.md`.** One stood in the docstring of `_holds_a_complete_http_request` at 28 words, and two stood in the module docstring of `tests/test_ja4l_header_block_terminator.py` at 27 and 32 words. Each one is now two sentences. The exemption of `.claude/rules/ste.md` covers the two Changelog records alone, and it reaches no docstring.

**Declined: the word `cache`.** The `state table` row of the `## Terms` table bars `cache` as a name for the state table of a fingerprinter. This prose names `http_cache` of `python/common.py:78`, which belongs to the FoxIO reference and holds no per-connection data of this project. **The docstring of the same function already carried that word before this change**, at `ja4plus/fingerprinters/ja4l.py:389`. A new word in the lines beside it would rotate a synonym, which rule 7 bars, so the reading keeps one word for one meaning.

**Declined: the word `gate`.** The issue body names `the JA4L request completeness gate`, and this pull request quotes that name. The `## Terms` table already holds `lint gate`, and `.claude/rules/batch-gate.md` names the batch gate and the merge gate. **A row of the `## Terms` table is a spec change**, and the plan of this issue puts the vocabulary out of scope, so this change adds no row.

**Nothing else.** The commit message reads `type(scope): summary`. Both records write the literal `TBD` and they agree. The test file follows the shape of `tests/test_ja4l_connection_state.py`.

Issue #609, opened 2026-08-14
Publish the four-part JA4TS value on a reset of a one-SYN-ACK connection


Crank-Git:
**The maintainer ruled this question on 2026-08-14, in `Crank-Git/ja4plus-go`, at
`Crank-Git/ja4plus-go#484`.** The Go half lands under `Crank-Git/ja4plus-go#495`, and this
issue holds the port half. `.claude/rules/parity.md` of the Go repository states that a
ruling lands in both repositories or in neither.

> **Answer 1.** The library publishes the stored four-part JA4TS value on a reset of a connection
> that holds one SYN-ACK.

## The mechanism

`ja4plus/fingerprinters/ja4ts.py:341` of tag `v1.1.0` calls `_reset_value` for a packet that
carries the reset flag. `SynAckTracker.reset_value` at
`ja4plus/fingerprinters/ja4ts.py:183-184` returns `None` until the connection holds two
SYN-ACK times, so a reset of a connection that holds one SYN-ACK produces no value.

**The Wireshark dissector writes the four-part value outside the delay guard.**
`wireshark/source/packet-ja4.c:1295-1299` sets `syn = 3` for the flag byte `0x004`.
`wireshark/source/packet-ja4.c:1599-1608` copies the window size, the maximum segment size,
the window scale and the option list from the stored connection state, and it writes
`hf_ja4ts`. `wireshark/source/packet-ja4.c:684` guards the delay list and the reset letter
on `conn->syn_ack_count > 1`, **and the four-part value sits outside that guard**.

Every citation reads the FoxIO reference at the pinned commit
`27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8`.

## The measured count, and it is 3 rather than 6

**6 per-packet comparisons of the FoxIO corpus name a reset frame that holds one SYN-ACK,
and the Go change closes 3 of them.** The Go conformance run measured every number below.

| Capture | Frame | Reset direction | Result |
|---|---|---|---|
| `ssh2.pcapng` | 849 | The server sent it. | The comparison matches. |
| `ssh2.pcapng` | 850 | The server sent it. | The comparison matches. |
| `browsers-x509.pcapng` | 174 | The server sent it. | The comparison matches. |
| `https3-301-get.pcap` | 20 | **The client sent it.** | The comparison stays a deviation. |
| `https3-301-get.pcap` | 21 | **The client sent it.** | The comparison stays a deviation. |
[...]

Issue #602, opened 2026-08-13
Record the JA4TS delay rounding in the divergence register


Crank-Git:
## The rule is in the code, and the register does not hold it

`ja4plus/fingerprinters/ja4ts.py` at `v1.1.0` rounds each part e delay to the nearest second,
half away from zero:

```
def _delay_seconds(later: float, earlier: float) -> int:
    delay = later - earlier
    return int(math.copysign(math.floor(abs(delay) + 0.5), delay))
```

**That choice settles a FoxIO reference split, and the divergence register records no row for
it.**

## The split

`docs/specs/foxio/JA4T.md` R24 and R25 of `Crank-Git/ja4plus-go` record the reading:

- **Zeek truncates** each part e delay to whole seconds, at `zeek/ja4t/main.zeek`.
- **Wireshark rounds** each delay to the nearest second, at `wireshark/source/packet-ja4.c:684`,
  and the reset delay at `wireshark/source/packet-ja4.c:694`.

**A delay of 1.6 seconds reaches `1` in Zeek and `2` in Wireshark.** R25 states that the same
split covers the reset delay.

**The deleted `technical_details/JA4T.md` separates the two readings not at all.** Its one
worked example is `RST 37.723966 - 31.714762 = RST 6.009 = R6`, and `6.009` reads `6` under both.

## Why this needs a row

The register rows beside it each end with a sentence of the form `The port must ... or the two
implementations disagree`. **Two rows already govern part e and the RST value**, and neither one
names the rounding. A reader who implements from the register alone therefore reaches a value
that differs from the shipped one whenever a delay carries a fractional part at or above one
half.

**The Go implementation adopted the rounding on 2026-08-13**, as a reading of the code above.
`Crank-Git/ja4plus-go#56` builds it and `Crank-Git/ja4plus-go#126` records the adoption rule.

## What this issue asks for

One row in `docs/specs/spec.md`, in the divergence register, that states the rounding, cites the
two implementations, and names the issue that holds the ruling. **No behaviour changes**, because
the code already holds the rule.

Crank-Git:
Landed on `dev` in batch #632, through pull request #655. It carries round 220.

`dev` is not the default branch of this repository, so a `Closes` keyword closes no issue from that merge. The project manager closes it here.

PR #644, opened 2026-08-15
Publish the four-part JA4TS value on a reset of a one-SYN-ACK connection

Crank-Git:
Refs #609.

## What this changes

`SynAckTracker.reset_value` of `ja4plus/fingerprinters/ja4ts.py` returned None until the
connection held two SYN-ACK times, so a reset of a connection the server answered once
produced no JA4TS value. It now returns the stored four-part prefix in that case.

**The delay list and the reset letter stay behind the two-SYN-ACK guard.** A connection
that holds two SYN-ACK times keeps the five-part value it produced before, and
`MAX_SYN_ACK_DELAYS` holds without a change. This change touches no rounding line, and
#602 owns the rounding of part e.

## Why

The maintainer ruled the question on 2026-08-14, at `Crank-Git/ja4plus-go#484`.
`Crank-Git/ja4plus-go#495` landed the port half, and this is the port half of the same
ruling. `wireshark/source/packet-ja4.c:1599-1608` copies the window size, the maximum
segment size, the window scale and the option list from the stored connection, and
`wireshark/source/packet-ja4.c:684` guards the delay list and the reset letter on
`conn->syn_ack_count > 1`.

## How this is tested

**Four cases came first and all four failed against the committed reader.** The classes
are `TestTheResetOfAOneSynAckConnection` of `tests/test_ja4ts_reset.py`.

```
AssertionError: None != '65535_2-1-3-1-1-4_65495_8'
AttributeError: 'NoneType' object has no attribute 'split'
```

**A mutation reversal proves each one bites.** With `return prefix` written back as
`return None`, the four cases fail again and three conformance comparisons fail with
them. The module was restored and both suites are green.

**A byte-identical replay of the 38 committed captures runs before and after.** The
replay produces 777 values before and 781 after, and no value of the 777 moved. All four
new values are JA4TS values.

| Capture | Frame | New value |
|---|---|---|
| `ssh2.pcapng` | 849 | `42600_2-1-1-4-1-3_1300_9` |
| `ssh2.pcapng` | 850 | `42600_2-1-1-4-1-3_1300_9` |
| `browsers-x509.pcapng` | 174 | `64400_2-1-3-4-0-0_1400_2` |
| `browsers-x509.pcapng` | 119 | `64240_2-1-3-4-0-0_1460_2` |

## The register

**The register falls from 140 keys to 137, against 137 `xfailed` cases**, so the
invariant holds at the new number. The three keys it loses are
`ssh2.pcapng/5:57368/JA4TS.2`, `ssh2.pcapng/5:57368/JA4TS.3` and
`browsers-x509.pcapng/2:54603/JA4TS.2`, and each one now compares as a pass.

**The three keys it keeps carry a client RST, and their cause changes.**
`wireshark/source/packet-ja4.c:1600` keys the connection by the stream, and
`_connection_key` keys it by the server endpoint, so a client RST reaches no stored
connection. `Crank-Git/ja4plus-go#502` holds that question.

**Frame 119 of `browsers-x509.pcapng` opens one divergence that no register row reaches.**
It is a server RST that carries ACK, and `wireshark/source/packet-ja4.c:1296` tests the
flag byte for equality, so the dissector writes no value. #246 adopted the Zeek reading
at `zeek/ja4t/main.zeek:167`, which tests the RST bit alone.
`TestTheAckResetDivergence` of `tests/test_foxio_wireshark_ja4ts.py` holds that
divergence as a comparison that runs, and `Crank-Git/ja4plus-go#503` holds the
consequence.

## The five gates

```
pytest tests/ -m "not spec_validation"  4718 passed, 8 skipped, 8 xfailed, 114 subtests passed
pytest tests/ -m spec_validation        1642 passed, 143 skipped, 137 xfailed
ruff format --check ja4plus/ tests/     227 files already formatted
mypy --strict ja4plus/                  Success: no issues found in 31 source files
```

The conformance suite reported 1635 passed, 143 skipped and 140 xfailed at the base.

Verified against: https://github.com/FoxIO-LLC/ja4/tree/main/wireshark (pinned commit `27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8`, retrieved 2026-08-15)


Crank-Git:
Self-review, two lenses, run in parallel against `ff725dc`. Both findings are repaired in `80a33ff`.

**Correctness lens — one real defect, accepted and repaired.** `docs/specs/spec.md` held a stale JA4TS count outside the exempt Changelog table. It read "61 of its 68 reference values compare as a pass" and "52 of the 58 Wireshark values match, 9 of the 10 Zeek values match, 6 reach". Those three numbers contradicted the divergence-register row of the same change set. The paragraph now reads 64, 55 and 3, and it states the four counts as a vertical list.

The lens found no defect in the control flow. It confirmed that `record()` appends on the call that creates the list, so `stamps` is never a present empty list and `len(stamps) < 2` separates the three cases correctly. It confirmed the frame-number arithmetic: frames 849, 850, 174 and 119 read as indices 848, 849, 173 and 118. It confirmed that no new case can pass vacuously, because each one asserts a non-empty reading before it asserts over it.

**Writing lens — one wrong mechanism statement and six overruns, all repaired.** `TestTheResetWithoutADelay` read "Both implementations nest the delay list inside the delay branch". That sentence contradicts the evidence under it, and it was an editing slip of mine. The class measures a RST that reaches no stored connection, so the docstring now states that reason. Six sentences of `ja4ts.py`, `docs/specs/foxio/JA4T.md` and the two test modules passed the 25-word limit; each one splits or becomes a vertical list. `test_the_client_sent_every_registered_reset` takes the present tense.

The lens confirmed that the three rewritten register entries keep `decided: true`, `capability: false` and `issue: 246`, which is the schema `.claude/rules/external-apis.md` states for a value decline.

**Nothing was rejected.** All five gates are green again after the repair: 4718 passed on the unit suite, and 1642 passed, 143 skipped, 137 xfailed on the conformance suite.

Issue #131, opened 2026-08-07
Compute the JA4H raw form and compare it against JA4H_ro


Crank-Git:
## What is wrong

The FoxIO expected-output files hold 89 `JA4H_ro` values. `ja4plus` computes no JA4H raw
form, so all 89 fail the conformance comparison. 11 occurrence-key cases fail with them,
one for each vector that holds a `JA4H_ro` key.

#121 made the comparison run for the first time. Before #121, `tests/conformance_index.py`
dropped every raw key, so no `JA4H_ro` value reached a test.

## Why it matters

A raw form exists so that a person can read what produced a hash. Every other method
`ja4plus` computes a raw form for now reports a measured number. JA4H reports zero of 89.

The Go port at `Crank-Git/ja4plus-go` holds a `Raw` field on its result struct for JA4 and
for JA4S. Parity rule 2 applies to the field name when this issue adds one for JA4H.

## What to build

1. Compute the JA4H original-order raw form in `ja4plus/fingerprinters/ja4h.py`.
2. Write it under the `raw_original_order` key of each JA4H entry, which is the name JA4
   and JA4S already use.
3. Add `"JA4H": (("JA4H_ro", "raw_original_order"),)` to `RAW_METHODS` in
   `tests/conformance_index.py`.
4. Remove the 100 register entries the fix makes pass, in the same change.

FoxIO publishes no `JA4H_r` key, so this issue adds no sorted raw form.

## Acceptance criteria

- [ ] `ja4plus` writes a JA4H raw form on every JA4H entry.
- [ ] The conformance suite compares 89 `JA4H_ro` values, and the measurement states how
      many match.
- [ ] `tests/foxio_deviations.json` holds no entry that this fix makes pass.
- [ ] `docs/implementation_notes.md` records the JA4H sort rule and the vector that
      supports it.

Measured on `epic/12-spec-conformance` at `03c7c02`. Part of #12.

Crank-Git:
## Plan — Epic 1 batch 6

**Approach.** `ja4plus` computes a raw form for JA4, JA4S and JA4SSH, and computes none
for JA4H. 89 `JA4H_ro` reference values therefore reach no comparison, and 100 register
entries record it. Build the JA4H raw form the way the other methods already build theirs.

**Read first.** `docs/specs/features/01-spec-conformance.md`, the FoxIO
`technical_details/JA4H.md`, and `ja4plus/fingerprinters/ja4.py` for the raw-form pattern
[...]

PR #647, opened 2026-08-15
Read a JA4H request path that holds a space, and hash an empty header list

Crank-Git:
Closes nothing on its own. This is a member of batch #631, and it targets
`batch/631-header-terminator`.

Refs #612.

## What this changes

**The two rules land together, and neither one closes a comparison alone.**

**Rule 1 repairs `REQUEST_LINE_PATTERN` of `ja4plus/utils/http_utils.py`.** The path group
read `(\S+)`, which matches no path that holds a space. Frame 4 of
`gre-erspan-vxlan.pcap` holds `GET /Hello Arkime HTTP/1.0`. The group read `/Hello`, and
the match then needed the version token where the line holds `Arkime`. The request reached
no JA4H value.

```
^([<method token characters>]+)[ \t]+([^ \t\r\n](?:[^\r\n]*?[^ \t\r\n])?)[ \t]+(HTTP/(?:\d+\.\d+|[23]))(?=[ \t\r\n]|$)
```

The pattern holds four rules, and each one has a case.

1. Each separator reads `[ \t]+` and never `\s`. `\s` matches a line feed, so the earlier
   pattern crossed into the second line of a payload. `SSH-2.0-OpenSSH_9.6\r\n/a HTTP/1.1\r\n`
   reached a request line that no line holds, and `is_http_request` gates a JA4L
   measurement point.
2. The path group reads `[^\r\n]` and never any character. `.` matches a bare carriage
   return, so the path of `GET /a\rFAKE HTTP/1.1` would hold two lines of the payload.
3. The group is lazy, so a first line that holds two version tokens reaches the earlier
   one. The non-space group read the earlier token, and this rule keeps it.
4. The path group reads neither a space nor a horizontal tab as its first character or as
   its last one.

**Rules 1 to 3 are the Go half's**, which `Crank-Git/ja4plus-go#575` merged on 2026-08-14.
**Rule 4 is this port's, and the self-review found it.** Without it the match costs the
square of the line length: the lazy path body accepts a space, the greedy separator after
it accepts a space, and a run of spaces then admits one split for each space. A version
token the line almost holds makes the match retry every one of them.

| Payload | Without rule 4 | With rule 4 |
|---|---|---|
| `GET a` + 32000 spaces + `HTTPX` | 3017.9 ms | 0.630 ms |
| `GET a` + 16000 `space tab` pairs + `HTTP/x` | 3320.0 ms | 0.593 ms |

`is_http_request` reads 8192 bytes of every TCP payload, so one packet paid about 200 ms
of processor time. **The base commit does not carry this**, because `\s` and `\S` are
disjoint. **The Go half needs no rule 4**, because `regexp` of Go backtracks nowhere. The
second PR comment holds the whole measurement.

Rule 4 moves no match, because the greedy separator already consumes every space of a run.
**A differential run over 400000 random request lines reports no disagreement** between
the two forms.

**The port keeps its own version token** `HTTP/(?:\d+\.\d+|[23])` with the lookahead,
which #35 records. The Go half holds `HTTP/\d+\.\d+`, and that difference is out of scope.

**Rule 2 removes the zero sentinel from part b of JA4H.**
`ja4plus/fingerprinters/ja4h.py` wrote `000000000000` where the joined header names were
empty. Part b now reads `hashlib.sha256(headers_str.encode()).hexdigest()[:12]` with no
guard, which is the form #619 landed on JA4X at `ja4plus/fingerprinters/ja4x.py:75` to
`:77`. An empty header list reads `e3b0c44298fc`.

**The maintainer ruled on 2026-08-14, and the deciding fact is a rank 1 image rule.** R12
of `docs/specs/foxio/JA4H.md` transcribes `Truncated SHA256 hash of Headers, in the order
they appear`, and it names no sentinel. R17 confines the sentinel to part c and part d.
R19 of that page now holds the ruling, and it quotes the superseded wording rather than
rewrite it.

## What it moves

**Exactly one comparison of the shared FoxIO vector set.** Frame 4 of
`gre-erspan-vxlan.pcap` moves from **no value** to
`ge10nn000000_e3b0c44298fc_000000000000_000000000000`, which is the value
`tests/foxio_vectors/wireshark_expected/gre-erspan-vxlan.pcap.json` holds.

**A replay of all 38 committed captures reads every method**, through
`tests/conformance_index.py:index_produced`, which is the reader the conformance suite
uses. It produces 1677 values on the base and 1679 here. **Two values move, both on that
one connection, and both from no value to a value**: `JA4H` and `JA4H_ro`. **No JA4L value
moves**, which matters because `is_http_request` gates a JA4L measurement point. The first
PR comment holds that record and both hashes.

**Rule 1 alone produces `ge10nn000000_000000000000_000000000000_000000000000`** against
the reference value, which reproduces the measurement the Go half reported.

## The register entry, which needs a reader

**The FoxIO Python expected-output file names no JA4H key on `gre-erspan-vxlan.pcap`.**
The value this project now produces therefore has nothing there to compare against, and
`test_the_produced_occurrence_keys_equal_the_reference` failed as an extra occurrence key.
`tests/foxio_deviations.json` gains the key `gre-erspan-vxlan.pcap/JA4H`, decided,
`capability: false`, owned by #612. The cause states that the FoxIO Wireshark dissector
holds the exact value.

**The Go half predicted this result**, in the second comment of #612: "the unaccepted
uncovered count rose by 3. The per-stream vector file of that capture names no JA4H key at
all". The register is the Python-side record of that state. **Read this entry before the
merge**, since it is the one piece of bookkeeping this change decides rather than copies.

`## The marker rule` of `tests/foxio_deviations.py` and the precedence-exception sentence
of `.claude/rules/external-apis.md` each move their denominator from 140 to 141.

## How it was tested

**Test first.** `tests/test_ja4h_request_line_and_empty_header_list.py` holds eleven cases.
**Six of the first eight failed on the base commit `07d906d`** before either code change,
and the self-review added three more.

Five reversals measure the rules apart.

|---|---|
| Rule 1 alone, sentinel restored | 3 |
| Rule 2 alone, old pattern | 3 |
| Both reverted, the base | 5 |
| The lazy group `(.+?)` | 2, and the run costs 41.92 s |
| A path group that ends on a space | 2, and the run costs 6.87 s against 0.29 s |

The five gates, on macOS with Python 3.14:

```
.venv/bin/pytest tests/ -m "not spec_validation"   4762 passed, 8 skipped, 8 xfailed
.venv/bin/pytest tests/ -m spec_validation         1635 passed, 142 skipped, 141 xfailed
.venv/bin/ruff format --check ja4plus/ tests/      231 files already formatted
```

**The conformance suite moves one skip to one xfail.** The base reports 1635 passed, 143
skipped and 140 xfailed. `tests/foxio_deviations.json` holds 141 keys against 141 xfailed
cases.

## Documentation

`docs/specs/foxio/JA4H.md` R19 and the agreement table, `docs/implementation_notes.md`
(two new sections), `docs/methods/ja4h.md`, `CHANGELOG.md` and the `## Changelog` table of
`docs/specs/spec.md`, both with the literal `TBD`.

Verified against: https://github.com/FoxIO-LLC/ja4/tree/main/technical_details (FoxIO pin
`27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8`, as the issue records)
Verified against: https://docs.python.org/3/library/re.html (atomic grouping `(?>...)` is
new in Python 3.11, retrieved 2026-08-15)


Crank-Git:
Self-review, measurement 1: the replay covers every method and not JA4H alone.

`is_http_request` of `ja4plus/utils/http_utils.py` gates a JA4L measurement point, so a
pattern change can move a JA4L value. The first replay of this branch read JA4H alone,
which measures the wrong thing. **This replay reads every method through
`tests/conformance_index.py:index_produced`**, which is the reader the conformance suite
itself uses, over all 38 committed captures.

| Read | Values | SHA-256 of the record |
|---|---|---|
| Base `07d906d` | 1677 | `b7c9e324f292bdacd8c590b697fe8154bd54e40029ac58e4e06cfd4bd1329dcd` |
| This branch | 1679 | `8dfc8582bd8412d28b90b94b759383f12a182a5c78020189d973632d39782db5` |

**Two values move, both on one connection of `gre-erspan-vxlan.pcap`, and both move from
no value to a value.**

```
gre-erspan-vxlan.pcap | (('100.20.9.1', '80'), ('100.20.9.2', '65174')) | JA4H    | before: None | after: ['ge10nn000000_e3b0c44298fc_000000000000_000000000000']
gre-erspan-vxlan.pcap | (('100.20.9.1', '80'), ('100.20.9.2', '65174')) | JA4H_ro | before: None | after: ['ge10nn000000__']
```

**No JA4L value moves, on that capture or on any other.** Every other value of all 1677 is
byte identical.

**The JA4H_ro value follows the FoxIO Python reference and not the dissector**, and that is
correct. `tests/foxio_vectors/https-connect.pcap.json` holds
`co10nn010000_User-Agent_` for a request that carries no cookie, so the reference ends the
raw form after the header names and one underscore.
`_generate_ja4h_raw_from_info` at `ja4plus/fingerprinters/ja4h.py:535` writes that form,
and #131 records the ruling. The FoxIO Wireshark dissector writes
`ge10nn000000___` in `tests/foxio_vectors/wireshark_expected/gre-erspan-vxlan.pcap.json`,
which carries the two empty cookie parts. **This change moves neither form**, and the
disagreement predates it.

**Note for a later reader: no case compares that JA4H_ro value.** `_method_params` of
`tests/test_spec_validation.py:155` builds one parameter set for each method of
`REPORTED_METHODS`, plus each method the expected file names. `JA4H_ro` is in neither list
for this capture, so the suite reports one extra occurrence key for JA4H and none for
JA4H_ro. **The register therefore holds one key and not two**, and a second key would name
a case that does not exist.

Crank-Git:
Self-review, measurement 2: the first push carried a processor-time defect, and `48f7322`
repairs it.

**The claim of the first push was wrong.** Its comment stated that the leading class
`[^ \t\r\n]` bounds the cost to the line length. That reasoning covers the split between
the first separator and the path, and it covers nothing after the path. **The second
separator is where the two quantifiers overlap.** The lazy path body `[^\r\n]*?` accepts a
space, and the greedy `[ \t]+` after it accepts a space, so a run of spaces admits one
split for each space. A version token that the line almost holds makes the match retry
every one of them.

**The case the first push shipped does not reach that shape.** `b"GET" + b" " * 3000`
holds no path and no near version token, so it stayed linear under both forms and it
reported a pass that measured nothing.

Measured on 2026-08-15, on the pattern object:

| Payload | Path group `[^ \t\r\n][^\r\n]*?` | The repaired group |
|---|---|---|
| `GET a` + 8000 spaces + `HTTPX` | 186.65 ms | 0.15 ms |
| `GET a` + 16000 spaces + `HTTPX` | 733.85 ms | 0.32 ms |
| `GET a` + 32000 spaces + `HTTPX` | 3017.9 ms | 0.630 ms |
| `GET a` + 16000 `space tab` pairs + `HTTP/x` | 3320.0 ms | 0.593 ms |

Doubling the run quadruples the first column and doubles the second.

**This reaches production code and not the pattern alone.** `is_http_request` reads
`REQUEST_LINE_LIMIT` = 8192 bytes of every TCP payload, and `ja4l.py:406` and
`ja4h.py:137` each call it. One packet whose payload starts with a method character
outside the nine fast-path tokens paid about 200 milliseconds of processor time.

**The base commit `07d906d` does not carry this.** `\s` and `\S` are disjoint, so the old
pattern held no overlap. **The defect was this branch's, and it is repaired before the
merge.**

## The repair

```
[ \t]+([^ \t\r\n](?:[^\r\n]*?[^ \t\r\n])?)[ \t]+
```

**The path reads neither a space nor a horizontal tab as its last character.** A path may
therefore end only at a character of that kind, so the candidates are those characters and
not every position of a space run. Each candidate reads the run after it once, and the
runs are disjoint, so the total cost is the line length. **The rule moves no match**,
because the greedy separator already consumed every space of a run, so no match ever ended
the path on a space.

**A differential run over 400000 random request lines reports no disagreement** between
the two forms. The alphabet holds the method characters, the separators, both line-ending
bytes, `HTTP/1.023`, the vertical tab and the form feed.

**The replay record is unchanged.** All 38 captures produce 1679 values under the repaired
form, and the record hashes
`8dfc8582bd8412d28b90b94b759383f12a182a5c78020189d973632d39782db5`, which is the same
SHA-256 the first push produced.

**Three cases now hold the bound**, and two of them are new.
`test_reads_a_line_that_almost_holds_a_version_token_in_linear_time` and
`test_reads_a_line_of_spaces_and_tabs_that_almost_holds_a_version_token_in_linear_time`
each fail under the shipped form. The whole file runs in **6.87 seconds** under that form
and in **0.29 seconds** under this one, and each timing case holds a limit of 1.0 second
against a good path of about 0.6 milliseconds.

**The Go half needs no such rule.** `regexp` of Go runs a finite automaton and it
backtracks nowhere, so `Crank-Git/ja4plus-go#575` is correct as it stands. An atomic group
would state the rule directly, and `(?>...)` needs Python 3.11 while `pyproject.toml`
reads `requires-python = ">=3.10"`.

## Two further findings of the same review

**The rule numbers of the test file were wrong.** Its docstrings named R18 and R27, which
belong to the image numbering the Go issue quotes. `docs/specs/foxio/JA4H.md` holds R12,
R17 and R19, and R18 of that page is a different rule. `48f7322` repairs the citation.

**Rule 1 carries a fourth behaviour change that the first push did not name.** Python `\s`
matches the vertical tab and the form feed, and `[ \t]` matches neither. Those two bytes
therefore separate no field now, and the path group reads each one as an ordinary
character. RFC 9112 names the space as the separator of a request line and it names
neither byte, so the repaired pattern is the closer reading. No vector reaches it, and
`docs/implementation_notes.md` records it.

## One finding declined

**The STE review asked for `read` in place of `reading` as a noun, under rule 14.** This
project uses `reading` as a domain noun throughout `.claude/rules/batch-gate.md`,
`docs/specs/spec.md` and `docs/specs/foxio/JA4H.md`. A change in this branch alone would
rotate a synonym against the corpus, which rule 7 refuses. The word stays, and a rename
belongs to the `## Terms` table rather than to one pull request.

Issue #628, opened 2026-08-15
Fuzz the ServerHello reader against a crafted extension block


Crank-Git:
**#617 repaired one crash of `_parse_server_hello`, and this port holds no fuzz target for that entry point.** The Go library found the same defect with a fuzz target in 1.14 seconds on its first run. **This project found it by reading the port**, so the defect was unfound rather than absent.

## What this issue asks for

**A fuzz target that reads a byte string and calls the ServerHello path.** `.github/workflows/test.yml` holds a `fuzz` job already, and a read of 2026-08-10 recorded 127 cases in it. The target belongs beside the ones that job runs.

**The seed corpus holds the two inputs #617 records.** The 53-byte record and the 64-byte input the Go fuzzer wrote. A reversal of the #617 repair then fails the ordinary test run.

## What this issue does not ask for

**No change to the reader.** #617 landed the repair, and this issue measures the reader rather than moves it.

## The rule this serves

`CLAUDE.md` states it. **Every packet is hostile input.** No parser trusts a length field it read from the packet, and a parser that cannot read a packet returns nothing rather than raises.

## Reversal path

#617, and `Crank-Git/ja4plus-go#556`.

Crank-Git:
Plan. Batch #642, integration branch `batch/642-request-line-and-tunnel`.

**Approach.** No file of `tests/fuzz/` names the ServerHello entry point. The five files there are `test_length_fields.py`, `test_malformed_captures.py`, `test_structural_validity.py`, `test_synthetic_hostile_input.py` and `test_truncated_captures.py`, and a grep for `server_hello` across them returns nothing. The Go port found the #617 crash in 1.14 seconds with a fuzz target, and this port found it by reading code.

Add a fuzz target that calls `_parse_server_hello` over arbitrary bytes.

**Files.** `tests/fuzz/` alone. **Do not edit any file under `ja4plus/`.**

**The shape.** This suite is hypothesis and pytest, and not go-fuzz. Follow the shape the five existing files already use rather than importing a new tool. Seed the two inputs the issue names: the well-formed record, and the 53-byte record of #617. Assert that the reader raises nothing and returns nothing it cannot justify.

**Out of scope.** Any repair of the reader. `4497e9f` closed the one branch that held the defect, and #629 recorded the audit that found no second branch. This issue builds the standing proof, and it fixes nothing.

[...]

Issue #617, opened 2026-08-14
_parse_server_hello raises IndexError on a crafted supported_versions extension


Crank-Git:
**`_parse_server_hello` raises `IndexError` on a crafted TLS record, and no caller catches it.** The Go library carried the identical defect, and `Crank-Git/ja4plus-go#556` repaired it on 2026-08-14.

**A fuzz target found the Go half in 1.14 seconds on its first run.** This port has no fuzz target for the same entry point, so the defect is unfound rather than absent.

## The defect

**Read at tag `v1.1.0`, `ja4plus/utils/tls_utils.py:278-279`, inside `_parse_server_hello`.**

The function clamps `ext_data_end` to `len(raw_data)`, **and then it never uses the clamped slice.** It tests the length the wire declares and it indexes the raw buffer:

```
if ext_len >= 2:
    ... raw_data[ext_data_start] ...
```

**`ext_len` is the declared length, and the clamp is what bounds the data that is actually present.** A record that declares a `supported_versions` length of 2 or more, and that ends at the extension header, makes `ext_len >= 2` true while no byte follows. **`raw_data[ext_data_start]` then raises `IndexError`.**

**Neither `_parse_server_hello` nor its only caller at `ja4plus/utils/tls_utils.py:101` holds a `try` or an `except`.** So the exception reaches the caller of the library.

## The input that reproduces it

**A minimal 53-byte record.** On these bytes the caller passes the full 53 bytes and `ext_data_start` is 53.

```
0x03 0x03                    the version
32 bytes of zero             the random
0x00                         the session identifier length
0x13 0x01                    the cipher suite
0x00                         the compression method
```

followed by an extension block that declares `supported_versions` with a length of 2 or more and supplies no byte of it.

**A second input reproduces it too**, and the Go fuzzer wrote it. It is 64 bytes, and the bytes `0x00 0x2b` are the `supported_versions` extension type:

```
go test fuzz v1
[]byte("\x1600\x009\x020000000000000000000000000000000000000\x000000000\x00\x0500000\x00+0000")
```

## The repair the Go library made

**One line.** The branch now reads the length of the data it indexes, and never the declared length:

```go
if len(extData) >= 2 {
```

[...]

PR #648, opened 2026-08-15
Fuzz the ServerHello reader against a crafted extension block

Crank-Git:
**`tests/fuzz/` named no ServerHello entry point.** A grep for `server_hello` across the
five case files of that directory returned nothing. #617 records that a fuzz target found
the identical defect of the Go library in 1.14 seconds, and that this project found its
own half by reading the port. **A read is not a standing proof**, so this change writes
the proof rather than repeats the read. #628 states the ask.

## What this change adds

`tests/fuzz/test_server_hello_extensions.py` holds 490 cases and it runs in 0.16 seconds.

**The target measures the class of defect and not the one instance.** A record declares
one length in an extension header, and it supplies another count of bytes. The corpus
holds five parts.

| Part | Count | What it varies |
|---|---|---|
| The crafted grid | 240 | Six extension types, eight declared lengths, five supplied counts |
| A random extension block | 512 | The whole block, under a fixed seed |
| Random bytes | 512 | The whole record, under a fixed seed |
| The byte-flip sweep | 165 | Three bit groups over each of the 55 bytes of a well-formed record |
| The seeds of #617 | 2 | The 53-byte record, and the 64-byte input the Go fuzzer wrote |

**Each part drives `_parse_server_hello`, and the grid drives `parse_tls_handshake` as
well.** #617 records that the exception reached the caller of the library, because neither
the reader nor its one caller holds an `except`. The entry-point cases read a `spy_on`
counter, so a case that reaches no reader fails.

**Every case states what the reader produced.** `tests/fuzz/README.md` records that a case
which asserts "no exception" passes where no parser runs.

## The target discriminates

**A reversal of the `4497e9f` guard at `ja4plus/utils/tls_utils.py:282`**, back to
`if ext_len >= 2:`, reads this.

```
27 failed, 463 passed, 1 warning in 0.32s
E                       IndexError: index out of range
ja4plus/utils/tls_utils.py:283: IndexError
```

**The restored file reads `490 passed, 1 warning in 0.16s`.** `git checkout -- ` restored
it, and `git diff --stat` reports no file under `ja4plus/`.

## What bounds the run time

**The suite imports no new tool.** The five existing files of `tests/fuzz/` each draw from
`random.Random` under a fixed seed, and this file follows that shape. The case count is
therefore fixed at 490, and no time budget or deadline decides it. `DRAWS` is 512 and
`SEED` is 628, so a failure repeats.

`.github/workflows/test.yml:417` runs `pytest tests/fuzz/ -v` as the `fuzz` job. That
selection moves from 127 cases to 617, and it runs in 1.41 seconds.

## What this change does not do

**It edits no file under `ja4plus/`**, so no fingerprint moves. #628 puts every repair of
the reader out of scope. The corpus found no new crash.

## How this was tested

| Gate | Result |
|---|---|
| `.venv/bin/pytest tests/ -m "not spec_validation"` | 5213 passed, 8 skipped, 8 xfailed, 114 subtests passed in 98.47s |
| `.venv/bin/pytest tests/ -m spec_validation` | 1635 passed, 143 skipped, 140 xfailed in 9.69s |
| `.venv/bin/ruff check ja4plus/ tests/` | All checks passed! |
| `.venv/bin/ruff format --check ja4plus/ tests/` | 229 files already formatted |
| `.venv/bin/mypy --strict ja4plus/` | Success: no issues found in 31 source files |
| `.venv/bin/pytest tests/fuzz/ -v` | 617 passed in 1.41s |

The conformance figures match the base commit exactly: 1635 passed, 143 skipped, 140
xfailed.

**Coverage falls nowhere.** This change adds cases and it removes none, and it edits no
source file.

## The Changelog

`CHANGELOG.md` and the `## Changelog` table of `docs/specs/spec.md` each carry the literal
`TBD`. The project manager assigns the round number at the batch gate.

## One note on the plan

**The plan named the suite as hypothesis and pytest, and no file of this repository
imports hypothesis.** `pyproject.toml` declares it nowhere. The plan also stated the rule
that decides the case: follow the shape the five existing files use, rather than import a
new tool. This change follows that rule, and it names the measurement here so no later
reader looks for a profile that does not exist.

Refs #628.


Crank-Git:
## Self-review, two lenses

**Correctness and the writing standard each read the diff.** `1d9bfac` holds every repair.

### The correctness lens found one wrong comment, and it is repaired

**`_parse_server_hello` branches on two extension types, which are 0x0010 and 0x002b.**
`ja4plus/utils/tls_utils.py:270` and `:277` hold the two. The comment above
`EXTENSION_TYPES` named four, which describes `_parse_client_hello` at
`ja4plus/utils/tls_utils.py:178` and `:194`. **The list of six types stands**, because a
type the loop skips belongs to the class this file measures: the loop advances on the
declared length whether a branch reads the data or not. The comment now states that
reason, and it states why 0x0000 and 0x000d stay in the list.

### The correctness lens read `assert silenced == 0` as true by construction, and that reading is right

**`_parse_server_hello` returns None on one condition, which is a record shorter than 11
bytes.** A flip of one bit moves no length of the record, so that count holds at zero for
the reader as it stands. **The assertion measures no corruption**, and the comment claimed
otherwise. The comment now states the shape of the reader. The case keeps its value in the
`handshake_type` assertion it makes on each of the 55 mutations, and in the raise it
refuses.

### The correctness lens found no vacuous case

It reran the corpus against a reader with the guard removed, and 35 cases turned red. That
mutation is not the one this pull request records. **This pull request reverses the guard
to `if ext_len >= 2:`**, which is the exact line `4497e9f` replaced, and that mutation
reads 27 failed and 463 passed. Both numbers are measurements of their own mutation, and
the restored reader reads 490 passed.

It confirmed the three record lengths by hand and by measurement: the seed record is 53
bytes, the well-formed record is 55, and `FUZZER_INPUT` is 64. It confirmed that
`assert produced == DRAWS` states a real invariant, because every record the builder
writes clears the three early guards whatever the extension block holds.

### The writing standard lens found two sentences past 25 words

**Rule 1 refuses both**, and neither file claims the exemption, which reaches the entries
of `CHANGELOG.md` and the Changelog rows of `docs/specs/spec.md` alone.

- The module docstring of `tests/fuzz/test_server_hello_extensions.py` held a 27-word
  sentence. It is now two sentences.
- `tests/fuzz/README.md` held a 35-word sentence that carried five items. **Rule 16 asks
  for a vertical list once a sentence carries more than two conditions**, so the five
  corpus parts are now a list.

One test name read `..._that_passes_the_record_...`, where `passes` carried a second
meaning beside the one rule 6 allows. The name now reads
`test_a_block_length_past_the_end_of_the_record_produces_a_server_hello`.

### The gates after the repair

| Gate | Result |
|---|---|
| `pytest tests/ -m "not spec_validation"` | 5215 passed, 8 skipped, 8 xfailed, 114 subtests passed in 131.36s |
| `pytest tests/ -m spec_validation` | 1635 passed, 143 skipped, 140 xfailed in 15.73s |
| `ruff format --check ja4plus/ tests/` | 229 files already formatted |
| `pytest tests/fuzz/ -v` | 617 passed in 1.48s |

**The unit suite moved from 5213 cases to 5215.** The vertical list of
`tests/fuzz/README.md` adds two parametrized cases to the prose readers, and the
conformance figures moved nowhere.

Crank-Git:
## Correction: the case count moved for a different reason

**The comment above stated that the vertical list of `tests/fuzz/README.md` adds two
cases. That statement is wrong, and I measured it rather than left it.**

**The two cases belong to the two prose readers, and they arrived when the new file became
tracked.** `tests/test_documented_method_count.py:434` and
`tests/test_ruling_vocabulary.py:370` each build their corpus with `git ls-files`, which
lists a tracked file and never an untracked one. The first run of the unit suite read 5213
passed, and `tests/fuzz/test_server_hello_extensions.py` was untracked at that moment, so
it reached neither reader. `5583478` tracked the file, and each reader then holds one more
parametrized case.

**The measurement that corrects it.** I took the previous `tests/fuzz/README.md` alone,
collected the suite, and compared the per-file counts against the head of this branch. The
two collections are identical. I then took the previous
`tests/fuzz/test_server_hello_extensions.py` alone and repeated it, and that comparison is
identical too. **Neither edit of `1d9bfac` moves one case.**

The suite reads 5231 collected at this head, which is 5215 passed, 8 skipped and 8
xfailed. `pytest tests/test_documented_method_count.py tests/test_ruling_vocabulary.py`
reads 787 passed.

**Read a case-count move against `git ls-files` before you read it against the prose.** A
new tracked file under `ja4plus/`, `tests/` or `examples/` adds one case to each of the
two readers, and the run that precedes the first commit of that file reports neither.

Issue #605, opened 2026-08-13
Add ja4ls as a --types token that selects the server values


Crank-Git:
## The proposal

**Add `ja4ls` to `VALID_TYPES` as a filter token that selects the JA4LS values alone.**
`--types ja4l` keeps printing both the JA4L values and the JA4LS values, so no existing behaviour
changes.

`ja4plus/cli.py:53-64` at `v1.1.0` holds ten tokens, and `_parse_types` at `:82-93` rejects any
token outside that list. **A caller who wants the server values alone therefore has no filter for
them**, and `docs/methods/ja4ls.md` tells the reader to filter on `ja4l` and then read the
`JA4L-S=` prefix by hand.

## Why the Go port is asking

**`Crank-Git/ja4plus-go` adopted the token on 2026-08-13**, under its issue #61 and R9 question 3
of its specification. **The two implementations produce identical values**, and the token adds a
filter over that identical output rather than changing it.

**The Go maintainer states the methodology as one to one or more.** This is one of the `or more`
cases: the output is the same, and one implementation offers a filter the other does not.

**This issue exists so the two converge rather than drift.** A filter token is not a fingerprint
value, so `.claude/rules/parity.md` of the Go repository does not bar the Go half from landing
first. **The register needs no row**, because no value moves.

## What it needs

- `ja4ls` joins `VALID_TYPES`.
- `ja4ls` selects a result whose value carries the `JA4L-S=` prefix. **The result type stays
  `ja4l`**, which `ja4plus/processor.py` sets from `_SPEC`, so the filter reads the value prefix
  and never the type.
- `ja4l` keeps selecting both, so no caller of `v1.1.0` sees a change.
- The help text and `docs/methods/ja4ls.md` state the two tokens.

**Decline it and the two implementations still agree on every value.** This is a convergence
request, and not a parity defect.

Crank-Git:
Plan. Batch #642, integration branch `batch/642-request-line-and-tunnel`.

**Approach.** `ja4plus/cli.py:53-63` holds `VALID_TYPES` with ten tokens, and no token selects the JA4L server values alone. `--types ja4l` selects the client values and the server values together. `ja4plus-go` adopted the `ja4ls` token on 2026-08-13, under its issue #61.

**Parity rule 2 decides this.** FoxIO specifies no command-line interface, so the port decides it, and the port has shipped. Adopt `ja4ls`.

**Files.** `ja4plus/cli.py`, and tests under `tests/`.

[...]

PR #651, opened 2026-08-15
Add ja4ls as a --types token that selects the server values

Crank-Git:
## What

`--types` accepts the token `ja4ls`, which selects the JA4LS values alone. Refs #605.

`VALID_TYPES` of `ja4plus/cli.py` moves from ten tokens to eleven, and `ja4ls` joins the
list after `ja4l`. **`--types ja4l` keeps selecting both latency values**, so no caller of
version 1.1.1 sees a change. This round adds a filter and it narrows none.

## Why the filter reads the value and not the type

**The filter cannot key on the result type.** `ja4plus/processor.py` sets the type from
`_SPEC`, and `JA4LFingerprinter` writes both latency methods, so a JA4L value and a JA4LS
value each carry the type `ja4l`. The new `_selecting_token` of `ja4plus/cli.py` therefore
tests `fingerprint.startswith("JA4L-S=")`, which `ja4plus/fingerprinters/ja4l.py:446`
writes.

**It reads `ja4l` before `ja4ls`.** A JA4LS value that both tokens select then reports at
the position of `ja4l`, so a run that names no `--types` option keeps the output order of
version 1.1.1.

## The reading of the Go port

**Parity rule 2 decided the token**, because FoxIO specifies no command-line interface and
the port shipped one on 2026-08-13 under its issue #61.

| Question | The answer the port states |
|---|---|
| Does `ja4l` still select both values? | Yes. `admitsResult` at `cmd/ja4plus/types.go:85-86` returns `filter["ja4l"] \|\| filter["ja4ls"]`. |
| What does `ja4ls` select? | The server value alone. `cmd/ja4plus/types.go:22` holds the token, and `:34` holds the same `JA4L-S=` prefix constant. |

**The port narrowed nothing**, so the token is additive in both implementations and
neither command line breaks. **The two token lists still order the methods differently**,
because the port lists `ja4t` and `ja4ts` before `ja4l` where this project lists `ja4l`
first. This change moves no existing token.

Verified against: `/Users/christiancrank/Development/Personal/ja4plus-go/cmd/ja4plus/types.go`
and `cmd/ja4plus/main.go:139-141` (read 2026-08-15). No file of the port was edited.

## How it is tested

**Eight cases came first and five of the eight failed.** The three that passed are the
guards this change must not break: `ja4l` selects both values, a default run selects every
method, and `_parse_types` declines an unknown token.

```
FAILED tests/test_cli_types_ja4ls.py::test_the_token_list_holds_ja4ls
FAILED tests/test_cli_types_ja4ls.py::test_the_ja4ls_token_selects_the_server_value_alone
FAILED tests/test_cli_types_ja4ls.py::test_the_ja4ls_token_declines_the_result_of_another_method
FAILED tests/test_cli_types_ja4ls.py::test_the_two_latency_tokens_together_select_both_values
FAILED tests/test_cli_types_ja4ls.py::test_the_parser_reads_the_ja4ls_token_in_upper_case
5 failed, 3 passed, 1 warning in 0.80s
```

`tests/test_cli_types_ja4ls.py` now holds eleven cases. Three of them run the committed
capture `tests/foxio_vectors/https-connect.pcap` through `main` and read the JSON lines,
so the token is proved end to end and not at the helper alone.

## The finding of this round

**A token is no longer a `type` value**, and one case of `tests/test_method_pages.py`
assumed the two were the same.
`test_the_example_of_each_method_page_comes_from_the_capture_it_names` compared the
`--types` token of a page against the `type` value of each result, so `ja4ls` matched
nothing:

```
AssertionError: tests/foxio_vectors/https-connect.pcap emitted no ja4ls fingerprint
```

That case now reads `_selecting_token`, so the token of a page means what it means on the
command line.

## Documentation

`docs/methods/ja4ls.md` states `ja4ls` as its `--types` token. `docs/methods/ja4l.md`,
`docs/methods/index.md`, `docs/output-schema.md`, `docs/api_reference.md` and `README.md`
each state the pair and the count of eleven. **The `type` field still holds one of ten
values**, and the `### Result` row of `docs/specs/spec.md` plus the docstring of
`ja4plus/types.py` now separate that count from the eleven tokens `--types` accepts.

`CHANGELOG.md` and the `## Changelog` table of `docs/specs/spec.md` each record the round
as `TBD`.

## The five gates

```
.venv/bin/pytest tests/ -m "not spec_validation"
4740 passed, 8 skipped, 1965 deselected, 8 xfailed, 16 warnings, 114 subtests passed in 121.16s

.venv/bin/pytest tests/ -m spec_validation
1642 passed, 143 skipped, 4799 deselected, 137 xfailed, 1 warning in 15.11s

.venv/bin/ruff check ja4plus/ tests/
All checks passed!

.venv/bin/ruff format --check ja4plus/ tests/
229 files already formatted

.venv/bin/mypy --strict ja4plus/
```

**No fingerprint moves**, because this round changes selection and never production. The
conformance counts equal the counts of the base, and `tests/foxio_deviations.json` holds
137 keys against 137 xfailed cases.


Crank-Git:
## Self-review

Two lenses ran over the diff. **Correctness reported no finding. Documentation reported three, and this branch takes two of them.**

### Correctness — clean

The lens traced `_selecting_token` over each combination of `{ja4l present, ja4ls present, neither}` against `{a JA4L-C value, a JA4L-S value, another method}`, and it read four other questions.

- **`order[token]` raises no `KeyError`.** The helper returns a token it has just tested `in order`, or None, and each caller drops the None before it indexes.
- **The sort stays stable.** `FingerprintResult` is a frozen dataclass with no `order=True`, so it holds no `__lt__`. `selected.sort(key=lambda pair: pair[0])` compares the key alone, so the sort never reaches the second element of the tuple and an equal key keeps the emission order.
- **`close_open_windows` always writes `fingerprint`.** `ja4plus/processor.py:270-277` sets the key on every dict it emits.
- **The default output order does not move.** A default run holds both tokens, and the helper reads `ja4l` first, so every latency value keeps the position of `ja4l`.

### Documentation — two findings taken, one declined

**Finding 1, taken. The `## Terms` table barred the word this change needs.** The `method name` row of `docs/specs/spec.md` read `token` in its `Do not use` column, and it defined a method name as "The lowercase token that names one fingerprinter". **`ja4ls` is a value `--types` accepts and it names no method**, so the two concepts are no longer one. The table now holds the term `types token`, and the `method name` row no longer bars the word.

**Finding 2, taken. One sentence of the round record used a non-finite clause.** The `being the guards` construction breaks rule 11 and rule 12, which the Changelog exemption does not cover. `CHANGELOG.md` and the `## Changelog` row each carry three finite sentences now.

**Finding 3, declined as no finding.** The lens read `docs/methods/ja4ls.md:9`, which tells a reader to read the `JA4L-S=` prefix to tell the two methods apart, and it asked whether `ja4ls` makes that advice stale. **It does not.** The sentence describes the `type` field of an output line, which still reads `ja4l` for both methods, and this change moves no `type` value. The filter advice that `ja4ls` replaces stood in a different paragraph, and this branch rewrote it.

### What neither lens checked

- **No provider run covers this branch**, because every commit ends with the skip keyword by design. The batch pull request is the first run of these cases on Linux.
- **The `mkdocs build --strict` step did not run here.** This change edits six published pages, and the `docs-build.yml` workflow filters on those paths, so the batch pull request reaches it.

Issue #657, opened 2026-08-15
Batch: two tunnel and quoted-header repairs, and two stale citations


Crank-Git:
This batch holds four members. They write four disjoint file sets, and none of them writes a file that `batch/631-header-terminator` still holds.

- [x] #610 — read the TCP header that an ICMP error message quotes for JA4T. `ja4plus/fingerprinters/ja4t.py`.
- [x] #646 — read the inner UDP layer on a tunneled DHCP packet. `ja4plus/fingerprinters/ja4d.py:198` and `ja4plus/fingerprinters/ja4d6.py:267`. #594 repaired the QUIC branches and filed this one rather than widen its own scope.
- [x] #616 — the Wireshark pin reads 4.4.2 and the FoxIO pin records v4.6.0. `.claude/rules/external-apis.md:35`.
- [x] #637 — the R10 citation of the JA4T transcription names the comment and not the test. `docs/specs/foxio/JA4T.md:162`.

Continuous integration runs once, on the batch pull request into `dev`.

Crank-Git:
finding: batch cross-check, batch #657

**What was compared.** The plans of #610, #646, #616 and #637, against each other, on file, function, interface, migration, configuration key and route.

**The pairs found.**

1. **#616 and #637 both read the Wireshark dissector at a pinned version, and one of them may move that pin.** #616 may write v4.6.0 in place of 4.4.2, and #637 corrects a line citation read at the pin `27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8`. **A line number is meaningless without the version it was read at.** They are sequenced: #637 corrects the citation at the present pin first, and #616 then reads every citation against whatever pin it writes. #616 carries the instruction to report any citation that moves.

**The clean negatives.**

| Member | Files it writes |
|---|---|
| #610 | `ja4plus/fingerprinters/ja4t.py` |
| #646 | `ja4plus/fingerprinters/ja4d.py`, `ja4plus/fingerprinters/ja4d6.py` |
| #616 | `.claude/rules/external-apis.md` |
| #637 | `docs/specs/foxio/JA4T.md` |

- **No two members write one file.**
- **#610 and #637 both concern JA4T and they do not meet.** #610 changes which packets the fingerprinter reads, and #637 corrects a citation of the transcription. Neither reads a shape the other changes.
- **No member adds a public name, a configuration key, a route or a migration.**
[...]

Issue #646, opened 2026-08-15
Read the inner UDP layer on a tunneled DHCP packet


Crank-Git:
Found while #594 routed the JA4 and JA4S QUIC branches through `innermost_layer`. **This issue is not that repair**, and #594 declined it as out of scope.

## The reading

`ja4plus/utils/tunnels.py:50` defines `innermost_layer`, and `ja4plus/utils/packet_utils.py:107` calls it to read the inner port layer. The docstring at `packet_utils.py:85-86` states the rule: the address pair names the outer address layer, and the port pair names the innermost port layer.

**The JA4D and JA4D6 branches call `packet.getlayer(UDP)` instead.**

- `ja4plus/fingerprinters/ja4d.py:198` — `udp = packet.getlayer(UDP)`
- `ja4plus/fingerprinters/ja4d6.py:267` — `udp = packet.getlayer(UDP)`

`Packet.getlayer` takes the layer count as its second parameter `nb`, which defaults to 1, so it returns the first layer of the class counting from the outside. On a tunneled packet that layer is the UDP header of the tunnel. `scapy/packet.py:1317-1327` at scapy 2.7.0 holds the signature and the docstring `Return the nb^th layer that is an instance of cls, matching flt values.`.

**These are the two branches that #594 left.** A read of 2026-08-15 found `getlayer(UDP)` on four lines of `ja4plus/`, and #594 repaired two of them.

## What it means

A DHCP message inside a VXLAN, GRE, ERSPAN or Geneve tunnel reaches the outer header. The DHCP reader then reads the tunnel header bytes rather than the DHCP message, so it produces no value. A relay agent in an overlay network sends such a message.

## Why it is not urgent

**No capture of the FoxIO corpus carries DHCP inside a tunnel**, so no value moves and no vector separates the two behaviours. `tests/foxio_vectors/dhcp.pcapng` and `tests/foxio_vectors/dhcpv6.pcap` each hold an untunneled message. The replay #594 measured produced the same values before and after its change, over all 38 committed captures.

## What this issue asks for

Read the two branches against `innermost_layer(packet, (UDP,))`. Read the port ranges first: `ja4d.py` matches the DHCPv4 ports and `ja4d6.py` matches the DHCPv6 ports, and a tunnel port is neither, so the shape of the defect may differ from the QUIC one.

[...]

Issue #642, opened 2026-08-15
Batch: the request line, the tunnel layer, the reset value and a fuzz target


Crank-Git:
This batch holds four members. Every one is verified against the code before it was scheduled, and the four write four different files.

- [x] #605 — add `ja4ls` as a `--types` token that selects the server values. `ja4plus/cli.py:53-63`.
- [x] #594 — read the inner UDP layer on a tunneled QUIC packet. `ja4plus/fingerprinters/ja4.py:422` and `ja4plus/fingerprinters/ja4s.py:83`.
- [x] #609 — publish the four-part JA4TS value on a reset of a one-SYN-ACK connection. `ja4plus/fingerprinters/ja4ts.py:183`.
- [x] #628 — a fuzz target for the ServerHello reader. `tests/fuzz/` alone.

Continuous integration runs once, on the batch pull request into `dev`. Every member pull request is a draft that carries the skip keyword.

Crank-Git:
finding: batch cross-check, `batch/642-request-line-and-tunnel`

**What was compared.** The plans of #594, #609, #628 and #605, against each other, on file, function, interface, migration, configuration key and route.

**The pairs found.** None. The four members write four disjoint file sets.

| Member | Files it writes |
|---|---|
| #594 | `ja4plus/fingerprinters/ja4.py`, `ja4plus/fingerprinters/ja4s.py` |
| #609 | `ja4plus/fingerprinters/ja4ts.py` |
| #628 | `tests/fuzz/` alone |
| #605 | `ja4plus/cli.py` |

**The clean negatives.**

- **No two members write one file**, as the table states.
- **No two members own the same new name.** #605 adds the token `ja4ls` to `VALID_TYPES`. No other member adds a public name or a token.
- **No member reads a shape another member changes.** #605 reads the value prefix `JA4L-S=`, which #609 does not write; #609 writes a JA4TS value, which #605 does not read.
- **No member changes a configuration key, a route or a migration.**
- **Two members move a fingerprint and two do not.** #609 publishes a value where the port published none, and #594 changes which layer two fingerprinters read. Each carries its own byte-identical replay proof, and neither reads the other's path. #628 and #605 move no value.

**One cross-batch reading, because two other integration branches are live.**

[...]

Issue #615, opened 2026-08-14
Record the JA4D option 81 rule as a divergence-register row


Crank-Git:
The maintainer ruled `Crank-Git/ja4plus-go#371` answer 2 on 2026-08-14. **The port already ships the ruled behavior, and this issue asks for the record and not for a code change.**

## What was ruled

**JA4D part a writes `n` when option 81 carries no domain name, and `d` when it carries one.** The mere presence of option 81 does not decide the flag.

## The deciding fact

**The rank 1 image label reads `Has a Domain name (d) or No domain (n)`.** It names the domain name and it does not name the option.

| Implementation | What decides the flag | Evidence |
|---|---|---|
| Wireshark | the name | `wireshark/source/packet-ja4.c:1521` |
| **The port** | **the name** | **`ja4plus/fingerprinters/ja4d.py:161-165` at tag `v1.1.0`** |
| Zeek | the presence | `zeek/ja4d/main.zeek:73` |

**Zeek contradicts the image, and the image outranks it.** The Zeek `d` reads as an upstream dissector artifact and not as a JA4 disagreement.

## What the port does

**No line of `ja4d.py` changes.** The port already writes `n` on an option 81 that carries no name.

**The divergence register holds no row for D3.** The answer is recorded only as a round 78 reading, and a reading is not a register row. **So the Go repository could not adopt the port answer on that basis**, and the maintainer ruled the question from the image instead.

**This issue asks for the register row**, so that both repositories carry the same record and the next reader finds it.

## What it moves

**Zero values of the shared FoxIO vector set.** The corpus holds 4 DHCPv4 messages and `option81_occurrences` is 0. **The rule reaches every option 81 outside the corpus.**

**The Go repository builds the test that separates the two answers**, because no vector reaches the value. The port carries the same kind of test if it holds none.

## Reversal path

**`Crank-Git/ja4plus-go#371`**, and this issue.

Crank-Git:
confirmation: the Go half landed on 2026-08-14, and the two implementations answer alike for every input.

**`Crank-Git/ja4plus-go#371` merged the Go half.** The worker read this port at tag `v1.1.0` and
it reports the identical comparison.

| Implementation | The condition | Location |
|---|---|---|
| This port | `has_fqdn = has_fqdn or len(opt_data) > _DHCP_FQDN_NAME_OFFSET`, offset 3 | `ja4plus/fingerprinters/ja4d.py:165` and `:64` |
[...]

PR #659, opened 2026-08-15
Read the inner UDP layer on a tunneled DHCP packet

Crank-Git:
Closes nothing on its own. Refs #646, and the batch pull request of #657 closes the issue.

## What this changes

`ja4plus/fingerprinters/ja4d.py:198` and `ja4plus/fingerprinters/ja4d6.py:267` each read
`udp = packet.getlayer(UDP)`. `Packet.getlayer` takes the layer count as its second
parameter `nb`, which defaults to 1, so it returns the first layer of the class counting
from the outside. On a tunneled packet that layer is the UDP header of the tunnel.

Both branches now read `innermost_layer(packet, (UDP,))`, which
`ja4plus/utils/tunnels.py:50` publishes and which `ja4plus/utils/packet_utils.py:107`
already reads for the reported port pair.

## Why the defect shape differs from the QUIC one

The issue body asked for this reading, and the measurement confirms it.

**The defect shape of these two methods is an absent value, and it is not a wrong one.**
Each method reads a port range before it reads the payload. `ja4d.py` matches 67, 68 and
4011, and `ja4d6.py` matches 546 and 547. A tunnel port is none of those five, so the
outer header failed the port test and the method returned None. The QUIC branches that
#594 repaired held no port test, so they passed the tunnel header bytes to the QUIC
decoder instead.

## How this was tested

**The mutation reversal.** The proving case was written first and it failed against
`getlayer`.

```
FAILED tests/test_dhcp_tunnel_inner_udp.py::test_the_ja4d_fingerprinter_reads_the_dhcp_layer_of_a_tunneled_packet
FAILED tests/test_dhcp_tunnel_inner_udp.py::test_the_ja4d6_fingerprinter_reads_the_dhcpv6_layer_of_a_tunneled_packet
FAILED tests/test_dhcp_tunnel_inner_udp.py::test_the_ja4d_fingerprinter_reports_a_tunneled_message_on_the_inner_ports
FAILED tests/test_dhcp_tunnel_inner_udp.py::test_the_ja4d6_fingerprinter_reports_a_tunneled_message_on_the_inner_ports
4 failed, 2 passed
```

The untunneled control of each pair passed in that run, so two absent values pass no
case. One failure reads `assert None == 'disco1500id_57-55_1-3-6-15'`, which is the
absent value the section above states. Against `innermost_layer` the file reads
`6 passed`.

**The byte-identical replay.** A replay of the 38 captures of `tests/foxio_vectors/`
through both methods, before the change and after it.

| Reading | Before | After |
|---|---|---|
| JA4D values | 4 | 4 |
| JA4D6 values | 6 | 6 |
| SHA-256 of the record | `c0f751b74793e997b8e184c4fed7b76374af71a0f349458b51571a02b0aa7e8b` | `c0f751b74793e997b8e184c4fed7b76374af71a0f349458b51571a02b0aa7e8b` |

**The corpus holds 4 DHCPv4 messages and 6 DHCPv6 messages**, each on the ports its
method matches. #615 measured the first count and this round measured the second. The set
holds `gre-sample.pcap`, `gre-erspan-vxlan.pcap` and `tcpdump-geneve.pcap`, which are the
three tunneled captures, and none of the three carries DHCP.

**The five gates, all green.**

```
4796 passed, 8 skipped, 1961 deselected, 8 xfailed, 16 warnings, 114 subtests passed in 96.36s
1635 passed, 143 skipped, 4855 deselected, 140 xfailed, 1 warning in 10.53s
All checks passed!
232 files already formatted
```

## The bound of the descent

`innermost_layer` walks `layer.payload` until `NoPayload`, it indexes nothing and it reads
no length field of the packet, so the chain is finite whenever the dissection finished.
#594 established that reading and this round read the function again and confirms it.

## Scope

`git grep -n "getlayer(UDP)" ja4plus/` returns two lines on this branch.

```
ja4plus/fingerprinters/ja4.py:422:        udp = packet.getlayer(UDP)
ja4plus/fingerprinters/ja4s.py:83:            udp = packet.getlayer(UDP)
```

**Those two are the sites #594 already repaired, and the base of this branch predates that
repair.** `batch/657-tunnel-icmp-citations` was cut from `dev` at `6338d12`, and batch #642
merged #594 into `dev` afterwards at `36d9fd5`. The two lines therefore leave this tree
when the integration branch takes `dev`. This branch repairs neither one, because another
batch owns them.

## Records

`CHANGELOG.md` carries `Round TBD.` and the `## Changelog` table of `docs/specs/spec.md`
carries one appended row with the literal `TBD`. `docs/implementation_notes.md` gains the
section `### The DHCP layer of a tunneled packet` under the JA4D chapter, which is a
different anchor from the one #594 used, so it merges with `dev` without a conflict.

Verified against https://scapy.readthedocs.io/ (scapy 2.7.0, the installed release, whose
`scapy/packet.py:1317-1327` holds the `getlayer` signature and its `nb` parameter).


Crank-Git:
## Self-review

Three lenses ran over the diff: correctness, robustness against hostile input, and the
writing standard with the round records. Each one read the diff and the changed files.

### Correctness — no finding

- **`innermost_layer` and `getlayer` return one layer on an untunneled packet**, so no
  committed value moves. The replay measures that, and the conformance suite passes.
- **`bytes(udp.payload)` reads the same bytes whether scapy leaves the payload `Raw` or
  dissects it into `BOOTP`.** The review measured 269 bytes on each path and compared them
  equal. That was the one shape that could have produced a wrong value rather than a newly
  correct one, and it does not.
- **The new cases are not vacuous.** Each value case asserts a literal, and the untunneled
  control of each pair passed against `getlayer` while the tunneled half failed.
- **The import adds no cycle.** `ja4plus/utils/tunnels.py` imports `scapy.packet` and
  `logging`, and it imports nothing from `ja4plus`.
- **The import sits where `ja4.py` and `ja4s.py` put it on `dev`**, immediately above
  `from ja4plus.fingerprinters.base import BaseFingerprinter`.

### Robustness — no finding

- **The walk terminates.** `innermost_layer` reads `layer.payload` until `NoPayload`, it
  holds no recursion, and the chain is the one scapy already built. The review measured
  2000 calls at 0.7 ms.
- **The change opens no new raise path.** `sport` and `dport` stand on a `UDP` instance
  whichever function returned it. A fuzz of 500 truncated and random payloads through
  `generate_ja4d` raised nothing.
- **It retains no packet reference and it adds no state table.**

### The writing standard — five findings, all fixed in `4cf6e6f`

Rule 1 of `.claude/rules/ste.md` sets 25 words for a description sentence, and five
sentences of the new prose ran past it: the two code comments, and three docstrings of
`tests/test_dhcp_tunnel_inner_udp.py`. The longest read 36 words. Each one now splits into
two sentences and states the same facts. The exemption of rule 1 covers `CHANGELOG.md` and
the `## Changelog` table of `docs/specs/spec.md` alone, so neither record changed.

**Every `file:line` citation of the new prose was read at its target and each one is
correct**, including `scapy/packet.py:1317-1327` against the installed scapy 2.7.0, which
holds `def getlayer(self, cls, nb=1, ...)` and the docstring
`Return the nb^th layer that is an instance of cls, matching flt values.`

### One item left unchanged, for the reviewer to rule on

The new prose writes **`behaviours`**, which matches the surrounding text of
`docs/implementation_notes.md` and the row #594 wrote. The project owner's standing
instruction is US English. A single new paragraph in US spelling inside a document that
spells it the other way would read as an inconsistency, so this round kept the document
consistent and records the question here rather than deciding it. **A sweep of the spelling
belongs to its own issue, and this round opened none.**

### The gates after the fix

```
4798 passed, 8 skipped, 1961 deselected, 8 xfailed, 16 warnings, 114 subtests passed in 101.74s (0:01:41)
1635 passed, 143 skipped, 4857 deselected, 140 xfailed, 1 warning in 12.18s
All checks passed!
232 files already formatted
```

The replay hash after the fix still reads
`c0f751b74793e997b8e184c4fed7b76374af71a0f349458b51571a02b0aa7e8b`.

Issue #637, opened 2026-08-15
Correct the stale reference citation of R10 in the JA4T transcription


Crank-Git:
## The citation names the comment and not the test

`docs/specs/foxio/JA4T.md:162` reads:

> - `wireshark/source/packet-ja4.c:1265` tests `tcp_flags == 0x02` for equality, so a SYN
>   that carries an ECN flag produces no JA4T value in the dissector.

**A read of the pinned commit `27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8` puts the test at line 1266.** Line 1265 holds the comment above it.

```
1265:                 // SYN for this stream - signal JA4T
1266:                 if (tcp_flags == 0x02) {
```

## The port carries the same shape of defect

R29 of `docs/specs/foxio/JA4T.md` in `Crank-Git/ja4plus-go` cites `rust/ja4/src/tcp.rs:153` for the assertion `is_initial_syn(0xC2)`. The same read puts that assertion at line 154, and line 153 holds the comment `// SYN + ECN (ECE + CWR)`.

**This repository edits no file of the port.** The port half needs its own issue there.

## What this issue asks for

Read every `file:line` citation of `docs/specs/foxio/JA4T.md` against the pinned commit, and correct the ones that name a comment or a blank line. **A `file:line` citation is evidence**, so a reader who follows one and lands on a comment cannot tell a stale citation from a moved reference.

#603 found this while it wrote the divergence register row for the JA4T SYN selection. That row states the two verified lines and it names both stale citations. #603 changed neither page, because its scope is one register row.

Verified against: https://github.com/FoxIO-LLC/ja4 (`wireshark/source/packet-ja4.c`, `rust/ja4/src/tcp.rs`, at the pinned commit `27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8`, retrieved 2026-08-15)

Crank-Git:
Plan. Batch #657.

**Approach.** `docs/specs/foxio/JA4T.md:162` cites `wireshark/source/packet-ja4.c:1265` for the equality test `tcp_flags == 0x02`. A read of the pinned commit `27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8` puts that test at line 1266, and line 1265 holds the comment above it.

Correct the citation.

**Files.** `docs/specs/foxio/JA4T.md`, and any case that reads the citation.

**Read the citation at the pin yourself.** Do not take the line number from this issue or from this plan.

[...]

Issue #610, opened 2026-08-14
Read the TCP header that an ICMP error message quotes for JA4T


Crank-Git:
**The Go library's maintainer ruled split T1 on 2026-08-14, and the port carries the other half of that ruling.** The ruling reads:

> **Answer 1.** The library reads a TCP header that an ICMP error message quotes.

`Crank-Git/ja4plus-go#484` holds the ruling, and `Crank-Git/ja4plus-go#494` built the Go half. **This issue asks the port for the same rule.**

## The mechanism

`ja4plus/fingerprinters/ja4t.py:153` holds `if not packet.haslayer(TCP): return None`, and `ja4plus/fingerprinters/ja4t.py:156` holds `tcp = packet[TCP]`. **`scapy` decodes a TCP header that an ICMP error message quotes as `IPerror` and `TCPerror`**, so `haslayer(TCP)` reports false. **No file of `ja4plus/` names ICMP**, measured at the tag `v1.1.0` with `git grep -il icmp -- ja4plus`.

So the port produces no JA4T value for a frame that carries an ICMP error message of a TCP SYN.

## The reference

**The Wireshark dissector produces a value.** `wireshark/source/packet-ja4.c:1261` matches the field abbreviation `tcp.flags` anywhere in the protocol tree, `wireshark/source/packet-ja4.c:1266` sets `syn = 1` for the flag byte `0x02`, and `wireshark/source/packet-ja4.c:1583-1584` writes `hf_ja4t` for that frame. The Rust reference, the Zeek package and the FoxIO Python script each produce none. `Crank-Git/ja4plus-go#484` states each citation, at the FoxIO commit `27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8`.

## The count

**31 per-packet values, all in `ssh2.pcapng`.** Each of the 31 frames carries the protocol stack `eth:ethertype:ip:icmp:ip:tcp`, and the vector holds `64240_2-1-3-1-1-4_1460_8` for every one. The Go run measured 1716 matches and 331 deviations before the change, and 1747 matches and 300 deviations after it. **The tracked report of the Go repository records the JA4T row of `ssh2.pcapng` at 44 matches and 31 deviations before, and 75 matches and 0 deviations after.**

**So a parity difference of 31 values is open until this issue lands.**

## What the Go half built

[...]

PR #662, opened 2026-08-15
Read the TCP header that an ICMP error message quotes for JA4T

Crank-Git:
Refs #610.

## What

`ja4plus/fingerprinters/ja4t.py:153` read `if not packet.haslayer(TCP): return None`.
`scapy` decodes a TCP header that an ICMP error message quotes as `IPerror` and
`TCPerror`, so that test reported false and the reader produced nothing.

New file `ja4plus/utils/icmp_quoted.py` reads the quoted datagram, and
`_generate_from_quoted_header` of `ja4plus/fingerprinters/ja4t.py` calls it where the
packet carries no TCP layer.

## Why

The maintainer ruled on 2026-08-14, under `Crank-Git/ja4plus-go#484`, and
`Crank-Git/ja4plus-go#494` built the Go half in `internal/parser/icmp_quoted.go`.
`wireshark/source/packet-ja4.c:1261` matches the field abbreviation `tcp.flags` anywhere
in the protocol tree, so the dissector writes a JA4T value for such a frame. #484 is the
reversal path, and a reversal changes both repositories.

## The five ICMP types

The reader reads `3` destination unreachable, `4` source quench, `5` redirect, `11` time
exceeded and `12` parameter problem. RFC 792 gives the quoted header to those five types
alone, and an informational message carries an identifier field in place of it. The Go
half reads the same five, in `quotesAPacket` of `internal/parser/icmp_quoted.go`.

```
Verified against: https://www.rfc-editor.org/rfc/rfc792.txt (RFC 792, retrieved 2026-08-15)
```

## Hostile input — what bounds each field

| Field | The bound |
|---|---|
| IP header length | `read_quoted_header` refuses a value below 20 bytes, and it refuses a value above the end of the datagram. |
| IP total length | A declared length above the carried length keeps the carried length. A declared length below it ends the datagram. |
| TCP data offset | The reader refuses a value below 20 bytes, and it refuses a value above the quoted transport bytes. |
| TCP option list | `read_options` of `ja4plus/utils/tcp_options.py` stops on an option length below 2, and it stops on one that reaches past the option field. |

The reader also returns nothing for an IP version other than 4, for a protocol other than
TCP, and for a later fragment. A reader that cannot read the quoted header returns
nothing, and it raises nothing.

**The reader calls no `scapy` dissector on the quoted datagram.** `TCP(bytes)` raises
`struct.error`, which `generate_ja4t` catches nowhere. A run of 200000 random 20-byte to
60-byte headers raised it 9 times, and
`tests/fuzz/test_icmp_quoted_header.py::TestTheReaderCallsNoScapyDissector` holds one
measured region. `readQuotedTCPFields` of the Go port declines `DecodeFromBytes` for the
matching defect of `gopacket`, which `Crank-Git/ja4plus-go#510` records.

## R9 keys the connection on the quoted header

R9 gives one value to one connection, from its first SYN. The outer address pair of an
ICMP message names a router and the client, so it names no endpoint of the connection the
value describes. `_generate_from_quoted_header` therefore builds the key from the quoted
IPv4 header and the quoted TCP ports.

## The measurement — a replay moves no value

A replay of the 38 captures of `tests/foxio_vectors/` reads **the same 64 JA4T values
before and after**, each on the same frame number. The result is byte-identical.

`ssh2.pcapng` carries 32 ICMP messages. 31 of them quote a SYN, and frame 197 quotes an
ACK. Those 31 messages report **10 connections**, and a captured SYN of every one of the
10 already reached the reader, before its first ICMP message. R9 therefore suppresses all
31, and the capture keeps its 19 values. Frame 197 alone carries an 11th connection, and
its quoted header carries no SYN.

`tests/foxio_deviations.json` holds 140 keys, it holds no JA4T key of `ssh2.pcapng`, and
this change moves no entry of it. The conformance suite reads 140 xfailed cases before
and after, which equals the key count.

**Warning: the count of 31 that #610 states is a per-frame count, and this project writes
one value for each connection.** `tests/foxio_vectors/wireshark_expected/ssh2.pcapng.json`
holds 75 `ja4.ja4t` values over 136 frames, and `ja4plus` writes 19 for that capture. No
case of `tests/` compares a per-frame JA4T value, so the difference costs no unmatched
case here. The Go port holds no R9, which is why its per-frame report moved from 44 to 75.
The capability this change adds is a value where the monitor reads the ICMP message and
never the SYN itself.

## How it was tested

- `tests/test_ja4t_icmp_quoted.py` — 30 cases. A restore of the `haslayer(TCP)` guard
  fails 11 of them.
- `tests/fuzz/test_icmp_quoted_header.py` — 102 cases. Crafted length fields, every
  truncation of one whole quote, 512 random payloads, a byte-flip sweep, and the region
  that `TCP(bytes)` refuses. A restore of the guard fails 50 of them.
- The five gates, all green.

```
pytest tests/ -m "not spec_validation"  4950 passed, 8 skipped, 1961 deselected, 8 xfailed
pytest tests/ -m spec_validation        1635 passed, 143 skipped, 5009 deselected, 140 xfailed
ruff format --check ja4plus/ tests/     236 files already formatted
mypy --strict ja4plus/                  Success: no issues found in 32 source files
```

## Records

`CHANGELOG.md` and the `## Changelog` table of `docs/specs/spec.md` each carry one entry
with the literal `TBD`. The Changelog edit is a pure append, and the three `## Terms`
rows append to the end of that table.

## The self-review

Two review passes read the diff. The correctness pass fuzzed `read_quoted_header` with
300500 random byte strings of every length from 0 to 200, and it fuzzed
`quoted_tcp_header` with 3176 crafted frames: an IPv6 outer header, `ICMPv6DestUnreach`,
a nested ICMP message, a 9000-byte payload, an empty payload, and every cut point of one
whole quote under each of the five error types. **No input raised.**

The conformance pass found three defects, and commit `6d738b9` repairs all three.

1. The two records read 11 connections where a recount reads 10. The 32nd message quotes
   an ACK, and that message alone carries the 11th.
2. The prose named the ruling `split T1`, which no record of this project defines.
3. Four cases asserted a type where the input decides an exact value.

`docs/specs/spec.md` gains three rows of the `## Terms` table for the words this change
introduces: `ICMP error message`, `quoted datagram` and `quoted header`. The Changelog
table edit stays a pure append.

## The merge of the integration branch

#646 and #637 landed while this branch was open, so the provider read this pull request
as `CONFLICTING` and it built no merge commit. `.claude/rules/batch-gate.md` records that
state. Commit `2a7686d` takes `batch/657-tunnel-icmp-citations` again.

Both sides append one record to `CHANGELOG.md` and one row to the `## Changelog` table of
`docs/specs/spec.md`, and the merge keeps every record of both sides. **The five gates
above are the run of the merged tree**, and the replay of the 38 captures reads the same
64 JA4T values on it.


Crank-Git:
Self-review, two passes, both read `git diff origin/batch/657-tunnel-icmp-citations...HEAD`.

**The correctness pass found no defect.** It fuzzed `read_quoted_header` with 300500 random byte strings of every length from 0 to 200, and `quoted_tcp_header` with 3176 crafted frames. The frames carried an IPv6 outer header, `ICMPv6DestUnreach`, a nested ICMP message, a 9000-byte payload, an empty payload, and every cut point of one whole quote under each of the five error types. **No input raised.** It also compared `tcp_prefix_from_header` against `tcp_prefix` over 50000 random headers and read 0 differences, which is the reading that keeps JA4T and JA4TS on one rule.

**The conformance pass found three defects, and `6d738b9` repairs all three.**

1. **The two records read 11 connections where a recount reads 10.** `ssh2.pcapng` carries 32 ICMP messages, 31 of them quote a SYN, and those 31 report 10 connections. Frame 197 quotes an ACK, and that message alone carries the 11th. The three places that held the wrong count now hold 10.
2. **The prose named the ruling `split T1`, and no record of this project defines that name.** Rule 10 of `.claude/rules/ste.md` refuses an abbreviation the project has not defined. Each of the five places now names the date and the issue.
3. **Four cases asserted a type where the input decides an exact value.** `assert result is None or isinstance(result, str)` cannot fail on any single-line change of the reader. Each of the four now states the value the crafted field produces, measured against the reader. A restore of the `haslayer(TCP)` guard now fails 50 of the 102 cases, where it failed 48 before.

**One reading the conformance pass raised, and this branch takes it.** The prose rotated `quoted bytes` against `quoted header`, which rule 7 refuses. `docs/specs/spec.md` gains three rows of the `## Terms` table: `ICMP error message`, `quoted datagram` and `quoting header`. Each row names the words this project declines.

**One reading it raised that this branch declines, and it records the decline here.** It reported the four crafted-field cases as a legitimate crash oracle. They are, and a crash oracle beside an exact value costs nothing, so the branch takes the exact value as well.

Crank-Git:
Two corrections to the self-review comment above.

1. The third `## Terms` row is `quoted header`, and the comment above wrote `quoting header`. `docs/specs/spec.md` holds `quoted header`.
2. The restored guard fails 50 of the 102 cases, and it failed 50 before the four cases took an exact value as well. The comment above wrote 48, and no run measured that number. The strengthening therefore changed no reversal count, and it changes what each of the four cases proves on a run that passes.

Issue #664, opened 2026-08-15
Batch: the sorted raw form, an uncompared snapshot, and two falsified records


Crank-Git:
This batch holds four members, and they write four disjoint file sets.

- [x] #600 — fill the sorted raw form so JA4H_r reaches a value. `ja4plus/fingerprinters/ja4h.py`.
- [x] #638 — the FoxIO Rust JA4L snapshot value is committed and no case compares it. `tests/test_foxio_rust_parity.py`.
- [x] #611 — two falsified sentences: a register cause line number, and a `wireshark_expected` file count.
- [x] #618 — one packet dated far in the future evicts every entry of a state table. The maintainer ruled the property is accepted, so this member states it in the documentation and changes no line of the age pass.

Continuous integration runs once, on the batch pull request into `dev`.

Crank-Git:
finding: batch cross-check, batch #664

**What was compared.** The plans of #600, #638, #611 and #618, on file, function, interface, migration, configuration key and route.

**The pairs found.** None. The four write four disjoint file sets.

| Member | Files |
|---|---|
| #600 | `ja4plus/fingerprinters/ja4h.py` |
| #638 | `tests/test_foxio_rust_parity.py`, and the register |
| #611 | `tests/foxio_deviations.json` cause text, `docs/specs/foxio/JA4L.md` |
| #618 | `docs/concurrency.md` or `docs/specs/features/03-concurrency-safety.md` |

**One pair needs a statement, because it looks like a collision and is not.**

1. **#638 and #611 both touch `tests/foxio_deviations.json`.** #611 edits the **cause text of an existing entry** and changes no key. #638 **adds an entry** for the JA4L snapshot comparison. The two do not touch one entry in common. **The register invariant binds both**: the file holds 138 keys today, #611 must leave it at 138, and #638 raises it by the entries it adds. Each states its own count. **They are not sequenced**, because the resolution is an append against a text edit and the project manager resolves it once at sub-merge.

**The clean negatives.**

- **No two members write one source file under `ja4plus/`.** #600 alone writes there.
- **No member adds a public name, a configuration key, a route or a migration.**
- **#600 alone can move a published field**, and no other member reads the JA4H raw forms.
- **#618 changes no executable line at all**, and #611 changes none either.
- **No member reads a shape another member changes.**

[...]

Issue #618, opened 2026-08-14
One packet dated far in the future evicts every entry of a state table


Crank-Git:
**One packet dated far in the future removes every entry of a bounded state table.** The self-review of `Crank-Git/ja4plus-go#565` found it on 2026-08-14, and **this port holds the identical comparison.**

**Neither project introduced it in a recent change.** It is the shape the eviction has always had, in both.

## The mechanism

**An age pass reads the capture timestamp of the packet that arrives, and it compares that ONE timestamp against the last-seen time of EVERY key of the table.** So one timestamp decides the age of every other connection.

**In this port, at tag `v1.1.0`, `ja4plus/utils/state_table.py`:**

```python
if now - entry[_LAST_SEEN] > self.max_connection_age:
    self.evict_key(key)
```

**`now` comes from the packet.** The Go library holds the same comparison in four places: its new `state_bound.go` `agedKeys`, `ja4ts.go` `evictAgedConnections` and `ja4ssh.go` `evictAgedConnections`.

## What an attacker gets

**Every packet is untrusted input, and the packet carries the timestamp.** A sender that dates one packet in the year 9999 ages every key of the table at once, and the pass then removes all of them.

**The cost is one packet for the WHOLE table, rather than one packet for one entry.** A bounded table already lets a sender evict one entry with one connection, and that cost is expected. **This is a different order of thing.**

**What it destroys is fingerprinting state**, so the effect is that a run answers with fewer fingerprints, or with none, rather than answering wrongly.

## Why the Go repository filed rather than repaired

**A repair invents a rule that no FoxIO source states.** Neither the images nor any reference implementation says what an implementation must do with a packet timestamp that runs backwards or far forwards.

**And a repair applied to one repository alone gives the two implementations two age clocks**, which is exactly the divergence that `.claude/rules/parity.md` exists to prevent. **So it is a maintainer ruling and a two-repository change**, and `Crank-Git/ja4plus-go#577` holds the Go half with four candidate answers.

## The candidates, so both repositories rule once

1. **Accept it.** State the property in the documentation and change no line. A capture is usually trusted input, and a live capture takes its timestamp from the host clock rather than from the sender.
2. **Use a monotonic clock for the age pass**, and keep the packet timestamp for the fingerprint. **The two clocks then differ on a replayed capture**, and an offline run of an old capture would evict nothing or everything.
[...]

PR #674, opened 2026-08-15
docs(concurrency): state that one future-dated packet evicts every entry

Crank-Git:
Refs #618. Batch #664, integration branch `batch/664-raw-form-and-records`.

## What

`docs/concurrency.md` gains the section `## One packet dated in the future evicts every entry`. **No line of `ja4plus/` changes.**

## Why

`ja4plus/utils/state_table.py:414` reads `if now - entry[_LAST_SEEN] > self.max_connection_age:` inside the loop at line 411, and `now` is the timestamp of the packet that arrives. One packet dated far in the future therefore ages every entry at once, and the age pass evicts all of them. A packet carries its own timestamp, so a sender influences that input, and a capture file the operator did not produce is the reachable case.

**The maintainer ruled answer 1 on 2026-08-14**, on `Crank-Git/ja4plus-go#577` and on #618. Both projects accept the property and document it, and no line of the age pass changes. The section states the property, names the reachable case, gives the reason this project accepts it, and states what the property costs: the eviction removes fingerprinting state, so a run answers with fewer fingerprints rather than with wrong ones, and the entry count bound holds the memory whatever timestamp a packet carries.

## How tested

New file `tests/test_future_dated_packet_eviction.py` holds four cases, and they read the behaviour rather than the sentence alone.

**One reversal measures them.** A bound of `now` to the newest last-seen time of the entries, written into `evict_aged` before the loop, fails `test_one_packet_dated_far_in_the_future_evicts_every_entry` and `test_one_packet_dated_in_the_future_evicts_an_entry_the_next_packet_needs`. It leaves `test_a_packet_inside_the_maximum_age_evicts_no_entry` and the documentation case green. Following #51, `inspect.getsource` of the loaded method in a fresh interpreter reported the mutation present before the run and absent after the restore.

`git diff --stat ja4plus/` reports nothing.

## Records

`CHANGELOG.md` and the `## Changelog` table of `docs/specs/spec.md` each carry the literal `TBD` in place of the round number.


Crank-Git:
Self-review. Three lenses: correctness of the documented property, the discrimination of the cases, and the scope boundary.

**Two findings, both repaired on this branch.**

1. **The documentation case read a phrase that spans a line break.** `"A packet dated far in the future\ntherefore ages every entry at once"` fails on any reflow of the paragraph, and a reflow says nothing about the property. Each phrase the case reads now sits on one line, and a comment states the reason. Commit `5389f22`.

2. **"Every entry" needed the shard qualifier.** `ja4plus/utils/state_table.py:412` reads `if entry[_SHARD] is not None and entry[_SHARD] != shard: continue`, so the age pass of one thread holds the entries of another thread. A caller that runs one thread loses every entry, because that thread owns them all. A sharded caller loses the entries of the thread that reads the packet, and it loses every entry that the wall clock dated. The section now states that, and it points at `docs/specs/features/03-concurrency-safety.md`. Commit `4a1bc1d`.

**Scope held.** `git diff --stat origin/batch/664-raw-form-and-records -- ja4plus/` reports nothing, so no line of the age pass changes, which is what the ruling of 2026-08-14 requires. The four cases and the two records are the whole change beside the page.

**No second unruled property surfaced.** The shard filter above is documented behaviour of #461 and it is no new finding.

Issue #671, opened 2026-08-15
Compare the two FoxIO Rust JA4SSH values that no case reads


Crank-Git:
A sweep of 2026-08-15, run under #638, read every value-bearing field of the 11 snapshots under `tests/foxio_vectors/rust_expected/`. It found a third committed value class that no case compares. #670 holds the second one.

## The measurement

`tests/foxio_vectors/rust_expected/ja4__insta@ssh2.pcapng.snap:215` opens a `ja4ssh:` block on stream 14, and that block holds two values.

```
  ja4ssh:
  - c36s36_c76s124_c74s5
  - c36s52_c42s76_c51s2
```

`read_rust_snapshot` of `tests/test_foxio_rust_parity.py` enters the `tls_certs` block alone, so it reads neither value. `SNAPSHOT_METHODS` names no JA4SSH method either.

## Why it matters

**`ssh2.pcapng` stream 14 carries a decided register entry.** `tests/foxio_deviations.json` holds `ssh2.pcapng/14:57377/JA4SSH.2` against #97, and `.claude/rules/external-apis.md` records that the FoxIO Rust value and the Zeek value agree on it. **That agreement rests on a reading, and no case measures it.**

## What this issue asks

Read whether the two values reach a comparison, and record the answer either way.

**Warning: the FoxIO Python expected-output file holds JA4SSH values for this capture**, so read the precedence rule of `.claude/rules/external-apis.md` before you compare against the snapshot. The precedence exception of that file may reach the second occurrence, because #97 declines the Python value there.

Issue #676, opened 2026-08-15
Batch: the citation guard, two uncompared snapshot classes, and a stale comment


Crank-Git:
This batch holds four members.

- [ ] #668 (moved out: ruled 2026-08-15, rescheduled) — generalize the citation guard over `docs/specs/foxio/`. 37 citations of `ja4h.py` were already wrong before this session. `priority:high`.
- [x] #665 — two comments name a Wireshark version #616 left behind. `ja4plus/fingerprinters/ja4d.py:58` and `ja4plus/fingerprinters/ja4d6.py:279`.
- [x] #670 — the Rust snapshot holds 6 uncompared JA4H values.
- [x] #671 — the Rust snapshot holds 2 uncompared JA4SSH values. **Sequenced after #670**, because both write `tests/test_foxio_rust_parity.py`.

Continuous integration runs once, on the batch pull request into `dev`.

Crank-Git:
finding: batch cross-check, batch #676

**What was compared.** The plans of #668, #665, #670 and #671, on file, function, interface, migration, configuration key and route.

**The pairs found.**

1. **#670 and #671 both write `tests/test_foxio_rust_parity.py`**, and each adds one method class to the snapshot comparison. **Sequenced.** #671 launches after #670 sub-merges. #638 measured both in one sweep, so the two are one reading split across two issues.
2. **#668 and #665 both concern a stale citation, and they do not meet.** #668 reads `docs/specs/foxio/` and holds citations of files this repository owns. #665 repairs two comments under `ja4plus/`, which name an external version rather than a line. **No action**, and the pair is stated so that a reader does not take them for one issue.

**The clean negatives.**

| Member | Files it writes |
|---|---|
| #668 | `tests/`, and the pages of `docs/specs/foxio/` it corrects |
| #665 | `ja4plus/fingerprinters/ja4d.py`, `ja4plus/fingerprinters/ja4d6.py` |
| #670, #671 | `tests/test_foxio_rust_parity.py` |

- **#665 is the one member that writes under `ja4plus/`**, and it changes comment lines alone.
- **No member moves a fingerprint.** #670 and #671 add comparisons and produce no value; #668 reads documentation; #665 changes no executable line.
- **No member adds a public name, a configuration key, a route or a migration.**
- **The register invariant binds #670 and #671**, because a new comparison may report a deviation. Each states its own count.

**One cross-batch reading, and it is the reason #658 is not in this batch.**

[...]

Issue #668, opened 2026-08-15
Replace the stale ja4h.py line citations of the JA4H reading page


Crank-Git:
`docs/specs/foxio/JA4H.md` cites this project's `ja4plus/fingerprinters/ja4h.py` by line number in 37 distinct places, and the citations name lines the file no longer holds. **The staleness predates #600**, and #600 found it rather than caused it.

## The measurement

A read of the base commit of #600, `ee37347`, measures three rows of the `### Part b, part c and part d agree` table.

| Row | Citation | The line the base commit holds |
|---|---|---|
| `b, join and hash` | `ja4h.py:354` | `headers = parsed.get("headers", {})` |
| `c, the sort` | `ja4h.py:364` | `cookie_values.append(v)` |
| `d, the sort key` | `ja4h.py:373` | `"language": headers.get("accept-language", ""),` |

**Each verdict of the table is still correct**, and each citation names the wrong line. `sorted(name for name, _ in cookie_pairs)` exists today, in `_ja4h_sorted_cookie_strings`.

**`D6 — `ja4h.py:62` states that FoxIO publishes no `JA4H_r` key` carries the same defect.** The docstring it quotes sat at line 70 of the base commit, and line 62 held `class JA4HFingerprinter(BaseFingerprinter):`.

## Why a line number goes stale here

Every change to `ja4h.py` above a cited line moves that line, and no case reads the citation. #600 added two functions to that file, so 37 citations moved again.

## What closes this

**Cite a stable target rather than a line.** `.claude/rules/ste.md` of `Crank-Git/ja4plus-go` states the same convention under `## How a citation names its target`: name the identifier, the rule number or the section heading, because those survive an edit above them. `_ja4h_sorted_cookie_strings` finds the sort after every later edit, and `ja4h.py:364` finds it until the next one.

**A citation of the FoxIO material keeps its `file:line`.** `.claude/rules/rulings.md` binds a reading of a FoxIO source to a file and a line, and `testdata/foxio.pin` holds those lines still. This issue reaches no `python/ja4h.py:<line>` citation and no `wireshark/source/packet-ja4.c:<line>` citation.

**A case that reads the citations would stop the next regression.** Without one, the next edit to `ja4h.py` moves them again.

Found by #600.

Crank-Git:
Raised in priority, and the project manager records the class this belongs to.

**A stale line citation is now the most common defect this session has found, and three issues hold instances of it.**

| Issue | Page | Finding |
[...]

PR #682, opened 2026-08-15
Compare the two FoxIO Rust JA4SSH values that no case reads

Crank-Git:
The batch pull request of #676 closes #671. This pull request builds #671.

`read_rust_snapshot` of `tests/test_foxio_rust_parity.py` entered the `tls_certs` block
and the `http` block, so the `ja4ssh` block reached no case. This change enters that block
too, and it compares both values the block holds.

One of the eleven local snapshots holds such a block.
`tests/foxio_vectors/rust_expected/ja4__insta@ssh2.pcapng.snap:215-217` writes 2 values on
stream 14.

**The block names no field.** It writes each value as a bare list item at the two-space
level, so the reader takes a value from the block it stands in. `- x509` and `- ja4h` each
carry a field name, and this one carries none.

**JA4SSH emits one value for each window.** One entry of the block is one window, and the
occurrence is the position of that window. The register keys a value by that same
position.

## The measurement

| Source | First window | Second window |
|---|---|---|
| `tests/foxio_vectors/ssh2.pcapng.json` | `c36s36_c76s124_c74s5` | `c36s36_c0s0_c2s0` |
| `ja4__insta@ssh2.pcapng.snap:215-217` | `c36s36_c76s124_c74s5` | `c36s52_c42s76_c51s2` |
| `ja4plus` | `c36s36_c76s124_c74s5` | `c36s52_c42s76_c51s2` |

**The FoxIO Python file publishes a JA4SSH value on the stream, and it publishes one for
each window.** The comparison therefore reaches both windows, and each one reaches it for
its own reason.

1. **The first window needs no ruling.** Both FoxIO references hold
   `c36s36_c76s124_c74s5`, and `tests/test_spec_validation.py` already compares it against
   the FoxIO Python file. This module is a second reference on it.
2. **The second window rests on the precedence exception.** #97 declines the FoxIO Python
   value `c36s36_c0s0_c2s0` under `"capability": false`, which is a value decline.
   `.claude/rules/external-apis.md` therefore lets another FoxIO implementation hold the
   reference value, and the FoxIO Rust snapshot and the Zeek baseline both hold
   `c36s52_c42s76_c51s2`. `tests/test_precedence_exception.py` records that row as one of
   the 6 the exception reaches.

**`ja4plus` matches the snapshot on both windows, so this pull request reports no
deviation and it moves no register entry.** #214 made this project emit the window a
connection holds open when the capture ends, and `docs/specs/foxio/JA4SSH.md` rests that
ruling on the agreement of the Rust snapshot and the Zeek baseline. **No case measured that
agreement**, and the new class is that measurement.

**This module keys no JA4SSH case against the register.** A mark on the second window would
xfail a case that passes, and `strict=True` fails the suite on such a pass.
`test_this_module_keys_no_ja4ssh_case_against_the_register` holds that rule.

## How this was tested

Every gate ran in a virtual environment of the worktree.

```
pytest tests/ -m "not spec_validation"  5519 passed, 8 skipped, 8 xfailed, 114 subtests passed
pytest tests/ -m spec_validation        1671 passed, 142 skipped, 138 xfailed
ruff format --check ja4plus/ tests/     243 files already formatted
mypy --strict ja4plus/                  Success: no issues found in 32 source files
```

**The register invariant holds.** `tests/foxio_deviations.json` holds 138 keys and the
conformance suite reports 138 xfailed. The base holds the same 138.

**The base reports 1660 passed on the conformance suite, so the 11 new cases are the whole
rise.** The unit suite reports the same counts before and after, because every new case
carries the `spec_validation` marker.

**No file under `ja4plus/` changed**, so no fingerprint moved. `git diff --stat` against
the base reads `CHANGELOG.md`, `docs/specs/spec.md` and `tests/test_foxio_rust_parity.py`.

**The comparison discriminates, measured both ways.** Write `c36s52_c42s76_c51s3` in place
of the second snapshot value and 2 cases fail:

```
Failed: ssh2.pcapng 172.16.225.48:57377 <-> 54.160.114.75:22 JA4SSH.2: rust=c36s52_c42s76_c51s3 ja4plus=c36s52_c42s76_c51s2
2 failed, 228 passed, 1 xfailed
```

Write `c36s36_c76s124_c74s6` in place of the first one and 2 cases fail the same way:

```
Failed: ssh2.pcapng 172.16.225.48:57377 <-> 54.160.114.75:22 JA4SSH.1: rust=c36s36_c76s124_c74s6 ja4plus=c36s36_c76s124_c74s5
2 failed, 228 passed, 1 xfailed
```

Restore the snapshot and the module reports `230 passed, 1 xfailed`.

## One finding this pull request does not repair

**`docs/specs/foxio/JA4SSH.md` holds a section named `Why the parity harness never compared
this value`, and this change makes that section false.** It also cites
`tests/test_foxio_rust_parity.py:72` for a two-entry `SNAPSHOT_METHODS`, which the file has
not held since #216. The page is a documentation repair rather than a test change, and #668
of this batch is the member that writes under `docs/specs/foxio/`. A follow-up issue records
it, and this pull request leaves the page alone.

Verified against: https://github.com/FoxIO-LLC/ja4/tree/main/rust/ja4/src/snapshots (FoxIO pin, commit dated 2026-07-21, read through `tests/foxio_vectors/rust_expected/`)


Crank-Git:
Self-review, correctness lens: no defect.

**The new branch reads two lines and no other line of the eleven snapshots.** `grep -nE '^  -' -r tests/foxio_vectors/rust_expected/` returns three shapes. `- x509:` and `- ja4h: <value>` each carry a field name, and `- c36s36_c76s124_c74s5` and `- c36s52_c42s76_c51s2` carry none. The two bare items stand in the `ja4ssh` block alone.

**Each of the three flags clears at three sites.** A `- stream:` line clears all three, the opener branch sets one and clears the other two, and a field line of the stream clears all three. `ssh_extras:` matches no opener, so it closes the `ja4ssh` block. The nested fields of `ssh_extras` sit at four spaces, so the indent filter drops them before the flag logic reads them.

**`held.ssh.extend` mirrors the `certs` and `http` merge**, so a stream identity that recurs across blocks accumulates its JA4SSH values the same way.

**The rewritten guard widens the old condition and narrows nothing.** A snapshot that passed `assert any(...)` before still passes `assert held`.

Crank-Git:
One finding outside this change set, filed as #684.

`tests/download_test_vectors.py:141` holds `RUST_COMPARED_FIELDS = (b"ja4: ", b"ja4s: ", b"ja4t: ", b"ja4x: ")`, and `:343` keeps a Rust snapshot only where one line starts with one of those four. **`ja4h` and `ja4ssh` reach no entry.** #670 added the first comparison and this pull request adds the second, and neither one moved that file.

**It costs nothing today.** Every local snapshot that holds a `ja4h` block or a `ja4ssh` block also holds a `ja4t` value, so the condition keeps all eleven. #684 records the repair and the case that would have caught it.

Crank-Git:
Self-review, conformance lens: one reading to state and one case to name.

**Every new case can fail, and three of them are measured.**

| Case | The mutation that reddens it |
|---|---|
| `test_the_produced_ja4ssh_equals_the_rust_snapshot_value` | Either snapshot value, measured in the body above |
| `test_the_two_foxio_references_part_on_the_second_window` | The same mutation, which moves the reference the class names |
| `test_the_register_declines_no_capability_on_either_value` | `"capability": true` on `ssh2.pcapng/14:57377/JA4SSH.2`, which fails 2 cases |
| `test_the_register_declines_the_second_python_value_as_a_value` | The same register move |
| `test_the_suite_collects_one_case_for_every_value_it_compares` | A reader that drops the `ja4ssh` branch, which empties the list |
| `test_the_local_snapshots_hold_the_two_values_the_reading_counts` | A snapshot that leaves the repository |

The register mutation reports:

```
AssertionError: ssh2.pcapng/14:57377/JA4SSH.2
2 failed, 228 passed, 1 xfailed
```

**One case passes by construction today, and it is a guard rather than a measurement.**
`test_this_module_keys_no_ja4ssh_case_against_the_register` reads `register_keys()`, which returns JA4T keys and JA4X keys alone. It reddens the day a writer keys a JA4SSH case here, which is the state it exists to refuse. `TestTheJa4xValuesTheRustSnapshotHolds` holds the same shape in `test_no_register_key_of_this_module_belongs_to_the_spec_validation_suite`.

**The precedence reading rests on the rule and not on which value looks right.** `.claude/rules/external-apis.md` states five facts, and the row holds all five. The key is the value form, the FoxIO Python file holds `c36s36_c0s0_c2s0` at that connection and that occurrence, the entry reads `"decided": true` under #97, it reads `"capability": false`, and the FoxIO Rust snapshot and the Zeek baseline hold one value between them.

Crank-Git:
Self-review, writing-standard lens: no finding.

**Every new sentence of `tests/test_foxio_rust_parity.py` is 25 words or fewer.** A read of the added lines, split on sentence boundaries, returns no sentence past the limit.

**The two records hold no `-ing` form at all.** A read of the added lines of `CHANGELOG.md` and `docs/specs/spec.md` matches none, so rule 14 reaches nothing here. Rules 1 and 3 exempt those two records, and every other rule reaches them.

**Each domain word stands in the `## Terms` table of `docs/specs/spec.md`.** `window` reads "The count of packets JA4SSH reads before it emits a fingerprint", and this change rotates none of its three rejected words. `vector`, `reading`, `ruling`, `baseline`, `raw form` and `method` each carry their table meaning.

**Every quoted value, key and path is verbatim.** `c36s36_c0s0_c2s0`, `c36s52_c42s76_c51s2`, `ssh2.pcapng/14:57377/JA4SSH.2` and `"capability": false` are copies.

Crank-Git:
Self-review, second lens: five findings taken and one declined. `caf0b12` holds the repair.

**Taken.**

1. **The module comment of `SSH_BLOCK_OPENER` ran 31 words in one sentence.** It now states the shape in one sentence and the consequence in a second.
2. **The docstring of `read_rust_snapshot` ran 28 words**, because the third block pushed a sentence that was inside the limit past it. It now names the three blocks in a second sentence.
3. **The inline comment of the new branch restated the module comment.** It now names one fact and cites the comment that holds the shape.
4. **The docstring of `_ssh_value_params` held a 37-word sentence and a 27-word sentence.** Each reading now stands in its own paragraph, one fact to a sentence.
5. **`CHANGELOG.md` carried the four-word noun cluster `FoxIO Rust parity module`.** The entry now opens with `tests/test_foxio_rust_parity.py`, which is the form the Changelog row of `docs/specs/spec.md` already used.

**One finding of the conformance lens is taken as well.** `test_this_module_keys_no_ja4ssh_case_against_the_register` read `register_keys()` alone, and that function holds no JA4SSH branch, so no edit of this change set could redden it. The case now reads the parameter list too:

```python
marked = [param.id for param in _ssh_value_params() if param.marks]
assert not marked, "\n".join(marked)
```

A parameter that carries the key of the second window reports `marks` as `True`, measured against `_deviation_marks(value_key("ssh2.pcapng", "14", "57377", "JA4SSH", 2))`, and the two live parameters report `False`.

**Declined: the phrase `as loud as a mismatch`.** Rule 9 of `.claude/rules/ste.md` bars a metaphor, and four sibling classes of this file already carry that exact sentence. A seventh wording for one concept breaks rule 7, which bars a synonym rotation. **The phrase belongs to the whole file or to none of it**, so a repair of it is one change across five classes and not a line of this one.

**The five gates stay green after the repair.** The unit suite reports 5519 passed, 8 skipped, 8 xfailed and 114 subtests passed. The conformance suite reports 1671 passed, 142 skipped and 138 xfailed. `ruff check`, `ruff format --check` and `mypy --strict ja4plus/` report no issue.

Issue #686, opened 2026-08-15
Batch: the mapping lookup, the citation guard, and the emission divergence row


Crank-Git:
This batch holds three members, each built on a ruling the maintainer made on 2026-08-15.

- [x] #639 — the mapping lookup reads both forms of an empty list. **It closes the detection loss that #619 and #612 opened**, and it carries the constraint that parts c and d of JA4H keep the sentinel as a ruled value.
- [x] #668 — the citation guard, under answer A. The pins stand and the guard reads each citation at the commit its section declares.
- [x] #667 — one divergence register row: `ja4plus` emits one JA4T value for one connection and `ja4plus-go` emits one for each frame.

**#658 is sequenced after #668**, because #668 repairs eight citations of `docs/specs/foxio/JA4T.md` and #658 corrects the JA4TS comparison table of that same page.

**#663 runs alone and last.** A repository-wide spelling sweep conflicts with every open branch, so it is scheduled when no other member is live.

Continuous integration runs once, on the batch pull request into `dev`.

Crank-Git:
finding: batch cross-check, batch #686

**What was compared.** The plans of #639, #668 and #667, on file, function, interface, migration, configuration key and route.

**The pairs found.**

1. **#668 and #667 both write `docs/specs/spec.md`, and they write two different tables.** #667 appends one row to the divergence register, and #668 appends one row to the Changelog table as every member does. **No action.** The two edits are appends into different tables of one file, and the project manager resolves the append once at sub-merge.

**The clean negatives.**

| Member | Files it writes |
|---|---|
| #639 | `ja4plus/ja4db.py`, and tests |
| #668 | `docs/specs/foxio/*.md`, and a new guard under `tests/` |
| #667 | `docs/specs/spec.md`, the divergence register |

- **#639 is the one member that writes under `ja4plus/`**, and it changes the lookup and no fingerprinter.
- **No member moves a fingerprint.** #639 changes which rows a lookup matches and no value this library produces.
- **No two members own the same new name.**
- **#668 and #639 both concern the empty-list ruling and they do not meet.** #639 reads the values that ruling produces, and #668 reads citations of the pages that record it.

**Two sequencing rules stand outside this batch, and the tracking issue states both.** #658 follows #668 on one page, and #663 runs alone because a repository-wide spelling sweep conflicts with every open branch.

Launching all three now.

Crank-Git:
finding: #639 writes two documentation pages the cross-check did not list for it

[...]

Issue #639, opened 2026-08-15
Three JA4X mapping rows hold a sentinel no value matches


Crank-Git:
**Three JA4X rows of the bundled FoxIO mapping file carry the zero sentinel
`000000000000`, and `ja4plus` produces no value that holds it after #619.** A lookup of a
certificate that carries an empty object identifier list therefore matches no row, and the
three rows below reach no `ja4plus` value at all.

Measured in the worktree of #619, on branch `issue/619-ja4x-empty-list`:

```
$ grep -n "000000000000" ja4plus/data/ja4plus-mapping.csv | head -3
26:Sliver/Havoc C2 Server,,,,,,,000000000000_4f24da86fad6_bf0f0589fc03,,,
27:Sliver/Havoc C2 Server,,,,,,,000000000000_7c32fa18c13e_bf0f0589fc03,,,
35:Qakbot C2,,,,,,,2bab15409345_af684594efb4_000000000000,,,
```

**The other seven rows the same command reports hold JA4H values, and #619 rules nothing on
JA4H.** `Crank-Git/ja4plus#612` carries that half, and the same reading reaches those rows
on the day it lands.

## Why the rows carry that form

**FoxIO builds the mapping file from its own implementations, and the Rust one writes the
sentinel.** `rust/ja4x/src/lib.rs` returns `000000000000` for an empty input, under the doc
comment `Returns "000000000000" (12 zeros) if the input string is empty.` R8 of
`docs/specs/foxio/JA4X.md` holds the whole reading, and the maintainer ruled on 2026-08-14
that `ja4plus` hashes instead.

## What this issue does not rule

**The mapping file is third-party data, and this project rewrites no row of it.**
`.claude/rules/external-apis.md` names it a FoxIO interface at a pinned commit. The
question is what `ja4plus` does at lookup time, and three readings stand open.

1. Leave the rows as they stand, and record the reach in the divergence register.
2. Read a JA4X value against the sentinel form as well, inside `ja4plus/ja4db.py` alone.
3. Ask FoxIO to move the rows.

**No measurement here names the cost of any reading**, and each one is a question for the
maintainer.

## What raised it

The self-review of #619 raised the finding. #619 changes no lookup path and no mapping row.

Crank-Git:
Raised in priority, and the project manager records why.

**This is a detection consequence of the empty-list ruling, and not only a data defect.**

[...]

PR #688, opened 2026-08-15
Read both forms of an empty-list value at lookup time

Crank-Git:
Refs #639. Batch member of #686. Base is `batch/686-lookup-and-guard`.

## What this changes

**FoxIO builds `ja4plus-mapping.csv` from its own implementations, and the Rust one writes
`000000000000` for an empty list.** This project hashes an empty list instead, under the
ruling of 2026-08-14, so it writes `e3b0c44298fc` where a row of that file holds the zero
sentinel. Three rows reached no value this library produces.

| Row | Value it holds | Application |
|---|---|---|
| `ja4plus/data/ja4plus-mapping.csv:26` | `000000000000_4f24da86fad6_bf0f0589fc03` | Sliver/Havoc C2 Server |
| `ja4plus/data/ja4plus-mapping.csv:27` | `000000000000_7c32fa18c13e_bf0f0589fc03` | Sliver/Havoc C2 Server |
| `ja4plus/data/ja4plus-mapping.csv:35` | `2bab15409345_af684594efb4_000000000000` | Qakbot C2 |

**The mapping file is FoxIO published data, so this change corrects no row of it.** The
maintainer ruled on 2026-08-15 that the lookup reads the two forms as one value.

## The constraint decides the shape

**`000000000000` does not name one thing across JA4+.**

| Part | What the sentinel names there | Does the lookup read both forms |
|---|---|---|
| JA4X part a, part b and part c | an empty object identifier list | Yes |
| JA4H part b | an empty header list | Yes |
| JA4H part c and part d | `no cookie` | No |

**R18 is a rank 1 rule and it names no sentinel for the hashed part. R27 confines the
sentinel to part c and part d.** The two forms were never synonyms there, so a blanket
fallback would map a real value onto an unrelated one.

## How the code refuses part c and part d, by construction

New table `_EMPTY_LIST_PARTS` of `ja4plus/ja4db.py` holds the part count and the part
indices of `ja4x` and `ja4h`, and it holds no other method.

```python
_EMPTY_LIST_PARTS: dict[str, tuple[int, tuple[int, ...]]] = {
    "ja4x": (3, (0, 1, 2)),
    "ja4h": (4, (1,)),
}
```

**The key is the column name the mapping file gives each row**, so no reading of a
fingerprint string decides which part the ruling reaches. **The index tuple for `ja4h`
names index 1 alone**, so no code path can reach index 2 or index 3. A method the table
omits gains no alias, and JA4 and JA4S are two such methods, because each one still writes
the zero sentinel for an empty list.

New function `_empty_list_aliases` builds one alias mapping when the client reads the
mapping file. `_do_lookup` reads that mapping after the mapping file and before the lookup
service.

## Every lookup path

| Path | Reads the alias | Why |
|---|---|---|
| `JA4DBClient.lookup` | Yes | It calls `_do_lookup` on a lookup cache miss |
| `JA4DBClient.lookup_many` | Yes | It calls `JA4DBClient.lookup` for each fingerprint |
| `JA4DBClient._do_lookup` | Yes | It holds the read |
| module-level `lookup` | Yes | It calls `JA4DBClient.lookup` of the shared client |
| `JA4DBClient._remote_lookup` | No | See below |

**The remote lookup reads no alias, and that is deliberate.** `ja4db.com` holds its own
database rather than the bundled mapping file, and a second request for the other form
would double the disclosure that #57 exists to bound.

**The alias mapping stands apart from the entries of the file.** `db info` reports
`len(db)` under FR-db-enrichment-11, so an alias inside `self._db` would report an entry
count the file does not hold.

## Proofs

**TDD, the failing case first.** On the base commit, against the unchanged lookup:

```
>       assert bundled_client.lookup("2bab15409345_af684594efb4_e3b0c44298fc") is not None
E       AssertionError: assert None is not None
E        +  where None = lookup('2bab15409345_af684594efb4_e3b0c44298fc')

FAILED test_an_empty_issuer_list_matches_the_first_sliver_row
FAILED test_an_empty_extension_list_matches_the_qakbot_row
2 failed, 1 passed
```

**The one that passed is the guard.** `test_a_sentinel_in_part_c_of_ja4h_gains_no_match`
passed on the base and it passes now, so the change adds the match and moves no other one.

**Both directions.** `e3b0c44298fc_4f24da86fad6_bf0f0589fc03` now reaches the Sliver row,
and `000000000000_4f24da86fad6_bf0f0589fc03` still reaches it.

**No false match.** These three return None, and each one is a value a blanket fallback
would have matched:

- `ge11nn08enus_050dd5cfb971_e3b0c44298fc_000000000000`
- `ge11nn08enus_050dd5cfb971_000000000000_e3b0c44298fc`
- `ge11nn08enus_050dd5cfb971_e3b0c44298fc_e3b0c44298fc`

**Corpus measurement, read on 2026-08-15.**

| Measurement | Count |
|---|---|
| Entries of the bundled mapping file | 70 |
| Entries that gain an alias | 3, all `ja4x` |
| JA4H rows of the file holding the sentinel | 7, every one in part c or part d |
| JA4H rows that gain an alias | 0 |
| Distinct JA4X corpus values | 31, of which 0 carry either form |
| JA4H corpus values | 151 |
| JA4H corpus values holding the sentinel in part c or part d | 114 |
| JA4H corpus values that gain a match | 0 |
| Lookups of a committed capture that change | 0 |

The 0 of the JA4X corpus agrees with the 0 of 120 that #619 measured. One JA4H value
carries the empty-list hash in part b, which agrees with the 1 of 215 that #612 measured,
and it gains no match because no row of the file holds the sentinel in part b. **The 114
values are the measure of the defect the constraint prevents.**

## No fingerprint moves

`git diff --stat ja4plus/` names `ja4plus/ja4db.py` alone, and this change touches no
fingerprinter. The conformance suite reports the counts of its base.

## The five gates

| Gate | Base | This branch |
|---|---|---|
| `pytest tests/ -m "not spec_validation"` | 5567 passed, 8 skipped, 8 xfailed, 114 subtests | 5603 passed, 8 skipped, 8 xfailed, 114 subtests |
| `pytest tests/ -m spec_validation` | 1647 passed, 142 skipped, 138 xfailed | 1647 passed, 142 skipped, 138 xfailed |
| `ruff check ja4plus/ tests/` | clean | All checks passed! |
| `ruff format --check ja4plus/ tests/` | clean | 247 files already formatted |
| `mypy --strict ja4plus/` | clean | Success: no issues found in 32 source files |

**The register invariant holds.** `tests/foxio_deviations.json` holds 138 keys against 138
xfailed cases, on the base and on this branch.

**The case count rises by 36, and every one is accounted for.** New file
`tests/test_ja4db_empty_list_alias.py` holds 33 cases, and three repository-wide checks
each parametrize over every Python file of `tests/` and each gains one case for the new
file: `tests/test_documented_method_count.py`, `tests/test_ruling_vocabulary.py` and
`tests/test_statistics_thread_term.py`.

**Coverage rises.** `ja4plus/ja4db.py` holds 144 statements on the base and 177 on this
branch, and the 19 uncovered lines are the same 19 on both sides, so the file moves from
125 of 144 to 158 of 177. Every statement this change adds is covered.

## Documentation

- `docs/specs/features/07-db-enrichment.md` gains FR-db-enrichment-18, FR-db-enrichment-19
  and the section `### The lookup reads both forms of an empty-list value`.
- `docs/api_reference.md` states the same table for a caller, under `### ja4plus.ja4db`.
- `CHANGELOG.md` records `Round TBD.` and `docs/specs/spec.md` records `| TBD | 2026-08-15 |`.

Verified against: https://github.com/FoxIO-LLC/ja4/blob/main/ja4plus-mapping.csv (FoxIO
mapping file at the pinned commit, retrieved 2026-08-15)


Crank-Git:
## Self-review, two lenses

**Correctness lens.** No defect found. The reviewer probed the four shapes that would
break the ruling, and each one held.

- A JA4H row that holds the sentinel in part b and in part c produced one alias, and that
  alias flipped part b alone. Part c and part d stayed unchanged.
- A JA4X row that holds the sentinel in all three parts produced exactly 7 aliases, which
  is 2³ − 1. The set held no duplicate and it held no copy of the original value.
- A row whose key equals the alias of another row keeps its own result.
  `_empty_list_aliases` writes no key that the mapping already holds, and `_do_lookup`
  reads the mapping file before the alias mapping, so the row wins twice over.
- `self._empty_list_aliases` is built once in `__init__` and no code mutates it, so the
  shared client needs no second lock.

**Writing lens.** Seven findings, and this branch takes all seven. Commit `ab904e8` holds
them.

- Six sentences passed the 25-word limit of rule 1 of `.claude/rules/ste.md`, in
  `ja4plus/ja4db.py`, `tests/test_ja4db_empty_list_alias.py` and
  `docs/specs/features/07-db-enrichment.md`. None of the six belongs to a record the
  exemption covers, so each one becomes two sentences.
- `docs/specs/features/07-db-enrichment.md` read `third-party data` where `CHANGELOG.md`
  and the Changelog row of `docs/specs/spec.md` each read `FoxIO published data`. Rule 7
  gives one concept one word, so the page takes the second form.

**The writing lens read two further questions and it raised neither as a finding.** The
`## Terms` table names `zero sentinel`, and the diff shortens it to `the sentinel` after
the first mention, which rounds 77, 228 and 231 already do. `C2` reaches this branch inside
`Sliver/Havoc C2 Server` and `Qakbot C2`, which are application names of the mapping file,
and the verbatim list of the standard covers them.

**The lens also swept `docs/` for a claim this change would leave stale, and it found
none.** `docs/migration-0.6-to-1.0.md`, `docs/output-schema.md`,
`docs/specs/features/00-foundation.md`, `docs/specs/features/03-concurrency-safety.md`,
`docs/specs/features/09-release.md` and `docs/specs/features/11-pre-release-validation.md`
each name the lookup, and no one of them states an entry count or a match rule that this
change moves.

## The case count and the coverage, after the second commit

**The first reading of the case count was taken while the new file was untracked, and this
comment corrects it.** Three repository-wide checks read the tracked file list of git, so
each one gained its case only after the commit. The suite reports 5603 passed against the
5567 of the base, and 33 of the 36 are the cases of the new file.

```
tests/test_documented_method_count.py::test_no_python_file_states_the_count_of_classes_as_a_count_of_methods[tests/test_ja4db_empty_list_alias.py]
tests/test_ruling_vocabulary.py::test_no_python_source_names_a_determination_of_the_user_by_the_rejected_word[tests/test_ja4db_empty_list_alias.py]
tests/test_statistics_thread_term.py::test_no_python_file_names_the_statistics_thread_by_the_rejected_word[tests/test_ja4db_empty_list_alias.py]
```

**Coverage rises and it falls nowhere.** `ja4plus/ja4db.py` holds 144 statements on the
base and 177 on this branch. The uncovered set is the same 19 lines on both sides, so the
file moves from 125 of 144 to 158 of 177, and every statement this change adds is covered.

## The five gates, after the second commit

```
All checks passed!
247 files already formatted
Success: no issues found in 32 source files
5603 passed, 8 skipped, 1970 deselected, 8 xfailed, 16 warnings, 114 subtests passed in 94.12s
1647 passed, 142 skipped, 5662 deselected, 138 xfailed, 1 warning in 9.69s
```

`tests/foxio_deviations.json` holds 138 keys, against the 138 xfailed cases of the
conformance suite. The conformance suite reports the counts of its base, so no fingerprint
moves.

**The provider holds no run for either head commit**, which the batch model requires of a
member. `gh api "repos/Crank-Git/ja4plus/actions/runs?head_sha=<head>"` reads `total_count`
0 for `a965b59` and for `ab904e8`.

Issue #709, opened 2026-08-16
Read the blank and comment citations of docs/specs/foxio/


Crank-Git:
## What this issue holds

**#690 widened `tests/foxio_citation_lines.py` so that it refuses a citation whose lines hold
no statement.** A blank line holds none, and a line that holds a comment alone holds none
either. #690 repaired the six citations of `docs/implementation_notes.md` that named such a
line, and it found 27 more on the pages of `docs/specs/foxio/`.

**#668 measured those 27 and passed every one of them**, because the condition of that round
read one question: does the file hold the cited line at the read the citation declares. A
blank line answers yes.

`tests/citation_line_allowlist.json` records all 27, as 25 entries. Each entry names this
issue. **This issue reads each entry and it removes the file.**

## The measurement

A read of 2026-08-15 against the tip of `batch/705-citation-records-a` measures the corpus.

| Class | Count |
|---|---|
| Citations of `docs/specs/foxio/` | 282 |
| Citations that name a blank line | 12 |
| Citations whose lines hold a comment alone | 15 |

## Two readings this issue must part

**A citation that names a blank line names nothing, and it is wrong in every class.**

**A citation that names a comment line is not always wrong**, because a page may cite the
comment that states a reason. `tests/test_foxio_rust_parity.py:72` is such a line, and six
pages cite it. `ja4plus/fingerprinters/ja4x.py:65` holds `# Create a signature based on
structural elements`, which restates the code below it, and a page that cites it names no
reason.

**Read each of the 27 and take one of two actions.** Repair the citation to the line that
holds the statement the page describes, or record the reading that keeps it.

## Warning: eight of the 12 blank citations stand under a pin

`docs/specs/foxio/JA4D.md` pins `ce2fa544` and `docs/specs/foxio/JA4X.md` pins `edfc7b2`.
**The maintainer ruled on 2026-08-15 that those pins stand.** A repair therefore reads the
pin and it names a line of that commit. A repair to a working-tree line falsifies the
citation.

## Warning: one citation of `docs/specs/foxio/JA4SSH.md` is wrong beyond its class

`docs/specs/foxio/JA4SSH.md:405` states that `ja4plus` closes the last window on a FIN+ACK
packet, at `ja4plus/fingerprinters/ja4ssh.py:221-222`. Those two lines hold a comment about
the message tracker and a retransmission. `ja4plus/fingerprinters/ja4ssh.py:268` holds the
[...]

Issue #705, opened 2026-08-15
Batch: the citation and documentation records, group A


Crank-Git:
The four members of this batch write different files, so they run in parallel.

- [x] #650 — a regular expression ported from ja4plus-go carries no statement of its backtracking cost — merged as #707
- [x] #690 — read the repository-owned citations of `docs/implementation_notes.md` — merged as #711
- [x] #691 — `docs/specs/foxio/` holds no JA4 transcription page — merged as #713
- [ ] #683 — retire the stale JA4SSH parity-harness section of the reading page

**#658 was a member and it is closed with no code.** Its premise is false at the tip.

**Two issues this batch filed, and neither is a member.** #709 removes
`tests/citation_line_allowlist.json`, which #690 wrote against 27 pre-existing citations. #714
repairs the divergence register row that #691 falsified.

**#691 edited `.github/workflows/test.yml`**, adding the pin commit `47ad216` to the one existing
`git fetch` command of the `Fetch the two recorded change sets` step. **Read that step at the batch
gate.** `.claude/rules/batch-gate.md` records #586: two fetch commands in one step fail on a
shallow clone, and the rule is one command that names every refspec.

**Read the cross-check addendum before you plan the remaining member.**

Round numbers stay `TBD` in every member.

Crank-Git:
finding: batch cross-check, batch/705-citation-records-a

Compared the four plan comments of #650, #690, #691 and #683 against each other, on five
dimensions: the file each member writes, the guard module each member touches, the issue numbers
each one cites as evidence, whether one plan assumes a shape another plan changes, and the round
record.

**No collision between the four.** Each writes a file no sibling writes.

| Member | The files it writes |
|---|---|
| #650 | `.claude/rules/conformance.md`, `CLAUDE.md` |
| #690 | `docs/implementation_notes.md`, `tests/foxio_citation_lines.py` |
| #691 | the new `docs/specs/foxio/JA4.md` |
| #683 | `docs/specs/foxio/JA4SSH.md` |

**One pair that shares a module and does not collide, stated because it is the pair a reader
would suspect.** #690 widens `tests/foxio_citation_lines.py` so that the guard reaches
`docs/implementation_notes.md`. #691 and #683 each write a page the guard already reaches, and
neither one edits the guard. **#690 is the one member that may edit that file**, and the plan of
each other member says so.

**One ordering consequence of that pair.** #690 makes the guard refuse a blank line and a comment
[...]

Issue #704, opened 2026-08-15
Batch: the sequenced divergence register rows, group A


Crank-Git:
The four members of this batch each write one row of the divergence register of
`docs/specs/spec.md`. That file is one table, so the members run one at a time and never
in parallel. Each ruling is already recorded on its own issue, and the work is the row.

- [x] #595 — rule whether the QUIC path evicts its JA4L points at point D — merged as #706
- [x] #607 — rule the JA4H emission frame and the incomplete body — merged as #710
- [x] #608 — read the JA4SSH packet selection against the ssh.direction label — merged as #712
- [ ] #613 — read every QUIC packet of a coalesced datagram for the JA4L points

**#691 of batch #705 falsified a live sentence of this table**, and #714 holds the repair. It is
not a member of this batch. Read the comment above before you assign rounds.

Round numbers stay `TBD` in every member. The project manager assigns them in one pass
after every live batch merges.

Crank-Git:
finding: batch cross-check, batch/704-register-rows-a

Compared the four plan comments of #595, #607, #608 and #613 against each other, on six
dimensions: the file each one writes, the table inside that file, the section of the code each
row describes, the parity rule each row claims, the issue numbers each row cites, and the round
record.

**One collision, and it is the collision this batch is built around.** All four members write the
divergence register table of `docs/specs/spec.md`, and all four write `CHANGELOG.md` and the
Changelog table of the same file. **The members therefore run one at a time**, each branching from
`batch/704-register-rows-a` after its predecessor sub-merges. That is the whole reason this batch
exists, and it is a schedule and not a defect.

**Warning to every member: `docs/specs/spec.md` holds two tables and one resolver does not serve
both.** The Changelog table takes the numbered rows of the base and appends the rows a branch adds.
The divergence register takes both sides. Batch #689 of session 16 lost the rows of #653 and #654
to a resolver written for the first table and applied to the second. **Count the hunks and read
which table each hunk sits in.**

**No other collision.** The four rows describe four disjoint areas of the code and no plan assumes
a shape another plan changes.

| Member | The code the row describes | Parity rule |
|---|---|---|
| #595 | the QUIC JA4L point eviction of `ja4l.py` | 1 |
| #607 | the emission frame of `ja4h.py` | 1 |
| #608 | the packet selection of `ja4ssh.py` | 1 |
| #613 | the first-packet read of `quic_utils.py` | 1 |

**Two readings the members must not merge with their own.**

[...]

Issue #691, opened 2026-08-15
docs/specs/foxio/ holds no JA4.md transcription page


Crank-Git:
A worker on #653 found this while it measured the empty-list question. The project manager files it, because it is wider than that issue.

## The gap

`docs/specs/foxio/` holds a transcription page for JA4S, JA4H, JA4T, JA4X, JA4L, JA4SSH and JA4D. **It holds none for JA4.**

**A reader who searches that directory for a JA4 rule finds silence**, and silence reads as "this project transcribed nothing" rather than "look elsewhere".

## Why the gap matters more for JA4 than for any other method

**JA4 is the one method of this family that carries a text specification beside its image.** `technical_details/JA4.md` of the FoxIO pin names the zero sentinel at line 121 for JA4_b and at line 176 for JA4_c, each with a stated reason.

**#653 turned on exactly that file.** The image names no empty case, so a reading that stopped at the image would have concluded that JA4 should hash an empty list — and 4 of the 167 JA4 vectors of the shared set hold `000000000000` in part c, so that change would have failed four conformance cases.

**The evidence that prevented a wrong ruling lives in a file this repository transcribes nowhere.**

## What this issue asks

1. **Write `docs/specs/foxio/JA4.md`**, transcribing the JA4 rules from the image **and** from `technical_details/JA4.md`, in the shape the sibling pages use.
2. **State plainly that JA4 carries a text specification and that the other methods do not**, so the next reader does not stop at the image.
3. **Record the empty-list rule**, with the citations #653 measured.
4. **Read whether the same shape reaches any other method.** `deleted-text-specifications.md` records what FoxIO removed; **read whether a live text specification exists for a method beside JA4** and report the answer, including the clean negative.

## Warning

**Read #668 before you add citations.** The maintainer ruled on 2026-08-15 that a citation under a pinned comparison section is read at the commit that section declares. **A new page must state its pin**, or it stands outside the guard that #668 builds.

Crank-Git:
## Plan, session 17

**Premise verified at `dda5ae7`.** `docs/specs/foxio/` holds `JA4D.md`, `JA4H.md`, `JA4L.md`,
`JA4S.md`, `JA4SSH.md`, `JA4T.md`, `JA4X.md`, `README.md`, `zeek.md` and
`deleted-text-specifications.md`. **There is no `JA4.md`.**

**Approach.** Transcribe the JA4 text specification into `docs/specs/foxio/JA4.md`, in the shape
[...]

PR #713, opened 2026-08-16
Transcribe the JA4 image and the JA4 text specification

Crank-Git:
Refs #691. Member of batch #705, so this pull request is a draft into `batch/705-citation-records-a` and every head commit ends with the skip marker.

`docs/specs/foxio/JA4.md` now transcribes the JA4 image and the JA4 text specification. **JA4 is the one method of the twelve that carries a complete text specification beside its image**, and the transcription directory held a page for every other method. A reader who searched it for a JA4 rule found silence.

The page reads two files and it states which one carries each rule. It holds 16 rules with two corroborations each, 3 disagreements against this project, the register reading and the rulings it raises. **It changes no file under `ja4plus/` and it moves no fingerprint.**

## Why the page reads two files

The procedure of `docs/specs/foxio/README.md` transcribes one image. JA4 also holds `technical_details/JA4.md`, 9153 bytes, at the pinned commit. That file is rank 1 material and not a deleted file, so the page states which of the two carries each rule.

Both hashes reproduce the inventory of `docs/specs/foxio/README.md` exactly.

```
14a9623ad05d6f8b5ccbff2023dc6fce10ff012dc2d202b497e3bc029aa75c94  technical_details/JA4.md
1bd63c14b3b96c2b70bfa8e85632450c9396af9a13e274489c0cb02f2a7e9615  technical_details/JA4.png
```

## Three findings this round measured

**The image draws a value that its own caption does not describe.** It captions `JA4_b` as the sorted cipher hash and it draws `acb858a92679`, which is the hash of the unsorted list. `technical_details/JA4.md:219` names that same value as the part b of the original-order form. The sorted list hashes to `8daaf6152771`. **No tool writes the string the image draws**: `t13d1516h2_acb858a92679_e5627efa2ab1` appears in no file of `tests/` and in no file of `docs/`.

**The text file states two example values and they disagree.** `technical_details/JA4.md:36` reads `b186095e22b6` for part c and `technical_details/JA4.md:194` reads `e5627efa2ab1`. The second is the value the stated rule produces, reproduced from the input the file prints at `technical_details/JA4.md:154`. `b186095e22b6` appears in no file of this repository.

**The Wireshark dissector writes no JA4 value at all.** `wireshark/source/packet-ja4.c:1723-1741` registers fifteen header fields and none reads `ja4.ja4`. Three references therefore corroborate each rule rather than four, and the page names the Python implementation, the Rust implementation and the Zeek package.

## The second question of the issue, answered

**No method beside JA4 holds a live text specification, so this round files no further issue.** A read of the provider on 2026-08-15 lists `technical_details/` at the pinned commit: twelve files, nine images and three text files. `JA4H.md` is 278 bytes, it builds no fingerprint, and `docs/specs/foxio/JA4H.md` already transcribes it under #203.

## One record this page falsifies

The divergence register of `docs/specs/spec.md` states that `docs/specs/foxio/` holds no `JA4.md` transcription. #653 wrote that sentence and it was true then. **Batch #704 owns that table, so this pull request edits no row of it**, and it records the constraint on #704 instead. The Changelog row of round 248 states the same sentence and it stays as written, because a Changelog row records a past measurement.

## The pin

The comparison section declares its pin, which #668 requires, and `.github/workflows/test.yml` fetches `47ad2163a15d93fed3864f4732e3f1e00b08d6b2` so the clone of depth 1 reaches it. `tests/foxio_citation_lines.py` reads every citation of the page under that pin. **This pull request adds no entry to `tests/citation_line_allowlist.json`**, which #709 owns.

## How this was tested

New file `tests/test_foxio_ja4_transcription.py` holds twelve cases. **Nine of the twelve failed before the page existed**, including `test_the_directory_holds_a_transcription_page_for_ja4`. The cases read the page against the inventory, against the divergence register and against the vector set, so a page that drifts from any one of them fails here.

The citation guard reported one wrong citation before the change landed: `ja4plus/utils/tls_utils.py:443` holds a blank line and `:444` holds the GREASE condition.

```
pytest tests/ -m "not spec_validation"   5724 passed, 8 skipped, 8 xfailed, 114 subtests passed
pytest tests/ -m spec_validation         1676 passed, 142 skipped, 138 xfailed
ruff format --check ja4plus/ tests/      254 files already formatted
mypy --strict ja4plus/                   Success: no issues found in 32 source files
```

**No fingerprint moves.** The conformance counts are identical before and after, and `tests/foxio_deviations.json` holds 138 keys against those 138 xfailed cases.

Verified against: https://github.com/FoxIO-LLC/ja4/blob/main/technical_details/JA4.md (retrieved 2026-08-15, commit `27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8`)
Verified against: https://github.com/FoxIO-LLC/ja4/blob/main/technical_details/JA4.png (retrieved 2026-08-15, commit `27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8`)


Crank-Git:
## Self-review

A review pass read the whole change through four lenses. It re-derived every quantitative claim of `docs/specs/foxio/JA4.md` from the data rather than from the page: the register, the vector set, the Rust snapshots and `ja4plus/` at the pin. **Every count and every citation came back exact.** It also rebuilt a tree without the page and ran the new file there, which reports 9 failed and 3 passed and reproduces the number this round states.

**It found four items and this branch repaired all four.**

1. **`docs/specs/foxio/JA4.md`, R5, stated one rule in the passive voice.** The line now reads that the reader ignores the handshake version at the top of the packet.
2. **R16 claimed DTLS coverage that no corroboration carried.** No implementation that writes JA4 marks a DTLS client hello. The rule now states the QUIC case, it names two corroborations for it, and R3 holds the whole reading of that character.
3. **`test_the_page_and_the_register_agree_on_the_sentinel_value` could not fail for the right reason.** It read the whole page and the whole specification, and the register names that sentinel for three methods, so a reader could delete the R13 section and the case would still pass. New reader `rule_section` returns one rule of the page, and the case now reads R13 alone and requires both source lines.
4. **`test_no_expected_output_file_holds_the_string_the_image_draws` asserted over a corpus it never measured.** An empty corpus passes every negative assertion. The case now states the floor first, and a read of 2026-08-15 returns 167 JA4 values and 167 JA4_o values.

**The lens raised one reading this branch declines.** Several corroboration bullets pass 25 words once a `file:line` citation counts as a word. `docs/specs/foxio/JA4S.md` and every other merged page hold that same shape, so a repair here would part this page from its siblings. The lens named it as a pre-existing convention and not a defect of this change.

## The gates at the head commit

```
pytest tests/ -m "not spec_validation"   5724 passed, 8 skipped, 8 xfailed, 114 subtests passed
pytest tests/ -m spec_validation         1676 passed, 142 skipped, 138 xfailed
ruff format --check ja4plus/ tests/      254 files already formatted
mypy --strict ja4plus/                   Success: no issues found in 32 source files
```

**The conformance counts are identical to the base**, which reads 1676 passed, 142 skipped and 138 xfailed. `tests/foxio_deviations.json` holds 138 keys against those 138 xfailed cases, so the register invariant holds. **No fingerprint moves and no file under `ja4plus/` changes.**

**No provider run exists for this branch, and that is by design.** Every head commit ends with the skip marker, so the batch pull request stays the first run of these cases on Linux.

Issue #684, opened 2026-08-15
Name the JA4H and JA4SSH snapshot blocks in the vector download filter


Crank-Git:
`tests/download_test_vectors.py:141` holds the list of snapshot fields the parity module
compares, and that list names four of the six.

```python
RUST_COMPARED_FIELDS = (b"ja4: ", b"ja4s: ", b"ja4t: ", b"ja4x: ")
```

`tests/download_test_vectors.py:343` keeps a Rust snapshot only where one line of it starts
with one of those four. **`ja4h` and `ja4ssh` reach no entry of the list.** #670 added the
JA4H comparison and #671 added the JA4SSH comparison, and neither one moved this file.

## What it costs today

**Nothing, and that is a measurement rather than a reason to leave it.** Every local
snapshot that holds a `ja4h` block or a `ja4ssh` block also holds a `ja4t` value, so the
condition keeps all eleven. A refresh that finds a capture whose snapshot holds an HTTP
request and no TCP handshake drops that snapshot, and the two comparisons then read one
value fewer with no report.

## What this issue asks

Add the two field forms to the list, and read the two block shapes before you write them.

1. The `http` block writes `- ja4h: <value>`, so the field form carries the list marker.
2. The `ja4ssh` block writes each value as a bare list item and it names no field at all,
   so a field form reaches it nowhere. The condition needs the block opener `ja4ssh:`
   instead.

**Write a case that reads the list against `tests/test_foxio_rust_parity.py`.** No case
holds the two files together today, which is why two rounds moved one and left the other.

## Warning

`tests/download_test_vectors.py` writes into `tests/foxio_vectors/`. Read what the script
does before you run it, and never let a run of it move a committed vector.

Crank-Git:
## Plan, session 17

**Premise verified at `dda5ae7` and re-read at `4d00e09`.**

```
tests/download_test_vectors.py:141  RUST_COMPARED_FIELDS = (b"ja4: ", b"ja4s: ", b"ja4t: ", b"ja4x: ")
tests/download_test_vectors.py:343  if not any(line.strip().startswith(RUST_COMPARED_FIELDS) for line in snapshot.splitlines()):
```

**Four fields of six. `ja4h` and `ja4ssh` reach no entry**, and #670 and #671 each added a
comparison without moving this file.

**Approach.** Add `b"ja4h: "` and `b"ja4ssh: "` to the tuple, so the download filter keeps a
snapshot that holds either block.

[...]

Issue #670, opened 2026-08-15
Compare the six FoxIO Rust JA4H values that no case reads


Crank-Git:
A sweep of 2026-08-15, run under #638, read every value-bearing field of the 11 snapshots under `tests/foxio_vectors/rust_expected/`. It found a second committed value class that no case compares.

## The measurement

Six `- ja4h:` values sit under an `http:` block of a Rust snapshot. `read_rust_snapshot` of `tests/test_foxio_rust_parity.py` enters the `tls_certs` block alone, so it reads none of them.

```
ja4__insta@ssh2.pcapng.snap:  http:
ja4__insta@ssh2.pcapng.snap:  - ja4h: ge11nn030000_9ab90a797ba7_000000000000_000000000000
```

`CERTIFICATE_BLOCK_OPENERS` names `tls_certs` and `- x509`, and it names no `http` block.

## The census the sweep produced

| Field | Values | Reaches a case |
|---|---|---|
| `ja4` | 191 | yes |
| `ja4s` | 110 | yes |
| `ja4t` | 39 | yes |
| `ja4x` | 43 | yes |
| `ja4l_c` | 35 | yes, after #638 |
| `ja4l_s` | 35 | yes, after #638 |
| `ja4h` | 6 | **no** |
| `ja4ssh` | 2 | **no**, and #GAP holds it |

## What this issue asks

Read whether the six values reach a comparison, and record the answer either way. Read each one against the FoxIO Python expected-output file first, because that file decides where both references hold a value.

**Warning: the six values carry the zero sentinel `000000000000` in part c and part d.** #619 ruled on the sentinel, and this issue decides nothing about it.

Crank-Git:
The  row of the census names #671, which holds that finding. The body reads `#GAP` because the two issues opened in one pass and this one opened first.

Crank-Git:
The `ja4ssh` row of the census names #671, which holds that finding. The body above reads `#GAP` because the two issues opened in one pass and this one opened first.

Crank-Git:
Plan. Batch #676, integration branch `batch/676-citation-guard`.

**Approach.** #638 swept all eight value-bearing classes of the eleven FoxIO Rust snapshots and found that `read_rust_snapshot` of `tests/test_foxio_rust_parity.py` enters the `tls_certs` block alone. **Six JA4H values under the `http` block therefore reach no case.**

Read the `http` block, and compare the JA4H values it holds.

[...]

PR #721, opened 2026-08-16
Name the JA4H, JA4SSH and JA4X snapshot blocks in the vector download filter

Crank-Git:
## What this changes

`tests/download_test_vectors.py:141` held `RUST_COMPARED_FIELDS` with four entries, and
`tests/test_foxio_rust_parity.py` compares six methods against a FoxIO Rust snapshot. #670
added the JA4H comparison and #671 added the JA4SSH comparison, and neither round moved the
field list. This change adds the two forms and holds the two files against each other.

## Why the two forms carry the shapes they do

**The issue body and the plan comment stated different forms, so a measurement decided
between them.** A read of 2026-08-15 applied each candidate to the eleven committed
snapshots under `tests/foxio_vectors/rust_expected/`, with the same strip the filter runs.

| Form | Snapshots it matches |
|---|---|
| `b"ja4h: "` | 0 of 11 |
| `b"- ja4h: "` | 5 of 11 |
| `b"ja4ssh: "` | 0 of 11 |
| `b"ja4ssh:"` | 1 of 11 |

The `http` block writes the value as a list item, so the stripped line opens with the list
marker. The `ja4ssh` block names no field at all, so its opener is the one line that carries
the method.

```
  http:
  - ja4h: ge11nn07enus_3e3b55d61660_000000000000_000000000000
```

```
  ja4ssh:
  - c36s36_c76s124_c74s5
```

**The issue body is correct and the plan comment's two literals are not.** The plan itself
required this reading before the forms were written.

## The two measurements the issue asks for

1. **The kept-snapshot count stands at 11 before the change and 11 after it.** Every
   committed snapshot that holds a JA4H block or a JA4SSH block holds a JA4T value as well,
   so the filter refuses no snapshot it kept before.
   `test_the_filter_keeps_every_committed_rust_snapshot` measures it, and it passes on both
   sides of the change.
2. **The case builds a snapshot that holds a JA4H block and no JA4T value.**
   `test_the_filter_keeps_a_snapshot_that_holds_one_method_alone[JA4H]` and its JA4SSH twin
   each failed on the base commit `4d00e09` and each passes now.

```
FAILED tests/test_download_filter_fields.py::test_the_filter_keeps_a_snapshot_that_holds_one_method_alone[JA4H]
FAILED tests/test_download_filter_fields.py::test_the_filter_keeps_a_snapshot_that_holds_one_method_alone[JA4SSH]
2 failed, 7 passed, 1 xfailed in 0.09s
```

A run of the ten final cases against the field list of `4d00e09` names three:

```
cases that fail under the base field list: ['JA4H', 'JA4SSH', 'JA4X'] count 3
```

## The third form, repaired on the same measurement

**`b"ja4x: "` matched 0 of the eleven snapshots.** The snapshot writes
`    - ja4x: <value>` inside the `tls_certs` block, exactly as the `http` block writes its
own value, so that entry of the list was dead.

| Form | Snapshots it matches |
|---|---|
| `b"ja4x: "` | 0 of 11 |
| `b"- ja4x: "` | 5 of 11 |

**The list now holds `b"- ja4x: "`.** This is the defect the JA4H entry carried, on an entry
that stood in the list already, and the maintainer's rule of 2026-08-15 says a worker repairs
what its own measurement falsifies, in the pull request that measured it. The issue this
worker first filed for it is closed under that rule.

**The kept-snapshot count reads 11 after this repair too.** A filter that began to keep a
snapshot it used to refuse would be a finding rather than a repair, and it keeps the same
eleven.

`test_the_filter_keeps_a_snapshot_that_holds_one_method_alone[JA4X]` carried
`pytest.mark.xfail(strict=True)` and it passes now, so the module carries no mark.

**A measured correction to the issue body.** It reads that the filter "drops" a snapshot it
refuses, "with no report". `tests/download_test_vectors.py` raises `ValueError` on that path,
so a refresh that meets such a snapshot fails and reports. The repair is the same either way,
and a reader who takes the issue's wording literally reads the failure mode wrong.

## How this is tested

New file `tests/test_download_filter_fields.py`, ten cases. `keeps_rust_snapshot` now holds
the condition that stood inside `download`, so a case reads the condition the refresh runs
rather than a copy of it. The extraction moves no behaviour.

One case reads the method names of `SNAPSHOT_METHODS`, `SNAPSHOT_CERT_METHOD`,
`SNAPSHOT_HTTP_METHOD` and `SNAPSHOT_SSH_METHOD` against the six the filter is expected to
name, so a rename in either file fails a case here.

**The two latency methods stand outside the six on purpose.** #684 read the field list as
four entries of six and named neither one, and all seven committed snapshots that hold a JA4L
value hold a JA4T value as well.

## Gates

| Gate | Result |
|---|---|
| `pytest tests/ -m "not spec_validation"` | 5745 passed, 8 skipped, 8 xfailed, 114 subtests |
| `pytest tests/ -m spec_validation` | 1676 passed, 142 skipped, 138 xfailed |
| `ruff format --check ja4plus/ tests/` | 256 files already formatted |
| `mypy --strict ja4plus/` | no issues in 32 source files |

**The conformance counts are identical before and after the change**: 1676 passed, 142
skipped and 138 xfailed on the base commit `4d00e09` and the same three counts here.
`tests/foxio_deviations.json` holds 138 keys against those 138 xfailed cases.

**This pull request changes no file under `ja4plus/`, so it moves no fingerprint.** It
downloads nothing and it moves no committed vector.

Round number: the literal `TBD` in the `CHANGELOG.md` entry and in the Changelog table row of
`docs/specs/spec.md`.

Refs #684


Crank-Git:
## Self-review, two lenses

- **The extraction preserves the prior behaviour.** `keeps_rust_snapshot` holds the
  expression that stood at the call site, and the call site reads `if not
  keeps_rust_snapshot(snapshot):`.
- **`b"ja4ssh:"` collides with no other entry and with no other line.**
  `ja4__insta@ssh2.pcapng.snap:215` is the one `ja4ssh` line of the corpus, and
  `ssh_extras:`, `ja4s: ` and `ja4l_s: ` each differ from it before the colon.
- **The synthetic snapshots reproduce the real indentation.** Checked against
  `ja4__insta@ssh2.pcapng.snap:215-216`, `:232-234` and
  `ja4__insta@latest.pcapng.snap:15-17`.
- **The JA4X case xfails for the stated reason.** Its value line strips to
  `- ja4x: a373a9f83c6b_2bab15409345_0f2217ba412e`, which starts with no entry of the
  tuple, so the assertion fails rather than an unrelated error.
- **No case passes for the wrong reason**, and none asserts an empty condition.

### The writing standard — three breaches, all repaired in `b31bc38`

1. **Rule 1**, five description sentences past 25 words: the comment above
   `RUST_COMPARED_FIELDS`, the module docstring of the new file and the
   `EXPECTED_METHODS` comment. The first now holds two paragraphs, and the other two each
   split one sentence.
2. **Rule 11**, the passive clause `the field list is expected to name`, in the new file
   and in the Changelog row of `docs/specs/spec.md`. Both now read `the six the filter
   names`.
3. **Rule 7**, synonym rotation on the refusal.
   `test_the_filter_drops_a_snapshot_that_holds_no_fingerprint_value` said `drops` where
   every other sentence says `refuses`, and the docstring of `keeps_rust_snapshot` said
   `rejects`. All three now say `refuses`.

**One finding of the second lens is declined.** It read the duplicate sentence `This round
changes no file under `ja4plus/` and it moves no fingerprint.` across the two records as
redundant. That sentence stands in every round record of this project, so a removal here
would break the pattern rather than repair it.

### The count moved, and the cause is the tracked file list

The first gate run reported 5744 unit cases where the round records said 5742. **Two cases
parametrize over the git-tracked file list**, and `tests/test_download_filter_fields.py`
reached that list at the commit rather than at the run. A collection of the working tree
against `d7ee3de` reports no difference, which proves the cause. Both records now state
5744 and name it.

Issue #727, opened 2026-08-16
Adopt the five-case filing rule of the port, and scope a round to a batch


Crank-Git:
**The maintainer ruled all three parts of this on 2026-08-15.** This issue carries them together
because they interlock: the filing rule stops the backlog regenerating, and the round model stops
every pull request writing the two most contended files of the repository.

## The measurement

A read of 2026-08-15 compares this repository against `Crank-Git/ja4plus-go`.

| Reading | `ja4plus` | `ja4plus-go` |
|---|---|---|
| Commits of the last 300 that touch library code | 56 | 152 |
| Commits of the last 300 that touch `docs/specs/spec.md` | 192 | 45 |
| Lines that file gains across them | 731, which is 3.8 a commit | 1564, which is 35 a commit |
| Commits of the last 300 that touch `CHANGELOG.md` | 186 | 39 |
| Rounds in the Changelog table | 249 | 71 |
| Library lines | 12353 | 15694 |
| Lines of the bookkeeping guards | 13540 | 4899 |

**This project edits its specification in 64 percent of commits and gains 3.8 lines each time. That
is not authoring. That is appending a table row.** The port edits the same file a third as often and
writes nine times more at each edit.

**The guards that enforce the bookkeeping now hold more lines than the library they guard.**

## Part 1. Adopt the five-case filing rule, verbatim

**`Crank-Git/ja4plus-go/CLAUDE.md:244` holds this rule and the maintainer adopted it there on
2026-08-13. This repository never adopted it.**

**Parity rule 2 governs: where FoxIO specifies nothing, the port has already shipped a choice, and
this project adopts it.** FoxIO specifies no issue-filing policy. **This is a parity defect and not a
proposal.**

Copy the rule into `CLAUDE.md` of this repository **verbatim**, including its measurement. A finding
earns a tracker issue in five cases and never otherwise.

- Behavior.
- A fingerprint value.
- A guard that guards nothing.
- A blocked epic.
- A question the maintainer must rule.

**Every other finding is repaired in the batch that found it, and it files no issue.**

Add the valve the maintainer set on 2026-08-15, which `.issue-flow.json` records under
`filingRule`: **no issue is filed whose only symptom is a stale citation, or a record that describes
a past state**, and a worker repairs what its own change falsifies in the same pull request.

**Warning: do not reword the rule to suit this repository.** Two files that state one rule in two
wordings is how drift starts, and the maintainer chose the verbatim adoption for that reason.

[...]

Issue #621, opened 2026-08-15
Decide whether to reverse the decryption non-goal of #129, because the Go port is building it


Crank-Git:
## Why this issue exists

**The Go port is building a route for key material into a fingerprinter. This project declined decryption as a non-goal at #129, and that decline is recorded as reversible. This issue asks whether the maintainer reverses it.**

**Nothing in this repository changes until the maintainer answers.** This issue proposes no code.

## What this project decided, and the reason it gave

`docs/specs/spec.md` `Non-goals` at the tag `v1.1.0`:

> - Decryption is out of scope. `ja4plus` reads no key material and reads no Decryption
>   Secrets Block. It decrypts no TLS record and no QUIC 1-RTT packet. Traffic that a
>   capture carries only in encrypted form therefore produces no fingerprint, even when
>   the same capture file carries the key material that decrypts it. #129 holds the

**Round 26 of the `## Changelog` states the reason, and the reason is a security boundary and never a scope preference:**

> The reason is a security boundary: a program that reads a key log file reads the plain text of the traffic the operator captured, and that capability is larger than fingerprint production. The decision is reversible.

**17 JA4H register entries carry #129 as their cause**, and round 51 records 43 more marked entries that state the reference decrypts a capture the operator captured secrets for.

## What the Go port already ships, and it predates this issue

**`Crank-Git/ja4plus-go` already exports eight names for key material**, and its maintainer accepted them on 2026-08-11, at round 11 of its own changelog.

| Name | Signature |
|---|---|
| `ErrNoSecret` | `var ErrNoSecret error` |
| `KeyLog` | `type KeyLog struct{ ... }` |
| `ParseKeyLog` | `func ParseKeyLog(r io.Reader) (*KeyLog, error)` |
| `ReadKeyLogFromCapture` | `func ReadKeyLogFromCapture(r io.Reader) (*KeyLog, error)` |
| `KeyLog.Secret` | `func (k *KeyLog) Secret(clientRandom []byte, label string) ([]byte, error)` |
| `KeyLog.ClientRandoms` | `func (k *KeyLog) ClientRandoms() [][]byte` |
| `KeyLog.Len` | `func (k *KeyLog) Len() int` |
| `DecryptQUICPacket` | `func DecryptQUICPacket(payload, secret []byte, connectionIDLength int) ([]byte, error)` |

[...]

Issue #593, opened 2026-08-11
Adopt or decline the key-log interface that ja4plus-go ships first


Crank-Git:
## The Go implementation ships a key-log interface, and this port holds none

`Crank-Git/ja4plus-go` issue #40 built the reader for a pcapng Decryption Secrets Block, and
it exports an interface so that a caller who reads a capture through another reader can
supply TLS secrets. **This port exposes no equivalent name.**

`.claude/rules/parity.md` rule 2 states that the port decides the interface where FoxIO
specifies nothing and the Go project shipped nothing, **and that where the Go project ships
a name first the rule runs the other way.** This is that case. The maintainer accepted the
names on 2026-08-11.

## The eight names

The Go project exports these from its root package:

| Name | What it is |
|---|---|
| `KeyLog` | The type that holds the secrets of one capture. |
| `ParseKeyLog(r io.Reader) (*KeyLog, error)` | Reads the NSS key-log format. |
| `ReadKeyLogFromCapture(r io.Reader) (*KeyLog, error)` | Reads a Decryption Secrets Block from a pcapng capture. |
| `KeyLog.Secret(clientRandom []byte, label string) ([]byte, error)` | Returns one secret. |
| `KeyLog.ClientRandoms() [][]byte` | Names every connection the key log covers. |
| `KeyLog.Len() int` | The count of secrets. |
| `DecryptQUICPacket(payload, secret []byte, connectionIDLength int) ([]byte, error)` | Decrypts one QUIC packet. |
| `ErrNoSecret` | The error a caller reaches when no secret covers the connection. |

## One measurement worth carrying across

`gopacket` at `v1.1.19` **cannot surface a Decryption Secrets Block at all.**
`pcapgo/pcapng.go:35-40` defines no constant for block type 10, and `pcapgo/ngread.go:504`
discards it in the `default` case. The Go project therefore wrote its own reader. **Check
whether scapy has the same gap before you plan this.**

## What this issue asks

Decide whether this port adopts the same interface, adopts it under different names, or
declines it. **A decline is a legitimate answer** — the two languages need not expose the
same helper set — but the register on both sides must record which way it went.

The Go side records this in the `## Parity with ja4plus` register of `docs/specs/spec.md`.

Opened by the issue-flow loop on the Go side. No code change is requested here.

Crank-Git:
## The Go side's eight names are final, and Epic 10 freezes them

This is the information this issue needs in order to adopt or decline. Recorded from
[...]

PR #739, opened 2026-08-16
Record the key-material interface the port ships first

Crank-Git:
## What this changes

`docs/specs/spec.md` carries one new row of the `## Parity with ja4plus-go` divergence
register. The row records the eight names `Crank-Git/ja4plus-go` exports for key
material, and it declines all eight. `tests/test_key_log_interface_ruling.py` holds the
row and every measurement it states.

This change edits no line under `ja4plus/`. It writes no `CHANGELOG.md` entry and no
`## Changelog` row, because #727 moved the round record to the batch pull request.

## Why the answer is a decline

#621 declines the decryption capability, and an interface that carries key material has
no consumer while that decline stands. Parity rule 2 decides the shape of a name this
project exposes, and it compels no capability. The row states the two-part reversal
condition of https://github.com/Crank-Git/ja4plus/issues/621#issuecomment-5306066957,
and it states the spent condition of 2026-08-15 nowhere as pending.

## The four measurements, and two premises they moved

1. **The eight names.** A read of 2026-08-16 found them at `types.go:111`, `:126`,
   `:139`, `:157`, `:201`, `:218`, `:239` and `:256`, on the `dev` branch of the port, on
   its `master` branch, and on its `v1.0.0` tag. **A release of the port therefore
   already carries all eight names**, and no release carries the capability #621
   declines. The row states both.
2. **`scapy` surfaces a Decryption Secrets Block.** `scapy/utils.py:1645` maps block type
   10 to `_read_block_dsb`, and `scapy/utils.py:1958` defines it, at scapy 2.7.0.
   **The gap the issue body reports for `gopacket` does not transfer.** The port pins
   `github.com/gopacket/gopacket v1.7.1` today, which reads the block at
   `pcapgo/pcapng.go:43` and `pcapgo/ngread.go:360`. The `v1.1.19` reading of the issue
   body covers `github.com/google/gopacket` alone.
3. **What this project decrypts today.** `ja4plus/utils/quic_utils.py:96` holds
   `derive_initial_secrets` and `ja4plus/utils/quic_utils.py:208` holds
   `decrypt_initial_payload`. Neither one reads key material.
4. **The corpus carries key material.**
   `tests/foxio_vectors/chrome-cloudflare-quic-with-secrets.pcapng` holds one block of
   type 10, of secrets type `0x544c534b`, over 10 lines and 2 connections. A read of the
   `analyze` command over that capture writes the `scapy` warning
   `PcapNg: TLS Key Log available, but the TLS layer is not loaded!`, leaves
   `conf.tls_nss_keys` empty, and produces the usual values. The row records that finding
   as a question for the maintainer.

## How this was tested

Test-driven: `tests/test_key_log_interface_ruling.py` failed 15 of its 21 cases before
the row existed, and the 5 live measurement cases passed from the start.

- `pytest tests/ -m "not spec_validation"`
- `pytest tests/ -m spec_validation`
- `ruff check ja4plus/ tests/`
- `ruff format --check ja4plus/ tests/`
- `mypy --strict ja4plus/`

The change moves no fingerprint value and no entry of `tests/foxio_deviations.json`.

Verified against: https://github.com/Crank-Git/ja4plus-go/blob/dev/types.go (retrieved 2026-08-16)
Verified against: https://github.com/Crank-Git/ja4plus-go/blob/v1.0.0/types.go (retrieved 2026-08-16)

Refs #593.


Crank-Git:
## Self-review, by lens

A `general-purpose` reviewer read the new row against the row above it, against every
citation, against `.claude/rules/ste.md`, and against the new test module. It reported
three repairs and one gap, and `2c87da1` takes all four.

1. **The row stated part 2 of the reversal condition in other words than the row of
   #621.** That row reads "this project measures the count of entries the capability
   closes, against its own corpus", and this row read "the yield of the capability". A
   reader could not tell one condition from two. The row now uses the words of the row
   above.
2. **The row wrote `secret` where the `## Terms` table requires `key material`.**
   `docs/specs/spec.md:132` declines that synonym. The Go identifier `Secret` stays as it
   is, because the standard reproduces an identifier verbatim.
3. **The row wrote `finding` where the `## Terms` table requires `reading`.**
   `docs/specs/spec.md:138` declines that synonym.
4. **The row states 2 connections and no case held that number.**
   `test_the_corpus_capture_holds_the_secrets_block_the_row_states` now counts the client
   randoms of the block, and `test_the_register_states_what_the_corpus_capture_carries`
   reads both counts out of the row.

The reviewer confirmed every `file:line` of the row by reading the file, and it confirmed
that `_read_block_dsb` is a method of `RawPcapNgReader`. It found no case that reads a Go
file, no network call and no machine-specific path.

**One reading the row records and this pull request does not act on.** The command-line
program surfaces a `scapy` warning about key material it declines to read. That is a
question for the maintainer, and https://github.com/Crank-Git/ja4plus/issues/702#issuecomment-5306189215
carries it to the issue that holds the three gaps.

Issue #697, opened 2026-08-15
get_stream applies no byte-cap truncation, and the Go GetStream does


Crank-Git:
A worker on #654 found this while it measured the byte-cap admission difference. It records the finding rather than repairs it, because nothing reaches it today.

## The difference

`get_stream` of `ja4plus/utils/tcp_stream.py` returns the stored bytes with no byte-cap truncation. `GetStream` of the port truncates.

**It can never matter under the present admission rule of this project.** That rule refuses the segment that would cross the cap, so the stored bytes never exceed it and a truncation would remove nothing.

**It becomes live the moment the admission rule changes.** The reversal run of #654 wrote the port admission condition into `ja4plus/utils/tcp_stream.py:151` and then measured `get_stream` returning **1048826 bytes, past the cap of 1048576**.

## Why this is filed and not repaired

**A truncation that removes nothing is untestable**, so a repair today ships a line no case can exercise and no reader can justify.

**The register row of #654 records the admission split**, and this issue records the consequence that sits behind it. **Read them together.** A future round that changes the admission rule reads this issue before it ships.

## What this issue asks

1. **Decide whether `get_stream` truncates.** The answer costs nothing today and it removes a trap from whoever changes the admission rule.
2. **Where it should truncate, add the truncation and the case that proves it** — the case needs the admission rule reversed, so state how it exercises the path.
3. **Where it should not, record why in the module**, so the difference from the port is deliberate rather than absent.

Crank-Git:
## Plan, batch #738, branch `batch/738-guards-and-parity`

### The half that is settled

**Parity rule 2 settles the interface. The port truncates, so this project adopts the truncation.**
`CLAUDE.md` states the rule: where FoxIO specifies nothing and the port has already shipped a
choice, this project adopts it. FoxIO specifies no byte cap on a stream reader, and `GetStream` of
`Crank-Git/ja4plus-go` truncates.

Answer 1 of the issue body is therefore `get_stream` truncates.

### The half that is open, and which this issue closes

**A truncation that removes nothing is untestable under the present admission rule.** The admission
rule of `ja4plus/utils/tcp_stream.py` refuses the segment that would cross the cap, so the stored
bytes never exceed the cap and the truncation removes nothing.

**#654 recorded the method that exercises the path.** Its reversal run wrote the port admission
condition into `ja4plus/utils/tcp_stream.py:151` and then measured `get_stream` returning 1048826
[...]

Issue #654, opened 2026-08-15
The two ports count 757 and 750 segments on one stream of http2-with-cookies.pcapng


Crank-Git:
A measurement of #620 found one difference between the two ports that it did not explain. This issue holds it.

## The measurement

#620 replayed every capture of `tests/foxio_vectors/` through the reassembler on 2026-08-15, to settle which stream holds the largest segment count.

| Reading | This project | `ja4plus-go` |
|---|---|---|
| The stream of `http2-with-cookies.pcapng` on port 443, under the byte cap of 1048576 | 757 segments | 750 segments |

`internal/parser/tcp_stream.go:62-65` holds the port reading. This project measured 757 twice, by two independent methods: one replays `TCPStreamReassembler`, and the other counts distinct `(seq, len)` pairs with no reassembler at all. The two agree.

**#620 measured no cause for the difference of seven, and it closed on the reading it was opened for.** The 788-segment attribution was the question of that issue, and this project now agrees with the port on it: the 788 belongs to `ssh-scp-1050.pcap` on port 22, and never to `http2-with-cookies.pcapng`.

## Why seven segments matter, and why they do not

**They move no fingerprint value.** Both counts sit far below the bound of 4096 that #620 recorded, so no stream of the corpus reaches the bound under either reading, and the bound holds whichever count is right.

**They mean the two reassemblers disagree about what one capture holds.** That is a difference in reading and not in policy, and a difference of that kind reaches a fingerprint the moment a capture puts it near a boundary. The corpus holds no such capture today.

## What this issue asks

1. **Find the seven segments.** Enumerate the `(seq, len)` pairs each port counts for `142.250.187.206:443->192.168.2.200:58847`, and name the seven the port does not.
2. **Read which port is right.** The candidates worth reading first are a retransmission that one port folds and the other counts, a zero-length segment, a segment the byte cap cuts at a different point, and a difference in where each port starts the stream.
3. **Record the answer.** Where this project is wrong, repair it. Where the port is wrong, open the issue on `Crank-Git/ja4plus-go`. Where the two readings are each defensible, the divergence register records the split.

**This issue changes no fingerprinter before step 2 answers.** No value moves under either count today, so there is no repair to rush.

Issue #738, opened 2026-08-16
Batch: the guards that guard less than a reader believes


Crank-Git:
**Four members, and no member of this batch moves a fingerprint value.** #736 and #720 each read a guard that guards less than a reader believes. #697 and #742 each repair a difference the guards did not read. **#742 joined after the cross-check**, and the addendum comment of this issue compares it.

- [x] #736 — the conformance suite does not discriminate an emission frame move — `priority:high`
- [x] #720 — a review that never reported must be reported as absent — `priority:high`
- [ ] #742 — the command prints unknown for every value close_open_windows publishes — `priority:high`
- [x] #697 — `get_stream` applies no byte-cap truncation

**This batch runs under the round model of #727.** No member writes a `CHANGELOG.md` entry and no
member writes a `## Changelog` row. The project manager writes one round entry at the batch gate.

**The batch takes the round after batch #731.** `dev` holds 263 rounds today, batch #731 takes 264,
and this batch takes 265 where #731 merges first.

**Warning: #663 is not a member and it runs alone.** The maintainer ruled the US English sweep on
2026-08-16, and a repository-wide sweep conflicts with every open branch at once.

Crank-Git:
finding: batch cross-check, `batch/738-guards-and-parity`

**The project manager compared the three plan comments of this batch before any member launched.**
The three plans are https://github.com/Crank-Git/ja4plus/issues/736#issuecomment-5306161922,
https://github.com/Crank-Git/ja4plus/issues/720#issuecomment-5306161968 and
https://github.com/Crank-Git/ja4plus/issues/697#issuecomment-5306162008.

## What the comparison read

Each plan was read against every other plan for a shared file, a shared function, a shared
interface, a shared migration, a shared configuration key and a shared route. It was also read for
a plan that assumes a shape another plan changes.

## The file sets

| Member | Files the plan names |
|---|---|
| #736 | `tests/test_spec_validation.py`, `.claude/rules/conformance.md`, `tests/foxio_vectors/` read-only |
| #720 | one file under `.claude/rules/`, and a census that reads the tracker and writes nothing |
| #697 | `ja4plus/utils/tcp_stream.py`, one new test module |

## The findings

**One pair shares a directory and no file.** #736 writes `.claude/rules/conformance.md` and #720
writes a rule file of its own under `.claude/rules/`. **#720 states no file name yet**, because step
2 of its plan reads where the rule belongs. **Warning to #720: name a file that is not
[...]

PR #743, opened 2026-08-16
Truncate the reassembled TCP run at the byte cap

Crank-Git:
Refs #697.

## The answer, and the rule that decides it

**`get_stream` truncates.** FoxIO states no byte cap on a stream reader, so parity rule 2
of `CLAUDE.md` gives the port the interface. `GetStream` of `Crank-Git/ja4plus-go`
truncates the run at `MaxBytes`, and this project adopts that reading.

`internal/parser/tcp_stream.go:225` holds `GetStream`, and its own comment states the
rule: `Returns data up to the first gap or MaxBytes, whichever comes first.`

## Every premise of the brief measured true

| Premise | The measurement |
|---|---|
| `get_stream` truncates nothing | The base form reads `return bytes(result)`, and its comment states that the method needs no bound of its own. |
| The admission rule refuses the crossing segment | `ja4plus/utils/tcp_stream.py:151` reads `if stream["bytes"] + len(data) > self.max_stream_bytes`. |
| The register row of #654 records the split | `docs/specs/spec.md:331` holds the row `The segment one TCP stream refuses at its byte cap`, and it states that `GetStream` truncates the run at `MaxBytes`. |

## What the truncation removes today, which is nothing

`add_segment` refuses the segment that would cross the byte cap, so one stream stores no
more bytes than the cap and the run never reaches it. The module states that reason above
the return, because a reader cannot see it from the code.

**#654 left the admission rule with the user.** The port admits the crossing segment and
stores 250 bytes past the cap. A move to that admission rule reaches this line, and this
change makes that move cost no byte of the value the method returns.

## How the cases reach a path the admission rule blocks

**Each case stores its segments under a byte cap that admits every one of them, and it
then lowers the cap to the value the reader reads.** The stored run of 150 bytes then
stands above a cap of 100.

The shape uses the public writer alone. No case writes `reassembler.streams`, and no case
reverses the admission rule of `add_segment`. The plan named two other shapes, and each
one costs more: a direct write of the stored state couples the case to four private keys,
and a reversal of the admission rule edits the line that the register row of #654 pins.

## The red-to-green

The five cases were written first. Against the unchanged module, four failed and one
passed.

```
FAILED tests/test_get_stream_byte_cap_truncation.py::test_get_stream_returns_no_more_bytes_than_the_byte_cap
FAILED tests/test_get_stream_byte_cap_truncation.py::test_get_stream_returns_the_earliest_bytes_of_the_run
FAILED tests/test_get_stream_byte_cap_truncation.py::test_get_stream_returns_no_byte_where_the_byte_cap_stands_below_zero
FAILED tests/test_get_stream_byte_cap_truncation.py::test_the_module_states_the_parity_rule_that_decides_the_truncation
4 failed, 1 passed
```

**The one case that passed is `test_get_stream_removes_no_byte_under_the_present_admission_rule`.**
It holds the reading of the base form, and it holds the same reading after the change.

## The mutation reversal

Two mutations ran live, and each one discriminates in both directions.

| Mutation | The reading |
|---|---|
| `return bytes(result)`, which is the base form | 3 failed, 2 passed |
| `return bytes(result[: self.max_stream_bytes])`, which drops the clamp | 1 failed, 4 passed |
| The committed form | 5 passed |

The second mutation returns 145 bytes of a 150-byte run for a cap of -5, because a plain
slice reads the negative value as an offset from the end.

## This change moves no fingerprint value

**A replay of the 38 captures of `tests/foxio_vectors/` through the command-line program
writes 788 values before this change and 788 after, and 0 differ.** The replay reads
`ja4plus analyze --format json` on each capture and compares every field of every value.

`tests/foxio_deviations.json` holds 138 keys before and after.

## The five gates

| Gate | Before | After |
|---|---|---|
| `pytest tests/ -m spec_validation` | 1684 passed, 142 skipped, 138 xfailed | 1684 passed, 142 skipped, 138 xfailed |
| `pytest tests/ -m "not spec_validation"` | 5872 passed, 8 skipped, 8 xfailed, 114 subtests passed | 5880 passed, 8 skipped, 8 xfailed, 114 subtests passed |
| `ruff check ja4plus/ tests/` | All checks passed | All checks passed |
| `ruff format --check ja4plus/ tests/` | 264 files already formatted | 264 files already formatted |
| `mypy --strict ja4plus/` | no issue | no issue in 32 source files |

The unit suite rises by 8, which is the 5 cases of the new file and 3 cases that
parametrize over the tracked test files.

## What this change does not touch

- **The admission rule of `add_segment`.** #654 left it with the user, and no line of it
  moves. `ja4plus/utils/tcp_stream.py:151` still holds the condition the register row
  cites, on the same line.
- **`docs/specs/spec.md`.** The register row of #654 records the admission rule, and this
  change moves no row of it.
- **`CHANGELOG.md` and the `## Changelog` table.** #727 moved the round record to the
  batch pull request.

Verified against: `Crank-Git/ja4plus-go`, `internal/parser/tcp_stream.go:225`, at commit
`a83caa56ef39ff6b4b524a104e4ea30b9e2bdffa` on branch `dev`, read 2026-08-16. This pull
request edits no file of the port, and no case of this repository builds, runs or imports
it.

### DIFF
--- a/ja4plus/utils/tcp_stream.py
+++ b/ja4plus/utils/tcp_stream.py
@@ -223,8 +223,8 @@ class TCPStreamReassembler(StateTable):
                 break
 
         # This truncation removes no byte today. `add_segment` refuses the segment that
-        # would cross the byte cap, so one stream stores no more bytes than the cap and
-        # the run this method builds never reaches it. #697 records that reading.
+        # would cross the byte cap. One stream therefore stores no more bytes than the
+        # cap, and the run this method builds never reaches it. #697 records the reading.
         #
         # The truncation stands here because `GetStream` of `Crank-Git/ja4plus-go`
         # truncates. FoxIO states no byte cap on a stream reader, so parity rule 2 of
--- a/ja4plus/utils/tcp_stream.py
+++ b/ja4plus/utils/tcp_stream.py
@@ -196,8 +196,9 @@ class TCPStreamReassembler(StateTable):
     def get_stream(self, key: Any) -> bytes:
         """Reassemble and return contiguous stream data from base_seq.
 
-        Returns data from the earliest sequence number up to the first gap. The order
-        holds across a wrap of the 32-bit sequence number.
+        Returns data from the earliest sequence number up to the first gap, or up to
+        `max_stream_bytes`, whichever comes first. The order holds across a wrap of the
+        32-bit sequence number.
         """
         if key not in self.streams:
             return b""
@@ -221,9 +222,22 @@ class TCPStreamReassembler(StateTable):
             else:
                 break
 
-        # `add_segment` bounds the bytes the stream stores, and the result never holds
-        # more than the stream stores, so this method needs no bound of its own.
-        return bytes(result)
+        # This truncation removes no byte today. `add_segment` refuses the segment that
+        # would cross the byte cap, so one stream stores no more bytes than the cap and
+        # the run this method builds never reaches it. #697 records that reading.
+        #
+        # The truncation stands here because `GetStream` of `Crank-Git/ja4plus-go`
+        # truncates. FoxIO states no byte cap on a stream reader, so parity rule 2 of
+        # `CLAUDE.md` gives the port the interface.
+        #
+        # #654 records the admission rule of each port in the divergence register, and the
+        # user decides which form this project keeps. The port admits the crossing segment
+        # and stores 250 bytes past the cap. A move to that admission rule therefore
+        # reaches this line, and it moves no byte of the value this method returns.
+        #
+        # A byte cap below zero holds no byte, and the port states that rule too. A slice
+        # on the negative value would drop the last bytes of the run instead.
+        return bytes(result[: max(self.max_stream_bytes, 0)])
 
     def base_seq(self, key: Any) -> int | None:
         """Return the sequence number the reassembled stream starts at, or None.
--- a/tests/test_get_stream_byte_cap_truncation.py
+++ b/tests/test_get_stream_byte_cap_truncation.py
@@ -1,0 +1,109 @@
+"""`get_stream` truncates the contiguous run at the byte cap.
+
+#654 found that `get_stream` returned the stored bytes with no truncation, and that
+`GetStream` of the port truncates. #697 holds the answer: this project truncates, and
+parity rule 2 of `CLAUDE.md` decides it. FoxIO specifies no byte cap on a stream reader,
+so the port holds the interface and this project adopts it.
+
+**The truncation removes no byte today.** `add_segment` refuses the segment that would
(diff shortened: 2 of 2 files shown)


Crank-Git:
## Self-review

**Two lenses ran and both reported.** A correctness reviewer read the diff against the port
and against the admission rule. A writing reviewer read every new sentence against
`.claude/rules/ste.md`. **No third lens ran**, because this change touches no transport
layer, no user interface and no external interface.

### The correctness lens found no defect

**It measured the equivalence rather than asserted it.** The port truncates inside its loop,
the moment the run passes `MaxBytes`, and this project slices the whole run at the end. The
reviewer emulated the loop of the port in Python and read it against `get_stream` over 2000
randomized trials. Each trial varied the segment count, the arrival order, the overlaps and
the cap, and the caps included a negative value, zero and the exact run length. **It reported
0 mismatches.**

It read five other questions and each one answered clean.

- `trim_stream` recomputes `stream["bytes"]` from the surviving segments, so it lowers the
  count and never raises it past the cap.
- The admission condition still stands on line 151, which is the line the register row of
  #654 cites. `tests/test_tcp_byte_cap_admission_ruling.py` reads that line and it passes.
- Each of the five cases fails against a stated wrong reader, and no assertion is vacuous.
- The empty stream, one segment, a cap of exactly the run length, a cap of 0 and a wrap of
  the sequence number each read correctly.
- The `assert` of the helper is a precondition of the setup. A failure there names the
  helper, so it separates a broken setup from a broken reader.

### The writing lens found four sentences past the word limit, and each one is repaired

Commit `3976ae0` splits them. Rule 1 of `.claude/rules/ste.md` holds a description sentence
at 25 words.

| Location | Words |
|---|---|
| `ja4plus/utils/tcp_stream.py`, the comment above the return | 29 |
| `tests/test_get_stream_byte_cap_truncation.py`, the module docstring | 31 |
| The docstring of `_stream_past_the_byte_cap` | 28 |
| The docstring of `test_get_stream_returns_no_byte_where_the_byte_cap_stands_below_zero` | 27 |

**Two comments above a constant stated what the constant holds rather than the reason.** One
of them goes out and the other one now states the reason. Every other rule of the standard
read clean: the active voice, the present tense, the articles, the test names and the
docstring openers.

### One finding this pull request declines

**`byte cap` is not a row of the `## Terms` table of `docs/specs/spec.md`.** The writing lens
raised it, and this pull request writes no row.

Three readings decline it. The gap predates this change, and `docs/specs/spec.md` already
uses the words in the divergence register and in several Changelog rounds. **The cross-check
of batch #738 states that no plan of this batch writes `docs/specs/spec.md`**, and it names
that file as the one this batch would collide on. The filing rule of `CLAUDE.md` reaches a
record with no issue, so this finding earns no issue either. **The verdict of this member
carries it to the project manager.**

Issue #742, opened 2026-08-16
The command prints unknown for every value close_open_windows publishes


Crank-Git:
**This issue earns a place under the five-case filing rule of `CLAUDE.md`, as a user-visible
output.** The batch review of pull request #741 found it on 2026-08-16.

## The measurement

**The command-line program prints `unknown` in place of the address pair for every JA4L value
that `close_open_windows` publishes.**

```
$ ja4plus analyze tests/foxio_vectors/tls3.pcapng
unknown                                             ja4l        JA4L-S=3583_57_quic
```

**The fingerprinter attributes that value correctly**, and the loss happens in the command.

```python
{'fingerprint': 'JA4L-S=3583_57_quic',
 'connection': 'udp_104.21.234.234:443_192.168.1.169:61884',
 'src': '104.21.234.234', 'dst': '192.168.1.169',
 'srcport': 443, 'dstport': 61884}
```

## The cause

`_close_open_windows` at `ja4plus/cli.py:343` drops the four endpoint fields of the entry and reads
them back from the connection key instead. `_endpoints_from_connection` at `ja4plus/cli.py:126`
reads one key form, which its docstring states at `ja4plus/cli.py:130-131`.

```
The key form is `<address>:<port>-<address>:<port>`.
```

**The JA4L connection key holds no hyphen.** It is `udp_104.21.234.234:443_192.168.1.169:61884`, so
`connection.partition("-")` at `ja4plus/cli.py:143` finds no separator and the function returns
`"", 0, "", 0` at `ja4plus/cli.py:145`. `_connection_label` at `ja4plus/output.py:101` then returns
`unknown`.

## Why the defect arrived now

**The docstring of `_close_open_windows` states the premise that was true when it was written.**

```
JA4SSH is the only method that holds a window
```

**#606 ended that premise on 2026-08-16.** It moved the QUIC `JA4L-S` emission to the packet that
fills point `D`, and `JA4LFingerprinter.close_open_windows` publishes every value point `D` never
releases. The register row of #606 records that **22 values of the committed captures reach that
path**, and 20 of them sit on `tls-handshake.pcapng`.

## Why the conformance suite reports nothing

**The conformance suite reads the fingerprinter and never the command.** The entry carries the
address pair, so the suite compares the value at the key `tls3.pcapng/25:61884/JA4L-S.1` and it
reports 1676 passed, 142 skipped and 138 xfailed with the defect live.

[...]

Issue #606, opened 2026-08-13
The QUIC JA4L-S value is emitted from the point B packet, and the reference uses point D


Crank-Git:
## The difference

**`ja4plus-go` moved its QUIC `JA4L-S` emission to the packet that fills point D. `ja4plus` still emits it from the packet that fills point B. A user who runs both implementations now gets the value on two different frames.**

## What the Go side measured

`Crank-Git/ja4plus-go` issue #447 read every FoxIO reference at the pinned commit `27f0cbf9fd3000c072f82a0f7d0361dc99acf6c8` and found them unanimous: **`wireshark/source/packet-ja4.c` emits both `ja4ls` and `ja4l` on the frame that fills `timestamp_D`.**

**The value is identical and the frame differs**, so a per-packet vector holds it at a different frame and the comparison fails.

Four pairs state the shape:

| Capture | The old frame | The reference frame | The value |
|---|---|---|---|
| `ssh2.pcapng` | 1042 | 1046 | `16192_57_quic` |
| `ssh2.pcapng` | 1140 | 1147 | `5389_57_quic` |
| `tls3.pcapng` | 144 | 147 | `4213_59_quic` |
| `tls3.pcapng` | 293 | 297 | `3051_57_quic` |

**The Go change closed 14 deviations and raised the whole-corpus match count from 1658 to 1664.** It also removed 21 register entries that the change made stale, and the register fell from 459 keys to 438.

## Where the port emits it today

**`ja4plus/fingerprinters/ja4l.py:549-551` returns the `JA4L-S` value from `_quic_server_initial`**, which the server Initial packet reaches. **That is the point B packet.**

**Read that line at the current tip rather than trusting this citation** — it was read on 2026-08-13 against tag `v1.1.0`.

## Why this is a defect and not a ruling

**`.claude/rules/parity.md` row 1 governs it: the FoxIO implementations agree and the implementation differs, so the code changes and the register gains nothing.** No reference split, and no maintainer ruling is needed.

**The rule that makes this urgent is the parity rule itself**: a change that lands on one side alone produces exactly the disagreement parity exists to prevent. **The Go side has landed. This is the other half.**

## What this issue builds

- The QUIC `JA4L-S` value is emitted from the packet that fills point D, not point B.
- **A test holds the emission packet and fails when it moves back.**
- **The conformance measurement states how many values moved and on which captures**, before and after.
- **Any register entry the change makes stale is removed in the same change.**

## One difference the Go side could not close, for reference

[...]

PR #745, opened 2026-08-16
Print the address pair of every window a capture leaves open

Crank-Git:
## What this change does

`ja4plus analyze` printed `unknown` in place of the address pair for every JA4L value that
`close_open_windows` publishes. 22 values of the committed captures printed that word. The
command now prints the pair.

```
$ ja4plus analyze tests/foxio_vectors/tls3.pcapng   # before
unknown                                             ja4l        JA4L-S=3583_57_quic

$ ja4plus analyze tests/foxio_vectors/tls3.pcapng   # after
104.21.234.234:443 -> 192.168.1.169:61884           ja4l        JA4L-S=3583_57_quic
```

## Warning: the issue names the wrong function, and this change measured it

**The issue body states that `_close_open_windows` at `ja4plus/cli.py:343` drops the four
endpoint fields of the entry. `Processor.close_open_windows` at `ja4plus/processor.py:254`
drops them, and the command never receives them.** The fingerprinter entry carries all
four, and the processor built a result dict of three keys.

```python
# ja4plus/processor.py, before this change
{"type": fp_type, "fingerprint": entry["fingerprint"], "connection": entry.get("connection")}
```

**The plan therefore names one file where the repair needs two.** The cause chain of the
issue holds otherwise, and so does every acceptance criterion. The change adds the four
keys to the result dict of the processor. That addition breaks no caller: every case of
`tests/test_processor.py`, `tests/test_method_pages.py` and `tests/test_thread_safety.py`
reads a named key and none reads the key set.

**The Go port already carries the endpoints through this path, and parity rule 2 adopts
it.** `Processor.CloseOpenWindows` of `processor.go:239` returns `[]FingerprintResult`, and
`JA4SSHFingerprinter.CloseOpenWindows` of `ja4ssh.go:590` fills the address pair from the
connection state. The port parses no key. This change adopts the source and not the
container type, so `close_open_windows` returns dicts as before.

## Which source the command reads, and why

**The entry of the fingerprinter is the first source, and the connection key is the
fallback.** The entry states the pair the fingerprinter attributed the value to. A key
restates that same pair in a form each fingerprinter picks for itself, and the two forms
drift as they did here.

**Two fingerprinters hold a window, and the fallback reads the key form of each one.**

| Fingerprinter | The key form it writes | Endpoint fields on the entry |
|---|---|---|
| `JA4LFingerprinter` | `<protocol>_<address>:<port>_<address>:<port>`, which `_reported_key` of `ja4plus/fingerprinters/ja4l.py:376` builds | All four |
| `JA4SSHFingerprinter` | `<address>:<port>-<address>:<port>` | None |

`JA4SSHFingerprinter` states no endpoint field on a closed window, so the fallback stays
and it serves that fingerprinter today.

## The count

**A census of 2026-08-16 read 28 values that `close_open_windows` publishes over the
committed captures: 22 from JA4L and 6 from JA4SSH.** The 22 are the count that the
register row of #606 states, and all 22 printed `unknown`. The 6 JA4SSH values printed a
correct pair, because the key parse read their form.

| Capture | Values |
|---|---|
| `ssh2.pcapng` | 2 |
| `ssh-scp-1050.pcap`, `ssh.pcapng`, `ssh2-malformed.pcap`, `ssh2-moloch-crash.pcap`, `tcpdump-geneve.pcap`, `tls3.pcapng` | 1 each |

## Why the present guards missed the defect

**Every guard that stood read a fingerprint entry, and no guard read a printed line.** The
entry of `JA4LFingerprinter` carries the correct address pair, so a reader of the entry
sees no defect. The conformance suite compares the value at the key
`tls3.pcapng/25:61884/JA4L-S.1` and it reads no line of the command. It reported 1684
passed with the defect live.

`tests/test_cli_closed_window_endpoints.py` reads the printed line of the command for the
eight captures that hold an open window.

## How this was tested

Test-driven. The new module failed 10 cases and 3 subtests against the defect, and the
first failure is the acceptance criterion:

```
FAILED TheCommandPrintsAnAddressPairForEveryClosedWindow::test_the_command_prints_the_address_pair_of_the_quic_server_value_of_tls3
```

**The change moves no fingerprint value.** A replay of all 38 files of
`tests/foxio_vectors/` read 788 values before the change and 788 after it, and the two
dumps compare byte-identical. `tests/foxio_deviations.json` holds 138 keys before and
after.

The five gates, run in this worktree:

| Gate | Result |
|---|---|
| `pytest tests/ -m "not spec_validation"` | 5892 passed, 8 skipped, 8 xfailed |
| `pytest tests/ -m spec_validation` | 1684 passed, 142 skipped, 138 xfailed |
| `ruff format --check ja4plus/ tests/` | 265 files already formatted |
| `mypy --strict ja4plus/` | no issues found in 32 source files |

The conformance reading equals the baseline of the base branch, which this worktree
measured before the change.

## What this change leaves

**Four other prose sites state the premise that #606 ended, and this change repairs the
two in the files it touches.** `ja4plus/cli.py` and `ja4plus/processor.py` no longer name
JA4SSH as the only method that holds a window. These four keep the older sentence, and no
file of this change touches them.

- `ja4plus/fingerprinters/base.py:166`
- `docs/api_reference.md:63`
- `tests/test_processor.py:99`, `tests/test_ja4ssh_windows.py:317` and
  `tests/conformance_index.py:243`, each in a docstring or a comment.

Refs #742

### DIFF
--- a/ja4plus/cli.py
+++ b/ja4plus/cli.py
@@ -123,14 +123,53 @@ def _packet_timestamp(packet: Packet) -> datetime | None:
         return None
 
 
+def _endpoints_of_entry(entry: dict[str, Any]) -> tuple[str, int, str, int]:
+    """Return the four endpoint values of one window the processor closed.
+
+    The entry of the fingerprinter is the first source, and the connection key is the
+    fallback. #742 records the reason for that order. The entry states the pair the
+    fingerprinter attributed the value to, and a key restates the same pair in a form
+    each fingerprinter picks for itself.
+
+    `JA4LFingerprinter` states all four fields on a closed window.
+    `JA4SSHFingerprinter` states none, so the fallback reads its key.
+
+    Args:
+        entry: One result dict of `Processor.close_open_windows`.
+
+    Returns:
+        A tuple of the source address, the source port, the destination address and the
+        destination port. An entry that states no address and carries a key this call
+        cannot read produces empty addresses and the port zero.
+    """
+    src_ip = entry.get("src")
+    dst_ip = entry.get("dst")
+    if src_ip or dst_ip:
+        return (
+            str(src_ip or ""),
+            int(entry.get("srcport") or 0),
+            str(dst_ip or ""),
+            int(entry.get("dstport") or 0),
+        )
+    return _endpoints_from_connection(str(entry.get("connection") or ""))
+
+
 def _endpoints_from_connection(connection: str) -> tuple[str, int, str, int]:
     """Return the four endpoint values that a connection key names.
 
-    A window that no packet closes carries the connection key and no packet, so the
-    command reads the endpoints back from the key. The key form is
-    `<address>:<port>-<address>:<port>`. An IPv6 address holds a colon and no hyphen, so
-    the first hyphen separates the two endpoints and the last colon of each half
-    separates the port.
+    This call is the fallback of `_close_open_windows`. It reads a window whose entry
+    states no endpoint, and `JA4SSHFingerprinter` writes every such entry today.
+
+    The two fingerprinters that hold a window write two key forms, and this call reads
+    both. #742 measured what one form costs: the call read the hyphen form alone, so it
+    returned nothing for the 22 JA4L values of the committed captures.
+
+    - `JA4SSHFingerprinter` writes `<address>:<port>-<address>:<port>`.
+    - `JA4LFingerprinter` writes `<protocol>_<address>:<port>_<address>:<port>`, which
+      `_reported_key` builds.
+
+    An IPv6 address holds a colon and neither separator, so the first hyphen or the two
+    underscores name the two halves. The last colon of each half separates the port.
 
     Args:
         connection: The connection key the fingerprinter reported.
@@ -142,7 +181,15 @@ def _endpoints_from_connection(connection: str) -> tuple[str, int, str, int]:
     """
     first, separator, second = connection.partition("-")
     if not separator:
-        return "", 0, "", 0
+        # The underscore form opens with the protocol name, and an address holds no
+        # underscore. The first underscore therefore ends the protocol and the next one
+        # separates the two halves.
+        _, separator, remainder = connection.partition("_")
+        if not separator:
+            return "", 0, "", 0
+        first, separator, second = remainder.partition("_")
+        if not separator:
+            return "", 0, "", 0
     src_ip, src_port = _split_endpoint(first)
     dst_ip, dst_port = _split_endpoint(second)
     return src_ip, src_port, dst_ip, dst_port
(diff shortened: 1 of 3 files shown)


Crank-Git:
## Self-review

**One agent reported and no agent is absent.** A general-purpose child at the Sonnet tier
read the commit under three lenses: correctness, hostile input and the writing standard. It
returned every lens. `.claude/rules/agent-reports.md` of this branch asks for this sentence.

### The findings it took, and the repair of each one

**A stale citation, `ja4plus/cli.py`.** The docstring of `_endpoints_from_connection` read
`This call is the fallback of _close_open_windows`. The caller is `_endpoints_of_entry`
after this change. The repair names that function.

**Three sentences past the 25-word limit of `.claude/rules/ste.md`.** One stands in each of
`ja4plus/cli.py`, `ja4plus/processor.py` and `tests/test_cli_closed_window_endpoints.py`,
and each one carries two facts. The repair splits each into two sentences.

### The finding I decline, and the reading

**The child asks `_endpoints_of_entry` to read the entry only where all four fields state a
value.** It reads the two addresses today, so an entry that states an address and no port
takes the entry and reports the port zero. The child asks for the connection key instead.

**I decline it on three readings.**

1. **No fingerprinter reaches that state.** `JA4LFingerprinter` stores `server_endpoints`
   from a packet that produced a QUIC value, and such a packet carries a port layer. A
   packet with no port layer produces no value to attribute.
2. **The output already defines the form.** `display_source` of `ja4plus/output.py:83`
   writes `src_ip -> dst_ip` where a result carries addresses and no port, and #742 asks
   for an address pair. The partial entry therefore prints the pair the fingerprinter
   attributed, and it prints no `unknown`.
3. **The repair costs the addresses where the key is unreadable.** A fallback that fires on
   a missing port returns `"", 0, "", 0` for a key of neither form, so a correct address
   pair reaches the user as `unknown`. That is the defect this pull request removes.

**The child raised the reading and it is a fair one**, because a partial entry is a shape
this code accepts. The three readings above decide against it, and no measurement of this
repository contradicts them.

### Two notes the child recorded, and neither one earns a change

**The added keys break no caller.** The child read every caller of `close_open_windows` in
this repository and in `examples/`, and none of them reads the key set. It ran 297 cases of
the seven modules that reach the method, and all passed.

**Three constants of the new module are census values.** `CAPTURES_THAT_HOLD_A_WINDOW`,
`CLOSED_WINDOW_VALUE_COUNT` and `TLS3_QUIC_SERVER_PAIR` state a reading of 2026-08-16
rather than a value a FoxIO file holds. A capture that changes moves them by hand. The
child names `tests/test_ja4l_quic_point_d_emission.py` as the same pattern, and this
project already accepts it.

### The gates after the repair

`pytest tests/test_cli_closed_window_endpoints.py` reports 11 passed and 8 subtests passed.
`ruff check`, `ruff format --check` and `mypy --strict` each pass. The repair edits prose
alone, so it moves no fingerprint value.

